PR #1, opened 2022-12-20
ci: prepare repo for development

behnazh-w:


.flake8
```diff
@@ -0,0 +1,30 @@
...
+# E266: too many leading '#' for block comment
+# W503: line break before binary operator
+# D105: Missing docstring in magic method
+# D104: Missing docstring in public package
+# D404: First word of the docstring should not be `This`
+# PT009: use a regular assert instead of unittest-style
+ignore = E203,E266,W503,D105,D404,PT009
+per-file-ignores =
+    __init__.py:D104
+    tests/*:D400,D200,D102,D104,D107
```
tromai, line 18: These ignores have missing description in the comment block if you were planning to add it:

- D400
- D200
- D102
- D104
- D107

.gitattributes
```diff
@@ -0,0 +1,12 @@
+# Set default behavior to automatically normalize line endings.
```
tromai, line 1: The copyright header is missing from this file.

.github/codeql/codeql-config.yaml
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) 2022 - 2022, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+name: CodeQL configuration
+paths:
+- src/macaron
```
tromai, line 6: Should we also enable CodeQL on our Go package at `golang/` as well?

.github/dependabot.yaml
```diff
@@ -0,0 +1,38 @@
...
+  commit-message:
+    prefix: chore
+    prefix-development: chore
+    include: scope
+  open-pull-requests-limit: 13
+  target-branch: staging
+  # Add additional reviewers for PRs opened by Dependabot. For more information, see:
+  # https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#reviewers
+  # reviewers:
+  # -
```
tromai, line 38: I think we could enable Dependabot for our Go module as well.

.github/workflows/_build.yaml
```diff
@@ -0,0 +1,134 @@
...
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+
+    # Generate the requirements.txt that contains the hash digests of the dependencies and
+    # generate the SBOM using CyclonDX SBOM generator.
+    - name: Generate requirements.txt and SBOM
+      if: matrix.os == env.ARTIFACT_OS && matrix.python == env.ARTIFACT_PYTHON
+      run: make requirements sbom
+
+    # Remove the old requirements.txt file (which includes _all_ packages) and generate a
```
tromai, line 91: Should we only perform the removal before generating the `requirements.txt` and the SBOM in the above step?

.github/workflows/codeql-analysis.yaml
```diff
@@ -0,0 +1,71 @@
...
+  push:
+    branches:
+    - main
+    - staging
+  pull_request:
+    branches:
+    - main
+    - staging
+    # Avoid unnecessary scans of pull requests.
+    paths:
```
tromai, line 18: Same comment on whether we should also scan our Go packages.

Makefile
```diff
@@ -0,0 +1,270 @@
...
+	flit build --setup-py --format sdist
+dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-docs-html.zip: docs
+	python -m zipfile -c dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-docs-html.zip docs/_build/html
+dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-build-epoch.txt:
+	echo $(SOURCE_DATE_EPOCH) > dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-build-epoch.txt
+
+# Build the HTML documentation from the package's source.
+.PHONY: docs
+docs: docs/_build/html/index.html
+docs/_build/html/index.html: check test
```
tromai, line 232: I don't think we should run the check and test whenever we want to build the documentation as it will take a lot of time. For example, a contributor when wanting to update the RST files in the `docs` directory shouldn't worry about getting the `check` and `test` but only care about whether the building the documentation pass. 

However, always try to build the docs before pushing might be okay as we want to avoid inconsistent in API documentation.

src/macaron/code_analyzer/call_graph.py
```diff
@@ -41,17 +43,18 @@ def has_callee(self) -> bool:
...
-    Parameters
-    ----------
-    root : Node
-        The root call graph node.
-    repo_path : str
-        The path to the repo.
-    """
+    """This is the generic class for creating a call graph."""
 
     def __init__(self, root: Node, repo_path: str) -> None:
```
tromai, line 48: Putting the documentation inside `__init__` might not be rendered in the generated Sphinx documentation at the moment. I think we could tell Sphinx to display it according to the conversation here - https://stackoverflow.com/a/5599712

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -74,6 +74,20 @@ def _has_deploy_command(self, commands: list[list[str]], build_tool: BaseBuildTo
         return ""
 
     def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResultType:
+        """Implement the check in this method.
```
tromai, line 77: Discussion needed: Whether we should put pydoc for child class method that implements the method from the parent class. As far as I know, the pydoc can be inherited, that is why initially I didn't put the pydoc on inherited child class method (so that we are not duplicating the pydoc).

tromai (changes requested):

src/macaron/slsa_analyzer/checks/base_check.py
```diff
@@ -14,35 +15,36 @@
...
-    eval_reqs : list[ReqName]
-        The list of SLSA requirements that this check addresses.
-    result_on_skip : CheckResultType
-        The status for this check when it's skipped based on another check's result.
-    """
+    """This abstract class is used to implement Checks in Macaron."""
 
+    # The dictionary that contains the data of all SLSA requirements.
     SLSA_REQ_DATA = get_requirements_dict()
-    """The dictionary that contains the data of all SLSA requirements."""
```
tromai, line 36: The comment was put under the this class variable so that it would appear in the generated Sphinx API documentation. I wonder if flake8 flagged this one as error?

tests/parsers/yaml/test_yaml_loader.py
```diff
@@ -56,7 +56,7 @@ def test_load(self) -> None:
         schema_file = os.path.join(self.RESOURCES_DIR, "schema.yaml")
         schema: Schema = yamale.make_schema(schema_file)
 
-        assert not YamlLoader.load(os.path.join(self.RESOURCES_DIR, "invalid.yaml"))
+        assert YamlLoader.load(os.path.join(self.RESOURCES_DIR, "invalid.yaml")) == {None: None}
```
tromai, line 59: I wonder why we would not use `not ...` in this case?

tromai (changes requested):

.github/dependabot.yaml
behnazh-w: I've created an issue [here](https://github.com/oracle-samples/macaron/issues/2).

.github/workflows/pr-conventional-commits.yaml
```diff
@@ -0,0 +1,58 @@
...
+      uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
+      with:
+        fetch-depth: 0
+
+    - name: Set up Python
+      uses: actions/setup-python@13ae5bb136fac2878aff31522b9efb785519f984 # v4.3.0
+      with:
+        python-version: '3.11'
+
+    # Install Commitizen without using the package's Makefile: that's much faster than
```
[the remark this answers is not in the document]
behnazh-w: > I see that the `run` command does not use the package's Makefile. Is this as expected?

Yes, it's expected. Doesn't the comment clarify that we don't want to use `Makefile` here due to performance issues?

.github/workflows/pr-conventional-commits.yaml
```diff
@@ -0,0 +1,58 @@
...
+      run: |
+        pip install --upgrade pip wheel
+        pip install 'commitizen ==2.37.1'
+
+    # Run Commitizen to check the title of the PR which triggered this workflow, and check
+    # all commit messages of the PR's branch. If any of the checks fails then this job fails.
+    - name: Check PR title
+      run: echo "$PR_TITLE" | cz check
+      env:
+        PR_TITLE: ${{ github.event.pull_request.title }}
```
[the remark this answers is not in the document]
behnazh-w: > Is it safe for us to provide the title of the PR to the cz tool here (potential unsanitized user input)?

We use environment variables precisely for this reason:
> For inline scripts, the preferred approach to handling untrusted input is to set the value of the expression to an intermediate environment variable.

See https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable.

.github/workflows/_build.yaml
```diff
@@ -0,0 +1,134 @@
+# Copyright (c) 2022 - 2022, Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
behnazh-w: > Could this workflow be named `build.yaml` instead? Because it is a reusing-workflow, I don't think we should name it in a similar manner as a private workflow thinking

This is a private workflow, which is also a reusable workflow. So, keeping `_build.yaml` is preferred.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,214 @@
+# Copyright (c) 2022 - 2022, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+# We run checks on pushing to the specified branches.
+# Pushing to main also triggers a release.
+
+name: Check and Release
+on:
+  workflow_dispatch:
+  # push:
```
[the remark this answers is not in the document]
behnazh-w: > Why is the `push`event being commented here?

This is temporary as we don't want to release when we merge to `main`. Once the dev environment is set and merged, I will uncomment the `push` event.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,214 @@
...
+    - name: Create changelog and bump
+      run: cz bump --changelog --yes
+
+    - name: Push the release
+      run: |
+        git push
+        git push --tags
+
+  # When triggered by the version bump commit, build the package and publish the release artifacts.
+  build:
```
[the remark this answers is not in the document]
behnazh-w: > I don't quite get this step. According to the conditions, it is the same as the `check` and the `bump` steps defined above. Is running the `./.github/workflows/_build.yaml` workflow also redundant as well?

This is how the release workflow works:

1. We merge to `main` and first we check the build.
2. If build succeeds, we create and push a `bump:` commit to `main` and stop.
3. Now the workflow is running again but this time with `bump:` commit. Now the package is actually built, provenances are created and release is made.

Makefile
behnazh-w: > I don't think we should run the check and test whenever we want to build the documentation as it will take a lot of time. For example, a contributor when wanting to update the RST files in the `docs` directory shouldn't worry about getting the `check` and `test` but only care about whether the building the documentation pass.
> 
> However, always try to build the docs before pushing might be okay as we want to avoid inconsistent in API documentation.
Yes, I see your point. But we want to make sure we generate docs when all checks and tests pass. What we can do locally is to comment out the `test` and `check` here if we just want to test creating `docs`.

src/macaron/code_analyzer/call_graph.py
behnazh-w: > Putting the documentation inside `__init__` might not be rendered in the generated Sphinx documentation at the moment. I think we could tell Sphinx to display it according to the conversation here - https://stackoverflow.com/a/5599712

Thanks, I'll add the change to be able to render the docstrings in `__init__` methods.

src/macaron/slsa_analyzer/checks/build_as_code_check.py
behnazh-w: > Discussion needed: Whether we should put pydoc for child class method that implements the method from the parent class. As far as I know, the pydoc can be inherited, that is why initially I didn't put the pydoc on inherited child class method (so that we are not duplicating the pydoc).

I think we should include the docstring in child classes for now and not silence Flake8. It helps readability of the code itself and we won't need o jump around to find the docstrings of the parent class while developing.

src/macaron/slsa_analyzer/checks/base_check.py
behnazh-w: > The comment was put under the this class variable so that it would appear in the generated Sphinx API documentation. I wonder if flake8 flagged this one as error?

Yes, it was flagged as error.

tests/parsers/yaml/test_yaml_loader.py
behnazh-w: > I wonder why we would not use `not ...` in this case?

`YamlLoader.load` is returning `{None: None}` which is not `None`. The test doesn't pass with `not ...`.

.github/workflows/pr-conventional-commits.yaml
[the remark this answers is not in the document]
tromai: Ah sorry, I missed the comment. It's all good.

.github/workflows/_build.yaml
tromai: > Should we only perform the removal before generating the `requirements.txt` and the SBOM in the above step?

I think I misunderstand while looking at this step. I believe the `requirements.txt` in the `Generate requirements.txt and SBOM` step will contain the packages that are not runtime dependencies for Macaron. It is different from the `requirements.txt` file used in the `Prune packages and generate required requirements.txt` step. I will resolve the conversation if you can confirm that I understand it correctly 👍

behnazh-w: >I believe the `requirements.txt` in the `Generate requirements.txt and SBOM` step will contain the packages that are not runtime dependencies for Macaron. It is different from the `requirements.txt` file used in the `Prune packages and generate required requirements.txt`

Yes, that's correct.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,214 @@
...
+    permissions:
+      contents: read
+
+  # Create a new Release on Github from the verified build artifacts, and optionally
+  # publish the artifacts to a PyPI server.
+  release:
+    needs: [build]
+    name: Release
+    outputs:
+      release-tag: ${{ steps.upload-assets.outputs.release-tag }}
```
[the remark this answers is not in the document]
behnazh-w: > I wonder where the values of `release-tag` and `release-url` (e.g. `${{ steps.upload-assets.outputs.release-tag }}`) came from. Are they coming from the reuseable workflow `./.github/workflows/_build.yaml` ?

They come from here: https://github.com/oracle-samples/macaron/blob/4d8d5ab2cd0e35918ae735a5e9b537a1d31ee9d1/.github/workflows/release.yaml#L133-L134

.github/workflows/release.yaml
```diff
@@ -0,0 +1,214 @@
...
+      release-url: ${{ steps.upload-assets.outputs.release-url }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write # To publish release notes.
+    steps:
+
+    - name: Check out repository
```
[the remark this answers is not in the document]
behnazh-w: > As in the release step, I believe that we only reuse the built artifacts without running the build again. I wonder why do we need to checkout the repository again?

We need it to create the `RELEASE_NOTES.md` by commitizen [here](https://github.com/oracle-samples/macaron/blob/4d8d5ab2cd0e35918ae735a5e9b537a1d31ee9d1/.github/workflows/release.yaml#L123).

>The same comment for line 185.

We need it for `gh` to run [here](https://github.com/oracle-samples/macaron/blob/4d8d5ab2cd0e35918ae735a5e9b537a1d31ee9d1/.github/workflows/release.yaml#L196).

Makefile
```diff
@@ -0,0 +1,270 @@
...
+  $(findstring docs,$(MAKECMDGOALS)), \
+  $(findstring prune,$(MAKECMDGOALS)), \
+)
+ifeq ($(NEED_VENV),)
+  # None of the current goals requires a virtual environment.
+else
+  ifeq ($(origin VIRTUAL_ENV),undefined)
+    $(warning No Python virtual environment found, proceeding anyway)
+  else
+    ifeq ($(wildcard .venv/upgraded-on),)
```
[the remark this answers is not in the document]
behnazh-w: > Python virtual environment not yet set up

In this case `.venv/upgraded-on` does not exist which means that `make setup` has not run yet.

Makefile
```diff
@@ -0,0 +1,270 @@
...
+      $(warning Python virtual environment not yet set up, proceeding anyway)
+    endif
+  endif
+endif
+
+# If the project configuration file has been updated (package deps or
+# otherwise) then warn the user and suggest resolving the conflict.
+ifeq ($(shell test pyproject.toml -nt .venv/upgraded-on; echo $$?),0)
```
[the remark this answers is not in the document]
behnazh-w: > When I run `make venv` the first time, this warning is issued. I think we should only issue when the virtual environment existed but `make setup` hasn't been run in the current directory?

When you run `make venv` two times without running `make setup`,  `.venv/upgraded-on` does not exist (because `make setup` has not run yet), Makefile concludes that`pyproject.toml` is newer than (`-nt`) `.venv/upgraded-on`, so we generate the error to guide the user.

Makefile
```diff
@@ -0,0 +1,270 @@
...
+# Check, test, and build artifacts for this package.
+.PHONY: all
+all: check test dist docs
+
+# Create a virtual environment, either for Python3.11 (default) or using
+# the Python interpreter specified in the PYTHON environment variable. Also
+# create an empty pip.conf file to ensure that `pip config` modifies this
+# venv only, unless told otherwise.
+.PHONY: venv
+venv:
```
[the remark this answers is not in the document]
behnazh-w: > Is this behavior intended ?

Yes, `Makefile` by default prints all the recipe.

tromai (approved):

PR #17, opened 2023-01-04
chore: update the copyright header at a specific line

behnazh-w:
Signed-off-by: behnazh-w <behnaz.hassanshahi@oracle.com>

### DIFF
--- a/scripts/dev_scripts/copyright-checker.sh
+++ b/scripts/dev_scripts/copyright-checker.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright (c) 2022 - 2022, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 #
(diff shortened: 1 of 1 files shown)


scripts/dev_scripts/copyright-checker.sh
```diff
@@ -29,25 +30,46 @@ done
...
+            if [ ${#missing_license_note} -eq 0 ]; then
+                expected="$expected\n# $license_note"
+            fi
         fi
 
-        if ! grep -i -e "Copyright (c) .* Oracle and/or its affiliates. All rights reserved" "$f" 1>/dev/null;then
-            echo "Copyright header missing for $f"
-            sed -i "1s/^/$expected\n\n/" "$f"
+        # Find the first matching copyright line.
+        line_number=$(grep -m 1 -n -i -e "Copyright (c) .* Oracle and/or its affiliates. All rights reserved" "$f" | cut -d : -f 1)
```
tromai, line 57: Because we only match this line but not the `license_note` line, does it means that if the `lincense_note` line if missing won't be added if this line is already existed?. For example with this file:
```
#!/usr/bin/env bash

# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.

# other stuffs
```
The `copyright-checker.sh` script won't raise any issue.

tromai (changes requested):

scripts/dev_scripts/copyright-checker.sh
behnazh-w: The script currently mainly checks if the copyright is missing or needs an update. To check if the license is missing only we would need an extra check at the beginning of the script, but that was not the focus of this fix. It should be fine for now as anytime a new file is added the license will be added too, and right now all the files have licenses. I will add an issue to specifically check for the missing license.

tromai (approved):
LGTM

PR #28, opened 2023-01-09
build: remove paths outside the package

behnazh-w:
This PR moves the data resources and path to binaries, which the final `macaron` Python package needs, under the package path and removes paths that reach outside the package.

Signed-off-by: behnazh-w <behnaz.hassanshahi@oracle.com>


tromai:
Running the integration tests (i.e `make integration-test` is not passing as we are still using the old `resources` path) [here](https://github.com/oracle-samples/macaron/blob/07b2c5d1c7b47591cb4f854d4bb841f522b83d02/scripts/dev_scripts/integration_tests.sh#L20).

behnazh-w:
> Running the integration tests (i.e `make integration-test` raise an error message as we are still using the old `resources` path) [here](https://github.com/oracle-samples/macaron/blob/07b2c5d1c7b47591cb4f854d4bb841f522b83d02/scripts/dev_scripts/integration_tests.sh#L20). The tests still passed, however. The error message: `cp: cannot stat '.../macaron/resources/settings.xml': No such file or directory`.

See commit b0f45230df3adfe7e469ae08cd86250979073d97.

tromai (approved):
LGTM

PR #29, opened 2023-01-23
feat!: Improve justification output for provenance_l3_check

ailrst:
Check now displays all provenance artefacts which passed, failed, or were skipped.

The check now fails when _any_ artefact mentioned in the provenance fails to verify, 
or must be skipped.

### DIFF
--- a/src/macaron/slsa_analyzer/checks/provenance_l3_check.py
+++ b/src/macaron/slsa_analyzer/checks/provenance_l3_check.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2022 - 2022, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """This modules implements a check to verify a target repo has intoto provenance level 3."""
@@ -11,6 +11,8 @@ import subprocess  # nosec B404
 import tarfile
 import tempfile
 import zipfile
+from dataclasses import dataclass
+from enum import Enum
 from pathlib import Path
 
 from macaron.config.defaults import defaults
@@ -25,6 +27,27 @@ from macaron.slsa_analyzer.registry import registry
 from macaron.slsa_analyzer.slsa_req import ReqName
 
 logger: logging.Logger = logging.getLogger(__name__)
+
+
+class _VerifyArtefactResultType(Enum):
+    """Result of attempting to verify an asset."""
+
+    PASSED = "verify passed"
+    FAILED = "verify failed"
+    ERROR = "verify error"
+    NO_DOWNLOAD = "unable to download asset"
+    TOO_LARGE = "asset file too large to download"
+
+
+@dataclass
+class _VerifyArtefactResult:
+    """Dataclass storing the result of verifying a single asset."""
+
+    result: _VerifyArtefactResultType
+    artefact_name: str
+
+    def __str__(self) -> str:
+        return str(self.result.value) + ": " + self.artefact_name
 
 
 class ProvenanceL3Check(BaseCheck):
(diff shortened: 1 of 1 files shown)


src/macaron/slsa_analyzer/checks/provenance_l3_check.py
```diff
@@ -287,12 +344,34 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
...
+                result for ci_name, prov_url, result in all_feedback if result not in passed and result not in failed
+            ]
+
+            if failed or skipped:
+                result_value = CheckResultType.FAILED
+
+            if passed:
+                check_result["justification"].append("Successfully verified level 3: ")
+            else:
+                check_result["justification"].append("Failed verification for level 3: ")
```
behnazh-w, line 370: I think `result_value = CheckResultType.PASSED` is missing when verification is successful?

Also, if `passed` is not empty, this would override the results even if `failed` or `skipped` are not empty.

```suggestion
            if failed or skipped:
                result_value = CheckResultType.FAILED
                check_result["justification"].append("Failed verification for level 3: ")
            else:
                result_value = CheckResultType.PASSED
                check_result["justification"].append("Successfully verified level 3: ")
```

src/macaron/slsa_analyzer/checks/provenance_l3_check.py
```diff
@@ -287,12 +344,34 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
...
+
+            check_result["justification"].append(",".join(map(str, all_results)))
+            return result_value
+
         check_result["justification"].append("Could not verify level 3 provenance.")
         return CheckResultType.FAILED
```
behnazh-w, line 376:
```suggestion
        return result_value
```

behnazh-w (changes requested):

behnazh-w:
Is there any particular reason to use `!` in the PR title? Are you expecting a breaking change?

ailrst:
I used `!` in the commit since it changes the check result for an identical check where some assets verify and some don't

src/macaron/slsa_analyzer/checks/provenance_l3_check.py
ailrst: I applied these changes in [155b695](https://github.com/oracle-samples/macaron/pull/29/commits/155b695102190fac8d6607baf12dfd60d465dd78), thank you.

behnazh-w:
> I used `!` in the commit since it changes the check result for an identical analysis target where some assets verify and some don't

Yes, that makes sense. But because we haven't released Macaron yet, we don't need to add `!` for now.

src/macaron/slsa_analyzer/checks/provenance_l3_check.py
```diff
@@ -27,6 +29,27 @@
 logger: logging.Logger = logging.getLogger(__name__)
 
 
+class _VerifyArtefactResultType(Enum):
+    """Result of attempting to verify an asset."""
+
+    PASSED = "verify passed"
+    FAILED = "verify failed"
+    ERROR = "verify error"
```
behnazh-w, line 37: Can you please add comments for these three types to clarify how they are concluded/differentiated?

behnazh-w (changes requested):

src/macaron/slsa_analyzer/checks/provenance_l3_check.py
```diff
@@ -202,10 +230,11 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
         """
         # TODO: During verification, we need to fetch the workflow and verify that it's not
         # using self-hosted runners, custom containers or services, etc.
-        all_feedback = []
+        all_feedback: list[tuple[str, str, _VerifyArtefactResult]] = []
```
behnazh-w, line 233: How about using [`typing.NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple) instead of a plain `tuple` to improve the readability and prevent potential mistakes in looking up fields?

behnazh-w (changes requested):

src/macaron/slsa_analyzer/checks/provenance_l3_check.py
ailrst: That's perfect thanks I didn't know that existed

ailrst: This is addressed in [78b65b2](https://github.com/oracle-samples/macaron/pull/29/commits/78b65b23f3b61676ad407205fc9d7bee39fa7b99)

src/macaron/slsa_analyzer/checks/provenance_l3_check.py
ailrst: This is addressed in [78b65b2](https://github.com/oracle-samples/macaron/pull/29/commits/78b65b23f3b61676ad407205fc9d7bee39fa7b99)

src/macaron/slsa_analyzer/checks/provenance_l3_check.py
behnazh-w: Thanks, LGTM. Let's close this PR and move the changes to https://github.com/oracle-samples/macaron/pull/46 as discussed.

PR #46, opened 2023-01-31
feat: add check output to database and implement souffle policy engine

ailrst:
<p>Preparing Macaron for policy engine which uses the Souffle datalog interpreter.</p>
<h2 id="architecture">Architecture</h2>
<p>The goal is to use souffle to evaluate the policy, while <a href="https://souffle-lang.github.io/directives#input-directive">loading the facts directly from the sqlite database</a>.</p>
<p>For this to work there is the following requirements, from the souffle docs:</p>
<blockquote>
<p>The data is expected to be stored in a table matching the relation name prefixed by an underscore and the sqlite3 database is expected to contain a view matching the relation name. For example, for the relation edge, the sqlite3 database should have a table named _edge.</p>
</blockquote>
<p>To input it there needs to be a relation declared that matches the view, and a corresponding input statement.</p>
<p>So broadly the way this pr works is:</p>
<ol type="1">
<li>Tables are declaratively defined in macaron and checks using SQLAlchemy</li>
<li>From the analyzer class, <code>DatabaseManager.create_tables()</code> creates the database, tables, and views if they don’t exist.</li>
<li>Checks store populated orm-mapped tables to <code>CheckResult["result-tables"]</code></li>
<li><code>Analyzer</code> stores these to the database after analysis is completed, along with the information macaron stores such as the analyzed repositories and dependency tree</li>
</ol>
<p>The policy engine is invoked from a separate script which is passed the database file and a policy file</p>
<pre><code>python -m macaron.policy_engine -h
usage: policy_engine [-h] -d DATABASE [-f FILE] [-s]

options:
  -h, --help            show this help message and exit
  -d DATABASE, --database DATABASE
                        Database path
  -f FILE, --file FILE  Replace policy file
  -s, --show-preamble   Show preamble
$ python -m macaron.policy_engine -d output/macaron.db  -f tests/policy_engine/resources/policies/testpolicy.dl -h</code></pre>
<p>At this stage what this does is</p>
<ol type="1">
<li>The database is opened and all the schema is reflected into the SQLAlchemy metadata</li>
<li>For each table beginning with an <code>_</code> a corresponding souffle declaration and import is generated</li>
<li>Some helper relations and rules are generated</li>
<li>The prelude is constructed by combining the import statements, helper relations, and some additional non-generated rules</li>
<li>A file is created with the generated prelude prepended to the actual policy file</li>
<li>Souffle is invoked on this file, and the results are printed</li>
</ol>
<h2 id="changes-summary">Changes Summary</h2>
<ul>
<li><p>Import SQLAlchemy to manage database connection</p></li>
<li><p>Refactor <code>DatabaseManager</code> to use SQLAlchemy (api change)</p>
<ul>
<li>and add corresponding unit tests</li>
</ul></li>
<li><p><code>AnalyzeContext</code> now returns orm-mapped tables to be inserted into the database, rather than constructing SQL queries</p></li>
<li><p><code>CheckResult</code> has a new field <code>"result_tables: list[Table]"</code></p></li>
<li><p><code>Analyzer</code> now populates tables to store the analysis, dependency, and slsa-level results, and check_results</p></li>
<li><p><code>AnalyzeContext</code> now stores a orm-mapped table to represent the repository being analyzed, which is stored to the database by the <code>Analyzer</code> object before analysis starts</p></li>
<li><p><code>Analyzer</code> stores all tables which checks insert into <code>CheckResult["result_tables"]</code> to the database after analysis</p>
<ul>
<li>Nearly all checks are modified to define and store result tables</li>
</ul></li>
<li><p><code>base_check.py</code> defines a table to store check results</p></li>
<li><p><code>base_check.py</code> defines an SQLAlchemy declarative mixin <code>CheckFactsTable</code> which defines check_result id, and repository id, foreign key fields which when result tables inherit from it, the analyzer will automatically populate.</p></li>
<li><p><code>provenance_l3_check</code> is stricter as per <a href="https://github.com/oracle-samples/macaron/pull/29">pull/29</a>.</p></li>
<li><p>add: <code>policy_engine/__main__.py</code> is the entry point for the policy engine</p></li>
<li><p>add: <code>policy_engine/souffle_code_generator.py</code> contains the logic for generating the souffle datalog for data import</p></li>
<li><p>add: <code>policy_engine/souffle.py</code> contains the wrapper for invoking souffle in a temporary directory</p>
<ul>
<li>and corresponding unit tests</li>
</ul></li>
<li><p><code>policy_engine/policy.py</code> has some changes due to a manually reverted refactor; it will likely have to be refactored again to integrate the policy engine</p></li>
</ul>

### To do

- [x] Policy engine: validate database version before proceeding
- [x] Integrate provenance policies using CUE or proof of concept policy engine
- [x] Have macaron run policy and include result in reports
- [x] Update build; add souffle to docker 
- [x] Filter duplicate analyze_ctx


Makefile
```diff
@@ -86,13 +86,15 @@ venv:
 # We also install cyclonedx-go to generate SBOM for Go, compile the Go modules,
 # install SLSA verifier binary, and download mvnw.
 .PHONY: setup
+.PHONY: souffle
```
behnazh-w, line 89: Can you please move this down closer to the target?

src/macaron/policy_engine/prelude/aggregate_rules.dl
```diff
@@ -0,0 +1,59 @@
...
+  * The build is verifiably automated and .
+  */
+aggregate_level_requirement(2, repo) :-
+    repository(repo, _,_,_,_,_,_),
+    check_passed(repo, "mcn_build_script_1"),
+    check_passed(repo, "mcn_build_service_1"),
+    check_passed(repo, "mcn_build_as_code_1").
+
+
+// The minimum aggregate level a reposotories dependencies have
```
behnazh-w, line 32:
```suggestion
 \**
  * The minimum aggregate level the dependencies of a repository should have.
  */
```

src/macaron/policy_engine/prelude/aggregate_rules.dl
```diff
@@ -0,0 +1,59 @@
...
+ */
+aggregate_level_requirement(1, repo) :-
+    repository(repo, _,_,_,_,_,_),
+    check_passed(repo, "mcn_build_service_1"),
+    check_passed(repo, "mcn_version_control_system_1").
+
+
+/**
+  * The build is verifiably automated and .
+  */
```
behnazh-w, line 24:
```suggestion
/**
 * The build is verifiably automated and deployable.
 */
```

src/macaron/policy_engine/prelude/helper_rules.dl
```diff
@@ -0,0 +1,61 @@
+/* Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved. */
+/* Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/. */
+
```
behnazh-w, line 3: Docs missing.

src/macaron/policy_engine/prelude/policy.dl
```diff
@@ -0,0 +1,20 @@
+/* Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved. */
+/* Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/. */
+
```
behnazh-w, line 3: Docs missing.

src/macaron/policy_engine/souffle.py
```diff
@@ -0,0 +1,177 @@
...
+import shutil
+import subprocess  # nosec B404
+import tempfile
+from types import TracebackType
+from typing import Optional
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class SouffleError(Exception):
```
behnazh-w, line 23: Similar comment as other errors. Please add the discussed TODO.

Makefile
ailrst: Applied in https://github.com/oracle-samples/macaron/pull/46/commits/1efa2349a079341ec72316b8c00d677e0a91e615

Makefile
```diff
@@ -105,6 +107,29 @@ $(MACARON_PATH)/resources/mvnw:
...
+            "Ubuntu")                                                                                                                                      \
+                sudo wget https://souffle-lang.github.io/ppa/souffle-key.public -O /usr/share/keyrings/souffle-archive-keyring.gpg; \
+                echo "deb [signed-by=/usr/share/keyrings/souffle-archive-keyring.gpg] https://souffle-lang.github.io/ppa/ubuntu/ stable main" | sudo tee /etc/apt/sources.list.d/souffle.list; \
+                sudo apt update; \
+                sudo apt install souffle; \
+                ;;                                                                                                                                         \
+	esac;                                                                                                                                                  \
+	fi && \
+	which souffle
+
```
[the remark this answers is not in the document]
ailrst: Applied in https://github.com/oracle-samples/macaron/pull/46/commits/1efa2349a079341ec72316b8c00d677e0a91e615

.flake8
```diff
@@ -22,6 +22,7 @@ ignore = E203,E266,W503,D105,D404,PT009
 per-file-ignores =
     __init__.py:D104
     tests/*:D400,D200,D102,D104,D107
+    src/macaron/database/views.py:B903
```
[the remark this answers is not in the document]
ailrst: Done, in f56ca9efddb5f03f86bdfa627f9d1a72f00db4cd

.pre-commit-config.yaml
```diff
@@ -79,6 +79,7 @@ repos:
     types: [text, python]
     args: [--rcfile, pyproject.toml]
 
+
```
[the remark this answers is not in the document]
ailrst: No it wasn't, I've removed it in f56ca9efddb5f03f86bdfa627f9d1a72f00db4cd

src/macaron/__main__.py
```diff
@@ -76,15 +77,20 @@ def analyze_slsa_levels_single(analyzer_single_args: argparse.Namespace) -> None
     sys.exit(status_code)
 
 
-def verify_prov(verify_args: argparse.Namespace) -> None:
+def verify_prov(verify_args: argparse.Namespace) -> NoReturn:
     """Verify a provenance against a user defined policy."""
     prov_file = verify_args.provenance
-    policy_file = verify_args.policy
+    policy_files = list(filter(lambda path: ".yaml" == os.path.splitext(path)[1], global_config.policy_paths))
```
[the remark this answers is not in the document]
ailrst: True. I also added a todo under the policy class; I think eventually it would be better to have the Policy class define what filetypes it can be constructed from (eg. with a method that is passed the filename)  so the configuration can just ask each policy and pick the first one that says yes. Currently though cue and YAML policies both use the same class, and souffle policies are another unrelated class since they need different interfaces, but the design could still be rationalized better.

src/macaron/dependency_analyzer/dependency_resolver.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2022 - 2022, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
ailrst: These should have been in recent rebase

src/macaron/output_reporter/results.py
```diff
@@ -237,6 +243,23 @@ def get_ctxs(self) -> Iterable[AnalyzeContext]:
             if record.context:
                 yield record.context
 
+    def get_dependencies(self, root_record: Optional[Record] = None) -> Iterable[tuple[AnalyzeContext, AnalyzeContext]]:
+        """Get the generator for all AnalyzeContext instances.
+
```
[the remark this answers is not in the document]
ailrst: Fixed in https://github.com/oracle-samples/macaron/pull/46/commits/7584d191a925261dd109b81512af671aa81d42aa

src/macaron/output_reporter/results.py
```diff
@@ -237,6 +243,23 @@ def get_ctxs(self) -> Iterable[AnalyzeContext]:
...
+        Yields
+        ------
+        Tuple[AnalyzeContext, AnalyzeContext]
+            Dependency
```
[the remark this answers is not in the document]
ailrst: Fixed in https://github.com/oracle-samples/macaron/pull/46/commits/7584d191a925261dd109b81512af671aa81d42aa

src/macaron/policy_engine/exceptions.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""Exceptions used by the policy engine."""
+
+
+class InvalidPolicyError(Exception):
```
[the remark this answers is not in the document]
ailrst: Fixed in https://github.com/oracle-samples/macaron/pull/46/commits/7584d191a925261dd109b81512af671aa81d42aa

src/macaron/policy_engine/policies.md
```diff
@@ -0,0 +1,252 @@
...
+.decl all_y()
+all_y() :- count : nums(x,2) = count : nums(_,_).
+```
+
+
+### Subsumption rule: Ordering
+
+If you want to define a partial order: for example to find a maximum or minimum.
+
+It has the syntax `atom <= atom2 :- disjunction`. It replaces atom1 with atom2 if disjunction holds. The fact `atom1` is
```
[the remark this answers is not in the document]
ailrst: Fixed in https://github.com/oracle-samples/macaron/pull/46/commits/7584d191a925261dd109b81512af671aa81d42aa

src/macaron/policy_engine/prelude/aggregate_rules.dl
```diff
@@ -0,0 +1,59 @@
+/* Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved. */
+/* Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/. */
+
```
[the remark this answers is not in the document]
ailrst: Fixed in https://github.com/oracle-samples/macaron/pull/46/commits/7584d191a925261dd109b81512af671aa81d42aa

src/macaron/policy_engine/prelude/aggregate_rules.dl
ailrst: Fixed in https://github.com/oracle-samples/macaron/pull/46/commits/7584d191a925261dd109b81512af671aa81d42aa

src/macaron/policy_engine/prelude/helper_rules.dl
ailrst: Fixed in https://github.com/oracle-samples/macaron/pull/46/commits/3e2695dbb0d81e9e9eb8f3c3d2f49405ccce4ac7

src/macaron/policy_engine/prelude/aggregate_rules.dl
ailrst: Fixed in https://github.com/oracle-samples/macaron/commit/7584d191a925261dd109b81512af671aa81d42aa

src/macaron/policy_engine/souffle.py
ailrst: Fixed in https://github.com/oracle-samples/macaron/commit/7584d191a925261dd109b81512af671aa81d42aa

src/macaron/policy_engine/prelude/policy.dl
ailrst: Fixed in https://github.com/oracle-samples/macaron/commit/7584d191a925261dd109b81512af671aa81d42aa

src/macaron/util.py
```diff
@@ -230,3 +230,17 @@ def copy_file_bulk(file_list: list, src_path: str, target_path: str) -> bool:
...
     return True
+
+
+JsonType = int | float | str | None | bool | list["JsonType"] | dict[str, "JsonType"]
+
+
+def get_if_exists(doc: JsonType, path: list[str | int]) -> JsonType | None:
+    """Get a json dict value if it exists."""
+    while len(path) > 0:
+        this: str | int = path.pop(0)
```
tromai, line 241: Is it okay if we don't check for the type of `this` here? The caller of this function could provide with a list of invalid members as the `path`.

ailrst: If they do that they're already violating the contract of the method per the hints in the signature. But the hint on this _is_ redundant as it just repeats the hint on path.

src/macaron/__main__.py
```diff
@@ -74,15 +75,20 @@ def analyze_slsa_levels_single(analyzer_single_args: argparse.Namespace) -> None
     sys.exit(status_code)
 
 
-def verify_prov(verify_args: argparse.Namespace) -> None:
+def verify_prov(verify_args: argparse.Namespace) -> NoReturn:
```
tromai, line 78: According to Python 3.11's documentation, `Never` is recommended to be used here instead - https://docs.python.org/3/library/typing.html#typing.NoReturn. I wonder if pyupgrade could detect such details :thinking:

behnazh-w: > According to Python 3.11's documentation, `Never` is recommended to be used here instead - https://docs.python.org/3/library/typing.html#typing.NoReturn (from Python 3.11). I wonder if pyupgrade could detect such details thinking

We have [ticket](https://github.com/oracle-samples/macaron/issues/52) to set `pyupgrade`'s version to `py 3.11. I'm expecting a lot of code rewrite when that Python version changes :wink:

pyproject.toml
```diff
@@ -25,8 +25,9 @@ dependencies = [
     "requests >=2.28.0,<3.0.0",
     "pydriller >=2.0,<3.0.0",
     "yamale >=4.0.3,<5.0.0",
-    "packaging >=21.3,<24.0.0",
-    "jinja2 >=3.1.2,<4.0.0"
+    "packaging >=21.3,<23.0.0",
```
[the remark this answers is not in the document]
ailrst: I'm not sure, it might have been introduced in a merge conflict, I've put it back to `    "packaging >=21.3,<24.0.0"` now and its fine

src/macaron/config/global_config.py
```diff
@@ -60,9 +64,23 @@ def load(
...
         self.local_repos_path = local_repos_path
         self.gh_token = gh_token
-        self.policy_path = policy_path
         self.resources_path = resources_path
 
+        # Find the policies
+        policy_files = []
+        for policy_path in policy_paths:
+            if os.path.isdir(policy_path):
+                for file_name in os.listdir(policy_path):
```
tromai, line 73: I wonder if using `glob` here would be possible?

For example:
```python
for file_path in glob.iglob(f"{policy_path}/**", recursive=False)
    ...
```

src/macaron/output_reporter/results.py
```diff
@@ -83,6 +83,8 @@ class Record(Generic[RecordNode]):
         The context instance for this repo.
     dependencies : list[RecordNode]
         The list of Records for the analyzed dependencies of this repo.
+    policy_violation: bool
```
tromai, line 86: I don't think this attribute is accessed/initialized anywhere in the code.

src/macaron/output_reporter/results.py
```diff
@@ -95,6 +97,8 @@ class Record(Generic[RecordNode]):
     description: str
     pre_config: Configuration
     status: SCMStatus
+    policies_passed: list
```
tromai, line 100: The code documentation for these two attributes do not exist (if those attributes are final).

src/macaron/policy_engine/__main__.py
```diff
@@ -0,0 +1,178 @@
...
+    """Invoke souffle and report result."""
+    with SouffleWrapper() as sfl:
+        copy_prelude(config.database_path, sfl)
+        with open(policy_file, encoding="utf-8") as file:
+            text = file.read()
+
+        try:
+            res = sfl.interpret_text(text)
+        except SouffleError as error:
+            print(error.command)
```
behnazh-w, line 115: Please use logger and avoid `print`. Please check other uses of `print` too.

src/macaron/policy_engine/prelude/aggregate_rules.dl
```diff
@@ -0,0 +1,87 @@
...
+/**
+ *  Souffle datalog rules defining an aggregate approximate security level for a repository.
+ *  This is not a SLSA level, but is based on Macaron checks which are inspired by SLSA requirements.
+ *
+ *  For a repository to meet one level, it must meet the requirements of all lower levels.
+ *  These levels do consider the level of a repository's dependencies.
+ */
+
+/* aggregate verifying a level */
+.decl aggregate_level(level: number, repo: number)
```
tromai, line 13: How are `aggregate_level` and `aggregate_level_requirement` differ from each other? :thinking:

ailrst: `aggregate_level` considers the dependencies of `repo` as well

tromai: Would it be okay to clarify it using a short comment?

src/macaron/policy_engine/prelude/helper_rules.dl
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Rules for referring to parts of a json type using the full path name.
+ */
+
+.decl json(name: symbol, id: number, root: JsonType)
+
+
+.decl json_path(j: JsonType, a: JsonType, key:symbol)
+
+json_path(a, b, key) :- a = $Object(k, b), json(name,_,a), key=cat(name, cat(".", k)).
```
tromai, line 60: Just a small note on the inconsistent spacing of the attributes in the relations. Perhaps we could create a ticket for resolving it later. Btw, have you encountered any linter for souffle :thinking: ?

src/macaron/policy_engine/prelude/aggregate_rules.dl
```diff
@@ -0,0 +1,87 @@
...
+    check_passed(repo, "mcn_provenance_level_three_1"),
+    check_passed(repo, "mcn_trusted_builder_level_three_1"),
+    check_passed(repo, "mcn_policy_check_1"),
+    not_self_hosted_git(repo, _).
+
+/**
+ * The minimum aggregate level the dependencies of a repository should have.
+ */
+.decl aggregate_level_min_dependency_level(level: number, dep_level:number)
+aggregate_level_min_dependency_level(1, 1).
```
tromai, line 63: Are these rules being used anywhere or it's left there as helper rules?

src/macaron/policy_engine/prelude/helper_rules.dl
behnazh-w: > Just a small note on the inconsistent spacing of the attributes in the relations. Perhaps we could create a ticket for resolving it later. Btw, have you encountered any linter for souffle thinking ?

I have come across this repo to lint Souffle Datalog: https://github.com/langston-barrett/souffle-lint
It could be something to add as a pre-commit hook locally, but haven't tried it out yet.

src/macaron/policy_engine/prelude/policy.dl
```diff
@@ -0,0 +1,79 @@
...
+ *      policy_id: The unique identifier for this policy.
+ *      repo: the id field of a repository relation.
+ */
+.decl Policy(policy_id: symbol, repo: number, message: symbol)
+
+/**
+ *  Specifies that a repository repo must satisfy the policy policy_id: that the fact Policy(policy_id, repo)
+ *  must exist.
+ */
+.decl apply_policy_to(policy_id: symbol, repo: number)
```
tromai, line 34: In the documentation, it is mentioned that `that the fact Policy(policy_id, repo)  must exists`. However, I don't see any rules defined for the relation between apply_policy_to and Policy, what would be the best way to understand this rule?

ailrst: These are the interface for defining policies, as used in the example in `tests/policy_engine/resources/policies/test_policy.dl`

src/macaron/policy_engine/souffle.py
```diff
@@ -0,0 +1,179 @@
...
+    output_dir: str
+    include_dir: str
+    fact_dir: str
+    library_dir: str
+    souffle_stdout: Optional[str]
+    souffle_stderr: Optional[str]
+
+    TEMP_SOURCEFILE_NAME = "source.dl"
+
+    def __init__(
```
tromai, line 58: Could you add some documentation on what the parameters are? I think the ones that might be confusing are `include_dir` vs `library_dir`.

src/macaron/policy_engine/souffle.py
```diff
@@ -0,0 +1,179 @@
...
+        self.souffle_stdout = str(result.stdout.decode("utf-8"))
+
+    def copy_to_includes(self, filename: str, text: str) -> None:
+        """Create a file with in the include directory.
+
+        Parameters
+        ----------
+        filename: str
+            The base name of the file
+        text:
```
tromai, line 123:
```suggestion
        text: str
```

src/macaron/policy_engine/souffle.py
```diff
@@ -0,0 +1,179 @@
...
+        with open(os.path.join(self.include_dir, filename), "w", encoding="utf-8") as file:
+            file.write(text)
+
+    def interpret_file(self, filename: str, with_prelude: str = "") -> dict:
+        """
+        Interpret a file.
+
+        Parameters
+        ----------
+            filename: str the file to run
```
tromai, line 135:
```suggestion
            filename: str 
                The file to run
            with_prelude: str 
                String literal to append to the start of the file before running it
```

tests/policy_engine/test_souffle.py
```diff
@@ -0,0 +1,65 @@
...
+        with SouffleWrapper(fact_dir=str(self.FACT_DIR)) as sfl:
+            result = sfl.interpret_text(self.TEXT)
+            assert result == {"path": [["1", "2"], ["1", "3"], ["2", "3"]]}
+
+    def test_error(self) -> None:
+        """Test throwing an error on an invalid souffle program."""
+        with SouffleWrapper() as sfl:
+            self.assertRaises(SouffleError, sfl.interpret_text, ".input edge")
+
+    def test_consecuitve(self) -> None:
```
tromai, line 45: Did you mean to use `test_consecutive` here?

src/macaron/policy_engine/prelude/aggregate_rules.dl
ailrst: They were supposed to be but the implementation of aggregate_level got deleted in the refactor by mistake, its added back now in https://github.com/oracle-samples/macaron/pull/46/commits/ee565267d152c2e1439bec8ed96c0730d4fc8f6d. In `aggregate_rules` everything aside from the `aggregate_level` relation and the `Policy` relations are internal helper methods.

src/macaron/policy_engine/prelude/aggregate_rules.dl
ailrst: Clarified in https://github.com/oracle-samples/macaron/pull/46/commits/ee565267d152c2e1439bec8ed96c0730d4fc8f6d

src/macaron/policy_engine/__main__.py
ailrst: Fixed in https://github.com/oracle-samples/macaron/pull/46/commits/ee565267d152c2e1439bec8ed96c0730d4fc8f6d

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -167,18 +171,46 @@ def run(self, user_config: dict, skip_deps: bool = False) -> int:
             find_ctx = report.find_ctx(dup_record.pre_config.get_value("path"))
             dup_record.context = find_ctx
 
-        # Store the analysis result into the SQLite database.
-        for ctx in report.get_ctxs():
-            self.store_result_to_db(ctx)
+        # Store the analysis result(s) into the SQLite database.
+        self.db_man.create_tables()
```
behnazh-w, line 175: Looks like here all we need to do is to create tables for the new check's results. Instead of calling `create_tables()` to create all the tables (if they don't exist), it would make sense to have a function in the db manager that only adds the new tables for the checks results in Macaron. WDUT?

src/macaron/slsa_analyzer/table_definitions.py
```diff
@@ -0,0 +1,136 @@
...
+
+class AnalysisTable(ORMBase):
+    """
+    ORM Class for the analysis information.
+
+    This information pertains to a single invocation of the macaron tool.
+    """
+
+    __tablename__ = "_analysis"
+    id = Column(Integer, primary_key=True, autoincrement=True)  # noqa: A003
```
behnazh-w, line 47: Using `autoincrement` for `Integer` types makes primary keys predictable, which is not usually a good security practice. Instead using `UUID`s is encouraged. Can you please add a TODO to revisit this for all the tables? If we need order, we should find other mechanisms to establish it, like timestamp or a separate `autoincrement` column just for keeping orders.

src/macaron/slsa_analyzer/table_definitions.py
```diff
@@ -0,0 +1,136 @@
...
+    """Table storing the SLSA requirements a repository satisfies."""
+
+    __tablename__ = "_slsa_requirement"
+    repository = Column(Integer, ForeignKey("_repository.id"), primary_key=True)
+    requirement = Column(String, primary_key=True)
+    requirement_name = Column(String, primary_key=False)
+    feedback = Column(String, nullable=True)
+
+
+class RepositoryDependency(ORMBase):
```
behnazh-w, line 31: Is this a many-to-many association table for two repositories? I'm trying to understand.

src/macaron/slsa_analyzer/database_store.py
```diff
@@ -0,0 +1,105 @@
...
+    db_man.insert(result_table, results)
+    return results
+
+
+def store_analysis_to_db(db_man: DatabaseManager, main_record: Record) -> AnalysisTable:
+    """Store the analysis to the database."""
+    db_man.create_tables()
+
+    analysis = AnalysisTable(
+        analysis_time=datetime.now().isoformat(sep="T", timespec="seconds"),
```
behnazh-w, line 95: Two issues with using `now` (in general):

1. It gives the local time of host machine and is not normalized to UTC.
2. It doesn't have a timezone attached.

```suggestion
        analysis_time=datetime.utcnow(tz=datetime.timezone.utc).isoformat(sep="T", timespec="seconds"),
```

src/macaron/slsa_analyzer/table_definitions.py
```diff
@@ -0,0 +1,136 @@
...
+class AnalysisTable(ORMBase):
+    """
+    ORM Class for the analysis information.
+
+    This information pertains to a single invocation of the macaron tool.
+    """
+
+    __tablename__ = "_analysis"
+    id = Column(Integer, primary_key=True, autoincrement=True)  # noqa: A003
+    analysis_time = Column(String, nullable=False)
```
behnazh-w, line 48: SQLAlchemy has a [DateTime](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.DateTime) type. Let's let SQLAlchemy handle the mapping for us.

src/macaron/slsa_analyzer/analyzer.py
ailrst: I am just trusting SQLAlchemy to be smart enough to only create what is needed, since all the tables are known by the same MetaData object it should be fine. I probably call create_tables more often than needed. Since the tables are defined declaratively, as long as all the modules containing ORM tables have been loaded/evaluated by the time Analyzer() is initialized they should all be created there. I don't _know_ this is the case so call it here (after all checks have run) as well to be sure. The exception is the  `analyze_result` table which is declared using the Core API rather than the ORM api: https://github.com/oracle-samples/macaron/blob/ee565267d152c2e1439bec8ed96c0730d4fc8f6d/src/macaron/slsa_analyzer/analyze_context.py#L198-L198, this will need to be created later. 

> only adds the new tables for the checks results

Figuring out what tables are 'new' would be complicated, and is probably something SQLAlchemy already models somehow internally. I don't really want to try and optimise that since I don't fully understand it or see a meaningful performance impact as is.

src/macaron/slsa_analyzer/table_definitions.py
ailrst: That's correct

src/macaron/policy_engine/policy.py
```diff
@@ -147,26 +146,227 @@ def parse(target: Any) -> bool:
...
+        The list of repositories, policy pairs failing
+    _passed:
+        The list of repository, policy pairs passing
+
+    """
+
+    text: str
+    sha: str
+    filename: str
+    _result: dict | None
```
tromai, line 171: I think putting the documentation here would be rendered in the Sphinx documentation for those private attributes. I think we should relocated it. How do you think?

pyproject.toml
```diff
@@ -25,8 +25,9 @@ dependencies = [
     "requests >=2.28.0,<3.0.0",
     "pydriller >=2.0,<3.0.0",
     "yamale >=4.0.3,<5.0.0",
+    "jinja2 >=3.1.2,<4.0.0",
     "packaging >=21.3,<24.0.0",
-    "jinja2 >=3.1.2,<4.0.0"
+    "SQLAlchemy==1.4.46"
```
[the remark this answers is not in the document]
ailrst: It looks like this will at least require rewriting all the orm-mapped classes to use `Mapped` and `mapped_column` instead of `Column`, the migration pathway is documented here https://docs.sqlalchemy.org/en/14/changelog/migration_20.html. We don't have too many tables so it should hopefully be relatively straightforward though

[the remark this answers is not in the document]
behnazh-w: Let's go for it then :+1:

src/macaron/slsa_analyzer/analyzer.py
behnazh-w: > Figuring out what tables are 'new' would be complicated, and is probably something SQLAlchemy already models somehow internally. I don't really want to try and optimise that since I don't fully understand it or see a meaningful performance impact as is.

We discussed about it offline. We can probably create all the tables early at load time.

src/macaron/slsa_analyzer/database_store.py
ailrst: It looks like using `datetime.now(tz=timezone.utc)` is the recommended way to represent UTC while retaining the timezone information,https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

src/macaron/slsa_analyzer/table_definitions.py
ailrst: We should use the DateTime type yes but I want to explicitly specify that sqlalchemy converts to [RFC3339](https://tools.ietf.org/html/rfc3339) to match SLSA/in-toto, since the timestamp won't just be consumed by python/sqlalchemy but the policy engine.

src/macaron/slsa_analyzer/database_store.py
behnazh-w: Makes sense, thanks.

src/macaron/slsa_analyzer/table_definitions.py
behnazh-w: And SQLite doesn't have DateTime types anyway. So, it's fine to keep it as `String` but can you please add a comment to explain why this decision has been made?

src/macaron/database/database_manager.py
```diff
@@ -1,112 +1,136 @@
...
+    """
+    This class handles and manages the connection to sqlite database during the session.
+
+    Note that since SQLAlchemy lazy-loads the fields of mapped ORM objects, if the database connection is closed any
+    orm-mapped objects will become invalid. As such the lifetime of the database manager must be longer than any of the
+    objects added to the database (using add() or add_and_commit()).
+    """
 
-    def __init__(self, db_path: str):
+    def __init__(self, db_path: str, base=ORMBase):  # type: ignore
```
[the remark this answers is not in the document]
ailrst: I did this to allow multiple bases to exist in the same program, in case we are using multiple databases. It is also so that in the tests we can more carefully control what is added to the base and we don't automatically register the ORM tables if we want to import and test some other functionality.

src/macaron/database/database_manager.py
```diff
@@ -1,112 +1,136 @@
...
             The path to the target database.
         """
-        self.db_path = db_path
-        self.is_init = False
-        self.db_con = None
-        self.db_cursor = None
+        self.engine = create_engine(f"sqlite+pysqlite:///{db_path}", echo=False, future=True)
+        self.db_name = db_path
+        self.session = Session(self.engine)
+        self._base = base
```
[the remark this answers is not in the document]
ailrst: I'd prefer to call it from analyzer or registry when we know all the tables have been registered, that fact isn't really related to constructing this object.

src/macaron/slsa_analyzer/table_definitions.py
ailrst: Implemented here https://github.com/oracle-samples/macaron/pull/46/commits/4e61112595339f7871796e707f604cc1ff1b5711

src/macaron/slsa_analyzer/database_store.py
ailrst: Fixed here https://github.com/oracle-samples/macaron/pull/46/commits/4e61112595339f7871796e707f604cc1ff1b5711

src/macaron/slsa_analyzer/table_definitions.py
ailrst: Fixed here https://github.com/oracle-samples/macaron/pull/46/commits/4e61112595339f7871796e707f604cc1ff1b5711

src/macaron/policy_engine/souffle.py
ailrst: Fixed here https://github.com/oracle-samples/macaron/pull/46/commits/4e61112595339f7871796e707f604cc1ff1b5711

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -167,18 +171,46 @@ def run(self, user_config: dict, skip_deps: bool = False) -> int:
...
+        analysis = store_analysis_to_db(self.db_man, main_record)
+
+        for record in report.get_records():
+            if not record.status == SCMStatus.DUPLICATED_SCM:
+                if record.context:
+                    store_analyze_context_to_db(self.TABLE_NAME, self.db_man, analysis, record.context)
+
+        # Store dependency relations
+        for parent, child in report.get_dependencies():
+            dependency = RepositoryDependency(
```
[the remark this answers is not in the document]
ailrst: This is now a Core API table as of https://github.com/oracle-samples/macaron/pull/46/commits/4e61112595339f7871796e707f604cc1ff1b5711, but will wait to upgrade to SQLAlchemy 2 to add the list since the 1.4 API is a lot more convoluted for that `relationship`

src/macaron/policy_engine/policy.py
ailrst: Fixed here https://github.com/oracle-samples/macaron/pull/46/commits/4e61112595339f7871796e707f604cc1ff1b5711

src/macaron/output_reporter/results.py
ailrst: Fixed https://github.com/oracle-samples/macaron/pull/46/commits/4e61112595339f7871796e707f604cc1ff1b5711

src/macaron/output_reporter/results.py
ailrst: Removed https://github.com/oracle-samples/macaron/pull/46/commits/4e61112595339f7871796e707f604cc1ff1b5711

src/macaron/config/global_config.py
ailrst: Fixed https://github.com/oracle-samples/macaron/pull/46/commits/4e61112595339f7871796e707f604cc1ff1b5711

src/macaron/config/global_config.py
```diff
@@ -60,9 +64,18 @@ def load(
         self.debug_level = debug_level
         self.local_repos_path = local_repos_path
         self.gh_token = gh_token
-        self.policy_path = policy_path
         self.resources_path = resources_path
 
+        # Find the policies
+        policy_files = []
+        for policy_path in policy_paths:
+            for policy_file_path in glob.iglob(f"{policy_path}/**", recursive=False):
```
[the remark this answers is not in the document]
tromai: For number 1 I don't think it is True as per the documentation [here](https://docs.python.org/3/library/glob.html#glob.glob), `iglob` is an Iterator version of `glob.glob` so I don't think the behavior is different.

I tried the following snippet in my Python 3.11 virtual environment:
```
import glob

list(glob.iglob("./test_dir/**", recursive=False)) # will only print out the first level of test_dir (including files and directories)
```

src/macaron/database/table_definitions.py
```diff
@@ -0,0 +1,139 @@
...
+For tables associated with checks see base_check.py.
+"""
+
+from sqlalchemy import Boolean, Column, ForeignKey, Integer, String
+from sqlalchemy.orm import declarative_mixin, declared_attr
+
+from macaron.database.database_manager import ORMBase
+from macaron.database.rfc3339_datetime import RFC3339DateTime
+
+# TODO: Use GUIDs as primary keys rather than incremental
```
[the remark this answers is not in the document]
ailrst: No reason, I use the terms interchangeably

src/macaron/config/global_config.py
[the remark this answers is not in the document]
ailrst: 1. It recurses 1 level which is the desired behaviour; if a folder is specified then it yeilds the folder and all the files in the folder, and if a file is specified it just yeilds that file.
2. Its not necessarily a YAML file it could also be a cue file. The filetype is checked when the policy is constructed so no need to do it twice.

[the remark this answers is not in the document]
behnazh-w: > 1. if a folder is specified then it yeilds the folder and all the files in the folder, and if a file is specified it just yeilds that file.

 The `f"{policy_path}/**"` pattern won't match a single file though, e.g., if `policy_path` is `policy.cue`.

src/macaron/policy_engine/policy_registry.py
```diff
@@ -0,0 +1,125 @@
...
+    def evaluate_souffle_policies(self, database_path: str, restrict_to_analysis: int | None = None) -> list[Any]:
+        """Evaluate all known souffle policies.
+
+        Parameters
+        ----------
+        database_path: str
+            The path to the database file to evaluate the policy against
+        restrict_to_analysis: int | None
+            Optional, if is not None then restrict policy evaluation to the repositories associated with the analysis id
+            restrict_to_analysis.
```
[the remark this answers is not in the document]
ailrst: Thanks, fixed in https://github.com/oracle-samples/macaron/pull/46/commits/766a2bf3d6b1d52cd639d68e59d4013e835c655f

src/macaron/slsa_analyzer/analyze_context.py
```diff
@@ -188,72 +195,69 @@ def bulk_update_req_status(self, req_list: list, status: bool, feedback: str) ->
...
 
-        Returns
-        -------
-        str
-            The generated query.
-        """
-        result = (
-            f"INSERT OR REPLACE INTO {table_name} VALUES "
-            + "(:full_name,:branch_name,:commit_sha,:commit_date,:slsa_level,:is_full_reach,"
+    def get_slsa_level_table(self) -> dict:
```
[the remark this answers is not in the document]
ailrst: Thanks, fixed in https://github.com/oracle-samples/macaron/pull/46/commits/766a2bf3d6b1d52cd639d68e59d4013e835c655f

src/macaron/policy_engine/policy_registry.py
```diff
@@ -0,0 +1,125 @@
...
+class PolicyRegistry:
+    """
+    The policy registry class stores policies and their results.
+
+    Parameters
+    ----------
+    macaron_path: str
+        The path to the macaron module
+    policy_paths: list[str]
+        The list of policy file paths. all((os.isfile(path) for path in policy_paths)) must be true.
```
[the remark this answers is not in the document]
ailrst: Thanks, fixed in https://github.com/oracle-samples/macaron/pull/46/commits/766a2bf3d6b1d52cd639d68e59d4013e835c655f

src/macaron/config/global_config.py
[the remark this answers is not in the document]
ailrst: Ah of course, thanks. I it back to just explicitly checking if its a folder or file in this commit  https://github.com/oracle-samples/macaron/pull/46/commits/766a2bf3d6b1d52cd639d68e59d4013e835c655f

src/macaron/slsa_analyzer/analyzer.py
ailrst: In https://github.com/oracle-samples/macaron/pull/46/commits/766a2bf3d6b1d52cd639d68e59d4013e835c655f all the other calls to create_tables are removed except for the one [here](https://github.com/oracle-samples/macaron/blob/766a2bf3d6b1d52cd639d68e59d4013e835c655f/src/macaron/slsa_analyzer/analyzer.py#L97-L97) at the end of `Analyzer`'s constructor. This is after `registry.prepare()` so all all the tables _must_ have been loaded by this point.

pyproject.toml
```diff
@@ -25,8 +25,9 @@ dependencies = [
     "requests >=2.28.0,<3.0.0",
     "pydriller >=2.0,<3.0.0",
     "yamale >=4.0.3,<5.0.0",
+    "jinja2 >=3.1.2,<4.0.0",
     "packaging >=21.3,<24.0.0",
-    "jinja2 >=3.1.2,<4.0.0"
+    "SQLAlchemy>=2.0.0",
```
behnazh-w, line 30: To be consistent:

```suggestion
    "SQLAlchemy >=2.0.3,<3.0.0",
```

src/macaron/config/global_config.py
```diff
@@ -60,9 +63,22 @@ def load(
         self.debug_level = debug_level
         self.local_repos_path = local_repos_path
         self.gh_token = gh_token
-        self.policy_path = policy_path
         self.resources_path = resources_path
 
+        # Find the policies
```
behnazh-w, line 68: Can you please make sure all the comment sentences end with `.`?

src/macaron/policy_engine/__main__.py
```diff
@@ -0,0 +1,187 @@
...
+        output.append(str(key))
+        for value in values:
+            output.append(f"    {value}")
+
+    logger.info("Policy results:\n%s", "\n".join(output))
+
+    return any(res["failed_policies"])
+
+
+def main() -> Never:
```
behnazh-w, line 147: Can you please add this as an entrypoint script in [`pyproject.toml`](https://github.com/oracle-samples/macaron/pull/46/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R47)?

src/macaron/policy_engine/__main__.py
```diff
@@ -0,0 +1,187 @@
...
+    """Policy engine configuration."""
+
+    database_path: str
+    interactive: bool = False
+    policy_id: int | None = None
+    policy_file: str | None = None
+    show_prelude: bool = False
+
+
+global_config = Config()
```
behnazh-w, line 40: Is this configuration used anywhere?
Also, I would use `policy_config` to avoid mixing it up with Macaron's `global_config`.

src/macaron/policy_engine/__main__.py
```diff
@@ -0,0 +1,187 @@
...
+    def __enter__(self) -> "Timer":
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:  # type: ignore
+        self.stop = time.perf_counter()
+        self.delta = self.stop - self.start
+        logger.debug("%s delta: %s", self.name, f"{self.delta:0.4f}")
+
+
+def get_generated(database_path: os.PathLike | str) -> SouffleProgram:
```
behnazh-w, line 61: The docstring misses parameters and return type.

src/macaron/policy_engine/souffle.py
```diff
@@ -0,0 +1,187 @@
...
+    # The directory to store outputted facts in
+    output_dir: str
+    # The directory for Souffle to search for datalog files in when using #include
+    include_dir: str
+    # The directory for Souffle to load facts from
+    fact_dir: str
+    # The directory to link Souffle functor shared libraries from
+    library_dir: str
+    souffle_stdout: Optional[str]
+    souffle_stderr: Optional[str]
```
behnazh-w, line 59: Why are these class variables defined here? Aren't you reassigning them later in the constructor as instance variables?

src/macaron/policy_engine/souffle_code_generator.py
```diff
@@ -21,7 +21,7 @@ class SouffleProgram:
         Set of declaration souffle statements (begin with .decl token)
     directives: None | set[str]
         Set of directives (begin with ".", e.g. .input)
-    rules: None | set[str]
+    rules: None | set[str]-
```
tromai, line 24: I wonder if this is a typo?

src/macaron/policy_engine/souffle.py
```diff
@@ -0,0 +1,187 @@
...
+    TEMP_SOURCEFILE_NAME = "source.dl"
+
+    def __init__(
+        self,
+        souffle_full_path: str = "souffle",
+        output_dir: Optional[str] = None,
+        include_dir: Optional[str] = None,
+        fact_dir: Optional[str] = None,
+        library_dir: str = os.curdir,
+    ):
```
behnazh-w, line 71: The docstring is missing here and I'm surprised that the checks are passing!!! :thinking:

src/macaron/policy_engine/souffle.py
```diff
@@ -0,0 +1,187 @@
...
+        self.souffle_stdout = None
+        self.souffle_stderr = None
+        self.temp_dir = tempfile.mkdtemp()
```
behnazh-w, line 74: How about using 
```
with tempfile.TemporaryDirectory()
```
context manager to create the temp dir?

Also good to catch `OSError`s.

src/macaron/policy_engine/souffle.py
```diff
@@ -0,0 +1,187 @@
...
+    include_dir: str
+    # The directory for Souffle to load facts from
+    fact_dir: str
+    # The directory to link Souffle functor shared libraries from
+    library_dir: str
+    souffle_stdout: Optional[str]
+    souffle_stderr: Optional[str]
+
+    # The temporary file to store the datalog program in when executing from the temporary directory
+    TEMP_SOURCEFILE_NAME = "source.dl"
```
behnazh-w, line 62: We should move these to [`defaults.ini`](https://github.com/oracle-samples/macaron/blob/database-policy/src/macaron/config/defaults.ini) and avoid hardcoding them.

src/macaron/policy_engine/souffle.py
ailrst: The souffle class this is defined in is expected to be used as a context manager, since it manages the contents of the temporary directory, the directory needs to live as long as the class. Is it possible to have a context manager persist across method-calls?

src/macaron/policy_engine/__main__.py
ailrst: Its not used, will delete it.

src/macaron/policy_engine/souffle.py
ailrst: Sorry, Java/C++ habit :sweat_smile:

src/macaron/policy_engine/souffle.py
ailrst: https://github.com/oracle-samples/macaron/pull/46/commits/db11e761070bbd8c28841a19da111817ef1432ea Am using a temporary file now to avoid having a default name at all.

src/macaron/policy_engine/souffle.py
ailrst: Fixed https://github.com/oracle-samples/macaron/pull/46/commits/db11e761070bbd8c28841a19da111817ef1432ea

src/macaron/policy_engine/souffle.py
ailrst: Fixed https://github.com/oracle-samples/macaron/pull/46/commits/db11e761070bbd8c28841a19da111817ef1432ea

src/macaron/policy_engine/__main__.py
ailrst: Fixed https://github.com/oracle-samples/macaron/pull/46/commits/db11e761070bbd8c28841a19da111817ef1432ea

src/macaron/policy_engine/__main__.py
ailrst: Fixed https://github.com/oracle-samples/macaron/pull/46/commits/db11e761070bbd8c28841a19da111817ef1432ea

src/macaron/policy_engine/__main__.py
```diff
@@ -0,0 +1,187 @@
+# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""
+Stand-alone policy engine.
```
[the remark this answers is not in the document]
ailrst: Fixed https://github.com/oracle-samples/macaron/pull/46/commits/db11e761070bbd8c28841a19da111817ef1432ea

src/macaron/policy_engine/souffle_code_generator.py
ailrst: Fixed https://github.com/oracle-samples/macaron/pull/46/commits/db11e761070bbd8c28841a19da111817ef1432ea

src/macaron/policy_engine/readme.md
```diff
@@ -0,0 +1,91 @@
...
+4. Some simple pre-written Policies
+
+These can be found in the folder [prelude/](prelude). The automatically generated rules can be found by running
+`macaron.policy_engine -d database.db -s`.
+
+A policy consists of two statements, a definition clause and an enforcement clause. For example, to write a policy that
+requires that the repository has verified authenticated provenance we can write:
+
+```c
+Policy("auth-provenance", repositoryid, "") :-    check_passed(repositoryid, "mcn_provenance_level_three_1").
```
tromai:
```suggestion
Policy("auth-provenance", repositoryid, "") :- check_passed(repositoryid, "mcn_provenance_level_three_1").
```

src/macaron/policy_engine/readme.md
```diff
@@ -0,0 +1,91 @@
...
+    ['3', 'slsa-framework/slsa-verifier', 'auth-provenance']
+    ['4', 'slsa-framework/slsa-verifier', 'auth-provenance']
+passed_policies
+    ['auth-provenance']
+failed_policies
+```
+
+For the implementation of this see [prelude/policy.dl](prelude/policy.dl).
+
+When  any row in the `failed_policies` relation means policy failure, and the policy engine will exit nonzero.
```
tromai:
```suggestions
When any row in the `failed_policies` relation means policy failure, and the policy engine will exit nonzero.
```

src/macaron/policy_engine/examples/simple_example.dl
```diff
@@ -0,0 +1,8 @@
+/* Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved. */
+/* Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/. */
+
+
+#include "prelude.dl"
+
+Policy("auth-provenance", repositoryid, "") :-    check_passed(repositoryid, "mcn_provenance_level_three_1").
```
tromai:
```suggestion
Policy("auth-provenance", repositoryid, "") :- check_passed(repositoryid, "mcn_provenance_level_three_1").
```

src/macaron/policy_engine/examples/aggregate.dl
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * This is an example of applying the aggregate level policy.
+ */
+
+apply_policy_to("aggregate_level_3", repo) :-
+    // if we have provenance, then require using a trusted builder,
+    // verifying the provenance attestations, and
+    // dependencies must use some kind of scripted build
+    repository(repo, name,_,_,_,_,_),
```
tromai, line 15:
```suggestion
    repository(repo, name, _, _, _, _, _),
```

src/macaron/policy_engine/__main__.py
ailrst: Done in https://github.com/oracle-samples/macaron/pull/46/commits/f73924020c8f8c3b629f31dcc4700d311450bd02

src/macaron/policy_engine/policy_engine.py
```diff
@@ -0,0 +1,112 @@
...
+from sqlalchemy import MetaData, create_engine
+
+from macaron.policy_engine.souffle import SouffleError, SouffleWrapper
+from macaron.policy_engine.souffle_code_generator import (
+    SouffleProgram,
+    get_souffle_import_prelude,
+    project_table_to_key,
+    project_with_fk_join,
+)
+from macaron.util import logger
```
behnazh-w, line 18: We should let the `logging` module detect the appropriate logger object for the module.

```suggestion
logger: logging.Logger = logging.getLogger(__name__)
```

ailrst: Addressed in https://github.com/oracle-samples/macaron/pull/46/commits/73c1c81b6bdb6c999112532dab89103685eca48b

src/macaron/policy_engine/examples/aggregate.dl
ailrst: Fixed in https://github.com/oracle-samples/macaron/pull/46/commits/73c1c81b6bdb6c999112532dab89103685eca48b

src/macaron/policy_engine/examples/simple_example.dl
ailrst: Fixed in https://github.com/oracle-samples/macaron/pull/46/commits/73c1c81b6bdb6c999112532dab89103685eca48b

src/macaron/policy_engine/readme.md
ailrst: Fixed in https://github.com/oracle-samples/macaron/pull/46/commits/73c1c81b6bdb6c999112532dab89103685eca48b

src/macaron/policy_engine/readme.md
ailrst: Fixed in https://github.com/oracle-samples/macaron/pull/46/commits/73c1c81b6bdb6c999112532dab89103685eca48b

tests/database/test_database_manager.py
```diff
@@ -9,67 +9,81 @@
...
 
+    TEST_VALUE = "Hello World"
+    TEST_IDENT = 10
+
     def setUp(self) -> None:
-        db_path = str(Path(__file__).parent.joinpath("macaron.db"))
-        self.db_man = DatabaseManager(db_path)
+        """Set up the database and ensure it is empty."""
+        self.db_path = str(Path(__file__).parent.joinpath("macaron.db"))
+        print(self.db_path)
```
behnazh-w, line 43: One more `print` left here :sweat_smile:

tromai (approved):

behnazh-w (approved):

PR #57, opened 2023-02-08
feat: add dependency analyzer for Gradle

behnazh-w:
This PR refactors the build tools and dependency analyzer and adds a feature to automatically detect Gradle dependencies and map them to their SCM repositories.

As part of the refactoring, this PR also fixes related issues to dependency analyzer and adds back two skipping dependencies.

For now the integration test for Gradle dependency analysis is disabled due to a non-deterministic behavior of CycloneDX Gradle plugin. This issue will be addressed separately in https://github.com/oracle-samples/macaron/issues/60.


Makefile
```diff
@@ -105,6 +105,15 @@ $(PACKAGE_PATH)/resources/mvnw:
 		&& rm -r maven-wrapper-distribution-3.1.1-bin.zip \
 		&& echo -e "distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.8.6/apache-maven-3.8.6-bin.zip\nwrapperUrl=https://repo.maven.apache.org/maven2/org/apache/maven/wrapper/maven-wrapper/3.1.1/maven-wrapper-3.1.1.jar" > .mvn/wrapper/maven-wrapper.properties \
 		&& cd $(REPO_PATH)
+$(PACKAGE_PATH)/resources/gradlew:
+	cd $(PACKAGE_PATH)/resources \
+	&& export GRADLE_VERSION=7.6 \
```
tromai, line 110: The indentation of this block is not consistent compare to `$(PACKAGE_PATH)/resources/mvnw`.

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -0,0 +1,206 @@
...
+    # TODO: use the official CycloneDX library to validate and deserialize BOM files
+    # once merged: https://github.com/CycloneDX/cyclonedx-python-lib/pull/290
+    if not os.path.exists(file_path):
+        raise CycloneDXParserError(f"Unable to locate the BOM file: {str(file_path)}.")
+
+    with open(file_path, encoding="utf8") as file:
+        try:
+            return dict(json.load(file))
+        except (ValueError, json.JSONDecodeError) as error:
+            raise CycloneDXParserError(f"Could not process the dependencies at {file_path}: {error}") from None
```
tromai, line 53: I wonder if it is better to raise the error as follow instead of appending the error message.
```suggestion
            raise CycloneDXParserError(f"Could not process the dependencies at {file_path}.") from error
```
This syntax is according to [PEP-3134](https://legacy.python.org/dev/peps/pep-3134/).

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -0,0 +1,206 @@
...
+    modules = set()  # Stores all module dependencies.
+    for child_path in child_bom_paths or []:
+        try:
+            bom_objects.append(deserialize_bom_json(child_path))
+        except CycloneDXParserError as error:
+            logger.error(error)
+            continue
+    for bom in bom_objects:
+        try:
+            bom_ref = bom.get("metadata").get("component").get("bom-ref")  # type: ignore
```
tromai, line 120: From my understanding, this part is for handling the child `bom.json` (where `bom-ref` exists), isn't it? This also means that on line 123, we are not getting the direct dependencies of the **main target** but the recursive dependencies instead. 
Would it be okay to have a small comment at the beginning of this part as well?

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -0,0 +1,206 @@
...
+        bom_objects.append(root_bom)
+    except (CycloneDXParserError, KeyError) as error:
+        logger.error(error)
+        return
+
+    if components is None:
+        logger.error("The BOM file at %s misses components.", str(root_bom_path))
+        return
+
+    dependencies = []
```
tromai, line 110: From what I see, each element in `dependencies` and `modules` are string (which are all of [`bom-ref`](https://cyclonedx.org/docs/1.4/json/#components_items_bom-ref) type). Would it be a good idea to put an annotation as `list[str]` here?

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -0,0 +1,206 @@
...
+    modules = set()  # Stores all module dependencies.
```
tromai, line 111: I am a bit confused about `modules dependencies`. Which type of dependencies does this variable store? From my understanding, the `modules` variable stores all available `bom-ref` of every components mentioned in `bom_objects`.

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -0,0 +1,206 @@
...
+            if bom_ref:
+                modules.add(bom_ref)
+            for node in bom.get("dependencies"):  # type: ignore
+                if node.get("ref") == bom_ref or recursive:
+                    dependencies.extend(node.get("dependsOn"))
+        except KeyError as error:
+            logger.debug(error)
+
+    for dependency in dependencies:
+        if dependency in modules:
```
tromai, line 130: I don't quite understand why do we need to ignore the dependencies that already existed in `modules`.

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -0,0 +1,206 @@
...
+
+    with open(file_path, encoding="utf8") as file:
+        try:
+            return dict(json.load(file))
+        except (ValueError, json.JSONDecodeError) as error:
+            raise CycloneDXParserError(f"Could not process the dependencies at {file_path}: {error}") from None
+
+
+def get_root_component(root_bom_path: Path) -> Optional[dict | None]:
+    """Get dependency components.
```
tromai, line 57: This code documentation is incorrect.

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -0,0 +1,206 @@
...
+    recursive: bool
+        Set to False to get the direct dependencies only (default).
+
+    Yields
+    ------
+    dict
+        The dependencies as CycloneDX components.
+    """
+    bom_objects: list[dict] = []
+    try:
```
tromai, line 98: Could we reuse the method `get_root_component` defined above to get the root component from `root_bom_path`?

src/macaron/dependency_analyzer/dependency_resolver.py
```diff
@@ -46,20 +52,19 @@ def __init__(self, resources_path: str, file_name: str, debug_path: str, tool_ve
...
         file_name : str
             The name of dependency output file.
-        debug_path : str
-            The file path where all the dependencies will be stored for debugging.
+        tool_name: str
+            The name of the dependency analyzer.
         tool_version : str
             The version of the dependency analyzer.
+        repo_path: str
+            The path to the target repo.
```
tromai, line 60: The documentation for `visited_deps` is missing.

src/macaron/slsa_analyzer/build_tool/base_build_tool.py
```diff
@@ -107,12 +111,57 @@ def prepare_config_files(self, wrapper_path: str, build_dir: str) -> bool:
...
+            The relative paths from the repo path that contain build scripts.
+        """
+        config_paths: set[str] = set()
+        for build_cfg in self.build_configs:
+            config_paths.update(glob.glob(os.path.join(repo_path, "**", build_cfg), recursive=True))
+
+        list_iter = iter(sorted(config_paths, key=lambda x: (str(Path(x).parent), len(Path(x).parts))))
+        try:
+            prefix = next(list_iter)
+            yield Path(prefix).parent.relative_to(repo_path)
```
tromai, line 156: If the elements in `list_iter` are already the path to the parent dir of build config files discovered in a repository, why do we need to find the parent here again?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -213,94 +220,74 @@ def resolve_dependencies(self, main_ctx: AnalyzeContext) -> dict[str, Dependency
...
+        )
+
+        # Clean up existing SBOM files.
+        dep_analyzer.remove_sboms(main_ctx.repo_path)
+
+        commands = dep_analyzer.get_cmd()
+        working_dirs: Iterable[Path] = build_tool.get_build_dirs(main_ctx.repo_path)
+        for working_dir in working_dirs:
+            # Get the absolute path to use as the working dir in the subprocess.
+            working_dir = Path(main_ctx.repo_path).joinpath(working_dir)
```
tromai, line 263: Would it be possible if we return the absolute `workfing_dir` directly from `build_tool.get_build_dirs` so that we don't need to recompute the absolute path here (unless the result of `build_tool.get_build_dirs` is also used elsewhere),

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -213,94 +220,74 @@ def resolve_dependencies(self, main_ctx: AnalyzeContext) -> dict[str, Dependency
...
+                    log_file.write(analyzer_output.stdout.decode("utf-8"))
 
             except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as error:
                 logger.error(error)
-                with open(log_path, mode="w", encoding="utf-8") as log_file:
-                    logger.info("Storing dependency resolver log to %s", log_path)
+                with open(log_path, mode="a", encoding="utf-8") as log_file:
                     log_file.write(error.output.decode("utf-8"))
             except FileNotFoundError as error:
                 # Only happen if the gradlew at the repo dir has an invalid format
```
tromai, line 282: We are now using the gradlew and mvnw directly from our `resources` directory (previously, we copied the wrapper files to the repository directory to run it) so I think this comment could be removed.

src/macaron/dependency_analyzer/dependency_resolver.py
```diff
@@ -46,20 +52,19 @@ def __init__(self, resources_path: str, file_name: str, debug_path: str, tool_ve
             The path to the resources directory.
         file_name : str
             The name of dependency output file.
-        debug_path : str
-            The file path where all the dependencies will be stored for debugging.
+        tool_name: str
+            The name of the dependency analyzer.
         tool_version : str
             The version of the dependency analyzer.
+        repo_path: str
```
tromai, line 59: This attribute is only needed for the `CycloneDxMaven` class for now. Do you think it is a good idea to move it there? :thinking:

tests/dependency_analyzer/cyclonedx/test_cyclonedx.py
```diff
@@ -0,0 +1,102 @@
...
+    """Test the CyclondeDX helper functions."""
+
+    RESOURCES_DIR = Path(__file__).parent.joinpath("resources")
+    EXPECTED_DIR = Path(__file__).parent.joinpath("expected_results")
+
+    def test_deserialize_bom_json(self) -> None:
+        """Test deserializing a bom.json file."""
+        # Deserialize a valid bom.json.
+        path = Path(TestCyclondeDX.RESOURCES_DIR, "valid_bom.json")
+        deserialize_bom_json(path)
```
tromai, line 29: I think this result is not asserted or verified anywhere in this test case.

tests/dependency_analyzer/cyclonedx/test_cyclonedx.py
```diff
@@ -0,0 +1,102 @@
...
+        # Deserialize a bom.json that does not exist.
+        path = Path(TestCyclondeDX.RESOURCES_DIR, "does_not_exist")
+        self.assertRaises(CycloneDXParserError, deserialize_bom_json, path)
+
+        # Deserialize an invalid bom.json.
+        path = Path(TestCyclondeDX.RESOURCES_DIR, "invalid_bom.json")
+        self.assertRaises(CycloneDXParserError, deserialize_bom_json, path)
+
+    @classmethod
+    def test_get_dep_components(cls) -> None:
```
tromai, line 40: Why would we need to make this a class method instead of using `self` to access the `RESOURCES_DIR` attribute?
For example:
```python
    def test_get_dep_components(self) -> None:
        self.RESOURCES_DIR ...
```

src/macaron/dependency_analyzer/cyclonedx.py
behnazh-w: The reason I'm using `None` is to avoid adding the previous error stack traces as the cause. That gets really messy. See the docs [here](https://docs.python.org/3/library/exceptions.html#exception-context).

src/macaron/dependency_analyzer/cyclonedx.py
behnazh-w: For projects that have sub-projects there will be more than one BOM files: a root BOM and children BOM files. Because the sub-projects also appear in the `dependencies` section, we need to filter them out to avoid mixing them up with the third-party dependencies. By adding the `bom-ref` of each BOM file to the `modules`, we can filter those out later.

src/macaron/dependency_analyzer/cyclonedx.py
behnazh-w: See my answer [here](https://github.com/oracle-samples/macaron/pull/57#discussion_r1107885322).

src/macaron/dependency_analyzer/cyclonedx.py
behnazh-w: It can't be used directly because `get_root_component` returns `root_bom.get("metadata").get("component")` whereas here we need `root_bom.get("components")`.

src/macaron/dependency_analyzer/dependency_resolver.py
behnazh-w: I didn't add `visited_deps` because it's not passed to the constructor as a parameter.

src/macaron/slsa_analyzer/build_tool/base_build_tool.py
behnazh-w: The items in `list_iter` are the full path to the config files. I think `prefix` is a confusing variable name. I'll change it to `cfg_path`.

src/macaron/slsa_analyzer/analyzer.py
behnazh-w: The reason I'm using relative paths here is for testing. We now use `syrupy` to take a snapshot of the returned paths from this function, and capturing absolute paths would add paths on my own local machines, which would not work.

src/macaron/dependency_analyzer/dependency_resolver.py
behnazh-w: I think it's a general enough attribute that would be useful for other dependency resolvers down the road?

src/macaron/dependency_analyzer/dependency_resolver.py
tromai: Oops, I didn't look properly at the constructor's signature. All good.

tests/dependency_analyzer/cyclonedx/test_cyclonedx.py
behnazh-w: `self.x` creates a new object and overrides the class variable `RESOURCES_DIR`. It's a common mistake that can be avoided by using `cls.RESOURCES_DIR`.

Anyway, I am going to refactor this whole test and use `pytest` and `syrupy` to simplify.

behnazh-w:
@tromai addressed comments in 78c34e891acd6ede6c5bfad51a2405eef43e4283.

tromai:
> @tromai addressed comments in [b72bde3](https://github.com/oracle-samples/macaron/commit/b72bde3f6e0598d049c2b165e7491928708da2d1)

LGTM

tromai (approved):

PR #67, opened 2023-02-15
feat: add python as a supported build tool

sophie-bates:
Currently able to detect whether a Python project uses Pip or Poetry to manage its dependencies. This should be expanded in the future.

- For Pip projects, we search for the existence of a`pyproject.toml`, `setup.py`, or `setup.cfg` file. 
- For Poetry projects, we search for the existence of a `pyproject.toml` file, that we parse using [tomllib](https://docs.python.org/3/library/tomllib.html) to search for configuration settings for `[tools.poetry]`. 

For the Build Service Check, the build commands that we currently support are:

- `pip install`
- `poetry build`
- `flit build`
- `python setup.py`

For the Build as Code Check, the deploy commands that we currently support are:

- `poetry publish`
- `flit publish`
- `twine upload`

As well as the use of the [pypa/gh-action-pypi-publish](https://github.com/pypa/gh-action-pypi-publish) reusable workflow.


tests/e2e/expected_results/urllib3/urllib3.json
```diff
@@ -1,15 +1,15 @@
...
     },
     "target": {
         "info": {
             "full_name": "urllib3/urllib3",
             "local_cloned_path": "git_repos/github_com/urllib3/urllib3",
             "remote_path": "https://github.com/urllib3/urllib3",
             "branch": "main",
-            "commit_hash": "87a0ecee6e691fe5ff93cd000c0158deebef763b",
-            "commit_date": "2022-10-04T07:59:23-05:00"
+            "commit_hash": "8619650a81a959b716b2f38b2a714ac1831e9a98",
```
tromai, line 11: This commit hash seems to be the latest commit on the urllib3 repository. In the integration test, we pin the hash of the repository [here](https://github.com/oracle-samples/macaron/blob/07d07ff006ce1230ffe2cb50f7230840e7110a6f/scripts/dev_scripts/integration_tests.sh#L57) so that our results are deterministic. I think you could either update the commit hash of urllib3 in the [integration_test.sh](https://github.com/oracle-samples/macaron/blob/07d07ff006ce1230ffe2cb50f7230840e7110a6f/scripts/dev_scripts/integration_tests.sh#L57) or update this expected result with the analysis on the commit hash specified in [here](https://github.com/oracle-samples/macaron/blob/07d07ff006ce1230ffe2cb50f7230840e7110a6f/scripts/dev_scripts/integration_tests.sh#L57).

sophie-bates: Updated the expected result with the analysis on the original commit hash, in [7da1e4a](https://github.com/oracle-samples/macaron/pull/67/commits/7da1e4a7a35f4b7d97abe2309fad8fd26780db72)

src/macaron/slsa_analyzer/build_tool/pip.py
```diff
@@ -0,0 +1,82 @@
+# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""This module contains the Pip class which inherits BaseBuildTool.
+
+This module is used to work with repositories that use Poetry for dependency management.
```
behnazh-w, line 6:
```suggestion
This module is used to work with repositories that use pip for dependency management.
```

src/macaron/slsa_analyzer/build_tool/base_build_tool.py
```diff
@@ -15,6 +15,29 @@
...
+    file_name : str
+        The name of the file to search.
+
+    Returns
+    -------
+    list:
+        The result list of strings or empty if the file doesn't exist anywhere lower than the path.
+    """
+    pattern = os.path.join(path, "**", file_name)
+    files_detected = glob.glob(pattern, recursive=True)
```
behnazh-w, line 36: I'm not sure if adding this wrapper function over `glob.glob` makes too much sense. All it's doing is calling `glob.glob()` really and I think this could just be done directly at `is_detected` function.

src/macaron/slsa_analyzer/build_tool/poetry.py
```diff
@@ -0,0 +1,103 @@
...
+                return True
+
+        for conf in self.entry_conf:
+            # Find the paths of all pyproject.toml files.
+            files_detected = find_files(repo_path, conf)
+
+            if files_detected:
+                try:
+                    # Take the highest level file (shortest file path)
+                    file_path = min(files_detected, key=len)
```
behnazh-w, line 57: Note that the `len` of the first level path part could be larger than a path with more parts, e.g., `a/b/c.txt` vs `aaaaaaaaaaaa/c.txt`. You would need to get the number of parts here probably:
```
len(Path(x).parts)
```

src/macaron/slsa_analyzer/build_tool/poetry.py
```diff
@@ -0,0 +1,103 @@
...
+
+                    # Parse the .toml file
+                    with open(file_path, "rb") as toml_file:
```
behnazh-w, line 60: What if the file is not a `.toml` file?

src/macaron/slsa_analyzer/build_tool/poetry.py
```diff
@@ -0,0 +1,103 @@
...
+                        data = tomllib.load(toml_file)
+                        if ("tool" in data) and ("poetry" in data["tool"]):
+                            return True
+                    return False
```
behnazh-w, line 64: What if there are more `pyproject.toml` files? Shouldn't we analyze the other ones? Or should we return `False` if the one with the shortest path doesn't have `poetry"` in its `data["tool"]`?

src/macaron/slsa_analyzer/build_tool/poetry.py
```diff
@@ -0,0 +1,103 @@
...
+                except FileNotFoundError:
+                    logger.error("Failed to read the %s file.", conf)
+                    return False
+
+        return False
+
+    def prepare_config_files(self, wrapper_path: str, build_dir: str) -> bool:
+        """Prepare the necessary wrapper files for running the build.
+
+        This method will return False if there is any errors happened during operation.
```
behnazh-w, line 74:
```suggestion
        This method returns False on errors.
```

src/macaron/slsa_analyzer/build_tool/poetry.py
```diff
@@ -0,0 +1,103 @@
...
+        ----------
+        wrapper_path : str
+            The path where all necessary wrapper files are located.
+        build_dir : str
+            The path of the build dir. This is where all files are copied to.
+
+        Returns
+        -------
+        bool
+            True if succeed else False.
```
behnazh-w, line 86:
```suggestion
            True if succeeds else False.
```

src/macaron/slsa_analyzer/build_tool/poetry.py
sophie-bates: Based on our discussion, my understanding was that we were looking for a single `pyproject.toml` file, and if there are multiple, take the one that is closest to the root, and only analyze that. This stemmed from our discussion about not supporting multiple build types yet. So I think we should return `False` if the one with the shortest path doesn't have a `tool.poetry` section in its `pyproject.toml` file.

src/macaron/slsa_analyzer/build_tool/poetry.py
sophie-bates: At the moment, we are only looking for a `pyproject.toml` file, as this is required by poetry projects. So for now, we'll only need to parse `.toml` files. But do you think I should add a check to make sure [ `conf` ](https://github.com/oracle-samples/macaron/pull/67/files/017c55d56101ec75ea720249556f79419f0de62f#diff-28f2413473de6808f14b6408518f95765b0cfa2b1b231cd55393142b25e05177R50) has a `.toml` extension before parsing?

tromai: I think what Behnaz means is that the content of the file might not be in TOML format. According to the documentation of [tomllib.load](https://docs.python.org/3/library/tomllib.html#tomllib.load), it will raises an Exception if that scenario happens. I

sophie-bates: Ohh gotcha, I'm on the same page now. Thanks

src/macaron/slsa_analyzer/build_tool/poetry.py
sophie-bates: Fixed in [f7bec7f](https://github.com/oracle-samples/macaron/pull/67/commits/f7bec7f96288f729de6b80f58e604a65549cea96)

src/macaron/slsa_analyzer/build_tool/poetry.py
sophie-bates: Fixed in [f7bec7f](https://github.com/oracle-samples/macaron/pull/67/commits/f7bec7f96288f729de6b80f58e604a65549cea96)

src/macaron/slsa_analyzer/build_tool/poetry.py
sophie-bates: Added additional `try .. catch` block around the `.toml` file parsing in [f7bec7f](https://github.com/oracle-samples/macaron/pull/67/commits/f7bec7f96288f729de6b80f58e604a65549cea96).

src/macaron/slsa_analyzer/build_tool/poetry.py
sophie-bates: Good point, I changed the `key` function to look for parts instead in [f7bec7f](https://github.com/oracle-samples/macaron/pull/67/commits/f7bec7f96288f729de6b80f58e604a65549cea96)

src/macaron/slsa_analyzer/build_tool/base_build_tool.py
sophie-bates: I agree. I've removed this function and put the functionality inside `is_detected` instead in [f7bec7f](https://github.com/oracle-samples/macaron/pull/67/commits/f7bec7f96288f729de6b80f58e604a65549cea96)

src/macaron/slsa_analyzer/build_tool/pip.py
sophie-bates: Fixed in [f7bec7f](https://github.com/oracle-samples/macaron/pull/67/commits/f7bec7f96288f729de6b80f58e604a65549cea96)

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -136,7 +218,7 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
                             os.path.basename(bash_cmd["CI_path"]),
                         )
 
-                        justification: list[str | dict[str, str]] = [
+                        justification_cmd: list[str | dict[str, str]] = [
                             {
                                 f"The target repository uses build tool {build_tool.name} to deploy": bash_source_link,
```
sophie-bates, line 223: For Python projects, rather than saying (i.e.) "...uses build tool pip to deploy", should this reflect the actual publishing tool used to be more descriptive? Something like: "...uses build tool pip, with Twine to deploy".

behnazh-w: > For Python projects, rather than saying (i.e.) "...uses build tool pip to deploy", should this reflect the actual publishing tool used, to be more descriptive? Something like: "...uses build tool pip, with Twine to deploy".

A more descriptive justification would be good, but the original message was logged with the assumption that the command would already contain the exact build tool and deploy command. Please feel free to adjust the message and push for review if you have a better message in mind.

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -75,13 +78,31 @@ def _has_deploy_command(self, commands: list[list[str]], build_tool: BaseBuildTo
...
+            check_build_commands = any(build_cmd for build_cmd in deploy_tool if build_cmd == cmd_program_name)
+
+            # Support the use of python modules, i.e. 'python -m pip install'
+            check_module_build_commands = any(
+                module == cmd_program_name and com[1] and com[1] == "-m" and com[2] in deploy_tool
+                for module in build_tool.builder_module
+            )
+            prog_name_index = 2 if check_module_build_commands else 0
+
+            # logger.info("com: %s", com[(prog_name_index + 1) :])
```
tromai, line 97: Would this logger line needed ? else we can remove it here.

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -75,13 +78,31 @@ def _has_deploy_command(self, commands: list[list[str]], build_tool: BaseBuildTo
             if not cmd_program_name:
                 logger.debug("Found invalid program name %s.", com[0])
                 continue
-            if any(build_cmd for build_cmd in build_tool.builder if build_cmd == cmd_program_name):
-                # Check the arguments in the bash command for the deploy goals.
-                # If there are no deploy args for this build tool, accept as deploy command.
+
+            # Account for Python projects having separate tools for packaging and publishing.
+            if build_tool.publisher:
```
tromai, line 83: I think this if block can be put before the for loop because we don't need to rerun it multiple times.

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -75,13 +78,31 @@ def _has_deploy_command(self, commands: list[list[str]], build_tool: BaseBuildTo
...
+                deploy_tool = build_tool.publisher
+            else:
+                deploy_tool = build_tool.builder
+
+            check_build_commands = any(build_cmd for build_cmd in deploy_tool if build_cmd == cmd_program_name)
+
+            # Support the use of python modules, i.e. 'python -m pip install'
+            check_module_build_commands = any(
+                module == cmd_program_name and com[1] and com[1] == "-m" and com[2] in deploy_tool
```
tromai, line 92: What would happen if `com[2]` does not exist in the first place?

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -75,13 +78,31 @@ def _has_deploy_command(self, commands: list[list[str]], build_tool: BaseBuildTo
...
+            prog_name_index = 2 if check_module_build_commands else 0
+
+            # logger.info("com: %s", com[(prog_name_index + 1) :])
+
+            if check_build_commands or check_module_build_commands:
                 if not build_tool.deploy_arg:
                     logger.info("No deploy arguments required. Accept %s as deploy command.", str(com))
                     return str(com)
-                for word in com[1:]:
+                logger.info(build_tool.deploy_arg)
```
tromai, line 103: Could you please include more details on what the value we print out to `INFO` is?.

src/macaron/slsa_analyzer/checks/build_as_code_check.py
sophie-bates: Yep, I meant to remove that.

src/macaron/slsa_analyzer/checks/build_as_code_check.py
sophie-bates: This was for my own debugging, and I meant to remove it

src/macaron/config/defaults.ini
```diff
@@ -211,6 +211,52 @@ jenkins =
...
+github_actions = pypa/gh-action-pypi-publish
+
+# This is the spec for trusted Poetry packaging tools.
+[builder.poetry]
+entry_conf = pyproject.toml
+build_configs = poetry.lock
+builder =
+    poetry
+    poetry-core
+builder_module =
```
behnazh-w, line 250:
```suggestion
interpreter = 
```

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -75,13 +84,22 @@ def _has_deploy_command(self, commands: list[list[str]], build_tool: BaseBuildTo
             if not cmd_program_name:
                 logger.debug("Found invalid program name %s.", com[0])
                 continue
-            if any(build_cmd for build_cmd in build_tool.builder if build_cmd == cmd_program_name):
-                # Check the arguments in the bash command for the deploy goals.
-                # If there are no deploy args for this build tool, accept as deploy command.
```
behnazh-w, line 80: Why are you removing these comments?

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -75,13 +84,22 @@ def _has_deploy_command(self, commands: list[list[str]], build_tool: BaseBuildTo
...
+
+            check_build_commands = any(build_cmd for build_cmd in deploy_tool if build_cmd == cmd_program_name)
+
+            # Support the use of python modules, i.e. 'python -m pip install'
+            check_module_build_commands = any(
+                module == cmd_program_name and com[1] and com[1] == "-m" and com[2] and com[2] in deploy_tool
```
behnazh-w, line 92: `"-m"`: We shouldn't hardcode the values here. Instead they should come from the spec `ini` file.

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -75,13 +84,22 @@ def _has_deploy_command(self, commands: list[list[str]], build_tool: BaseBuildTo
             if not cmd_program_name:
                 logger.debug("Found invalid program name %s.", com[0])
                 continue
-            if any(build_cmd for build_cmd in build_tool.builder if build_cmd == cmd_program_name):
-                # Check the arguments in the bash command for the deploy goals.
-                # If there are no deploy args for this build tool, accept as deploy command.
+
+            check_build_commands = any(build_cmd for build_cmd in deploy_tool if build_cmd == cmd_program_name)
+
+            # Support the use of python modules, i.e. 'python -m pip install'
```
behnazh-w, line 90:
```suggestion
            # Support the use of interpreters like Python that load modules, i.e., 'python -m pip install'
```

src/macaron/slsa_analyzer/checks/build_as_code_check.py
behnazh-w: `module` could change to `interpreter` because `Python` is not the module.

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -105,7 +123,7 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
         """
         # Get the build tool identified by the mcn_version_control_system_1, which we depend on.
         build_tool = ctx.dynamic_data["build_spec"].get("tool")
-        ci_services = ctx.dynamic_data["ci_services"]
+        ci_services = ctx.dynamic_data["ci_services"]  # keep storing things here
```
behnazh-w, line 126: Instead of `# keep storing things here` can you please add a more descriptive comment? Also always end a sentence with `.` :smile:

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -114,6 +132,67 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
                 # Checking if a CI service is discovered for this repo.
                 if isinstance(ci_service, NoneCIService):
                     continue
+
+                trusted_deploy_actions = defaults.get_list("builder.pip.ci.deploy", "github_actions", fallback=[])
+
+                # Check for use of a trusted Github Action to publish/deploy.
```
behnazh-w, line 138:
```suggestion
                # Check for use of a trusted Github Actions workflow to publish/deploy.
```

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -114,6 +132,67 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
...
+                            continue
+                        if workflow_name in trusted_deploy_actions:
+                            trigger_link = ci_service.api_client.get_file_link(
+                                ctx.repo_full_name,
+                                ctx.commit_sha,
+                                ci_service.api_client.get_relative_path_of_workflow(
+                                    os.path.basename(callee.caller_path)
+                                ),
+                            )
+                            # TODO: verify that caller_path and CI_PATH are the same
```
behnazh-w, line 158: Can you please elaborate what you mean here?

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -114,6 +132,67 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
...
+                            if ctx.dynamic_data["is_inferred_prov"] and ci_info["provenances"]:
+                                predicate = ci_info["provenances"][0]["predicate"]
+                                predicate["buildType"] = f"Custom {ci_service.name}"
+                                predicate["builder"]["id"] = deploy_action_source_link
+                                predicate["invocation"]["configSource"][
+                                    "uri"
+                                ] = f"{ctx.remote_path}@refs/heads/{ctx.branch_name}"
+                                predicate["invocation"]["configSource"]["digest"]["sha1"] = ctx.commit_sha
+                                predicate["invocation"]["configSource"]["entryPoint"] = trigger_link
+                                predicate["metadata"]["buildInvocationId"] = html_url
```
behnazh-w, line 193: You also need to add results to `check_result["result_tables"] ` to be picked up by the policy engine, e.g., 
```
check_result["result_tables"] = [
     BuildAsCodeTable(
         build_tool_name=build_tool.name,
         ci_service_name=ci_service.name,
         build_trigger=trigger_link,
         deploy_command=deploy_cmd,
         build_status_url=html_url,
    )
]
```

src/macaron/slsa_analyzer/checks/build_service_check.py
behnazh-w, line 1: Various comments for the `build_as_code_check.py` apply to `build_service_check.py`.

tests/slsa_analyzer/build_tool/__snapshots__/test_poetry.ambr
```diff
@@ -0,0 +1,14 @@
...
+  list([
+    PosixPath('.'),
+  ])
+# ---
+# name: test_get_build_dirs[mock_repo1]
+  list([
+  ])
+# ---
+# name: test_get_build_dirs[mock_repo2]
+  list([
```
behnazh-w, line 12: Shouldn't we detect a build here? Why is this list empty?

tests/slsa_analyzer/checks/test_build_as_code_check.py
```diff
@@ -116,6 +133,34 @@ def test_build_as_code_check(self) -> None:
         gradle_deploy.dynamic_data["ci_services"] = [ci_info]
         assert check.run_check(gradle_deploy, check_result) == CheckResultType.PASSED
 
+        # Use poetry publish to publish the artifact
```
behnazh-w, line 136:
```suggestion
       # Use poetry publish to publish the artifact.
```

Same for comments below.

tests/slsa_analyzer/checks/test_build_as_code_check.py
```diff
@@ -147,3 +192,28 @@ def test_build_as_code_check(self) -> None:
         bash_commands["commands"] = []
         maven_deploy.dynamic_data["ci_services"] = [ci_info]
         assert check.run_check(maven_deploy, check_result) == CheckResultType.FAILED
+
+        # This Github Actions workflow uses gh-action-pypi-publish to publish the artifact.
```
behnazh-w, line 196: Can you please add a new function to test GitHub Actions workflow deployment? This function is getting hard to read.

tests/slsa_analyzer/checks/test_build_service_check.py
```diff
@@ -116,6 +132,48 @@ def test_build_service_check(self) -> None:
...
+
+        # Use pip as a module incorrectly in CI to build the artifact.
+        pip_module_build_ci = AnalyzeContext("use_build_tool", os.path.abspath("./"), MagicMock())
+        pip_module_build_ci.dynamic_data["build_spec"]["tool"] = pip
+        bash_commands["commands"] = [["python", "pip", "install"]]
+        pip_module_build_ci.dynamic_data["ci_services"] = [ci_info]
+        assert check.run_check(pip_module_build_ci, check_result) == CheckResultType.FAILED
+
+        # Use pip as a module in CI with invalid goal to build the artifact.
+        pip_module_build_ci = AnalyzeContext("use_build_tool", os.path.abspath("./"), MagicMock())
```
behnazh-w, line 171: When you adjust the attribute names in the spec, please update the variable names here too.

tests/slsa_analyzer/checks/test_build_service_check.py
```diff
@@ -147,3 +205,5 @@ def test_build_service_check(self) -> None:
         ci_info["service"] = gitlab_ci
         maven_build_ci.dynamic_data["ci_services"] = [ci_info]
         assert check.run_check(maven_build_ci, check_result) == CheckResultType.FAILED
+
+        # TODO: Python module - maybe not for this context, just build
```
behnazh-w, line 209: Can you please elaborate?

behnazh-w:
@sophie-bates Please update the `README.md` and add Python support (minus dependency analysis for now).

src/macaron/slsa_analyzer/checks/build_as_code_check.py
sophie-bates: That was an accident :sweat_smile:.

src/macaron/slsa_analyzer/checks/build_as_code_check.py
sophie-bates: This was a comment that I meant to remove :sleepy:, my bad.

tests/slsa_analyzer/checks/test_build_as_code_check.py
sophie-bates: Completed in [9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d). Refactored this file to use functions (+ fixtures) rather than the class-based approach. I added several fixtures for the build tools and CI services into `conftest.py` so they can be shared across many tests.

tests/slsa_analyzer/checks/test_build_service_check.py
sophie-bates: This was meant to be removed, done in [9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d).

tests/slsa_analyzer/checks/test_build_service_check.py
sophie-bates: Done in [9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d).

tests/slsa_analyzer/build_tool/__snapshots__/test_poetry.ambr
sophie-bates: Made changes in `defaults.ini` as well as `Poetry.get_build_dirs()` to address this issue ([9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d)).

src/macaron/slsa_analyzer/build_tool/poetry.py
```diff
@@ -0,0 +1,118 @@
...
+            The path where all necessary wrapper files are located.
+        build_dir : str
+            The path of the build dir. This is where all files are copied to.
+
+        Returns
+        -------
+        bool
+            True if succeeds else False.
+        """
+        return False
```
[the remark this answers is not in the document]
sophie-bates: Fixed in [9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d).

src/macaron/slsa_analyzer/build_tool/pip.py
```diff
@@ -0,0 +1,87 @@
...
+        ----------
+        repo_path: str
+            The path to the target repo.
+
+        Returns
+        -------
+        DependencyAnalyzer
+            The DependencyAnalyzer object.
+        """
+        return NoneDependencyAnalyzer()
```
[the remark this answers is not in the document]
sophie-bates: Done in [9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d).

src/macaron/slsa_analyzer/build_tool/pip.py
```diff
@@ -0,0 +1,87 @@
...
+        wrapper_path : str
+            The path where all necessary wrapper files are located.
+        build_dir : str
+            The path of the build dir. This is where all files are copied to.
+
+        Returns
+        -------
+        bool
+            True if succeed else False.
+        """
```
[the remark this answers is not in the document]
sophie-bates: Done in [9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d).

src/macaron/config/defaults.ini
```diff
@@ -211,6 +211,52 @@ jenkins =
...
+packager =
+    pip
+    pip3
+    flit
+    conda
+publisher =
+    twine
+    flit
+    conda
+builder_module =
```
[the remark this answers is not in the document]
sophie-bates: Done in [9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d).

src/macaron/slsa_analyzer/build_tool/poetry.py
```diff
@@ -0,0 +1,118 @@
...
+        ----------
+        repo_path: str
+            The path to the target repo.
+
+        Returns
+        -------
+        DependencyAnalyzer
+            The DependencyAnalyzer object.
+        """
+        return NoneDependencyAnalyzer()
```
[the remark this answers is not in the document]
sophie-bates: Done in [9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d).

src/macaron/config/defaults.ini
sophie-bates: Changed in [9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d).

src/macaron/slsa_analyzer/checks/build_as_code_check.py
sophie-bates: Restored these comments in [9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d).

src/macaron/slsa_analyzer/checks/build_as_code_check.py
sophie-bates: Added additional variable in `defaults.ini` to store flag values, and changed `module` to `interpreter` in [9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d).

src/macaron/slsa_analyzer/checks/build_as_code_check.py
sophie-bates: Fixed in [9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d).

src/macaron/slsa_analyzer/checks/build_as_code_check.py
sophie-bates: Fixed in [9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d).

src/macaron/slsa_analyzer/checks/build_as_code_check.py
sophie-bates: Fixed in [9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d).

src/macaron/slsa_analyzer/checks/build_as_code_check.py
sophie-bates: Added in [9c167f2](https://github.com/oracle-samples/macaron/pull/67/commits/9c167f271c3b880c3db95d6de5fc7bb51bdfcf3d).

src/macaron/slsa_analyzer/build_tool/poetry.py
```diff
@@ -115,4 +122,38 @@ def get_dep_analyzer(self, repo_path: str) -> DependencyAnalyzer:
...
+        config_files = self.build_configs + self.package_lock
+        for build_cfg in config_files:
+            config_paths.update(glob.glob(os.path.join(repo_path, "**", build_cfg), recursive=True))
+
+        list_iter = iter(sorted(config_paths, key=lambda x: (str(Path(x).parent), len(Path(x).parts))))
+        try:
+            cfg_path = next(list_iter)
+            yield Path(cfg_path).parent.relative_to(repo_path)
+            while next_item := next(list_iter):
+                if str(Path(cfg_path).parent) in next_item:
```
tromai, line 153: I would recommend we very careful of the `in` between two strings because the behavior might be implicit. For example, if `cfg_path` is `/home/boo` and `next_item` is `/home/foo/home/boo/` , `if str(Path(cfg_path).parent) in next_item:` will be True and we will ignore the valid `/home/foo/home/boo/`. 
I think we could use `.startswith` here ?

tests/conftest.py
```diff
@@ -64,6 +71,23 @@ def setup_test(test_dir: Path, macaron_path: Path) -> NoReturn:  # type: ignore
     defaults.clear()
 
 
+@pytest.fixture()
+def check_result() -> CheckResult:  # pylint: disable=unused-argument
+    """Create a CheckResult instance.
+
+    Parameters
+    ----------
+    setup_test
```
tromai, line 80: The doc is inconsistent with the parameter provided to this fixture.

sophie-bates: I think this fixture should be passed the `setup_test` parameter.

src/macaron/slsa_analyzer/build_tool/poetry.py
sophie-bates: That's a good point. I based this abstraction off the implementation in [BaseBuildTool.get_build_dirs()](https://github.com/oracle-samples/macaron/blob/add-python-as-a-supported-build-tool/src/macaron/slsa_analyzer/build_tool/base_build_tool.py#L162-L165), so this function would need updating too if that's the case.

tromai: @behnazh-w How do you think about it?

behnazh-w: > That's a good point. I based this abstraction off the implementation in [BaseBuildTool.get_build_dirs()](https://github.com/oracle-samples/macaron/blob/add-python-as-a-supported-build-tool/src/macaron/slsa_analyzer/build_tool/base_build_tool.py#L162-L165), so this function would need updating too if that's the case.

That's not possible because first we sort the elements based on the [parent path](https://github.com/oracle-samples/macaron/blob/add-python-as-a-supported-build-tool/src/macaron/slsa_analyzer/build_tool/base_build_tool.py#L158). So `/home/foo/home/boo/` should never be filtered by `/home/boo`. But using `.startswith` would make it more explicit.

Regardless, I'm not sure if it makes sense to override this function to add `config_files = self.build_configs + self.package_lock`. As discussed, `poetry.lock` on its own cannot determine a build tool and `pyproject.toml` must always exist for poetry builds. So there shouldn't be any need to add `self.package_lock` here.

sophie-bates: @behnazh-w that's a good point. I had done it this way before we adjusted the config variable names, and now I agree that we don't need to override.

tests/slsa_analyzer/build_tool/__snapshots__/test_poetry.ambr
```diff
@@ -0,0 +1,16 @@
+# serializer version: 1
+# name: test_get_build_dirs[mock_repo0]
+  list([
+    PosixPath('.'),
+  ])
+# ---
+# name: test_get_build_dirs[mock_repo1]
+  list([
+    PosixPath('.'),
```
behnazh-w, line 9: This seems to be a false positive. The problem is that `get_build_dirs` function is not checking if the detected directory has a valid build. This PR should fix this issue: https://github.com/oracle-samples/macaron/pull/135

sophie-bates: Thank you for fixing that.

tests/slsa_analyzer/checks/test_build_as_code_check.py
```diff
@@ -1,149 +1,230 @@
...
-
-        # Test GitLab CI.
-        maven_deploy = AnalyzeContext("use_build_tool", os.path.abspath("./"), MagicMock())
-        maven_deploy.dynamic_data["build_spec"]["tool"] = maven
-        ci_info["service"] = gitlab_ci
-        bash_commands["commands"] = []
-        maven_deploy.dynamic_data["ci_services"] = [ci_info]
-        assert check.run_check(maven_deploy, check_result) == CheckResultType.FAILED
+
+@pytest.fixture()
```
behnazh-w, line 31: This `fixture` is not needed. Why not to directly call `BuildAsCodeCheck()` from the test function if nothing needs to be prepared before the test?

tests/conftest.py
```diff
@@ -63,6 +71,23 @@ def setup_test(test_dir: Path, macaron_path: Path) -> NoReturn:  # type: ignore
...
+        Depends on setup_test fixture.
+
+    Returns
+    -------
+    CheckResult
+        The CheckResult instance.
+    """
+    return CheckResult(justification=[])  # type: ignore
```
behnazh-w, line 88: It's better to instantiate `CheckResult` directly in tests to set the expected output. This fixture is not doing much and I'm not sure if it's helpful.

tests/slsa_analyzer/checks/test_build_as_code_check.py
```diff
@@ -1,149 +1,230 @@
...
+        name=os.path.basename(workflow_path),
+        node_type=GHWorkflowType.INTERNAL,
+        source_path=workflow_path,
+        parsed_obj=parsed_obj,
+        caller_path="",
+    )
+    root.add_callee(callee)
+    github_actions_service.build_call_graph_from_node(callee)
+    ci_info["callgraph"] = gh_cg
+    assert build_as_code_check.run_check(gha_deploy, check_result) == CheckResultType.PASSED
```
behnazh-w, line 230: Please add a failed case too, for example change the workflow name to None and a wrong name and check that they fail.

tests/conftest.py
sophie-bates: Fixed in [29cdd3a](https://github.com/oracle-samples/macaron/pull/67/commits/29cdd3a9e361a0f5f29c835805a2d9f186b92d0a).

src/macaron/slsa_analyzer/build_tool/poetry.py
sophie-bates: Removed this override in [29cdd3a](https://github.com/oracle-samples/macaron/pull/67/commits/29cdd3a9e361a0f5f29c835805a2d9f186b92d0a).

tests/slsa_analyzer/checks/test_build_as_code_check.py
sophie-bates: Fixed in [29cdd3a](https://github.com/oracle-samples/macaron/pull/67/commits/29cdd3a9e361a0f5f29c835805a2d9f186b92d0a).

tests/conftest.py
sophie-bates: Fixed in [29cdd3a](https://github.com/oracle-samples/macaron/pull/67/commits/29cdd3a9e361a0f5f29c835805a2d9f186b92d0a).

tests/slsa_analyzer/checks/test_build_as_code_check.py
sophie-bates: Fixed in [29cdd3a](https://github.com/oracle-samples/macaron/pull/67/commits/29cdd3a9e361a0f5f29c835805a2d9f186b92d0a).

behnazh-w (approved):

tromai (approved):

PR #76, opened 2023-02-27
build: build and deploy Macaron in Docker image

### LINKED ISSUE #3: Deploy Macaron as a Docker image
**Description/Requirements**

We need to setup the Docker image for Macaron. The Docker image will be use the latest Oracle linux 9 slim - `container-registry.oracle.com/os/oraclelinux:9-slim`. For more information, please see [h](https://container-registry.oracle.com/).

The required run time components in the deployed Docker are:

- Needed to be installed from source (because `oraclelinux:9-slim` only has up to Python3.9):
    - [Python 3.11.1](https://www.python.org/downloads/release/python-3111/)
- Run time libraries - Can be installed from the [yum repos of oracle linux 9](https://yum.oracle.com/oracle-linux-9.html).
    - sqlite-devel-3.34.1-5.el9
    - git-2.31.1-2.el9.2
    - java-11-openjdk-devel-1:11.0.17.0.8-2.0.1.el9_0
- The Go packages will be compiled and copied to the Docker image during build time. Therefore, we don't need to install Go in the final Docker image.
- The maven wrapper files in `resources/` will be downloaded and verified using GitHub Actions before copied to the Docker image.

**Other requirements**

- The output files generated by the Docker image must not have root owner as we want to make them available for the user.
- Make sure that it fits to be used in the release pipeline for Macaron (#1).
- The final Docker image must be signed (probably using [Sigstore](https://github.com/sigstore/cosign#quick-start)) and published to [GH Container registry
](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-docker-registry)
- Use Docker multi-stage build to minimize the final Docker image size.

tromai:
Close #3 

Tasks:

- [x] Make the PR title match the conventional commit message style.
- [x] Use docker CLI inside the Github Action runner instead of using the 3rd party Actions `docker/setup-buildx-action`
- [x] Reverting accidentally removed targets in Makefileakefile
- [x] Create a `workflow_dispatch` workflow to build and push the Macaron base image (which contains git, JDK 17, Souffle and Python3.11). The workflow that builds the final image will pull this base image from `ghcr.io/oracle-samples/macaron-base`.
- [x] Integrate the build and deploy Macaron Docker image to the existing release pipeline.
- [x] Generate SBOM for the final Docker image.
- [x] Use long cmd options in bash scripts for readability.
- [x] Add the feature for the new policy language in the run_macaron.sh (Solution TBD).
- [x] Use the same cmd options in `run_macaron.sh` as in `macaron` Python module.
- [x] Fix the incorrect JAVA_HOME in the Macaron base image.
- [x] Preserve the run time dependencies for Souffle
- [x] Add integration tests for the Docker image
- [x] Update Python v to 3.11.2
- [x] Remove pushing the Docker image in `_build_docker.yaml`
- [x] Add extra optimization for the macaron-base Docker image (by merging the RUN blocks together).
- [x] Remove the on push event for `build_base_image.yaml`


behnazh-w:
@tromai the PR title should match conventional commits.

.github/workflows/_build.yaml
```diff
@@ -90,6 +90,27 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ github.token }}
 
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@8c0edbc76e98fa90f69d9a2c020dcb50019dc325   # v2.2.1
```
behnazh-w, line 94: We try not to use third-party Actions. Can we use `docker` directly instead?

Makefile
```diff
@@ -249,9 +249,9 @@ integration-test:
 # set to the build date/epoch. For more details, see: https://flit.pypa.io/en/latest/reproducible.html
 .PHONY: dist
 dist: dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-py3-none-any.whl dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION).tar.gz dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-docs-html.zip dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-build-epoch.txt
-dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-py3-none-any.whl: check test integration-test
+dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-py3-none-any.whl:
```
behnazh-w, line 252: Why are you removing `check test integration-test` from the `dist` target? The `dist` target should only run when the tests pass.

Makefile
```diff
@@ -249,9 +249,9 @@ integration-test:
...
 	flit build --setup-py --format wheel
-dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION).tar.gz: check test integration-test
+dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION).tar.gz:
```
behnazh-w, line 254: Similar comment for the dependency targets of `dist`.

docker/user.sh
```diff
@@ -1,24 +1,47 @@
...
 
-groupmod -o -g $USER_GID macaron
-usermod -o -g $USER_GID -u $USER_UID macaron
+# We update the GID and UID of the existing macaron user in the container
+# with the GID and UID of the current user in the host machine.
+# This will make the files created during the runtime of the container would have
+# the current user of the host machine as the owner.
+if [[ -n "$USER_GID" ]] && [[ -n "$USER_UID" ]];
+then
+    groupmod -o -g "$USER_GID" macaron
```
behnazh-w, line 12: Can we use the long cmd options to improve the readability?

Makefile
tromai: Oops. I accidentally remove it. I will add it back.

.github/workflows/_build.yaml
tromai: I will switch to use docker CLI inside `run` instead.

docker/Dockerfile.base
```diff
@@ -0,0 +1,277 @@
...
+    && dnf clean all \
+    && rm -rf /var/cache/yum \
+    && rm -rf /var/cache/dnf \
+    && souffle --version
+
+# INSTALLING MACARON RUNTIME DEPENDENCIES AND OTHER UTILITIES PACKAGES.
+#   - java-17-openjdk-devel
+#   - git
+#   - findutils
+ENV JAVA_HOME="/usr/lib/jvm/java-17-openjdk"
```
tromai, line 262: This is not correct as we are now using the Oracle JDK 17.

tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/b12c92fbb627a4d3d995e124a6e5846439aa6fcd.

Makefile
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/f36424e2ba30eaf136e40616aa35fe71954fcc37

Makefile
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/f36424e2ba30eaf136e40616aa35fe71954fcc37

.github/workflows/_build.yaml
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/b3b86c7d2e1307ff3a76b10615886555ecd207a3

docker/user.sh
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/37e91e824c2d7a67e0071df08ce83048e0c4ab52

docker/Dockerfile.final
```diff
@@ -0,0 +1,77 @@
...
+
+# Installing the Python dependencies in the Python virtual environment.
+# We switch to user macaron so that when we install the dependencies using pip,
+# the warning of not having correct ownership of /home/macaron is not raised.
+USER macaron:macaron
+COPY --chown=macaron:macaron $WHEEL_PATH $HOME/dist/
+RUN : \
+    && python3 -m venv $HOME/.venv \
+    && . .venv/bin/activate \
+    && pip install --no-compile --no-cache-dir --upgrade pip \
```
tromai, line 64: There should be only one pipe install. This line is unnecessary.

.github/workflows/_build_docker.yaml
```diff
@@ -0,0 +1,55 @@
...
+      env:
+        # This will be picked up by run_macaron.sh.
+        MACARON_IMAGE_TAG: test
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        ./scripts/dev_scripts/integration_tests_docker.sh ${{ github.workspace }} ./scripts/release_scripts/run_macaron.sh
+
+    - name: Push the built Docker image
+      run: |
+        if [[ -n "${{ inputs.release-tag }}" ]];
```
behnazh-w, line 51: You can use an `if` condition on the step [as shown here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif) to prevent the step from running.

.github/workflows/build_base_image.yaml
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
```
behnazh-w, line 3: Please add documentation on top to describe the purpose of this workflow.

.github/workflows/build_base_image.yaml
```diff
@@ -0,0 +1,28 @@
...
+name: Build Macaron base image
```
behnazh-w, line 4:
```suggestion
name: Build base image
```

.github/workflows/build_base_image.yaml
```diff
@@ -0,0 +1,28 @@
...
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  build-macaron-base:
```
behnazh-w, line 13:
```suggestion
  build-base-image:
```

.github/workflows/build_base_image.yaml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+    - name: Check out repository
+      uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
+      with:
+        fetch-depth: 0
+
+    - name: Log in to GitHub Container Registry
+      run: docker login ghcr.io --username ${{ github.actor }} --password ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Build and push Macaron base image
```
behnazh-w, line 25:
```suggestion
    - name: Build and push the base image
```

.github/workflows/_build_docker.yaml
```diff
@@ -0,0 +1,55 @@
...
+      with:
+        fetch-depth: 0
+
+    - name: Download artifact
+      uses: actions/download-artifact@9782bd6a9848b53b110e712e20e42d89988822b7 # v3.0.1
+      with:
+        path: dist
+
+    - name: Log in to GitHub Container Registry
+      run: docker login ghcr.io --username ${{ github.actor }} --password ${{ secrets.GITHUB_TOKEN }}
```
behnazh-w, line 35: Don't we need to use `REPO_ACCESS_TOKEN` PAT here instead? I suspect `secrets.GITHUB_TOKEN` should not have the required scope access to login and later push the image.

.github/workflows/release.yaml
```diff
@@ -200,6 +200,48 @@ jobs:
       env:
         GH_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
 
+  build_push_docker_image:
+    needs: [build, release]
+    name: Build and push Docker image
+    permissions:
+      contents: read
+      packages: write # To push the Docker image to ghcr.io
+    uses: ./.github/workflows/_build_docker.yaml
```
behnazh-w, line 209: Don't we need to pass the required secrets down to the RWF? See the relevant documentation [here](https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow).

.github/workflows/_build_docker.yaml
behnazh-w, line 35: If the workflow is called from the `pr-change-set.yaml`, I don't understand why we would need to login because we won't push the image to the registry.

.github/workflows/release.yaml
```diff
@@ -200,6 +200,48 @@ jobs:
       env:
         GH_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
 
+  build_push_docker_image:
```
behnazh-w, line 203: Please add a comment above each step to explain what it does.

.github/workflows/release.yaml
```diff
@@ -200,6 +200,48 @@ jobs:
...
+
+  publish_docker_imabe_sbom:
+    name: Release the Docker image sbom
+    needs: [build_push_docker_image, release]
+    permissions:
+      contents: write # To publish the SBOM of the Docker image.
+    runs-on: ubuntu-latest
+    env:
+      SYFT_BIN: ${{ github.workspace }}/bin
+      SBOM_FILE_NAME: macaron-sbom-cyclonedx.json
```
behnazh-w, line 221: Please follow the naming convention we use for other SBOMs:
```
macaron-<release-version>-sbom-docker.json
```

e.g., 
`macaron-0.1.0-sbom-docker.json`

.github/workflows/release.yaml
```diff
@@ -200,6 +200,48 @@ jobs:
...
+    steps:
+    - name: Check out repository
```
behnazh-w, line 223: Why do we need to checkout the repository here?

.github/workflows/release.yaml
```diff
@@ -200,6 +200,48 @@ jobs:
...
+      uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+      with:
+        fetch-depth: 0
+
+    - name: Install Syft
+      # https://github.com/anchore/syft#installation
+      run: |
+        curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b ${{ env.SYFT_BIN }} v0.74.1
+        ${{ env.SYFT_BIN }}/syft --version
```
behnazh-w, line 232: You can just use `$SYFT_BIN` to read the environment variable.

tromai: Ah that makes sense. We are running with the shell commands.

.github/workflows/release.yaml
tromai: I believe we need to checkout the repository following what we already have for publishing the provenance with `gh relase upload` - https://github.com/oracle-samples/macaron/blob/9e89477f6ad91749bbb9bda1d7361ee0d1cb3305/.github/workflows/release.yaml#L180

.github/workflows/_build_docker.yaml
tromai: Currently, the image `ghcr.io/oracle-samples/macaron-base` is private. The login step is needed so that we could pull it for building the final Macaron Docker image.

behnazh-w: We need to make sure the workflows that are triggered with `pull_request` event have the minimal permissions and accesses because it can contain untrusted code. So we should not allow docker login for pull requests.

.github/workflows/release.yaml
behnazh-w: > I believe we need to checkout the repository following what we already have for publishing the provenance with `gh relase upload` -
> 
> https://github.com/oracle-samples/macaron/blob/9e89477f6ad91749bbb9bda1d7361ee0d1cb3305/.github/workflows/release.yaml#L180

Yes, that's right. `gh` needs a checked out git repo.

.github/workflows/release.yaml
```diff
@@ -200,6 +200,48 @@ jobs:
       env:
         GH_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
 
+  build_push_docker_image:
+    needs: [build, release]
+    name: Build and push Docker image
+    permissions:
+      contents: read
+      packages: write # To push the Docker image to ghcr.io
```
behnazh-w, line 208: Because `./.github/workflows/_build_docker.yaml` is also called from an untrusted context for testing PRs, we need to refactor the Docker build workflow and push the image either as a step here in the `release.yaml` or as a separate RWF. We want to tighten the permissions and access to pull requests.

.github/workflows/_build_docker.yaml
```diff
@@ -0,0 +1,55 @@
+# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+# This reuseable workflow builds the Macaron Docker image.
+# If release-tag is provided. The final image will be tagged and push to ghcr.io/oracle-samples/macaron:release-tag
```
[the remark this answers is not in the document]
tromai: This line will be removed according to our discussion.

.github/workflows/_build_docker.yaml
tromai: Because we are not going to push the Docker image here (from our latest discussion), `secrets.GITHUB_TOKEN` is sufficient for the integration tests.

.github/workflows/_build_docker.yaml
```diff
@@ -0,0 +1,63 @@
...
+        path: dist
+
+    # Verify hashes by first computing hashes for the artifacts and then comparing them
+    # against the hashes for the artifact.
+    - name: Verify the artifact hash
+      env:
+        ARTIFACT_HASH: ${{ inputs.artifact-sha256 }}
+      run: |
+        set -euo pipefail
+        cd dist
```
tromai, line 43: This line is needed because the artifact is downloaded into `dist/artifact_name/`. So the result of the download is
```
dist
└── artifact-ubuntu-latest-python-3.11
    ├── dist
    │   ├── macaron-0.0.0-build-epoch.txt
    │   ├── macaron-0.0.0-docs-html.zip
    │   ├── macaron-0.0.0-py3-none-any.whl
    │   ├── macaron-0.0.0-requirements.txt
    │   ├── macaron-0.0.0-sbom-go.json
    │   ├── macaron-0.0.0-sbom.json
    │   └── macaron-0.0.0.tar.gz
    └── src
        └── macaron
            └── bin
                ├── actionparser
                └── bashparser
```
Without changing the directory into dist, verifying the hash will fail as we don't reflect the exact directory structure that the hash is originally computed from. 
I tried following the existing [step](https://github.com/oracle-samples/macaron/blob/9e89477f6ad91749bbb9bda1d7361ee0d1cb3305/.github/workflows/release.yaml#L106) but it didn't work in this case.

.github/workflows/_build.yaml
```diff
@@ -140,3 +140,14 @@ jobs:
           ${{ env.PACKAGE_PATH }}/bin/bashparser
         if-no-files-found: error
         retention-days: 7
+
+  # For now only use the hard-coded artifact name.
```
behnazh-w, line 144: We don't need to hard-code it. You should be able to use:
```
artifact-${{ env.ARTIFACT_OS }}-python-${{ env.ARTIFACT_PYTHON }}
```
Also, please add a comment about the purpose of the following job.

.github/workflows/_build.yaml
```diff
@@ -140,3 +140,14 @@ jobs:
           ${{ env.PACKAGE_PATH }}/bin/bashparser
```
behnazh-w, line 140: After our discussion, we decided to remove these artifacts as they will be part of the wheel anyway?

behnazh-w, line 140: After our discussion, we decided to remove these artifacts as they will be part of the wheel anyway?

.github/workflows/_build_docker.yaml
behnazh-w: See my other comment. Didn't we decided to remove the Go binaries as they are included in the wheel? If we don't add those binaries, we shouldn't need to run `cd dist` I think.

.github/workflows/_build_docker.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+    # We need to login to access the private macaron-base image.
+    # TODO: remove it when macaron-base is public.
+    - name: Login to ghcr.io
+      run: docker login ghcr.io --username ${{ github.actor }} --password ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Build the Docker image
+      env:
+        IMAGE_NAME: ghcr.io/oracle-samples/macaron
+      run: ./scripts/dev_scripts/build_docker.sh "${IMAGE_NAME}" ${{ github.workspace }} ${{ inputs.release-tag }}
```
behnazh-w, line 56: We shouldn't need `inputs.release-tag` anymore because we never build a release image, right?

.github/workflows/_build_docker.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+    - name: Test the Docker image
+      env:
+        # This will be picked up by run_macaron.sh.
```
behnazh-w, line 60:
```suggestion
        # This environment variable will be picked up by run_macaron.sh.
```

.github/workflows/release.yaml
```diff
@@ -111,6 +113,34 @@ jobs:
...
         echo "$ARTIFACT_HASH" | base64 --decode | sha256sum --strict --check --status || exit 1
 
+    - name: Log in to GitHub Container Registry
+      run: docker login ghcr.io --username ${{ github.actor }} --password ${{ secrets.GITHUB_TOKEN }}
+
+    # The Docker image will be built with tags: latest and ${{ RELEASE_TAG }}
+    - name: Build the Docker image
+      run: |
+        RELEASE_TAG=$(git describe --tags --abbrev=0)
+        ./scripts/dev_scripts/build_docker.sh "${IMAGE_NAME}" ${{ github.workspace }} "${RELEASE_TAG}"
```
behnazh-w, line 123: Why not `"$IMAGE_NAME"` and `"$RELEASE_TAG"`?

behnazh-w, line 123: Why not `"$IMAGE_NAME"` and `"$RELEASE_TAG"`?

.github/workflows/_build_docker.yaml
tromai: > Didn't we decided to remove the Go binaries as they are included in the wheel?
Yes we did. However, because how the artifact hash is calculated [here](https://github.com/oracle-samples/macaron/blob/9e89477f6ad91749bbb9bda1d7361ee0d1cb3305/.github/workflows/_build.yaml#L118) so I decided to keep it in the uploaded artifact. Perhaps it's better if we remove the steps to calculate the hash for both `actionparser` and `bashparser`.

.github/workflows/_build.yaml
tromai: You can see my answer to this issue here - https://github.com/oracle-samples/macaron/pull/76#discussion_r1137896832

.github/workflows/_build_docker.yaml
tromai: Thanks for pointing it out. Yes, I forgot to remove it from here.

.github/workflows/release.yaml
tromai: I was following the convention for referencing environment variable in bash commands (`${ENV_NAME}`) so I used it here too. I could switch it to your suggestion to make it consistent.

.github/workflows/_build_docker.yaml
behnazh-w: > Perhaps it's better if we remove the steps to calculate the hash for both `actionparser` and `bashparser`.

Yes, please remove them.

.github/workflows/release.yaml
tromai: I think this comment is duplicated - https://github.com/oracle-samples/macaron/pull/76#discussion_r1137899352

behnazh-w: I usually use ShellCheck as a reference, which doesn't seem to recommend it. Yes, it's good to be consistent with other parts of our code.

.github/workflows/_build.yaml
tromai: I tried using this but GitHub Actions raises an error - https://github.com/oracle-samples/macaron/actions/runs/4432579381

behnazh-w: > I tried using this but GitHub Actions raises an error - https://github.com/oracle-samples/macaron/actions/runs/4432579381

Yeah, looks like `env` context still can only be used at `step` level: https://github.com/actions/runner/issues/2394
Please use the fixed image name and add a TODO to use `env` when/if it is supported here.

.github/workflows/release.yaml
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/d2fe70dc8bc12b52071d18e885bf63b86f9dcc25

.github/workflows/_build_docker.yaml
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/d2fe70dc8bc12b52071d18e885bf63b86f9dcc25

.github/workflows/_build_docker.yaml
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/d2fe70dc8bc12b52071d18e885bf63b86f9dcc25

.github/workflows/_build.yaml
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/d2fe70dc8bc12b52071d18e885bf63b86f9dcc25

.github/workflows/_build.yaml
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/d2fe70dc8bc12b52071d18e885bf63b86f9dcc25

.github/workflows/_build_docker.yaml
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/d2fe70dc8bc12b52071d18e885bf63b86f9dcc25

.github/workflows/release.yaml
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/24da909332996f30775f68257a844b0d2b9cbb6a

.github/workflows/release.yaml
```diff
@@ -200,6 +200,48 @@ jobs:
...
+        curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b ${{ env.SYFT_BIN }} v0.74.1
+        ${{ env.SYFT_BIN }}/syft --version
+
+    - name: Generate sbom
+      run: |
+        ${{ env.SYFT_BIN }}/syft \
+          ghcr.io/oracle-samples/macaron:${{ needs.release.outputs.release-tag }} \
+          -o cyclonedx-json=${{ env.SBOM_FILE_NAME }}
+
+    - name: Push to release asset
```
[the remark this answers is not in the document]
tromai: This part has been removed in https://github.com/oracle-samples/macaron/pull/76/commits/24da909332996f30775f68257a844b0d2b9cbb6a

.github/workflows/release.yaml
```diff
@@ -200,6 +200,48 @@ jobs:
...
+      env:
+        GH_TOKEN: ${{ github.token }}
```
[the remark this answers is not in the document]
tromai: This part is removed in https://github.com/oracle-samples/macaron/pull/76/commits/24da909332996f30775f68257a844b0d2b9cbb6a

.github/workflows/release.yaml
```diff
@@ -200,6 +200,48 @@ jobs:
...
+
+    - name: Install Syft
+      # https://github.com/anchore/syft#installation
+      run: |
+        curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b ${{ env.SYFT_BIN }} v0.74.1
+        ${{ env.SYFT_BIN }}/syft --version
+
+    - name: Generate sbom
+      run: |
+        ${{ env.SYFT_BIN }}/syft \
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/d2fe70dc8bc12b52071d18e885bf63b86f9dcc25

.github/workflows/release.yaml
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/d2fe70dc8bc12b52071d18e885bf63b86f9dcc25

.github/workflows/release.yaml
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/24da909332996f30775f68257a844b0d2b9cbb6a

.github/workflows/release.yaml
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/d2fe70dc8bc12b52071d18e885bf63b86f9dcc25

.github/workflows/_build_docker.yaml
```diff
@@ -0,0 +1,55 @@
...
+        path: dist
+
+    - name: Log in to GitHub Container Registry
+      run: docker login ghcr.io --username ${{ github.actor }} --password ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Build the Docker image
+      run: |
+        ./scripts/dev_scripts/build_docker.sh ${{ env.IMAGE_NAME }} ${{ github.workspace }} ${{ inputs.release-tag }}
+
+    - name: Test the built Docker image
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/d2fe70dc8bc12b52071d18e885bf63b86f9dcc25

scripts/dev_scripts/build_docker.sh
```diff
@@ -0,0 +1,38 @@
...
+WORKSPACE=$2
+RELEASE_TAG=$3
+
+DIST_PATH="${WORKSPACE}/dist"
+REPO_PATH="${WORKSPACE}"
+
+WHEEL_PATH=$(find "$DIST_PATH" -depth -type f -name 'macaron-*.whl' | head -n 1)
+if [[ -z "${WHEEL_PATH}" ]];
+then
+    echo "Cannot find Macaron wheel file in ${DIST_PATH}."
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/d2fe70dc8bc12b52071d18e885bf63b86f9dcc25

.github/workflows/_build.yaml
```diff
@@ -134,9 +132,20 @@ jobs:
...
         name: artifact-${{ matrix.os }}-python-${{ matrix.python }}
-        path: |
-          dist
-          ${{ env.PACKAGE_PATH }}/bin/actionparser
-          ${{ env.PACKAGE_PATH }}/bin/bashparser
+        path: dist
         if-no-files-found: error
         retention-days: 7
+
+  # This job called the reusable workflow _build_docker.yaml to build and test
```
behnazh-w, line 139:
```suggestion
  # This job calls the reusable workflow _build_docker.yaml to build and test
```

.github/workflows/_build_docker.yaml
```diff
@@ -0,0 +1,66 @@
...
+permissions:
+  contents: read
+
+jobs:
+  build-docker:
+    runs-on: ubuntu-latest
+    permissions:
+      packages: read
+    steps:
+    - name: Check out repository
```
behnazh-w, line 28: Please add a new line bellow `steps` to be consistent with other workflows.

.github/workflows/build_base_image.yaml
behnazh-w: @tromai Looks like this not addressed.

.github/workflows/build_base_image.yaml
behnazh-w: This comment is not addressed.

.github/workflows/build_base_image.yaml
behnazh-w: This issue is not addressed.

.github/workflows/pr-change-set.yaml
```diff
@@ -1,8 +1,8 @@
-# Copyright (c) 2022 - 2022, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
-# This workflow checks and tests the package code, and it builds all package
-# artifacts whenever there were changes to a pull request.
+# This workflow checks and tests the package code, builds all package
+# artifacts and the Docker image whenever there were changes to a pull request.
```
behnazh-w, line 5:
```suggestion
# artifacts and the Docker image whenever there are changes to a pull request.
```

.github/workflows/release.yaml
```diff
@@ -111,6 +113,38 @@ jobs:
         echo "Hash of package should be $ARTIFACT_HASH."
         echo "$ARTIFACT_HASH" | base64 --decode | sha256sum --strict --check --status || exit 1
 
+    # Log in to ghcr.io so that we could push the Macaron Docker image.
```
behnazh-w, line 116:
```suggestion
    # Log in to ghcr.io to push the Docker image.
```

docker/Dockerfile.base
```diff
@@ -0,0 +1,251 @@
+# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+# This Dockerfile is for building the base image which contains the necessary components
+# to build and run the Macaron Docker image.
+# This image is based on the container-registry.oracle.com/os/oraclelinux:9-slim image and has the main components:
```
behnazh-w, line 6:
```suggestion
# This image is based on the container-registry.oracle.com/os/oraclelinux:9-slim image and contains the following components:
```

docker/Dockerfile.base
```diff
@@ -0,0 +1,251 @@
...
+    PATH=/usr/local/bin:/usr/lib/jvm/jdk-17-oracle-x64/bin:$PATH
+
+# We run the installation of all components in one single RUN to minimize the final image size.
+RUN : \
+    && set -eux \
+# ---------------------------------------------------------------------------------------------------------------------
+    # INSTALL RUNTIME DEPENDENCIES.
+    && microdnf install -y \
+        # We switch to dnf to use some features of dnf that do not exist in microdnf.
+        dnf \
```
behnazh-w, line 39:
```suggestion
    # We switch to dnf to use some features of dnf that do not exist in microdnf.
    && microdnf install -y dnf \
```

docker/Dockerfile.base
```diff
@@ -0,0 +1,251 @@
...
+    # We can then safely remove the unused microdnf.
+    && dnf remove -y microdnf \
+    # Add the EPEL packages yum repo following the convention as existing repo definitions in OL9.
+    # The following yum repo will be resolved to this URL:
+    #       https://yum.oracle.com/repo/OracleLinux/OL9/developer/EPEL/x86_64/index.html
```
behnazh-w, line 44:
```suggestion
    # https://yum.oracle.com/repo/OracleLinux/OL9/developer/EPEL/x86_64/index.html
```

docker/Dockerfile.base
```diff
@@ -0,0 +1,251 @@
...
+        tzdata          \
+        ca-certificates \
+        # git and finutils are needed for running and building Macaron.
+        git             \
+        findutils       \
+        # Runtime libraries for Souffle. These are based on
+        # https://souffle-lang.github.io/build#software-requirements
+        # https://github.com/souffle-lang/souffle/blob/cc8ea091721fcfb60bed45a0edf571ad9d0c58a5/CMakeLists.txt#LL442C39-L442C129
+        # Exceptions (not installed):
+        #   mcpp - The package mcpp is not available oracle linux 9. However, we don't use Souffle's feature
```
behnazh-w, line 71:
```suggestion
        #   mcpp - The package mcpp is not available in Oracle Linux 9. However, we don't use Souffle's feature
```

docker/Dockerfile.base
```diff
@@ -0,0 +1,251 @@
...
+        #   python3 - We use the installed Python3.11.2 for this.
+        gcc-c++         \
+        libffi          \
+        libffi-devel    \
+        ncurses-devel   \
+        sqlite-devel    \
+        zlib-devel      \
+        # Oracle JDK17
+        https://download.oracle.com/java/17/latest/jdk-17_linux-x64_bin.rpm \
+    # Store the list of user installed packages so that we could preserve it when we clean up unused packages.
```
behnazh-w, line 82:
```suggestion
    # Store the list of user installed packages to preserve them when we clean up unused packages.
```

docker/Dockerfile.base
```diff
@@ -0,0 +1,251 @@
...
+    # https://dnf.readthedocs.io/en/latest/command_ref.html#history-command
+    && USER_MANUAL_INSTALLED="$(dnf history userinstalled | tail -n +2)" \
+# ---------------------------------------------------------------------------------------------------------------------
+    # INSTALL PYTHON FROM SOURCE.
+    # We enable the ol9_core_ready_builder repo for the package gdbm-devel
+    && dnf --enablerepo=ol9_codeready_builder install -y \
+        # Install Python build dependencies. These libraries are not needed during runtime and should be removed after
+        # the build complete.
```
behnazh-w, line 90:
```suggestion
        # the build completes.
```

docker/Dockerfile.base
```diff
@@ -0,0 +1,251 @@
...
+        # The required packages are listed here:
+        #   https://github.com/docker-library/python/blob/f568f56f28fab0fe87b34db777e2c2861cef002b/3.11/slim-buster/Dockerfile#L34
+        # Note that the link above describes the packages in Debian-based Linux distribution. The packages below might
+        # has different names because we are running in Oracle Linux (which is based on RedHat).
```
behnazh-w, line 94:
```suggestion
        # have different names because we are running in Oracle Linux (which is based on RedHat).
```

.github/workflows/_build_docker.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Build the Docker image
+      run: |
+        ./scripts/dev_scripts/build_docker.sh ${{ env.IMAGE_NAME }} ${{ github.workspace }} ${{ inputs.release-tag }}
+
+    - name: Test the built Docker image
+      env:
+        # This will be picked up by run_macaron.sh.
+        MACARON_IMAGE_TAG: test
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle-samples/macaron/commit/d2fe70dc8bc12b52071d18e885bf63b86f9dcc25

.github/workflows/_build_docker.yaml
```diff
@@ -0,0 +1,55 @@
...
+        ./scripts/dev_scripts/integration_tests_docker.sh ${{ github.workspace }} ./scripts/release_scripts/run_macaron.sh
+
+    - name: Push the built Docker image
```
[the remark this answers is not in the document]
tromai: Pushing the Docker image in this workflow is removed in https://github.com/oracle-samples/macaron/commit/24da909332996f30775f68257a844b0d2b9cbb6a

.github/workflows/_build_docker.yaml
tromai: This part is removed in https://github.com/oracle-samples/macaron/commit/24da909332996f30775f68257a844b0d2b9cbb6a

.github/workflows/build_base_image.yaml
tromai: Fixed in https://github.com/oracle-samples/macaron/commit/d2fe70dc8bc12b52071d18e885bf63b86f9dcc25

docker/Dockerfile.base
```diff
@@ -0,0 +1,251 @@
...
+	&& ldconfig \
+    && python3 --version \
+    # Create useful symlinks for python related tools.
+    # https://github.com/docker-library/python/blob/f568f56f28fab0fe87b34db777e2c2861cef002b/3.11/slim-buster/Dockerfile#L124
+	&& for src in idle3 pydoc3 python3 python3-config; do \
+		dst="$(echo "$src" | tr -d 3)"; \
+		[ -s "/usr/local/bin/$src" ]; \
+		[ ! -e "/usr/local/bin/$dst" ]; \
+		ln -svT "$src" "/usr/local/bin/$dst"; \
+	done \
```
behnazh-w, line 180:
```suggestion
	&& for src in idle3 pydoc3 python3 python3-config; do \
		dst="$(echo "$src" | tr -d 3)"; \
		[ -s "/usr/local/bin/$src" ]; \
		[ ! -e "/usr/local/bin/$dst" ]; \
		ln -svT "$src" "/usr/local/bin/$dst"; \
	   done \
```

docker/Dockerfile.base
```diff
@@ -0,0 +1,251 @@
...
+# ---------------------------------------------------------------------------------------------------------------------
+	# INSTALL PIP AND SETUPTOOLS
+    # https://github.com/docker-library/python/blob/f568f56f28fab0fe87b34db777e2c2861cef002b/3.11/slim-buster/Dockerfile#LL147C2-L166C15
```
behnazh-w, line 183:
```suggestion
        # https://github.com/docker-library/python/blob/f568f56f28fab0fe87b34db777e2c2861cef002b/3.11/slim-buster/Dockerfile#LL147C2-L166C15
```

docker/Dockerfile.base
```diff
@@ -0,0 +1,251 @@
...
+    && cmake -S . -B build -DSOUFFLE_DOMAIN_64BIT=ON -DCMAKE_INSTALL_PREFIX="/usr/local" \
+    && nproc="$(nproc)" \
+    && cmake --build build --target install -j "$nproc" \
+    && cd / \
+    && rm -rf /souffle-src \
+    && souffle --version \
+# ---------------------------------------------------------------------------------------------------------------------
+    # CLEANING UP.
+    # We mark all unecessary packages to be removed while preserving the user installed packages.
+	&& dnf list installed | tail -n +2 | cut -d' ' -f1 | xargs -r dnf mark remove > /dev/null \
```
behnazh-w, line 227:
```suggestion
    && dnf list installed | tail -n +2 | cut -d' ' -f1 | xargs -r dnf mark remove > /dev/null \
```

docker/Dockerfile.base
```diff
@@ -0,0 +1,251 @@
...
+	&& for src in idle3 pydoc3 python3 python3-config; do \
+		dst="$(echo "$src" | tr -d 3)"; \
+		[ -s "/usr/local/bin/$src" ]; \
+		[ ! -e "/usr/local/bin/$dst" ]; \
+		ln -svT "$src" "/usr/local/bin/$dst"; \
+	done \
+# ---------------------------------------------------------------------------------------------------------------------
+	# INSTALL PIP AND SETUPTOOLS
+    # https://github.com/docker-library/python/blob/f568f56f28fab0fe87b34db777e2c2861cef002b/3.11/slim-buster/Dockerfile#LL147C2-L166C15
+	&& wget -O get-pip.py "$PYTHON_GET_PIP_URL" \
```
behnazh-w, line 184: It would be better to use the default `pip` module installed with `Python` if possible. You mentioned installing it separately can reduce the final size of the image. Is that significant enough to be worth it?

docker/Dockerfile.base
```diff
@@ -0,0 +1,251 @@
...
+    # https://github.com/docker-library/python/blob/f568f56f28fab0fe87b34db777e2c2861cef002b/3.11/slim-buster/Dockerfile#LL111C2-L121C3
+    && find /usr/local -type f -executable -not \( -name '*tkinter*' \) -exec ldd '{}' ';' \
+		| awk '/=>/ { print $(NF-1) }' \
+		| sort -u \
+		| xargs -r dpkg-query --search \
+		| cut -d: -f1 \
+		| sort -u \
+		| xargs -r dnf mark install \
+    # We exclude systemd-udev because currently one of the build dependencies for Souffle pull this protected package.
+    # This command make sure that it won't break if systemd-udev is not pulled in the future.
```
behnazh-w, line 239: Why is it OK if in future `systemd-udev` is not pulled?

tromai: What I meant was that if one of the build dependencies of Souffle decides to not depend on `systemd-udev` anymore, we won't mark `systemd-udev` as user installed. If a library is not installed but we mark it as `userinstalled`, then `dnf` will raise errors.

docker/Dockerfile.base
tromai: I have just checked the differences. It only comes in around 4MB less when we use the default pip that comes with our Python installation.

behnazh-w: > I have just checked the differences. It only comes in around 4MB less when we use the default pip that comes with our Python installation.

So is it worth the separate installation?

tromai: I think the only benefit of doing what we have here is that we ensure that pip hash is checked properly before we install it. Otherwise, 4MB is not worth in terms of image size to walk this path. What do you think about it?

docker/Dockerfile.base
tromai: I think GitHub UI display this part a bit weirdly.
This is the original text:
```dockerfile
    && for src in idle3 pydoc3 python3 python3-config; do \
        dst="$(echo "$src" | tr -d 3)"; \
        [ -s "/usr/local/bin/$src" ]; \
        [ ! -e "/usr/local/bin/$dst" ]; \
        ln -svT "$src" "/usr/local/bin/$dst"; \
    done \
```
So I think if we want `done` to align with `for` , `done` must be moved 3 spaces to the left, which is not the usual 4 spaces indent :thinking:

tromai: It's very likely that tab and spaces are mixed in this Dockerfile. That's why GitHub UI display some parts differently. I will have a look at this.

tromai: Yes, that is true. I will also convert all tabs into 4 spaces in the next commit.

.github/workflows/_build_docker.yaml
tromai: Fixed https://github.com/oracle-samples/macaron/pull/76/commits/6d2ab69b14a7377c3735644354b069b9171b9f78

docker/Dockerfile.final
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/6d2ab69b14a7377c3735644354b069b9171b9f78

scripts/dev_scripts/build_docker.sh
behnazh-w, line 1: Can you run `shellcheck` on the shell scripts (except for integartion_tests_docker.sh) which need to be improved improve [in a separate PR](https://github.com/oracle-samples/macaron/issues/109).

scripts/release_scripts/run_macaron.sh
```diff
@@ -0,0 +1,394 @@
...
+#   -lr/--local-repos-path LOCAL_REPOS_PATH: The directory where Macaron looks for already cloned repositories.
+#   -po/--policy POLICY: The path to a policy file.
+#   -t/--personal_access_token PERSONAL_ACCESS_TOKEN: The GitHub personal access token, which is mandatory for running analysis.
+#   -v/--verbose: Run Macaron with more debug logs.
+# These are arguments for policy_engine entrypoint.
+#   -f/--file FILE: Replace policy file.
+#   -d/--database DATABASE: Database path.
+argv_main=()
+
+# Arguments whose values changed by this script before being passed to the Macaron Docker image.
```
behnazh-w, line 39: This sentence also appears at line 26. Can you please clarify the difference?

scripts/release_scripts/run_macaron.sh
```diff
@@ -0,0 +1,394 @@
...
+# These are arguments will be specific to action in macaron entrypoint.
```
behnazh-w, line 40:
```suggestion
# These are the sub-commands for a specific action.
```

Does that make sense?

scripts/release_scripts/run_macaron.sh
tromai: No, the two sentences are the same and they both descibe `argv_main` and `argv_action`. I will remove the duplicated one.

scripts/release_scripts/run_macaron.sh
tromai: Yep, it sounds good to me.

tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/8694a7fe8b5617a73598be3d26945989ee457684

scripts/release_scripts/run_macaron.sh
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/76/commits/8694a7fe8b5617a73598be3d26945989ee457684

scripts/dev_scripts/build_docker.sh
tromai: I have run and resolve the errors reported by `shellcheck` in https://github.com/oracle-samples/macaron/pull/76/commits/8694a7fe8b5617a73598be3d26945989ee457684

behnazh-w (approved):

tromai:
Rebase on staging and remove the on_push event for `build_base_image.yaml`

PR #96, opened 2023-03-20
fix: do not log check_module object to avoid info leakage

behnazh-w:
CodeQL has found [an info leakage bug](https://github.com/oracle-samples/macaron/security/code-scanning/1) in Macaron, which should be fixed in this PR :nerd_face:


tromai (approved):
Very interesting. Good job CodeQL :+1:

PR #104, opened 2023-03-21
fix: allow BOM component version and group be empty

### LINKED ISSUE #103: Error when resolving a CycloneDx SBOM component missing both version number and external references.
**PROBLEM**
At this [line](https://github.com/oracle-samples/macaron/blob/9094cf005b9f05920a1b24cbc8b23f480f648e93/src/macaron/dependency_analyzer/cyclonedx.py#L179), we are reading the value of key `version` in `item` which is a DependencyInfo (a TypedDict). However, if in the CycloneDx SBOM, a component does not have any defined `version` value , an `AttributeError` could be raised at this line.
For example, this component would be enough to cause the issue.
```json
{
    "bomFormat": "CycloneDX",
    "specVersion": "1.4",
    "serialNumber": "...",
    "version": 1,
    "metadata": {
        "timestamp": "...",
        "component": {
            ...
    },
    "components": [
        ...
        {
            "group": "far",
            "name": "bar",
            "type": "foo",
            "bom-ref": "boo"
        },
        ...
    ],
    ...
}
```

behnazh-w:
Closes #103

### DIFF
--- a/src/macaron/dependency_analyzer/dependency_resolver.py
+++ b/src/macaron/dependency_analyzer/dependency_resolver.py
@@ -4,6 +4,7 @@
 """This module processes and collects the dependencies to be processed by Macaron."""
 
 import logging
+import re
 from abc import ABC, abstractmethod
 from enum import Enum
 from typing import TypedDict
@@ -157,10 +158,12 @@ class DependencyAnalyzer(ABC):
             latest_deps[key] = item
         else:
             try:
+                # If version is found, compare the numeric part of the version string.
                 if (
-                    latest_value["version"]
-                    and item["version"]
-                    and version.Version(latest_value["version"].lower()) < version.Version(item["version"].lower())
+                    (latest_version := latest_value.get("version", ""))
+                    and (item_version := item.get("version", ""))
+                    and version.Version(re.sub(r"[^0-9.]", "", latest_version))
+                    < version.Version(re.sub(r"[^0-9.]", "", item_version))
                 ):
                     latest_deps[key] = item
             except ValueError as error:
--- a/src/macaron/dependency_analyzer/dependency_resolver.py
+++ b/src/macaron/dependency_analyzer/dependency_resolver.py
@@ -157,7 +157,6 @@ class DependencyAnalyzer(ABC):
             latest_deps[key] = item
         else:
             try:
-                # If version is found, compare the numeric part of the version string.
                 if (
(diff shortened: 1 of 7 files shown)


src/macaron/dependency_analyzer/dependency_resolver.py
```diff
@@ -157,10 +158,12 @@ def add_latest_version(
...
         else:
             try:
+                # If version is found, compare the numeric part of the version string.
                 if (
-                    latest_value["version"]
-                    and item["version"]
-                    and version.Version(latest_value["version"].lower()) < version.Version(item["version"].lower())
+                    (latest_version := latest_value.get("version", ""))
+                    and (item_version := item.get("version", ""))
+                    and version.Version(re.sub(r"[^0-9.]", "", latest_version))
```
tromai, line 165: With this regrex expression. Would it be an expected case when the version string is `1snapshot-3.1.4` would be converted to `13.1.4` :thinking: . I think in this example, it's obvious that the version string is dirty anyway.

behnazh-w: > With this regrex expression. Would it be an expected case when the version string is `1snapshot-3.1.4` would be converted to `13.1.4` thinking . I think in this example, it's obvious that the version string is dirty anyway.

That's not a valid semver. `major`, `minor` and `patch` parts should only contain digits according to the [spec](https://semver.org/) and only `pre-release` and `build` can contain ASCII alphanumerics and hyphens `[0-9A-Za-z-]` .

That said, my regex would consider `1.1.0` lower than `1.1.0-alpha1` which is not correct according to the spec.

One challenge is that developers do not always follow semver correctly, so the check here could easily fail and we might not pick the bigger version. I'll try to find a balance here and come up with something more accurate.

tromai: I agree.  Another example pop up in my mind is the usage of `rc` in the semver (e.g. v1.3.4-rc2).

behnazh-w: I've reverted the change, see commit 4b66c7e0997982621722f81ef967b5f68b91f449

 Turns out `version.Version` does compare semvers correctly and reports `1.1.0` > `1.1.0-alpha1`. So if the version string follows semver correctly, the current implementation is good enough.

tromai (approved):

PR #105, opened 2023-03-22
feat: support an existing SBOM as input

### LINKED ISSUE #95: Accept an SBOM as the input for Macaron
**PROBLEM**
Up until now, Macaron only accepts the repo URL of the main analysis target (e.g. `--rp https://github.com/micronaut-projects/micronaut-core`) as the entrypoint. This behavior, however, comes as a limitation when we want to support running the analysis against a generated SBOM of the analysis target. 

**SOLUTION**
TBD

tromai:
Closes #95 .

Tasks:

- [x] Add the parameters to accept the SBOM path.
- [x] Extract the dependencies from the SBOM content and run the analysis.
- [x] Add tests for this feature.
- [x] Update README.md


src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -234,21 +237,22 @@ def generate_reports(self, report: Report) -> None:
...
             reporter.generate(output_target_path, report)
 
-    def resolve_dependencies(self, main_ctx: AnalyzeContext) -> dict[str, DependencyInfo]:
+    def resolve_dependencies(self, main_ctx: AnalyzeContext, sbom_path: str) -> dict[str, DependencyInfo]:
         """Resolve the dependencies of the main target repo.
 
         Parameters
         ----------
         main_ctx : AnalyzeContext
             The context of object of the target repository.
```
behnazh-w, line 247: No need for newline here.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -268,6 +272,24 @@ def resolve_dependencies(self, main_ctx: AnalyzeContext) -> dict[str, Dependency
...
         # Start resolving dependencies.
+        deps_resolved: dict[str, DependencyInfo] = (
+            self._get_deps_from_sbom(sbom_path)
+            if sbom_path
+            else self._get_deps_from_dep_analyzer(main_ctx, dep_analyzer, build_tool.get_build_dirs(main_ctx.repo_path))
+        )
+
+        return deps_resolved
+
+    def _get_deps_from_sbom(self, sbom_path: str) -> dict[str, DependencyInfo]:
```
behnazh-w, line 283: Please add a unit test for this function.

tests/e2e/expected_results/maven-with-sbom/JavaHamcrest.json
behnazh-w, line 1: This seems to be more suitable to be added as a dependency analyzer test and not e2e test?

src/macaron/__main__.py
```diff
@@ -71,7 +71,9 @@ def analyze_slsa_levels_single(analyzer_single_args: argparse.Namespace) -> None
         # Get user config from yaml file
         run_config = YamlLoader.load(analyzer_single_args.config_path)
 
-    status_code = analyzer.run(run_config, analyzer_single_args.skip_deps)
+    # TODO(tromai): Support running the analysis with the SBOM without having to provide the repo path
```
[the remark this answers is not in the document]
tromai: The issue has been created here - https://github.com/oracle-samples/macaron/issues/108. I will remove this TODO line.

tests/e2e/expected_results/maven-with-sbom/JavaHamcrest.json
tromai: I agree that it's more related to the dependency analyzer tests. Do you think that we only need to check the `dependencies.json` file only in this case? :thinking:

src/macaron/slsa_analyzer/analyzer.py
tromai: Ah I see. I think it is better if I explain my design regarding these two functions. This comment also addresses other  [feedback](https://github.com/oracle-samples/macaron/pull/105/files#r1144380041):

- The reason I put these two functions as private functions because they _only_ exist at the moment to support extracting the dependencies from the SBOM if it is given by the user. With regard to https://github.com/oracle-samples/macaron/issues/108, these two will definitely be updated to better handle the SBOM and the analysis of Macaron. Therefore, I think it makes sense to have them as private functions because they are very likely to change. 
- From those reasons, it's also makes sense to only have integration tests to capture the behaviors that we want for this feature instead of having a unit tests to cover something that is likely to change (which I think will save us from rewriting that unit test). In additions, the `_get_deps_from_sbom` function only reuses existing APIs (`get_dep_components` and `convert_components_to_artifacts` which already have unit tests for them). 

Please let me know your thoughts on this Behnaz.

behnazh-w: > The reason I put these two functions as private functions because they only exist at the moment to support extracting the dependencies from the SBOM if it is given by the user. With regard to https://github.com/oracle-samples/macaron/issues/108, these two will definitely be updated to better handle the SBOM and the analysis of Macaron.

`get_deps_from_sbom` should be a fairly robust function and callable from external modules if necessary. That's independent of #108 where we need to change/add new workflows entirely to avoid requiring a target repo. That's why a unit test is necessary to make sure we can get final dependencies from this function. 

>the _get_deps_from_sbom function only reuses existing APIs (get_dep_components and convert_components_to_artifacts which already have unit tests for them).

That's fine. The goal here is to compare the final dependencies.

tests/e2e/expected_results/maven-with-sbom/JavaHamcrest.json
behnazh-w: > I agree that it's more related to the dependency analyzer tests. Do you think that we only need to check the `dependencies.json` file only in this case? thinking

Yes, `dependencies.json` would be good enough.

src/macaron/slsa_analyzer/analyzer.py
behnazh-w: >In this case, do you think it's better if we put this function inside the cyclonedx.py file to be an API instead ? 

Yes, that would make sense. It would be just a simple function like this
```
return convert_components_to_artifacts(get_dep_components(Path(sbom_path)))
```

tromai: > As for the function _get_deps_from_dep_analyzer below, it originally exists inside the public function Analyzer.resolve_dependencies() . I extract it out to prevent the code from clustering too much. So should it be a private function because it should be reached only from calling Analyzer.resolve_dependencies()? thinking Or it makes more sense to make it public.

After having a closer look, I think that `_get_deps_from_dep_analyzer` is not needed. We only need to add this block to the beginning of `Analyzer.resolve_dependencies()` function without changing else.
```
if sbom_path:
            logger.info("Getting the dependencies from the SBOM defined at %s.", sbom_path)
            return get_deps_from_sbom(sbom_path)
```

tests/e2e/expected_results/maven-with-sbom/JavaHamcrest.json
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/105/commits/bffab4f77cdaa79f7c7bfd91869a5a5d7a9515fc

src/macaron/slsa_analyzer/analyzer.py
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/105/commits/bffab4f77cdaa79f7c7bfd91869a5a5d7a9515fc

src/macaron/slsa_analyzer/analyzer.py
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/105/commits/e9d7a5f7cfc3cccfa399875c89958bebe940d4ce

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -268,6 +272,24 @@ def resolve_dependencies(self, main_ctx: AnalyzeContext) -> dict[str, Dependency
...
+        """Get the dependencies from the provided SBOM."""
+        logger.info("Getting the dependencies from the SBOM defined at %s.", sbom_path)
+        deps_components = get_dep_components(Path(sbom_path))
+        return convert_components_to_artifacts(deps_components)
+
+    def _get_deps_from_dep_analyzer(
+        self, main_ctx: AnalyzeContext, dep_analyzer: DependencyAnalyzer, working_dirs: Iterable[Path]
+    ) -> dict[str, DependencyInfo]:
+        """Get the dependencies by running the Dependency Analyzer for the target repo."""
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/105/commits/e9d7a5f7cfc3cccfa399875c89958bebe940d4ce

src/macaron/__main__.py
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/105/commits/e9d7a5f7cfc3cccfa399875c89958bebe940d4ce

scripts/dev_scripts/integration_tests.sh
```diff
@@ -74,6 +74,17 @@ python $COMPARE_DEPS $DEP_RESULT $DEP_EXPECTED || log_fail
 
 python $COMPARE_JSON_OUT $JSON_RESULT $JSON_EXPECTED || log_fail
 
+echo -e "\n----------------------------------------------------------------------------------"
+echo "apache/maven: Analyzing the repo path, the branch name and the commit digest with dependency resolution using a CycloneDx SBOM."
+echo -e "----------------------------------------------------------------------------------\n"
+SBOM_FILE=$WORKSPACE/tests/e2e/resources/apache_maven_root_sbom.json
```
behnazh-w, line 80: Please move this file to `tests/dependency_analyzer/resources`.

README.md
```diff
@@ -62,14 +62,15 @@ The GitHub token should be stored in an environment variable and supplied to Mac
...
+python -m macaron [-v] analyze -rp <path_to_target_repo> [-b <branch_name>] [-d <digest>] -c <config_path> [-sbom <sbom_path>]
 ```
 
 The main input parameters of the `analyze` command:
 - `-rp <path_to_target_repo>` specifies the path to the repository to analyze. This path can be both a local path (e.g. `/path/to/repo`) or a remote path (e.g. `git@github.com:organization/repo_name.git` or `https://github.com/organization/repo_name.git`).
 - `[-b <branch_name>]`: The name of the branch to analyze. If not specified, Macaron will checkout the default branch (if a remote path is supplied) or use the current branch that HEAD is on (if a local path is supplied).
 - `[-d <digest>]`: The hash of the commit to checkout in the current branch. This hash must be in full form. If not specified, Macaron will checkout the latest commit of the current branch.
 - `-c <config_path>`: The path to the configuration yaml file. This option cannot be used together with the `-rp` option.
+- `[-sbom <sbom_path>]`: The path to the CycloneDX format SBOM for analyzing the dependencies of the target repo.
```
behnazh-w, line 73:
```suggestion
- `[-sbom <sbom_path>]`: The path to the CyclondeDX SBOM of the target repo. 
```

README.md
```diff
@@ -97,6 +98,7 @@ The results of the examples above will be stored in ``output/reports/github_com/
 - Macaron automatically detects and analyzes direct dependencies for Java Maven projects. This process might take a while during the first run but will be faster during the subsequent runs. To skip analyzing the dependencies you can pass ``--skip-deps`` option.
 - If you supply a remote path, the repository is cloned to `git_repos/` before the analysis starts. If the repository has already been cloned to `git_repos/`, Macaron will not clone the repository and proceed to analyze the local repo instead.
 - The `branch` and `digest` in the config file or `-b` and `-d` in the CLI are all optional and can be omitted.
+- At the moment, if an SBOM is provided with `-sbom`, Macaron uses the dependency data within the SBOM instead of running the dependency plugins against the target repo.
```
behnazh-w, line 101:
```suggestion
- If an SBOM is provided via `--sbom-path` option, Macaron will not detect the dependencies automatically. Note that `--skip-deps` option disables dependency analysis even if an SBOM file is provided.
```

scripts/dev_scripts/integration_tests.sh
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/105/commits/1b3677a7093847f0462b857965bf96bda0b8a4a0

behnazh-w (approved):

PR #155, opened 2023-04-12
feat: add repo finding via scm metadata in artefact poms

benmss:
This adds the first part of the repo finding extension, which will search for repositories for artefacts that do not already have one associated with them. This is done by retrieving the related POM for these artefacts from the configured location (e.g. Maven central), and extracting the contained SCM URL metadata, if any. Found URLs allow for immediate further evaluation.
By default, the metadata must be located at:
```
<project>
   ...
   <scm>
       <url>HERE</url>
   </scm>
   ...
</project>
```


src/macaron/dependency_analyzer/repo_finder.py
```diff
@@ -0,0 +1,185 @@
...
+    for request_url in request_urls:
+        pom = retrieve_pom(session, request_url)
+        if pom != "":
+            break
+
+    if pom == "":
+        return []
+
+    # Parse XML data and return URL
+    return parse_pom(pom, tags)
```
tromai, line 185: This `parse_pom` does not remove the trailing spaces or new lines. So if a tag in the xml file is 
```
<url>
         https://example.example/example
</url>
```
The result would be something similar to 
```
['\n         https://example.example/example\n        ']
```
In this case, I think the returned URL might be incorrect.

src/macaron/dependency_analyzer/repo_finder.py
```diff
@@ -0,0 +1,185 @@
...
+    urls = []
+
+    version_pruned = ""
+    if "-" in version:
+        # Create a pruned version to fix some strangely labeled artifacts
+        version_pruned = re.sub("-[a-zA-Z]+", "", version)
+
+    # Create URLs using all configured repositories
+    for repo in global_config.artifact_repositories:
+        urls.append(f"{repo}/{group}/{artifact}/{version}/{artifact}-{version}.pom")
```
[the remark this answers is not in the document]
benmss: Yes. The detection method for strange versions isn't robust, and really only exists because I found one case where it was needed to succeed. I would say there is a still a chance the non pruned version is correct when the pruning is triggered.

src/macaron/dependency_analyzer/repo_finder.py
```diff
@@ -0,0 +1,185 @@
...
+        The POM as a string
+    tags : list[str]
+        The list of tags to try extracting from the POM
+
+    Returns
+    -------
+    list[str] :
+        The extracted contents of any matches tags
+    """
+    xml = fromstring(pom)
```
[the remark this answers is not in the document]
benmss: I think that's a good idea. I'll add it.

src/macaron/dependency_analyzer/repo_finder.py
benmss: Good catch. I'll fix that.

pyproject.toml
```diff
@@ -28,6 +28,7 @@ dependencies = [
     "packaging >=21.3,<24.0.0",
     "jinja2 >=3.1.2,<4.0.0",
     "SQLAlchemy >=2.0.0,<3.0.0",
+    "defusedxml >=0.7.1,<1.0.0"
```
behnazh-w, line 31: Please add `,` at the end of the line. That helps keeping git diff lines relevant in future.

pyproject.toml
```diff
@@ -186,6 +187,7 @@ module = [
     "pydriller.*",
     "gitdb.*",
     "yamale.*",
+    "defusedxml.*"
```
behnazh-w, line 190: Same comment here.

src/macaron/__main__.py
```diff
@@ -167,10 +167,18 @@ def main() -> None:
         "--policy",
         required=False,
         default=[],
-        help=("The path to a policy yaml file or directory."),
+        help="The path to a policy yaml file or directory.",
         action="append",
     )
 
+    main_parser.add_argument(
```
behnazh-w, line 174: Looks like the list of repositories can be provided both through args and `defaults.ini`. That would require consolidating the values if provided in both ways. To avoid these extra checking, I think it's better to just read it through `defaults.ini`.

tests/dependency_analyzer/repo_finder/resources/example_pom.xml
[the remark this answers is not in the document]
benmss: This should already be included as the meta data is inserted into the tags by the parsing library. E.g. `url` becomes `{http://maven.apache.org/POM/4.0.0}url`

src/macaron/dependency_analyzer/dependency_resolver.py
```diff
@@ -129,6 +130,16 @@ def add_latest_version(
         url_to_artifact: dict[str, set]
             Used to detect artifacts that have similar repos.
         """
+        if item["url"] == "" and item["version"] != "unspecified":
+            gav = f"{item['group']}:{item['name']}:{item['version']}"
```
behnazh-w, line 134: If `item['group']` or `item['name']` is empty, would the `gav` be valid?

src/macaron/dependency_analyzer/dependency_resolver.py
```diff
@@ -129,6 +130,16 @@ def add_latest_version(
...
+            urls = find_repo(gav, ["scm.url"])
+            if len(urls) != 0:
```
behnazh-w, line 136:
```suggestion
            if urls:
```
In Python this if condition would be false if the list is empty.

src/macaron/dependency_analyzer/repo_finder.py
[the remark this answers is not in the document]
behnazh-w: Do you remember the case where this was needed?

src/macaron/dependency_analyzer/repo_finder.py
behnazh-w, line 1: This module is mainly designed for Java projects. How about renaming it to `java_repo_finder.py`?

src/macaron/dependency_analyzer/repo_finder.py
```diff
@@ -0,0 +1,185 @@
...
+        The list of URLs created for the GAV by the create_urls trailing function call.
+    """
+    split = gav.split(":")
+    if len(split) < 3:
+        logger.error("Could not parse GAV: %s", gav)
+        return []
+
+    group = split[0]
+    artifact = split[1]
+    version = split[2]
```
behnazh-w, line 73:
```suggestion
    group, artifact, version = split
```

src/macaron/dependency_analyzer/repo_finder.py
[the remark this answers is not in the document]
benmss: I checked again and it seems none of the versions that were pruned led to a successful POM retrieval. I'll remove this.

src/macaron/dependency_analyzer/dependency_resolver.py
benmss: It would not. I've added additional checks.

src/macaron/dependency_analyzer/repo_finder.py
```diff
@@ -0,0 +1,185 @@
...
+def _find_element(parent: typing.Any, target: str) -> typing.Any:
+    # Attempt to match the target tag within the children of parent.
+    for child in parent:
+        # Account for raw tags and tags accompanied by Maven metadata enclosed in curly braces. E.g. '{metadata}tag'
+        if child.tag == target or child.tag.endswith("}" + target):
+            return child
+    return None
+
+
+def parse_pom(pom: str, tags: list[str]) -> list[str]:
```
behnazh-w, line 114: Can you please use an iterator instead of creating a list as the return value to improve the performance?

src/macaron/dependency_analyzer/repo_finder.py
```diff
@@ -0,0 +1,185 @@
...
+
+    Returns
+    -------
+    list[str] :
+        The extracted contents of any matches tags
+    """
+    xml = fromstring(pom)
+    results = []
+
+    # Try to match each tag with the contents of the POM
```
behnazh-w, line 133: nit: please add `.` at the end of comment sentences.

src/macaron/dependency_analyzer/repo_finder.py
```diff
@@ -0,0 +1,185 @@
...
+        An artifact represented as a GAV (group, artifact, version) in the format: G:A:V.
+    tags : list[str]
+        The list of XML tags to look for, each in the format: tag1[.tag2 ... .tagN]
+
+    Returns
+    -------
+    list[str] :
+        The URLs found for the passed GAV.
+    """
+    if len(global_config.artifact_repositories) == 0:
```
behnazh-w, line 164: No need to check the length.

src/macaron/dependency_analyzer/repo_finder.py
```diff
@@ -0,0 +1,185 @@
...
+        logger.warning("No repositories set in config artifact_repositories parameter for repo finder.")
+        return []
+
+    # Parse the GAV and create the URLs for its POM
+    request_urls = parse_gav(gav)
+    if len(request_urls) == 0:
```
behnazh-w, line 170: Same comment about length check.

tests/dependency_analyzer/repo_finder/test_repo_finder.py
```diff
@@ -0,0 +1,29 @@
...
+"""
+import os
+from pathlib import Path
+
+from macaron.config.global_config import global_config
+from macaron.dependency_analyzer.repo_finder import parse_gav, parse_pom
+from tests.macaron_testcase import MacaronTestCase
+
+
+class TestDependencyAnalyzer(MacaronTestCase):
```
behnazh-w, line 15: We have stopped using `MacaronTestCase` and custom classes for testing as `pytest` doesn't support it. You should be able to add tests as functions without using a class.

src/macaron/dependency_analyzer/dependency_resolver.py
```diff
@@ -130,13 +130,14 @@ def add_latest_version(
         url_to_artifact: dict[str, set]
             Used to detect artifacts that have similar repos.
         """
-        if item["url"] == "" and item["version"] != "unspecified":
+        if item["url"] == "" and item["version"] != "unspecified" and item["group"] and item["name"]:
             gav = f"{item['group']}:{item['name']}:{item['version']}"
             urls = find_repo(gav, ["scm.url"])
-            if len(urls) != 0:
+            # TODO decide how to handle multiple returned URLs
+            url = next((item for item in urls), "")
```
tromai, line 137: Because `urls` is already an `Iterator` instance, we could use it directly in `next()`: `url = next(urls, "")`
Reference: https://docs.python.org/3/library/functions.html#next

src/macaron/dependency_analyzer/dependency_resolver.py
```diff
@@ -129,6 +130,17 @@ def add_latest_version(
         url_to_artifact: dict[str, set]
             Used to detect artifacts that have similar repos.
         """
+        if item["url"] == "" and item["version"] != "unspecified" and item["group"] and item["name"]:
+            gav = f"{item['group']}:{item['name']}:{item['version']}"
+            urls = find_repo(gav, ["scm.url"])
```
tromai, line 135: Note that this method is performing a connection. This `add_latest_version` is called indirectly via this line in the test - https://github.com/oracle-samples/macaron/blob/1906f88c3d725ef01433301031b1bec0ea7829c6/tests/dependency_analyzer/cyclonedx/test_cyclonedx.py#L71
This is the call hierachy:
```
test_convert_components_to_artifacts -> convert_components_to_artifacts -> add_latest_version -> find_repo
```
Therefore, our unit tests still use network connection.

src/macaron/dependency_analyzer/java_repo_finder.py
```diff
@@ -0,0 +1,194 @@
...
+        logger.warning("Failed to retrieve pom from: %s, error code: %s", url, res.status_code)
+        return ""
+
+    logger.debug("Found artifact POM at: %s", url)
+
+    return res.text
+
+
+def _find_element(parent: typing.Optional[Element], target: str) -> typing.Optional[Element]:
+    if parent is None:
```
tromai, line 104:
```suggestion
    if not parent:
```

src/macaron/dependency_analyzer/java_repo_finder.py
```diff
@@ -0,0 +1,194 @@
...
+
+    results = []
+
+    # Try to match each tag with the contents of the POM.
+    for tag in tags:
+        element: typing.Optional[Element] = xml
+        tag_parts = tag.split(".")
+        for index, tag_part in enumerate(tag_parts):
+            element = _find_element(element, tag_part)
+            if element is None:
```
tromai, line 145:
```suggestion
            if not element:
```

src/macaron/dependency_analyzer/java_repo_finder.py
```diff
@@ -0,0 +1,194 @@
...
+                break
+            if index == len(tag_parts) - 1 and element.text is not None:
```
tromai, line 147:
```suggestion
            if index == len(tag_parts) - 1 and element.text:
```

src/macaron/dependency_analyzer/dependency_resolver.py
benmss: Good point. Is this something we want to change?

behnazh-w: How about using a flag (which by default is on) to find repos  via contacting the remote server? That way we could control making network calls.

src/macaron/dependency_analyzer/dependency_resolver.py
```diff
@@ -129,6 +130,17 @@ def add_latest_version(
...
+            # TODO decide how to handle multiple returned URLs
```
behnazh-w, line 136: Before picking a URL, we need to validate and check the URL, which happens in [find_valid_url](https://github.com/oracle-samples/macaron/blob/1906f88c3d725ef01433301031b1bec0ea7829c6/src/macaron/dependency_analyzer/dependency_resolver.py#L182). 

So you can just use this function to select a URL. Here is an [example](https://github.com/oracle-samples/macaron/blob/1906f88c3d725ef01433301031b1bec0ea7829c6/src/macaron/dependency_analyzer/cyclonedx.py#L194-L196).

I realized that while checking the changes in the expected outputs in `cyclonedx_apache_maven.json` that contained `http://svn.apache.org/repos/asf/commons/proper/jxpath/trunk` which we don't support as an scm.

src/macaron/dependency_analyzer/java_repo_finder.py
benmss: Interestingly, it turns out these are not equivalent in this case, as using `if not element` leads to different results.

behnazh-w: We shouldn't replace ` if element is None` with `if not element` if we need to check that `element` is `None`. Here if `element` is an empty list, `if not element` would still be `False`, even though it's not `None. See https://docs.python.org/3/library/stdtypes.html#truth-value-testing.

tromai: Hmmm. I see that `_find_element` returned `Optional[Element]` so checking it with `if not element` would be okay. However, because `Element` implement its `__bool__` :
```python
    def __bool__(self):
        warnings.warn(
            "The behavior of this method will change in future versions.  "
            "Use specific 'len(elem)' or 'elem is not None' test instead.",
            FutureWarning, stacklevel=2
            )
        return len(self._children) != 0 # emulate old behaviour, for now
```
This won't return the behavior that I was expecting (if the method return an `Element` instance but without any children `not element -> True` . I think we shouldn't change it to `if not element`).

src/macaron/config/defaults.ini
```diff
@@ -43,6 +43,12 @@ dep_tool_gradle = cyclonedx-gradle:1.7.4
 timeout = 2400
 recursive = False
 
+# This is the repo finder script.
+[repofinder]
```
behnazh-w, line 47: Since this is only going to target Java projects, can you please rename it to `repofiner.java` and adjust the comment above?

src/macaron/__main__.py
```diff
@@ -242,6 +242,15 @@ def main() -> None:
         help=("The path to the Jinja2 html template (please make sure to use .html or .j2 extensions)."),
     )
 
+    single_analyze_parser.add_argument(
```
behnazh-w, line 245: Are we still using this cmd arg or should we remove it?

benmss: It should have been removed.

behnazh-w (approved):

tromai (approved):

PR #159, opened 2023-04-13
chore(deps): pin Docker images with hash

behnazh-w:


docker/Dockerfile.final
```diff
@@ -10,7 +10,7 @@
 # Note that the local machine must login to ghcr.io so that Docker could pull the ghcr.io/oracle-samples/macaron-base
 # image for this build.
 
-FROM ghcr.io/oracle-samples/macaron-base
+FROM ghcr.io/oracle-samples/macaron-base:latest@sha256:b4b63d3be26fc00be06389b48d0bf49450cac402d64692d354f7ebdd40bb8828
```
tromai, line 13: I am still a bit concern about hash pinning the `macaron-base` image as the `macaron-base` is more likely to change than the base `container-registry.oracle.com/os/oraclelinux:9-slim`. I just want to point out that we need to update this hash every time we build a newer `macaron-base` image :sweat_smile:

behnazh-w: > I am still a bit concern about hash pinning the `macaron-base` image as the `macaron-base` is more likely to change than the base `container-registry.oracle.com/os/oraclelinux:9-slim`. I just want to point out that we need to update this hash every time we build a newer `macaron-base` image sweat_smile

I don't expect regular updates to the base image being frequent. If that's the case we need the change our workflows anyway. Also, it would be a good idea to thinking about tagging the base image more properly instead of using `latest` (a separate issue though).

tromai: > it would be a good idea to thinking about tagging the base image more properly instead of using latest
I wonder if we need to specify the tag when we have already pinned the sha checksum of the image ([references](https://docs.docker.com/engine/reference/commandline/pull/#/pull-an-image-by-digest-immutable-identifier))? I believe something like this:
```
ghcr.io/oracle-samples/macaron-base@sha256:b4b63d3be26fc00be06389b48d0bf49450cac402d64692d354f7ebdd40bb8828
```
would work.

behnazh-w: `ghcr.io/oracle-samples/macaron-base@sha256:b4b63d3be26fc00be06389b48d0bf49450cac402d64692d354f7ebdd40bb8828` would be using the `latest` tag anyway. So at least we are explicit here.

tromai (approved):

PR #164, opened 2023-04-17
fix: do not use download script for Syft

behnazh-w:
Based on Scorecard report we should avoid `downloadThenRun` patterns because we won't have a chance to review the download script before running it.

Also see this [discussion](https://serverfault.com/questions/226386/wget-a-script-and-run-it/890417).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -153,14 +153,12 @@ jobs:
           "https://github.com/anchore/syft/releases/download/v${SYFT_VERSION}/${CHECKSUMS}"
         test -s "$CHECKSUMS" || (echo "Unable to download $CHECKSUMS" && exit 0)
 
-        tar -zxvf "$ASSET_NAME" -C "$SYFT_BIN" syft
         EXPECTED=$(grep "${ASSET_NAME}" "${CHECKSUMS}" 2>/dev/null | tr '\t' ' ' | cut -d ' ' -f 1)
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -139,11 +139,35 @@ jobs:
...
+        test -s "$CHECKSUMS" || (echo "Unable to download $CHECKSUMS" && exit 0)
+
+        tar -zxvf "$ASSET_NAME" -C "$SYFT_BIN" syft
+        EXPECTED=$(grep "${ASSET_NAME}" "${CHECKSUMS}" 2>/dev/null | tr '\t' ' ' | cut -d ' ' -f 1)
+        SYFT_DIGEST=$(sha256sum "$ASSET_NAME" | cut -d ' ' -f 1)
+        rm -f "$ASSET_NAME"
+        rm -f "$CHECKSUMS"
+
+        # Check if artifact is valid.
+        if [ "$EXPECTED" == "$SYFT_DIGEST" ]; then
```
tromai, line 163: Should we perform this checking before extracting the tarball as we haven't trusted it yet?

behnazh-w: > Should we perform this checking before extracting the tarball as we haven't trusted it yet?

Goo point. See commit [46141d6](https://github.com/oracle-samples/macaron/pull/164/commits/46141d6701de9558c147b5371982092f8d1a3392).

tromai (approved):

PR #165, opened 2023-04-17
feat: support running the analysis with SBOM and the main software component with no repository

### LINKED ISSUE #157: Treat the main target as empty if it's not provided together with the SBOM.
**Description**
We allow the user to run the analysis for the dependencies listed in the SBOM file (provided via `-sbom`) but the main target software component doesn't have a repository available.
For example:
```bash
macaron analyze -sbom <path_to_sbom> -purl <no_repo_component_purl> 
```

tromai:
Closes #157 

# Feature Description

We want to allow the use case of: running the analysis on the dependencies listed in an SBOM generated for an artifact while the repository URL for that artifact is not available (e.g. the repository is in a private domain that Macaron cannot access).
Example usage:
```shell
macaron analyze -purl pkg:domain.com/org/name -sbom <path_to_sbom>
```

## Current state
- When Macaron cannot resolve the repository for the main software component, Macaron **will not** collect any dependencies and analyze them. This is because the implementation for both collecting dependencies from a provided SBOM (through `--sbom-path`) and collecting dependencies by running the SBOM generator are bundled in the same location (see [here](https://github.com/oracle/macaron/blob/71accbf81e849947ca238ebfa3a9a6ef88fd01f6/src/macaron/slsa_analyzer/analyzer.py#L256)) and this method is only run if the repository for the main software component is found (see [here](https://github.com/oracle/macaron/blob/71accbf81e849947ca238ebfa3a9a6ef88fd01f6/src/macaron/slsa_analyzer/analyzer.py#L158-L168)).
- For all software components, when Macaron cannot resolve the repository for it, the analysis will be skipped for it.
  - For dependencies, the skipping happens [here](https://github.com/oracle/macaron/blob/71accbf81e849947ca238ebfa3a9a6ef88fd01f6/src/macaron/slsa_analyzer/analyzer.py#L182-L191).
  - For the main software component, the skipping happens [here](https://github.com/oracle/macaron/blob/71accbf81e849947ca238ebfa3a9a6ef88fd01f6/src/macaron/slsa_analyzer/analyzer.py#L160-L162).

### Core engine
- For all software component, the analysis will not be skipped if the repository is not found. 
- When Macaron cannot resolve main software component's repository **and** a PURL string is provided by the user (via `-purl`), it will still collect the dependencies from the SBOM and run the analysis for those dependencies.
  - If Macaron cannot resolve the repository and no PURL is provided, it will exit on error (this has been already handle in #401 ). 
### HTML reports
- When the repository is not found for a software component, display a small message in the `Target Information` section to notify the users.
- When all the checks are failing for a software component, we collapse the check report table. This is because we have now only repository-based checks. Therefore, when the repository is not found for a software component, no check will pass. We hide the table for visibility.

## Example report
image


tromai:
This rebase is to get the changes for https://github.com/oracle-samples/macaron/pull/125

src/macaron/__main__.py
```diff
@@ -81,6 +81,11 @@ def analyze_slsa_levels_single(analyzer_single_args: argparse.Namespace) -> None
         # Get user config from yaml file
         run_config = YamlLoader.load(analyzer_single_args.config_path)
 
+    elif not analyzer_single_args.sbom_path:
+        logger.critical("No target provided for the analysis.")
+        logger.info("Run <macaron analyze --help> for help messages.")
+        sys.exit(1)
```
tromai, line 87: Change to `sys.exit(os.EX_USAGE)`

src/macaron/output_reporter/results.py
```diff
@@ -270,8 +280,15 @@ def find_ctx(self, record_id: str) -> AnalyzeContext | None:
...
         """Return the string representation of the Report instance."""
-        ctx_list = list(self.get_ctxs())
-        main_ctx: AnalyzeContext = ctx_list.pop(0)
+        main_ctx = self.root_record.context
+
+        if not main_ctx:
+            return "".join(
+                [
+                    "\nANALYSIS RESULT:\n\n",
+                    "\nThere is no information to display for the main target.\n\n",
```
tromai, line 289:
```suggestion
                    "\nThere is no information to display for this target.\n\n",
```

scripts/dev_scripts/integration_tests.sh
```diff
@@ -133,6 +133,18 @@ $RUN_MACARON analyze -rp https://github.com/apache/maven -b master -d 6767f2500f
...
 
+
+echo -e "\n----------------------------------------------------------------------------------"
+echo "apache/maven: Analyzing using a CycloneDx SBOM with PURL."
+echo -e "----------------------------------------------------------------------------------\n"
+SBOM_FILE=$WORKSPACE/tests/dependency_analyzer/cyclonedx/resources/apache_maven_root_sbom_minimal.json
+DEP_EXPECTED=$WORKSPACE/tests/dependency_analyzer/expected_results/apache_maven_with_sbom_minimal.json
+DEP_RESULT=$WORKSPACE/output/reports/maven/org_apache_maven/maven/dependencies.json
+
+$RUN_MACARON analyze -purl pkg:maven/org.apache.maven/maven -sbom "$SBOM_FILE" || log_fail
```
nicallen, line 144: Is this new test supposed to be testing the new use case of no main target repository? If that is the intention, then that should be made more clear. Also, isn't that purl a real, valid purl that in principle could be resolved to a repo (so there could be a main target repo found)?

src/macaron/output_reporter/templates/base_template.html
```diff
@@ -1,4 +1,4 @@
-<!-- Copyright (c) 2022 - 2022, Oracle and/or its affiliates. All rights reserved. -->
+<!-- Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved. -->
```
nicallen, line 1: Copyright notice should keep the start year, i.e. should be`2022 - 2023`.

src/macaron/output_reporter/templates/base_template.html
```diff
@@ -255,26 +255,30 @@
             padding-left: 2em
         }
 
-        .caret {
+        /*
+            The reason why we need to create a separate .togler class is because all .caret calss are set binded
```
nicallen, line 259: Typos:
`.togler` -> `.toggler`
`calss` -> `class`

nicallen (changes requested):

scripts/dev_scripts/integration_tests.sh
tromai: You are right, this test case is used for testing the new use case of no main target repository. I worked on this test case back then before the feature for mapping PURL string to repository path of the repo-finder is implemented. I will change this to explicitly indicate that this is a non-repo PURL.

tromai: Addressed in https://github.com/oracle/macaron/pull/165/commits/82ca751ab8b3e0ce79910ad413d05c4b8c16f07c

src/macaron/output_reporter/templates/base_template.html
tromai: Addressed in https://github.com/oracle/macaron/pull/165/commits/927bfe16e500d716e475ed7b04b33a7532fb98b8

src/macaron/output_reporter/templates/base_template.html
tromai: Fixed in https://github.com/oracle/macaron/pull/165/commits/3a18c96a78a3da6c6cea8d18a784cdaf0dd7bb9d

nicallen (approved):

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -170,17 +170,15 @@ def run(self, user_config: dict, sbom_path: str = "", skip_deps: bool = False) -
                     logger.info("Start analyzing the dependencies.")
                     for config in deps_config:
                         dep_status: SCMStatus = config.get_value("available")
-                        if dep_status != SCMStatus.AVAILABLE:
+                        if dep_status == SCMStatus.DUPLICATED_SCM:
```
behnazh-w, line 173: With this change, if a repository is missing, we will no longer report `MISSING REPO URL` and run the analysis anyway, reporting that result is available
image

While the HTML report for the dependency reports that the URL is missing
image

Is that intended?

tromai: Yes this is intended. I think I made it this way is because the data model allows a software component to have no repository. This would mean that if a software component doesn't have any repository available:
- All checks should still be run (even though they will all be failed because we only have repository-based checks for now).
- A [Component](https://github.com/oracle/macaron/blob/4282c6c6e5ffc080a094fec7622b6d07eedd8987/src/macaron/database/table_definitions.py#L120) instance is still created with its Repository field set to empty. 
Because we still create a software component and perform the analysis (even though all checks will fail) for the target software component (as agreed for this feature), I figure we should keep it consistent with the dependencies as well. Therefore, we don't skip dependency software component if they don't have a repository.

behnazh-w (approved):

PR #172, opened 2023-04-19
chore(deps): upgrade indirect Go dependencies

behnazh-w:


tromai (approved):

tromai:
I wonder why dependabot didn't notice and make any of these updates :thinking:

behnazh-w:
> I wonder why dependabot didn't notice and make any of these updates thinking

It doesn't seem to detect indirect dependencies.

PR #178, opened 2023-04-21
chore: turn off py warnings in prod

### LINKED ISSUE #177: Suppress Python warnings in production
We should not log [Python warnings](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONWARNINGS) in production.

behnazh-w:
Closes #177

### DIFF
--- a/scripts/release_scripts/run_macaron.sh
+++ b/scripts/release_scripts/run_macaron.sh
@@ -363,6 +363,10 @@ for v in "${proxy_var_names[@]}"; do
     [[ -n ${!v} ]] && proxy_vars+=("-e" "${v}=${!v}")
 done
 
+prod_vars=(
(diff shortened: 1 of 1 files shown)


scripts/release_scripts/run_macaron.sh
```diff
@@ -363,6 +363,10 @@ for v in "${proxy_var_names[@]}"; do
     [[ -n ${!v} ]] && proxy_vars+=("-e" "${v}=${!v}")
 done
 
+prod_var_names=(
+    "PYTHONWARNINGS=ignore"  # Turn off Python warnings in the production environment.
```
tromai, line 367: We are missing the `-e` flag here. It's because we are passing it the runtime container through `docker run`

tromai (approved):

PR #179, opened 2023-04-23
feat: multi build tool detection

timyarkov:
Basic multiple build tool detection (essentially converting the tool part of BuildSpec to a list)

TODO before final PR:

- Double triple check nothing was missed
- Rebase to staging (+squash commits)
- Update unit tests to multi-build tool
- Do general system testing to see if everything works as expected (did very quick test and at the very least it doesn't crash when ran on https://github.com/timyarkov/GuardianEmailClient)

### DIFF
--- a/src/macaron/slsa_analyzer/analyzer.py
+++ b/src/macaron/slsa_analyzer/analyzer.py
@@ -32,7 +32,6 @@ from macaron.output_reporter.results import Record, Report, SCMStatus
 from macaron.slsa_analyzer import git_url
 from macaron.slsa_analyzer.analyze_context import AnalyzeContext
 from macaron.slsa_analyzer.build_tool import BUILD_TOOLS
-from macaron.slsa_analyzer.build_tool.base_build_tool import NoneBuildTool
 
 # To load all checks into the registry
 from macaron.slsa_analyzer.checks import *  # pylint: disable=wildcard-import,unused-wildcard-import # noqa: F401,F403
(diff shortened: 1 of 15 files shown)


src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -258,71 +258,78 @@ def resolve_dependencies(self, main_ctx: AnalyzeContext, sbom_path: str) -> dict
 
         deps_resolved: dict[str, DependencyInfo] = {}
 
-        build_tool = main_ctx.dynamic_data["build_spec"]["tool"]
-        if not build_tool or isinstance(build_tool, NoneBuildTool):
-            logger.info("Unable to find a valid build tool.")
+        build_tools = main_ctx.dynamic_data["build_spec"]["tools"]
+        if len(build_tools) == 0:
```
behnazh-w, line 262: You can just use `if build_tools` in Python to check if a list is empty or not, no need to check the length.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -655,12 +662,12 @@ def perform_checks(self, analyze_ctx: AnalyzeContext) -> dict[str, CheckResult]:
 
             if build_tool.is_detected(analyze_ctx.git_obj.path):
                 logger.info("The repo uses %s build tool.", build_tool.name)
-                analyze_ctx.dynamic_data["build_spec"]["tool"] = build_tool
+                analyze_ctx.dynamic_data["build_spec"]["tools"].append(build_tool)
                 break
 
-        if not analyze_ctx.dynamic_data["build_spec"].get("tool"):
+        if len(analyze_ctx.dynamic_data["build_spec"].get("tools")) == 0:
```
behnazh-w, line 668: Same comment about checking lists.

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -281,24 +290,61 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
...
+
+        Returns
+        -------
+        CheckResultType
+            The result type of the check (e.g. PASSED).
+        """
+        # Get the build tool identified by the mcn_version_control_system_1, which we depend on.
+        build_tools = ctx.dynamic_data["build_spec"].get("tools")
+
+        if len(build_tools) == 0:
```
behnazh-w, line 325: Same comment about lists.

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -281,24 +290,61 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
...
+            check_result["justification"].append(failed_msg)
             return CheckResultType.FAILED
 
-        check_result["result_tables"] = [BuildAsCodeTable()]
-        failed_msg = "The target repository does not have a build tool."
-        check_result["justification"].append(failed_msg)
-        return CheckResultType.FAILED
+        ci_services = ctx.dynamic_data["ci_services"]
+
+        # Check for each build tool if its built as code
```
behnazh-w, line 333:
```suggestion
Check if "build as code" holds for each build tool.
```

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -107,14 +108,25 @@ def _has_deploy_command(self, commands: list[list[str]], build_tool: BaseBuildTo
                         logger.info("Found deploy command %s.", str(com))
                         return str(com)
         return ""
-
-    def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResultType:
-        """Implement the check in this method.
+    
+    def _tool_check(
```
behnazh-w, line 112:
```suggestion
    def _check_build_tool(
```

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -107,14 +108,25 @@ def _has_deploy_command(self, commands: list[list[str]], build_tool: BaseBuildTo
...
+            self, 
+            build_tool: BaseBuildTool, 
+            ctx: AnalyzeContext, 
+            ci_services: list[CIInfo], 
+            check_result: CheckResult
+        ) -> CheckResultType | None:
+        """
+        Run the check for a single build tool.
```
behnazh-w, line 120: Can you add why we need to do that?

src/macaron/slsa_analyzer/checks/build_service_check.py
```diff
@@ -94,122 +95,144 @@ def _has_build_command(self, commands: list[list[str]], build_tool: BaseBuildToo
...
-                for bash_cmd in ci_info["bash_commands"]:
-                    build_cmd = self._has_build_command(bash_cmd["commands"], build_tool)
-                    if build_cmd:
-                        # Get the permalink and HTML hyperlink tag of the CI file that triggered the bash command.
-                        trigger_link = ci_service.api_client.get_file_link(
-                            ctx.repo_full_name,
-                            ctx.commit_sha,
-                            ci_service.api_client.get_relative_path_of_workflow(os.path.basename(bash_cmd["CI_path"])),
+    
+    def _build_tool_check(
```
behnazh-w, line 99:
```suggestion
    def _check_build_tool(
```

src/macaron/slsa_analyzer/checks/build_service_check.py
```diff
@@ -94,122 +95,144 @@ def _has_build_command(self, commands: list[list[str]], build_tool: BaseBuildToo
...
+            self, build_tool: BaseBuildTool, 
+            ctx: AnalyzeContext, 
+            check_result: CheckResult, 
+            ci_services: list[CIInfo]
+        ) -> CheckResultType | None:
```
behnazh-w, line 104: Add docstring.

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -280,24 +283,59 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
...
-        check_result["justification"].append(failed_msg)
-        return CheckResultType.FAILED
+        ci_services = ctx.dynamic_data["ci_services"]
+
+        # Check if "build as code" holds for each build tool.
+        for tool in build_tools:
+            res = self._check_build_tool(tool, ctx, ci_services, check_result)
+
+            # Divide up build tool outputs
+            check_result["justification"].append("\n")
```
tromai, line 331: I don't think we need to add a  "\n" character to divide up the outputs. The content of each element of `check_result["justification"]` is not used for rendering the reports. So if you add "\n" here, the string `"\n"` will appear instead of a new line.
Furthermore, in the HTML reports each element in `check_result["justification"]` is rendered as a separated bullet point in the justification column.

src/macaron/slsa_analyzer/checks/build_script_check.py
```diff
@@ -59,20 +58,22 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
...
+            return CheckResultType.FAILED
+
+        # Check if any build tools are discovered for this repo.
         # TODO: look for build commands in the bash scripts. Currently
-        # we parse bash scripts that are reachable through CI only.
-        if build_tool and not isinstance(build_tool, NoneBuildTool):
-            pass_msg = f"The target repository uses build tool {build_tool.name}."
+        #       we parse bash scripts that are reachable through CI only.
+        for tool in build_tools:
+            pass_msg = f"The target repository uses build tool {tool.name}."
```
tromai, line 72: I think we also need to check if `build_tool` is not `NoneBuildTool` here too.

src/macaron/slsa_analyzer/checks/build_service_check.py
```diff
@@ -95,121 +96,160 @@ def _has_build_command(self, commands: list[list[str]], build_tool: BaseBuildToo
...
-                    continue
-                for bash_cmd in ci_info["bash_commands"]:
-                    build_cmd = self._has_build_command(bash_cmd["commands"], build_tool)
-                    if build_cmd:
-                        # Get the permalink and HTML hyperlink tag of the CI file that triggered the bash command.
-                        trigger_link = ci_service.api_client.get_file_link(
-                            ctx.repo_full_name,
-                            ctx.commit_sha,
-                            ci_service.api_client.get_relative_path_of_workflow(os.path.basename(bash_cmd["CI_path"])),
+        for ci_info in ci_services:
```
tromai, line 121: I think we also need to check if `build_tool` is not `NoneBuildTool` here too.

src/macaron/slsa_analyzer/checks/build_script_check.py
timyarkov: To confirm you mean as in if there is a NoneBuildTool within the list we treat that as an instant fail?

tromai: @timyarkov Hmmm, you are right. We check for `NoneBuildTool` before because we only support 1 build tool and `ctx.dynamic_data["build_spec"]["tool"]` was originally initialized as an instance of `NoneBuildTool`. 
Your changes already set `ctx.dynamic_data["build_spec"]["tools"]` to be an empty list so this statement `represent a lack of any build tools, so wouldn't an empty list do just fine for that` holds. 
@behnazh-w Please correct me if I am wrong. I think we still need to check for NoneBuildTool here as a safeguard because printing out the result for `NoneBuildTool` wouldn't make any sense for the user :thinking: .

behnazh-w: Looks like `NoneBuildTool` is not instantiated anywhere in the code anymore. Is that correct @timyarkov ?

If that's true, if we remove `NoneBuildTool` class altogether, we won't need to to check for it here.

timyarkov: Yeah far as I see that's the case, I'll remove it then

src/macaron/resources/.gitignore
```diff
@@ -3,7 +3,7 @@
 
 # Ignore Gradle project-specific cache directory
 .gradle
-gradle-*
+.gradle-*
```
behnazh-w, line 6:
```suggestion
gradle-*
```

tests/e2e/expected_results/micronaut-core/caffeine.json
behnazh-w, line 1: These updates can be reverted as they don't change the expected results anymore.

Same comment for `micronaut-core.json` and `sl4j.sjon`.

behnazh-w (approved):

tromai (approved):

PR #191, opened 2023-04-27
chore: allow disabling checks 

### LINKED ISSUE #133: Allow disabling a check
- [ ] We need to allow a list of checks to be disabled from the CLI and handle it properly in the report.
- [ ] Add an option to disable all the default checks.

tromai:
Closes #133


tromai:
This force push is to revert an unnecessary commit -> It automatically close the PR because this branch is similar to staging. I will re open it again after I push new commit.

PR #192, opened 2023-04-27
refactor: run policy engine using macaron entrypoint

### LINKED ISSUE #189: Run policy engine using the main entry point script
- [x] Add `verify-policy` action to `macaron`.
- [x] Update tests.
- [x] Add the cmd argument to `run_macaron.sh` for Docker run.

behnazh-w:
Closes #189 

The documentation will be further enhanced in a separate [ticket](https://github.com/oracle-samples/macaron/issues/190).


src/macaron/__main__.py
```diff
@@ -217,6 +197,15 @@ def main() -> None:
         ),
     )
 
+    single_analyze_parser.add_argument(
+        "-pe",
+        "--provenance-expectation",
+        required=False,
+        default=[],
+        help=("The path to provenance expectation file or directory."),
+        action="append",
```
tromai, line 206: Note: without providing `nargs="*"`, the user couldn't do something like:
```
macaron ... --provenance-expectation boo foo bah
```
Would this behavior be as expected ?

src/macaron/policy_engine/policy_engine.py
```diff
@@ -112,3 +114,64 @@ def policy_engine(database_path: str, policy_file: str) -> dict:
...
+        The SQLite database file to evaluate the policy against
+    show_prelude: bool
+        Just show the policy prelude and exit.
+    policy_file: str
+        The policy file to evaluate
+
+    Returns
+    -------
+    dict
+        The policy engine result.
```
tromai, line 157: Can you please add some documentation on what are the key and value of the returned dictionary? I think it would be easier to understand what is the internal representation of this dictionary.

src/macaron/__main__.py
```diff
@@ -301,14 +290,15 @@ def main() -> None:
     logger.info("The logs will be stored in debug.log")
 
     # Set Macaron's global configuration.
+    # The path to provenance expectation files will be updated if
+    # set through analyze sun-command.
```
tromai, line 294:
```suggestion
    # set through analyze sub-command.
```
Unless it's meant to be `sun-command` :smile:

src/macaron/__main__.py
```diff
@@ -244,14 +233,14 @@ def main() -> None:
     )
 
     # Dump the default values.
-    sub_parser.add_parser(name="dump_defaults", description="Dumps the defaults.ini file to the output directory.")
+    sub_parser.add_parser(name="dump-defaults", description="Dumps the defaults.ini file to the output directory.")
 
-    # Verifying a provenance against a policy.
-    verify_parser = sub_parser.add_parser(name="verify")
+    # Verify the Datalog policy.
+    vp_parser = sub_parser.add_parser(name="verify-policy")
```
tromai, line 239: I think we could add the `description` for this command, similar to `dump-defaults` above. 
In additions, the `analzye` command is missing the `description` too. Would you prefer us to address it in another PR?

scripts/release_scripts/run_macaron.sh
```diff
@@ -168,7 +161,7 @@ while [[ $# -gt 0 ]]; do
             ;;
         # This flag is duplicated for digest and database.
```
tromai, line 162: If agreed, I would want to create a ticket for addressing this duplication too. A flag shouldn't be duplicated with different meaning across our tool.

scripts/release_scripts/run_macaron.sh
```diff
@@ -314,25 +307,24 @@ if [[ -n "${sbom_path}" ]]; then
...
-if [[ -n "${provenance}" ]]; then
-    provenance="$(ensure_absolute_path "${provenance}")"
-    mounts+=("-v" "${provenance}:${MACARON_WORKSPACE}/provenance/${file_name}:ro")
+if [[ -n "${prov_exp}" ]]; then
+    prov_exp="$(ensure_absolute_path "${prov_exp}")"
+    mounts+=("-v" "${prov_exp}:${MACARON_WORKSPACE}/prov_expectations/${file_name}:ro")
 fi
 
-# POLICY_ENGINE entrypoint.
+# MACARON action - verify-policy
```
tromai, line 326:
```suggestion
# MACARON entrypoint - verify-policy action argvs
```
So it would be consistent with other actions.

scripts/release_scripts/run_macaron.sh
```diff
@@ -314,25 +307,24 @@ if [[ -n "${sbom_path}" ]]; then
...
 fi
 
-# MACARON entrypoint - Verify action argvs
-# Determine the provenance path to be mounted into ${MACARON_WORKSPACE}/provenance/${file_name}
-# This is for macaron verify action.
-if [[ -n "${arg_provenance}" ]]; then
-    provenance="${arg_provenance}"
-    err=$(check_file_exists "${provenance}" "-pr/--provenance")
+# Determine the provenance expectation path to be mounted into ${MACARON_WORKSPACE}/prov_expectations/${file_name}
+if [[ -n "${arg_prov_exp}" ]]; then
```
tromai, line 311: I think we should also handle the case where `arg_prov_exp` is a directory.

tests/policy_engine/resources/policies/slsa_verifier.yaml
```diff
@@ -10,10 +10,6 @@ definition:
   predicateType: https://slsa.dev/provenance/v0.2
 
   predicate:
-    builder:
```
tromai, line 13: Why is this part removed in the policy?

src/macaron/__main__.py
```diff
@@ -244,14 +233,14 @@ def main() -> None:
...
 
-    verify_parser.add_argument(
-        "-pr", "--provenance", required=True, type=str, help=("The path to the provenance file.")
-    )
+    vp_parser.add_argument("-d", "--database", required=True, type=str, help="Path to the database.")
+    vp_parser.add_argument("-f", "--file", required=False, type=str, default="", help="Path to the Datalog policy.")
```
tromai, line 242: This flag is not required. However, in this method https://github.com/oracle-samples/macaron/blob/8aa0d6769c01a809a0f381950209750583ba92d1/src/macaron/policy_engine/policy_engine.py#L106, we are still treating it as not `None`. A file `open(...)` with the default value `""` would return a `FileNotFound` error. 

To replicate, we can run this command:
```
macaron verify-policy -d <path_to_dabase>
```

src/macaron/config/global_config.py
```diff
@@ -65,19 +62,27 @@ def load(
         self.gh_token = gh_token
         self.resources_path = resources_path
 
-        # Find the policies.
-        policy_files = []
-        for policy_path in policy_paths:
-            if os.path.isdir(policy_path):
-                for policy_file_path in os.listdir(policy_path):
+    def load_expectation_paths(self, expectation_paths: list[str]) -> None:
```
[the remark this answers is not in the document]
behnazh-w: > I'm okay with having a separate method to load he expectation path. But I was wondering what was the reason behind this change? thinking

Previously the path to expectation files (for example YAML files) was passed as the optional argument of `macaron` for all actions, and loaded via `global_config.load`. But now we only set them for `analyze` action. So we need to load the values separately.

src/macaron/__main__.py
```diff
@@ -75,58 +77,45 @@ def analyze_slsa_levels_single(analyzer_single_args: argparse.Namespace) -> None
...
+
+        case "verify-policy":
+            sys.exit(verify_policy(action_args))
+
         case "analyze":
+            # Check that the GitHub token is enabled.
+            if not action_args.personal_access_token:
+                logger.error("GitHub access token not set.")
+                sys.exit(os.EX_USAGE)
             analyze_slsa_levels_single(action_args)
```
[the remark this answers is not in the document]
behnazh-w: > I think we need to have a separate PR to update `analyze_slsa_levels_single` so that it return the status code instead of handling `os.exit` by itself.

There is already a [ticket](https://github.com/oracle-samples/macaron/issues/175) where this can be handled.

> In additions, do you think that the check for existence of the access token could be relocated inside `analyze_slsa_levels_single` ?

There is already a [ticket](https://github.com/oracle-samples/macaron/issues/199) for this.

src/macaron/config/global_config.py
```diff
@@ -23,7 +23,7 @@ class GlobalConfig:
     resources_path: str = ""
 
     def __init__(self) -> None:
-        self.policy_paths = []
+        self.expectation_paths = []
```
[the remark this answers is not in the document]
behnazh-w: See commit [649081b](https://github.com/oracle-samples/macaron/pull/192/commits/649081b04e1df8698e544df6bd65fefd5be995fd).

src/macaron/__main__.py
```diff
@@ -75,58 +77,45 @@ def analyze_slsa_levels_single(analyzer_single_args: argparse.Namespace) -> None
     sys.exit(status_code)
 
 
-def verify_prov(verify_args: argparse.Namespace) -> Never:
-    """Verify a provenance against a user defined policy."""
-    prov_file = verify_args.provenance
-    policy_files = list(filter(lambda path: os.path.splitext(path)[1] in (".yaml", ".yml"), global_config.policy_paths))
+def verify_policy(action_args: argparse.Namespace) -> int:
```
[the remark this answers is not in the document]
behnazh-w: See commit [649081b](https://github.com/oracle-samples/macaron/pull/192/commits/649081b04e1df8698e544df6bd65fefd5be995fd).

src/macaron/__main__.py
behnazh-w: See commit [649081b](https://github.com/oracle-samples/macaron/pull/192/commits/649081b04e1df8698e544df6bd65fefd5be995fd).

src/macaron/__main__.py
behnazh-w: > add the `description`

There is a [ticket](https://github.com/oracle-samples/macaron/issues/211) to address this.

scripts/release_scripts/run_macaron.sh
behnazh-w: > If agreed, I would want to create a ticket for addressing this duplication too. A flag shouldn't be duplicated with different meaning across our tool.

There is a [ticket](https://github.com/oracle-samples/macaron/issues/211) to address this.

tests/policy_engine/resources/policies/slsa_verifier.yaml
behnazh-w: Because the YAML policy doesn't support regex, so we need to keep adding exact tag versions for the builder, e.g.,
` https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builder_go_slsa3.yml@refs/tags/v1.2.0`

src/macaron/__main__.py
behnazh-w: See commit [c09c5a3](https://github.com/oracle-samples/macaron/pull/192/commits/c09c5a3379ef0733cd926c0b767480d77809a220).

scripts/release_scripts/run_macaron.sh
behnazh-w: See commit [c09c5a3](https://github.com/oracle-samples/macaron/pull/192/commits/c09c5a3379ef0733cd926c0b767480d77809a220).

src/macaron/__main__.py
behnazh-w: See commit [c09c5a3](https://github.com/oracle-samples/macaron/pull/192/commits/c09c5a3379ef0733cd926c0b767480d77809a220).

scripts/release_scripts/run_macaron.sh
behnazh-w: See commit [6a71cf8](https://github.com/oracle-samples/macaron/pull/192/commits/6a71cf8de74d95541daa6ada412d8f6e2f64661d).

src/macaron/__main__.py
behnazh-w: I have decided disallowing a list of file paths altogether as it would complicate the `run_macaron.sh` script arg parsing. Instead users can put all the expectations in a directory.

See commit [6a71cf8](https://github.com/oracle-samples/macaron/pull/192/commits/6a71cf8de74d95541daa6ada412d8f6e2f64661d).

src/macaron/__main__.py
tromai: I wonder why don't we make the `file` argument `required=True` instead? We always exit when the user doesn't provide the path to the file so I think it's better if we make it mandatory.

behnazh-w: > I wonder why don't we make the `file` argument `required=True` instead? We always exit when the user doesn't provide the path to the file so I think it's better if we make it mandatory.

Good point.

behnazh-w: See commit [f58d698](https://github.com/oracle-samples/macaron/pull/192/commits/f58d6980b5b7c6649d07d753ced78ed31126cb19).

tromai (approved):

PR #208, opened 2023-05-09
test: return error when defaults.ini provided by user does not exist

behnazh-w:


benmss (approved):
I think the problem has been fixed nicely.

tromai (approved):

PR #210, opened 2023-05-09
chore(deps): do not upgrade indirect Go deps

### LINKED ISSUE #209: Do not update indirect Go dependencies
Using `go get -u` to upgrade Go dependencies, which [updates](https://go.dev/ref/mod#go-mod-file-updates) all dependencies (direct and indirect), can result in compatibility issues, ignoring the pinned dependencies of direct ones.

behnazh-w:
Closes #209 

For Go dependencies, instead of using `go get -u` we can use the code snippet suggested [here](https://github.com/golang/go/issues/28424#issuecomment-1101896499) to avoid updating indirect dependencies, which can result in a broken state.

Note that Dependabot [by default](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file) only updates the direct dependencies, so with this change `make upgrade-go` will be compatible with Dependabot.


tromai:
> so with this change make upgrade-go will be compatible with Dependabot.

I still not quite get this part. By `compatible`, do you mean that it reflects the same mechanism as Dependabot in updating only direct dependencies?

tromai (approved):

behnazh-w:
> I still not quite get this part. By `compatible`, do you mean that it reflects the same mechanism as Dependabot in updating only direct dependencies?

It means that by forcefully updating the indirect dependencies, the direct dependencies that depend on them (by pinning their version) might not be compatible with the new versions.

PR #216, opened 2023-05-12
fix: restore the runner if an uncaught exception happens in a check (#132)

### LINKED ISSUE #132: An Exception happens in a check will make the Runner not available for the rest of the analysis
**Description**
This issue is discovered together with issue #131 . When an exception happens, we could not properly re use the Runner which run that check. This will make the rest of the analysis failed with the following error message:
```
2023-03-29 12:04:07,600 [CRITICAL] Could not find any available runners. Stop the analysis...
```
**Reproduce**
We can run the analysis against a sample config where the repository mentioned in #131 (https://github.com/java-json-tools/json-patch) is one of the dependencies. For example:
```
target:
  id: apache/maven
  path: https://github.com/apache/maven

dependencies:
- id: exception-happen-repo
  path: https://github.com/java-json-tools/json-patch
- id: The analysis will failed to run against this dependency because the Runner is not available.
  path: https://github.com/ben-manes/caffeine.git
```

tromai:
Closes #132

### DIFF
--- a/src/macaron/slsa_analyzer/registry.py
+++ b/src/macaron/slsa_analyzer/registry.py
@@ -390,6 +390,7 @@ class Registry:
                                 logger.error("Exception in check %s: %s.", current_check_id, current_future.exception())
                                 logger.info("Check %s has failed.", current_check_id)
                                 current_future.cancel()
+                                self.runner_queue.put(current_runner)
                                 return results
                         except concurrent.futures.TimeoutError:
                             # The check is still running, put the future back into the queue.


src/macaron/slsa_analyzer/registry.py
behnazh-w, line 1: As discussed when a check throws an exception we need to stop the analysis as mark it as failed. However, the issue is that currently we don't allow any subsequent analysis of other dependencies. I think a more fundamental issue is that we use the same `registry` instance for the following analysis runs. [Here](https://github.com/oracle-samples/macaron/issues/222) is a ticket to keep track of this issue.

behnazh-w (approved):

PR #217, opened 2023-05-12
feat: add additional repo finding via parent POMs

benmss:
This adds the second part of the repo finding extension. Due to the hierarchical nature of POM files, artifacts can inherit SCM metadata from parent artifacts. In these cases, we can match additional artifacts to repositories via this method.
A configuration value exists to disable this feature during analysis:
```
[repofinder.java]
find_parents = False
```


src/macaron/dependency_analyzer/java_repo_finder.py
```diff
@@ -171,24 +199,52 @@ def find_repo(gav: str, tags: list[str]) -> Iterator[str]:
...
-
-    # Parse XML data and return URL
-    return parse_pom(pom, tags)
+    # Perform the following in a loop:
+    # - Create URLs for the current artifact POM
+    # - Parse the POM
+    # - Try to extract SCM metadata and return URLs
+    # - Try to extract parent information and change current artifact to it
+    # - Repeat
+    limit = 10
```
tromai, line 208: @behnazh-w  I wonder if we should put this variable into `defaults.ini`?

src/macaron/dependency_analyzer/java_repo_finder.py
```diff
@@ -148,17 +150,43 @@ def parse_pom(pom: str, tags: list[str]) -> Iterator[str]:
                 # Add the contents of the final tag
                 results.append(element.text.strip())
 
-    return iter(results)
+    return iter(results), len(results)
```
tromai, line 153: I think we could remove using Iterator here because later on in `dependency_resolver.py` we will convert it back to a list anyway.
```
urls = find_repo(
    item["group"],
    item["name"],
    item["version"],
    ["scm.url", "scm.connection", "scm.developerConnection"],
)
item["url"] = DependencyAnalyzer.find_valid_url(list(urls))
``` 

In additions, I don't think that using an Iterator would give us to much optimization as we create a list in the first hand. Creating an iterator from a list and iterate through it would give the same behavior as we are running a for loop through the original list itself :thinking:

behnazh-w: @tromai good point. I have created a [fix](https://github.com/oracle-samples/macaron/pull/230) so that using `Iterator` here would still make sense.

src/macaron/dependency_analyzer/java_repo_finder.py
```diff
@@ -148,17 +150,43 @@ def parse_pom(pom: str, tags: list[str]) -> Iterator[str]:
...
+def find_repo(group: str, artifact: str, version: str, tags: list[str]) -> Iterator[str]:
     """
     Attempt to retrieve a repository URL that matches the passed GAV artifact.
 
     Parameters
     ----------
-    gav : str
-        An artifact represented as a GAV (group, artifact, version) in the format: G:A:V.
+    group : str
+        The group identifier of an artifact
```
behnazh-w, line 185: Can you please add `.` at the end of these parameter descriptions?

src/macaron/dependency_analyzer/java_repo_finder.py
behnazh-w: > @behnazh-w I wonder if we should put this variable into `defaults.ini`?

Yes it should.

src/macaron/dependency_analyzer/java_repo_finder.py
```diff
@@ -171,24 +199,52 @@ def find_repo(gav: str, tags: list[str]) -> Iterator[str]:
...
+    # - Create URLs for the current artifact POM
+    # - Parse the POM
+    # - Try to extract SCM metadata and return URLs
+    # - Try to extract parent information and change current artifact to it
+    # - Repeat
+    limit = 10
+    while group and artifact and version and limit > 0:
+        # Create the URLs for retrieving the artifact's POM
+        group = group.replace(".", "/")
+        artifact = artifact.replace(".", "/")
```
behnazh-w, line 212: Is it possible that the artifact name contains `.`? :thinking:

benmss: It seems there is no exact naming scheme for artifact names. As they most likely represent the name of the related JAR file, or similar, we should assume any valid file name character can be used.
Source: [Maven](https://maven.apache.org/guides/mini/guide-naming-conventions.html)

src/macaron/dependency_analyzer/java_repo_finder.py
```diff
@@ -182,30 +182,33 @@ def find_repo(group: str, artifact: str, version: str, tags: list[str]) -> Itera
...
-    tags : list[str]
-        The list of XML tags to look for, each in the format: tag1[.tag2 ... .tagN]
+        The version number of an artifact.
+    tags : Iterator[str]
+        The list of XML tags to look for, each in the format: tag1[.tag2 ... .tagN].
 
     Returns
     -------
-    Iterator[str] :
+    list[str] :
```
tromai, line 195: Why would this be changed to `list` when we are still returning an Iterator (e.g `return (iter[])`) ?

benmss: I changed to a list as part of the earlier discussion, then reverted the change. Seems I missed a bit.

src/macaron/dependency_analyzer/java_repo_finder.py
```diff
@@ -148,47 +150,104 @@ def parse_pom(pom: str, tags: list[str]) -> Iterator[str]:
...
-    # Try each POM URL in order, terminating early if a match is found
-    with requests.Session() as session:
-        pom = ""
-        for request_url in request_urls:
-            pom = retrieve_pom(session, request_url)
-            if pom != "":
-                break
-
-    if pom == "":
+    if len(tags) == 0 or not any(tags):
```
behnazh-w, line 201: Isn't it enough to
```suggestion
    if not any(tags):
```

src/macaron/dependency_analyzer/java_repo_finder.py
```diff
@@ -148,47 +150,104 @@ def parse_pom(pom: str, tags: list[str]) -> Iterator[str]:
...
+    group : str
+        The group identifier of an artifact.
+    artifact : str
+        The artifact name of an artifact.
+    version : str
+        The version number of an artifact.
+    tags : Iterator[str]
+        The list of XML tags to look for, each in the format: tag1[.tag2 ... .tagN].
 
     Returns
```
behnazh-w, line 193:
```suggestion
   Yields
```

src/macaron/dependency_analyzer/java_repo_finder.py
```diff
@@ -148,47 +150,104 @@ def parse_pom(pom: str, tags: list[str]) -> Iterator[str]:
...
-        pom = ""
-        for request_url in request_urls:
-            pom = retrieve_pom(session, request_url)
-            if pom != "":
-                break
-
-    if pom == "":
+    if len(tags) == 0 or not any(tags):
+        logger.debug("No POM tags found for URL discovery.")
         return iter([])
```
behnazh-w, line 203: We don't need to return `iter([])` and can simply
```suggestion
        return
```

Same comment applies for other `return iter([])` in this function.

src/macaron/dependency_analyzer/java_repo_finder.py
```diff
@@ -148,47 +150,104 @@ def parse_pom(pom: str, tags: list[str]) -> Iterator[str]:
...
+        # Parse POM using defusedxml
+        pom_element = parse_pom(pom)
+        if pom_element is None:
+            return iter([])
+
+        # Attempt to extract SCM data and return URL
+        urls, url_count = find_scm(pom_element, tags)
+
+        if url_count > 0:
+            return urls
```
behnazh-w, line 242:
```suggestion
            yield from urls
```

src/macaron/dependency_analyzer/java_repo_finder.py
```diff
@@ -107,38 +83,64 @@ def _find_element(parent: typing.Optional[Element], target: str) -> typing.Optio
     # Attempt to match the target tag within the children of parent.
     for child in parent:
         # Account for raw tags, and tags accompanied by Maven metadata enclosed in curly braces. E.g. '{metadata}tag'
-        if child.tag == target or child.tag.endswith("}" + target):
+        if child.tag == target or child.tag.endswith(f"}}{target}"):
```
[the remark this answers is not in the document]
benmss: This is only for a slight performance increase using an f string over concatenation. The `}}` is how a single curly brace is represented in f strings.

tromai (approved):

src/macaron/dependency_analyzer/java_repo_finder.py
behnazh-w: > It seems there is no exact naming scheme for artifact names. As they most likely represent the name of the related JAR file, or similar, we should assume any valid file name character can be used. Source: [Maven](https://maven.apache.org/guides/mini/guide-naming-conventions.html)

Would that mean the URL would become invalid if the artifact name includes `.`? e.g. `foo.bar.baz` would change to `foo/bar/baz` :thinking:

benmss: I think you are right. I'll remove the replacement for artifact names.

behnazh-w (approved):

PR #228, opened 2023-05-16
chore: do not add a duplicate dependency to result record

### LINKED ISSUE #227: Remove duplicate dependencies from result records

behnazh-w:
This PR uses the result record ID  instead of the remote path as  identifier to prevent adding duplicate dependencies for same artifacts. 
Closes #227

### DIFF
--- a/src/macaron/output_reporter/results.py
+++ b/src/macaron/output_reporter/results.py
@@ -182,7 +182,7 @@ class Report:
         self.root_record: Record = root_record
         self.record_mapping: dict[str, Record] = {}
         if root_record.context:
-            self.record_mapping[root_record.context.remote_path] = root_record
+            self.record_mapping[root_record.record_id] = root_record
 
     def get_records(self) -> Iterable[Record]:
         """Get the generator for all records in the report.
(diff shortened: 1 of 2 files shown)


src/macaron/output_reporter/results.py
```diff
@@ -203,9 +203,11 @@ def add_dep_record(self, dep_record: Record) -> None:
         dep_record : Record
             The record of the dependency.
         """
-        self.root_record.dependencies.append(dep_record)
-        if dep_record.context:
-            self.record_mapping[dep_record.context.remote_path] = dep_record
+        # Do not add a dependency if it's a duplicate.
+        if dep_record.record_id not in self.record_mapping:
```
tromai, line 207: If we don't put the `dep_record` instance into `record_mapping` here, won't we lose the information for this duplicated dependency and it won't be printed out in the final result :thinking: ? I think we still add the dep_record to `self.root_record.dependencies` and we only need to change how the key of the `self.record_mapping` dictionary (to using record_id instead).

behnazh-w: I'm already using the `record_id` as the key to store records into `self.record_mapping`. But by adding the inclusion check the records in `self.root_record.dependencies` will also be consistent with `self.record_mapping`, which is necessary to avoid keeping two records with same `record_id`s in the dependencies (duplicates). Note that if two dependencies have different IDs but same remote paths, we still will add them both to dependencies and in the final report they will be shown as duplicates.

src/macaron/output_reporter/results.py
```diff
@@ -256,20 +258,20 @@ def get_dependencies(self, root_record: Optional[Record] = None) -> Iterable[tup
...
         ----------
-        remote_path : str
-            The remote path to look for the analyze context
+        record_id : str
+            The ID to look for the analyze context.
 
         Returns
         -------
         AnalyzeContext
             The analyze context for the given remote path or None if cannot find.
```
tromai, line 272:
```suggestion
            The analyze context for the given record ID or None if cannot find.
```

behnazh-w: See commit [9637268](https://github.com/oracle-samples/macaron/pull/228/commits/963726856bc6c711e3cccaa62871997bf1089b2b).

tromai (approved):

PR #238, opened 2023-05-23
fix: do not overwrite an existing check relationship when a check has no parent in the Registry

### LINKED ISSUE #237: Register a check with no parent could overwrite an existing relationship mapping in Registry
**Description**
In this part:
https://github.com/oracle-samples/macaron/blob/907eedf543131a2dd8d406bfa3ad54b3df8fae9a/src/macaron/slsa_analyzer/registry.py#LL75C1-L81C4
```python
        if not check.depends_on:
            self._check_relationships_mapping[check.check_id] = {}
        else:
            for parent_relationship in check.depends_on:
                if not self._add_relationship_entry(check.check_id, parent_relationship):
                    logger.error("Cannot load relationships of check %s.", check.check_id)
                    sys.exit(1)
```
We see that when registering a new check with no parent, an entry in `_check_relationships_mapping` for that check is initialized as an empty dictionary. This is correct if `_check_relationships_mapping` currently does not have an entry for it yet.
However, as all the checks might be loaded in a different order, the entry for the current check we are trying to add already exists.

For example: let's say we want to add `mcn_provenance_level_three_1` first and then we will add `mcn_provenance_available_1` (assume that `_check_relationships_mapping` is currently empty). 
Because `mcn_provenance_level_three_1` depends on `mcn_provenance_available_1` on `PASSED` (https://github.com/oracle-samples/macaron/blob/907eedf543131a2dd8d406bfa3ad54b3df8fae9a/src/macaron/slsa_analyzer/checks/provenance_l3_check.py#LL142C2-L144C4),  `_check_relationships_mapping` will become:
```python
{
    "mcn_provenance_available_1":  { "mcn_provenance_level_three_1" : CheckResultType.PASSED }
}
```
However, when we add `mcn_provenance_available_1`, because it does not have any parent, the code above will overwrite the entry for `mcn_provenance_available_1` in `_check_relationships_mapping`:
```python
{
    "mcn_provenance_available_1": {}
}
``` 

**Solution**
```python
        if not check.depends_on:
            if self._check_relationships_mapping.get(check.check_id, None) is None:
                self._check_relationships_mapping[check.check_id] = {}
        else:
            for parent_relationship in check.depends_on:
                if not self._add_relationship_entry(check.check_id, parent_relationship):
                    logger.error("Cannot load relationships of check %s.", check.check_id)
                    sys.exit(1)
```

tromai:
Closes #237

### DIFF
--- a/src/macaron/slsa_analyzer/registry.py
+++ b/src/macaron/slsa_analyzer/registry.py
@@ -69,10 +69,11 @@ class Registry:
             logger.error("The check registered is invalid.")
             sys.exit(1)
 
-        # Update the check relationship mapping. If this check does not depend on any check,
-        # we add it as a standalone check ONLY when _check_relationships_mapping does not have an entry for it.
+        # Update the check relationship mapping. Even though a check might not depend on any other checks, other
+        # checks can still depend on it, and therefore it might have been initialized and added to the mapping
+        # already. So we need to check if it already exists in `_check_relationships_mapping`.
         if not check.depends_on:
-            if self._check_relationships_mapping.get(check.check_id, None) is None:
+            if check.check_id not in self._check_relationships_mapping:
                 self._check_relationships_mapping[check.check_id] = {}
         else:
             for parent_relationship in check.depends_on:
--- a/src/macaron/slsa_analyzer/registry.py
+++ b/src/macaron/slsa_analyzer/registry.py
@@ -69,11 +69,11 @@ class Registry:
             logger.error("The check registered is invalid.")
             sys.exit(1)
 
-        # Update the check relationship mapping.
-        # If this check does not depend on any check,
-        # we add it as a standalone check.
+        # Update the check relationship mapping. If this check does not depend on any check,
+        # we add it as a standalone check ONLY when _check_relationships_mapping does not have an entry for it.
         if not check.depends_on:
-            self._check_relationships_mapping[check.check_id] = {}
+            if self._check_relationships_mapping.get(check.check_id, None) is None:
+                self._check_relationships_mapping[check.check_id] = {}
         else:
             for parent_relationship in check.depends_on:
                 if not self._add_relationship_entry(check.check_id, parent_relationship):


src/macaron/slsa_analyzer/registry.py
```diff
@@ -69,11 +69,11 @@ def register(self, check: BaseCheck) -> None:
...
             sys.exit(1)
 
-        # Update the check relationship mapping.
-        # If this check does not depend on any check,
-        # we add it as a standalone check.
+        # Update the check relationship mapping. If this check does not depend on any check,
+        # we add it as a standalone check ONLY when _check_relationships_mapping does not have an entry for it.
         if not check.depends_on:
-            self._check_relationships_mapping[check.check_id] = {}
+            if self._check_relationships_mapping.get(check.check_id, None) is None:
```
behnazh-w, line 75: Because `CheckResultType` [cannot](https://github.com/oracle-samples/macaron/blob/5615431c3d270c3fe1f40c82d1637637351806cd/src/macaron/slsa_analyzer/registry.py#L32) be `None`, to be more explicit it's better to directly check for inclusion here:
```suggestion
            if check.check_id not in self._check_relationships_mapping:
```

src/macaron/slsa_analyzer/registry.py
```diff
@@ -69,11 +69,11 @@ def register(self, check: BaseCheck) -> None:
             logger.error("The check registered is invalid.")
             sys.exit(1)
 
-        # Update the check relationship mapping.
-        # If this check does not depend on any check,
-        # we add it as a standalone check.
+        # Update the check relationship mapping. If this check does not depend on any check,
+        # we add it as a standalone check ONLY when _check_relationships_mapping does not have an entry for it.
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/238/commits/eda34112639e7a762dffde66b18f3749217f6b83

src/macaron/slsa_analyzer/registry.py
tromai: Fixed in https://github.com/oracle-samples/macaron/pull/238/commits/eda34112639e7a762dffde66b18f3749217f6b83

PR #241, opened 2023-05-25
fix: add 'packages: read' permission to release workflow

behnazh-w:


.github/workflows/release.yaml
```diff
@@ -27,6 +27,7 @@ jobs:
     uses: ./.github/workflows/_build.yaml
     permissions:
       contents: read
+      packages: read
```
tromai, line 30: Could you let me know why we need to have this fix? Is it because action-lint reports this as an error?

behnazh-w: > Could you let me know why we need to have this fix? Is it because action-lint reports this as an error?

No, that's because the nested `_build_docker.yaml` reusable workflow needs this permission. We need to pass the permissions here too.

tromai (approved):

PR #243, opened 2023-05-26
build: add a Makefile target for building the Docker image

behnazh-w:


.github/workflows/release.yaml
```diff
@@ -121,7 +121,8 @@ jobs:
     - name: Build the Docker image
       run: |
         RELEASE_TAG=$(git describe --tags --abbrev=0)
-        ./scripts/dev_scripts/build_docker.sh "$IMAGE_NAME" ${{ github.workspace }} "$RELEASE_TAG"
+        export RELEASE_TAG
```
tromai, line 124: Could we export the value directly?
```
export RELEASE_TAG=$(git describe --tags --abbrev=0)
```

behnazh-w: > Could we export the value directly?
> 
> ```
> export RELEASE_TAG=$(git describe --tags --abbrev=0)
> ```

Shellcheck recommends to keep them separate.

tromai (approved):

PR #244, opened 2023-05-26
build: use make setup install commitizen

behnazh-w:
Originally `commitizen` was installed individually for performance reasons. However, this meant that we had to always manually update `commitizen` in the GitHub Action. This PR runs `make setup` to install `commitizen`, which means installing all the packages but on the plus side we won't have to worry about updating it manually anymore.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -54,7 +54,7 @@ jobs:
     - name: Create empty virtual environment for Actions
       run: mkdir .venv
     - name: Install dependencies
-      run: make setup
+      run: make setup-github-actions
 
     - name: Set up user
       run: |
@@ -184,7 +184,7 @@ jobs:
     - name: Create empty virtual environment for Actions
       run: mkdir .venv
     - name: Install dependencies
-      run: make setup
+      run: make setup-github-actions
 
     # Create the Release Notes using commitizen.
     - name: Create Release Notes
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -50,10 +50,11 @@ jobs:
...
 
-    - name: Set up Commitizen
-      run: |
-        pip install --upgrade pip wheel
-        pip install 'commitizen ==2.37.1'
+    # Prepare the environment to run commitizen.
+    - name: Create empty virtual environment for Actions
+      run: mkdir .venv
+    - name: Install dependencies
+      run: make setup
```
tromai, line 57: > which means installing all the packages

Is it possible if we could define a target in the Makefile for installing only the components necessary for the `GitHub Action` only? For example that new target would only need to run:
```
python -m pip install --upgrade --upgrade-strategy eager --editable .[actions]
```

behnazh-w: > python -m pip install --upgrade --upgrade-strategy eager --editable .[actions]

This still installs the main dependencies, but still better than `make setup` that installs Go dependencies, tests, etc. :+1: 

See commit [77abbe2](https://github.com/oracle-samples/macaron/pull/244/commits/77abbe228cbb77a17177651d5f8ce3c1067536bb).

tromai (approved):

PR #249, opened 2023-05-29
ci(deploy): add a workflow to deploy docs to GitHub pages

### LINKED ISSUE #248: Add a workflow to deploy docs to GitHub pages

behnazh-w:
Closes #248 

By running this reusable workflow, the documentations will be published at https://oracle-samples.github.io/macaron/.

### DIFF
--- a/.github/workflows/_deploy-github-pages.yaml
+++ b/.github/workflows/_deploy-github-pages.yaml
@@ -19,11 +19,6 @@ permissions:
   contents: read
   pages: write
   id-token: write
-# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
-# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
-concurrency:
-  group: pages
-  cancel-in-progress: false
 
 jobs:
   # Download the generated docs archive, verify its hash, and deploy to GitHub pages.
--- a/.github/workflows/_deploy-github-pages.yaml
+++ b/.github/workflows/_deploy-github-pages.yaml
@@ -70,11 +70,13 @@ jobs:
 
     - name: Setup Pages
       uses: actions/configure-pages@f156874f8191504dae5b037505266ed5dda6c382 # v3.0.6
-    - name: Upload artifact
+
+    - name: Upload documentation artifact
       uses: actions/upload-pages-artifact@64bcae551a7b18bcb9a09042ddf1960979799187 # v1.0.8
       with:
         # Upload html/ directory.
         path: html
+
     - name: Deploy to GitHub Pages
       id: deployment
       uses: actions/deploy-pages@af48cf94a42f2c634308b1c9dc0151830b6f190a # v2.0.1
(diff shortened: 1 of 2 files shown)


.github/workflows/_deploy-github-pages.yaml
```diff
@@ -0,0 +1,80 @@
...
+        required: true
+        description: The sha of the artifact that contains docs content
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
+# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
+concurrency:
```
tromai, line 24: From my understanding, at any time, there can only be at most one in-progress run and at most one pending run for this workflow.
Trigger this workflow will removed any pending run. If there is an in-progress run, it will put itself to pending. If there is NO in-progress run, it will run. :thinking:

.github/workflows/_deploy-github-pages.yaml
```diff
@@ -0,0 +1,80 @@
...
+        if [[ -z "${DOCS_PATH}" ]];
+        then
+            echo "Unable to find Macaron docs files in ${DIST_PATH}."
+            exit 1
+        fi
+        unzip "$DOCS_PATH"
+
+    - name: Setup Pages
+      uses: actions/configure-pages@f156874f8191504dae5b037505266ed5dda6c382 # v3.0.6
+    - name: Upload artifact
```
tromai, line 73: I think we should name it `Upload GitHub pages` or `Upload Documentation`.

behnazh-w: See commit [3f2d6ff](https://github.com/oracle-samples/macaron/pull/249/commits/3f2d6ff8d45c06523bcfcfd2bb3fca5ab015256f).

.github/workflows/_deploy-github-pages.yaml
behnazh-w: Sorry I don't understand your question.

tromai: Oh I was trying to confirm if my understanding for the usage of `concurrency` here is correct.

tromai: Otherwise, everything LGTM.

tromai (approved):

.github/workflows/_deploy-github-pages.yaml
behnazh-w: > Oh I was trying to confirm if my understanding for the usage of `concurrency` here is correct.

Yes, you're correct. Thinking some more I decided to remove the `concurrency` condition as it's very unlikely to happen for us, and it's better to be consistent with other workflows.

See commit [43765be](https://github.com/oracle-samples/macaron/pull/249/commits/43765be9dea9ddf5f42f6e5c7f77b76e6a007c58).

PR #254, opened 2023-05-29
feat: add exclude and include check in ini config

### LINKED ISSUE #133: Allow disabling a check

tromai:
Close #133.

# Feature description

## Configuring exclude/include in Ini config.

We want to allow the users to:
- **exclude** checks that they don't want to run.
- **include** the checks that they want to run. 

Because checks in Macaron depend on each other (using the [`depends_on`](https://github.com/oracle-samples/macaron/blob/9217e9319ce4b3e08b4e4eb12c951f6a97b33c6e/src/macaron/slsa_analyzer/checks/base_check.py#L27) attribute). Therefore, when we want to disable a check (from users' exclude/include configuration), we also need to handle its transitive parents and children.
- **EXPLICIT_INCLUDE** = checks included from the user configuration
- **EXPLICIT_EXCLUDE** = checks included from the user configuration
- **EFFECTIVE_INCLUDED** = **EXPLICIT_INCLUDE** set plus transitive parents
- **EFFECTIVE_EXCLUDED** = **EXPLICIT_EXCLUDE** set plus transitive children
- **FINAL_INCLUDED** = **EFFECTIVE_INCLUDED** minus **EFFECTIVE_EXCLUDED**

The user should define the list of str with each element as **an ID of a check** or a **glob pattern** (similar to patterns used in [glob](https://docs.python.org/3/library/glob.html)) which will be used to match multiple check IDs. This configuration is located in the `defaults.ini` file. 

```ini
[analysis.checks]
# By default, we don't exclude any checks.
exclude =
# By default, we run all checks available.
include = *
```
Example on glob patterns:
- `mcn_*` would match all checks with IDs start with `mcn_`. 
- `*` would match all checks registered in Macaron.
- `mcn_boo_1` would match a single check with ID `mcn_boo_1`

## Displaying the run checks

- In the log messages, we want to display the list of checks that are going to run and excluded before the analysis.
- In the HTML reports, we want to have a section just before the Check report table 
  - Show all the checks in a "tree-like" view
  - The excluded checks are faded-out or crossed out indicating its disabled status.
  - This section is in a drop down section so that we could hide it when the user first open the report. 

For example, this could be what we display in the HTML reports given that mcn_build_as_code_1 is excluded.

image

# Implementation

## Background
When a check is registered in Macaron, we update the 3 following attribute of the `Registry` class:

- [`_all_checks_mapping`](https://github.com/oracle-samples/macaron/blob/907eedf543131a2dd8d406bfa3ad54b3df8fae9a/src/macaron/slsa_analyzer/registry.py#LL29C5-L29C24) : this is the dictionary that maps between the check id (type `str`) and the instance of a check (inherit from `BaseCheck`).
- [`_check_relationships_mapping`](https://github.com/oracle-samples/macaron/blob/907eedf543131a2dd8d406bfa3ad54b3df8fae9a/src/macaron/slsa_analyzer/registry.py#LL32C5-L32C33):
  - Has type `dict[str, dict[str, CheckResultType]]`:
    - Key (`str`): the ID of the parent check
    - Value (`dict[str, CheckResultType]`): the mapping between the children that depend on the parent check **to** the check status that each of them depend on (e.g `CheckResultType.PASSED`):
      - Key (`str`): the id of a child check
      - Val (`CheckResultType`): the status that this child check depends on.
  - This dictionary is used to make sure that the relationships are defined correctly (e.g. one check cannot depend on multiple parents).
- [`_graph`](https://github.com/oracle-samples/macaron/blob/907eedf543131a2dd8d406bfa3ad54b3df8fae9a/src/macaron/slsa_analyzer/registry.py#LL40C5-L40C11): this is the final graph where we traverse to get the next available check in the correct topological order when we run the analysis.
  - This graph is updated each time a new check is registered.
  - This graph will add the necessary edges between a newly registered check and its parent.

To describe those attributes conceptually:
- `_all_checks_mapping`: includes the nodes in the check relationship tree.
- `_check_relationships_mapping`: it is sort of an `adjacency list` that described the edges in the check relationship tree.
- `_graph`: is built from the 2 above attributes and is used for getting the topological order of the checks. 
 
## Solution

- The list of run checks are computed when the method `registry.prepare()` is run. This list will be stored in this [attribute](https://github.com/oracle/macaron/blob/0500c88ca52f54d640efd90a8a1921cc22a901a9/src/macaron/slsa_analyzer/registry.py#L70). 
- Even though we have known the list of run checks and excluded checks, `_all_checks_mapping`, `_check_relationships_mapping` and `_graph` still contain all of the registered checks.  
- When the checks are run within the `registry.scan()` method (which still goes through the check tree will all registered checks), checks which are not included in this list will *not* be run. Therefore, the result for those checks will not be included in the reports and the database.
- Please see commit https://github.com/oracle/macaron/pull/254/commits/bb3178f28e88800ae596e9b84ea18301073c0d98 for more details.


src/macaron/slsa_analyzer/registry.py
```diff
@@ -564,5 +752,31 @@ def _should_skip_check(check: BaseCheck, results: dict[str, CheckResult]) -> Ski
 
         return None
 
+    # TODO: Come up with a better name for this method.
```
tromai, line 755: Any suggestions would be great :grin:

tests/slsa_analyzer/checks/test_registry.py
```diff
@@ -8,12 +8,66 @@
...
+@pytest.fixture(name="check_registry")
+def check_registry_fixture() -> Registry:
+    """Return a registry instance with sample checks registered.
+
+    Returns
+    -------
+    Registry
+        The sample registry instance.
+    """
+    # Refresh Registry static variables before each test case
```
tromai, line 30: Should I put a visual representation of the checks as a tree to make it easier for us to understand it?

src/macaron/output_reporter/templates/macaron.html
```diff
@@ -164,6 +164,30 @@
 {% endcall %}
 {% endmacro %}
 
+{% macro render_run_checks(check_tree, excluded_checks) %}
+<ul class="tree-view-nested-list">
+    {% for key, first_level in check_tree.items() %}
```
tromai, line 169: The first level of the tree (i.e checks that don't have a parent) don't need to have the symbol └── before it
image

Therefore, we handle it separately.

src/macaron/output_reporter/templates/macaron.html
```diff
@@ -164,6 +164,30 @@
...
+    <li>
+        <div>{{ key }}</div>
+        <ul>
+            {% for key, next_level in first_level.items() recursive %}
```
tromai, line 173: This block handle the second level of checks onward (i.e. the checks that have parent).

src/macaron/output_reporter/templates/macaron.html
```diff
@@ -164,6 +164,30 @@
...
+            <li>
+                {% if key in excluded_checks %}
+                <div class="check-tree-node-excluded">{{ key }}</div>
```
tromai, line 176: `class="check-tree-node-excluded"` would give the content a ~strikethrough~ font style while `class="check-tree-node"` don't

src/macaron/slsa_analyzer/registry.py
```diff
@@ -564,5 +752,31 @@ def _should_skip_check(check: BaseCheck, results: dict[str, CheckResult]) -> Ski
...
+    def get_check_tree(self) -> CheckTree:
```
tromai, line 756: This method traverse the check tree as a whole without any include/exclude restrictions. Therefore, we could print out the whole run checks tree with all registered checks in the final HTML report.

tromai: - [ ] I think an example of the returned dictionary of this method would be useful.

tromai: Added in https://github.com/oracle/macaron/pull/254/commits/f6d7426d7b9f937477fbe5ab5cda809edd6dc18b

src/macaron/slsa_analyzer/registry.py
```diff
@@ -325,6 +342,165 @@ def _validate_check_relationship(relationship: Any) -> bool:
...
+
+                check = self._all_checks_mapping.get(ele)
+
+                for relation in check.depends_on:  # type: ignore
+                    parent = relation[0]
+                    if parent not in visited:
+                        stack.append(parent)
+                    elif parent in stack:
+                        # Circular Dependency.
+                        raise CheckCircularDependency("cycle nodes detected")
```
tromai, line 416: This Error is only raised when in this exact scenario:
- The user provide custom exclude/include config.
- Their are circular dependencies in the checks.
This Error was implemented because to obtain the transitive parents or children of an exclude/include check, we need to traverse the dependency tree. Therefore, we want to somehow report if there are any circular dependencies. Otherwise, if the user don't use the exclude/include feature, any circular dependencies are caught by the graphlib library [here](https://github.com/oracle/macaron/blob/f6d7426d7b9f937477fbe5ab5cda809edd6dc18b/src/macaron/slsa_analyzer/registry.py#L689).
To print out more details on the circular path, more logic could be add. Do you think it would complicate the method even more :thinking: ?

tests/slsa_analyzer/checks/test_registry.py
```diff
@@ -200,3 +254,112 @@ def test_validate_check_id_format(self) -> None:
...
+
+
+@pytest.mark.parametrize(
+    ("child_id", "has_error"),
+    [("mcn_a_1", True), ("mcn_b_1", True), ("mcn_d_1", False)],
+)
+def test_get_transitive_parent_circular(circular_registry: Registry, child_id: str, has_error: bool) -> None:
+    """This method test get_transitive_parent method with circular detection"""
+    if has_error:
+        with pytest.raises(CheckRegistryError):
```
tromai, line 362: Detecting `CheckRegistryError` here is not correct as it's the parent of `CheckCircularDependencyError` and `CheckNotExistError`. I will update the tests which detect `CheckRegistryError` to detect the specific exception they are testing for.

tromai: See https://github.com/oracle/macaron/pull/254/commits/7dde1a163871bb9d6029fe2635dd957b5c252721

src/macaron/slsa_analyzer/registry.py
```diff
@@ -331,7 +343,166 @@ def _validate_check_relationship(relationship: Any) -> bool:
 
         return False
 
-    def scan(self, target: AnalyzeContext, skipped_checks: list[SkippedInfo]) -> dict[str, CheckResult]:
+    def get_transitive_children(self, check_id: str) -> list[str]:
```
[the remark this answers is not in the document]
nicallen: Indeed, those two functions are doing exactly the same thing, just over different graphs. If you put them side-by-side, eliminate the superficial differences, and factor out the single meaningful difference between them (the edges to traverse), the result looks something like this:
```python
    def get_transitive_closure(self, check_id: str,
                               get_successors: Callable[[str], set[str]]) -> list[str]:
        visited = []
        stack = [check_id]

        while stack:
            current_check_id = stack[-1]

            if current_check_id not in visited:
                visited.append(current_check_id)

                if current_check_id not in self._all_checks_mapping:
                    raise CheckNotExistError(f"Check {current_check_id} does not exist.")

                for successor in get_successors(current_check_id):
                    if successor not in visited:
                        stack.append(successor)
                    elif successor in stack:
                        raise CheckCircularDependencyError("cycle nodes detected")
            else:
                stack.pop()

        return visited
```
Then `get_transitive_children` and `get_transitive_parents` can be defined as calling `get_transitive_closure` with an appropriately-defined `get_successors` function to get the children and the parents, respectively.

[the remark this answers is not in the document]
tromai: Thanks for the suggestions. I agree that the current implementation duplicate the graph traversal logic. This lead to duplicated code and unit tests. I will factor out the logic for obtaining parents or children and have one single implementation for the traversal.   

> Also, I don't think validating the graph (e.g. checking if some check_id is valid) should be part of the traversal logic.
@nathanwn I think this is a valid point. I will work on moving the such validation logic outside of the graph traversal implementation.  

> I suppose ensuring that the graph is well-formed can be done before any graph traversals happen.
The logic for obtaining transitive closures are run after the whole graph is validated and built by the registry (see [here](https://github.com/oracle/macaron/blob/0500c88ca52f54d640efd90a8a1921cc22a901a9/src/macaron/slsa_analyzer/registry.py#L689-L703) so it's safe to not performing such validation again. However, because there isn't anything that can stop `get_transitive_closure` being called at anytime, I think we need to ensure that `get_transitive_closure` must raise such errors if it were called when the graph is not in a valid state (for example, if not all checks have been registered) even though there isn't any reason to do so. :thinking: 
I think in this case we could have two solutions:
1. Raise an exception when `get_transitive_closure` is called incorrectly (before the graph is validated)
2. Perform validation similarly to my current implementation.
@nathanwn Please let you know what you think.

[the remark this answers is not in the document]
nathanwn: To clarify, what I mean here is to have the attribute `_all_checks_mapping` of the `Registry` validated before even running the exclude/include logic to avoid the `if current_check_id not in self._all_checks_mapping` branching in the traversal function. This `if` statement checks if there is any check that exists in a value but does not have a key in the mapping, which I would argue is essentially checking for an invalid state of this attribute. Some validation on this exact attribute has already been done [here](https://github.com/oracle/macaron/blob/0500c88ca52f54d640efd90a8a1921cc22a901a9/src/macaron/slsa_analyzer/registry.py#L685C1-L687C25), so I assume we might as well complete the validation in the same place. Otherwise, the traversal function also needs to validate the `_all_checks_mapping` attribute, which may be doing too many things.

I don't think the argument of "this function may be called anywhere so we need to validate the `_all_checks_mapping` attribute all the time" is a very strong argument, because the `Registry` class is already inherently stateful, evident by the fact that the `.prepare` method is always called first within a lifecycle of a `Registry` object. 

There is another reason why I don't quite like having this if statement in the function. If we follow the implementation that Nick gave earlier, we can see that the function is now implicitly dependent on the `_all_checks_mapping` attribute in all cases, which makes testing quite cumbersome: to test this we need to create a `Registry` object because the results of the function rely on the state of the `Registry` object (specifically, the state of the `_all_checks_mapping` attribute). We also have to create tests for cases where the `_all_checks_mapping` attribute is invalid to cover all branchings if we want to achieve 100% branch coverage for this function.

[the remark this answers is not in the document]
tromai: Thanks for the clarification. I think I better understand the picture now. I will work on:
- Ensure that the validation of `_all_checks_mapping` only happens once before any graph traversal (for exclude/include) happens.
- Remove the need to validating `_all_checks_mapping`  within the `get_transitive_closure` method.

tests/slsa_analyzer/checks/test_registry.py
```diff
@@ -200,3 +255,112 @@ def test_validate_check_id_format(self) -> None:
...
+        ([], ["mcn_c_1"], ["mcn_c_1", "mcn_d_1", "mcn_e_1"]),
+        (["mcn_d_1"], ["mcn_c_1"], ["mcn_e_1"]),
+        (["mcn_*"], ["*"], []),
+    ],
+)
+def test_get_final_checks(
+    check_registry: Registry, ex_pats: list[str], in_pats: list[str], final_checks: list[str]
+) -> None:
+    """This method tests the get_final_checks method."""
+    assert check_registry.get_final_checks(ex_pats=ex_pats, in_pats=in_pats).sort() == final_checks.sort()
```
tromai, line 302: This is incorrect as `check_registry.get_final_checks(ex_pats=ex_pats, in_pats=in_pats).sort()` and `final_checks.sort()` would always return None.

src/macaron/slsa_analyzer/registry.py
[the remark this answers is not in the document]
nicallen: To add to Nathan's comment, I would also argue that validating that the nodes of the graph is irrelevant to the task of that function, whether the checks are in `_all_checks_mapping` does not affect the ability to compute a transitive closure. That all the checks in the graph are valid is really an invariant that is the responsibility of the Registry to establish and maintain, before any invocation of `get_transitive_closure`, which is a generic method on graphs. In fact, removing that validation makes the function even more generic, removing the last reference to `self` and allowing `get_transitive_closure` to be moved out of the Registry class.

As for the other validation that `get_transitive_closure` is doing, checking for cycles, if the requirement that the graph be acyclic is also enforced elsewhere by Registry, and we consider `get_transitive_closure` to be a generic function on graphs, then it is entirely possible to just correctly produce the transitive closure for cyclic graphs as well, removing any need for the validation.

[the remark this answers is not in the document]
tromai: > if the requirement that the graph be acyclic is also enforced elsewhere by Registry
This is True. The requirements enforces by the Registry ensure that there would be no cycle when the check graph is built. 
With that being said, I think we could remove the validation for cycles entirely from `get_transitive_closure`, further generalize the function. I can now move `get_transitive_closure` out of the Registry class.

src/macaron/errors.py
```diff
@@ -54,3 +54,15 @@ class DuplicateError(MacaronError):
...
+class CheckRegistryError(MacaronError):
+    """The Check Registry Error class."""
+
+
+class CheckCircularDependencyError(CheckRegistryError):
+    """This error is raised when there is a circular dependency in the registered checks."""
+
+
+class CheckNotExistError(CheckRegistryError):
+    """This error is raised trying to obtain a un-registered check."""
```
benmss, line 68:
```suggestion
    """This error is raised when trying to obtain an un-registered check."""
```

src/macaron/slsa_analyzer/registry.py
```diff
@@ -331,7 +343,166 @@ def _validate_check_relationship(relationship: Any) -> bool:
...
+                        stack.append(parent)
+                    elif parent in stack:
+                        raise CheckCircularDependencyError("cycle nodes detected")
+            else:
+                stack.pop()
+
+        return visited
+
+    def get_final_checks(self, ex_pats: list[str], in_pats: list[str]) -> list[str]:
+        """Return a set of the checks' id to run from exclude and include glob patterns.
```
benmss, line 439:
```suggestion
        """Return a set of the check ids to run from the exclude and include glob patterns.
```

src/macaron/slsa_analyzer/registry.py
```diff
@@ -331,7 +343,166 @@ def _validate_check_relationship(relationship: Any) -> bool:
...
+
+        The exclude and include glob patterns are used to match against the id of registered checks.
+
+        Including a check would effectively include all transitive parents of that check.
+        Excluding a check would effectively exclude all transitive children of that check.
+
+        The final list of check to run would be the included checks minus the excluded checks.
```
benmss, line 446:
```suggestion
        The final list of checks to run would be the included checks minus the excluded checks.
```

src/macaron/slsa_analyzer/registry.py
```diff
@@ -498,25 +677,42 @@ def prepare(self) -> bool:
...
+        ex_pats = defaults.get_list(section="analysis.checks", item="exclude", fallback=[])
+        in_pats = defaults.get_list(section="analysis.checks", item="include", fallback=["*"])
+        try:
+            checks_to_run = registry.get_final_checks(ex_pats, in_pats)
+        except CheckRegistryError as error:
+            logger.error(error)
+            return False
+
+        if len(checks_to_run) == 0:
+            logger.info("There is no check to run according to the exclude/include configuration.")
```
benmss, line 706:
```suggestion
            logger.info("There are no checks to run according to the exclude/include configuration.")
```

src/macaron/slsa_analyzer/registry.py
```diff
@@ -579,5 +775,66 @@ def _should_skip_check(check: BaseCheck, results: dict[str, CheckResult]) -> Ski
...
         return None
 
+    def _get_check_tree_as_dict(self) -> CheckTree:
+        """Return a dictionary representation of the check relationships.
+
+        Returns
+        -------
+        CheckTree
+            A nested dictionary that represent the relationship between
+            checks. Each mapping (K, V) in the returned dictionary has K is the check id and
```
benmss, line 785:
```suggestion
            checks. For each mapping (K, V) in the returned dictionary, K is the check id and
```

src/macaron/slsa_analyzer/registry.py
```diff
@@ -579,5 +775,66 @@ def _should_skip_check(check: BaseCheck, results: dict[str, CheckResult]) -> Ski
...
+            V is a dictionary contains the children of that check.
```
benmss, line 786:
```suggestion
            V is a dictionary that contains the children of that check.
```

src/macaron/slsa_analyzer/registry.py
```diff
@@ -579,5 +775,66 @@ def _should_skip_check(check: BaseCheck, results: dict[str, CheckResult]) -> Ski
...
+
+        Examples
+        --------
+        Given the following checks and its relationships
```
benmss, line 790:
```suggestion
        Given the following checks and their relationships
```

src/macaron/slsa_analyzer/registry.py
```diff
@@ -579,5 +775,66 @@ def _should_skip_check(check: BaseCheck, results: dict[str, CheckResult]) -> Ski
...
+
+        .. code-block::
+
+            mcn_provenance_available_1
+            |-- mcn_provenance_level_three_1
+                |-- mcn_provenance_expectation_1
+            mcn_version_control_system_1
+            |-- mcn_trusted_builder_level_three_1
+
+        The result dictionary will be
```
benmss, line 800:
```suggestion
        The resulting dictionary will be:
```

src/macaron/slsa_analyzer/registry.py
```diff
@@ -579,5 +775,66 @@ def _should_skip_check(check: BaseCheck, results: dict[str, CheckResult]) -> Ski
...
+                },
+                'mcn_version_control_system_1': {
+                    'mcn_trusted_builder_level_three_1': {}
+                },
+            }
+        """
+
+        def _traverse(
+            relation_mapping: Mapping[str, dict[str, CheckResultType]],
+            root: str,
```
benmss, line 818: I would suggest renaming `root` to `node` or similar, as "root" is only really applicable for the initial call to this function.

src/macaron/slsa_analyzer/registry.py
```diff
@@ -434,7 +606,14 @@ def scan(self, target: AnalyzeContext, skipped_checks: list[SkippedInfo]) -> dic
                     return results
                 next_check: BaseCheck = check_queue.get()
 
-                # Look up check results to see if this check should be run
+                # Don't run excluded checks
+                if next_check.check_info.check_id not in self.checks_to_run:
+                    logger.debug("Check %s is disabled from user configutation.", next_check.check_info.check_id)
```
benmss, line 611:
```suggestion
                    logger.debug("Check %s is disabled by user configuration.", next_check.check_info.check_id)
```

src/macaron/errors.py
tromai: Thanks for the comment. However after this [discussion](https://github.com/oracle/macaron/pull/254#discussion_r1487025498) this Exception class has been removed in https://github.com/oracle/macaron/pull/254/commits/0be70cbcb7eba77a441802b447da8f4361050f32

src/macaron/slsa_analyzer/registry.py
tromai: I agree, I will change it to `node` instead.

src/macaron/slsa_analyzer/registry.py
tromai: Addressed in https://github.com/oracle/macaron/pull/254/commits/63b2883b703ee2b96584f00cb34f560f49a87db9

src/macaron/slsa_analyzer/registry.py
tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/63b2883b703ee2b96584f00cb34f560f49a87db9

src/macaron/slsa_analyzer/registry.py
tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/63b2883b703ee2b96584f00cb34f560f49a87db9

src/macaron/slsa_analyzer/registry.py
tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/63b2883b703ee2b96584f00cb34f560f49a87db9

src/macaron/slsa_analyzer/registry.py
tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/63b2883b703ee2b96584f00cb34f560f49a87db9

src/macaron/slsa_analyzer/registry.py
tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/63b2883b703ee2b96584f00cb34f560f49a87db9

src/macaron/slsa_analyzer/registry.py
tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/63b2883b703ee2b96584f00cb34f560f49a87db9

src/macaron/slsa_analyzer/registry.py
tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/63b2883b703ee2b96584f00cb34f560f49a87db9

src/macaron/slsa_analyzer/registry.py
tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/63b2883b703ee2b96584f00cb34f560f49a87db9

tests/slsa_analyzer/checks/test_registry.py
tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/0be70cbcb7eba77a441802b447da8f4361050f32

src/macaron/slsa_analyzer/registry.py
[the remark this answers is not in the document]
tromai: I have implemented the solution suggested in this commit - https://github.com/oracle/macaron/pull/254/commits/0be70cbcb7eba77a441802b447da8f4361050f32

src/macaron/slsa_analyzer/registry.py
```diff
@@ -29,6 +33,9 @@
 logger: logging.Logger = logging.getLogger(__name__)
 
 
+CheckTree = dict[str, "CheckTree"]
```
nathanwn, line 36: This is a bit nit-picking, but I don't think calling this type `CheckTree` is the most accurate. A tree by definition is a single connected component and usually has a "root node" in its representation. Rather, I think this is a `CheckForest`, i.e., a collection of check trees: each key is a root node name, and each value is a forest of subtrees under that root node.
In general, the system of checks in Macaron also makes up a forest of multiple trees, rooted at nodes having no parents (there is a `no_parent_checks` attribute in the `Registry` class storing these root nodes).

nathanwn:
I'm not sure if this is in-scope of this PR, but I am seeing some issues with the attributes `_graph` and `_is_graph_ready` in the `Registry` class:

https://github.com/oracle/macaron/blob/73962d8af12b0fcca08afe54323de266ac6c826b/src/macaron/slsa_analyzer/registry.py#L50C1-L60C28

These attributes serve the purpose of working out the correct execution order of checks. I think the class scope is too wide of a scope for them: they are only needed once we start executing the checks, and not needed when the `Registry` object is created and we are still population checks into the registry. The two attributes `_all_checks_mapping` and `_check_relationships_mapping` already serve the purpose of storing the check graph, so the `_graph` attribute is not needed for this purpose. One evidence of the `_graph` attribute being scoped too wide is the fact that we need to create a deep copy of it whenever we start executing the checks:

https://github.com/oracle/macaron/blob/73962d8af12b0fcca08afe54323de266ac6c826b/src/macaron/slsa_analyzer/registry.py#L471C1-L472C42

If we need to `deepcopy` like this, then it is a sign that this `_graph: TopologicalSorter` object is not supposed to be a long-living object, but rather something that should be used once and discarded. Hence, it is probably better to recreate a new topological sorter from the graph representation (i.e. `_all_checks_mapping` and `_check_relationships_mapping`) every time we need to work out check order.

I realized this issue when I looked at the unit tests that involve `Registry` objects: the creation of a `Registry` object also forces us to create a `TopologicalSorter`, which initially looks weird to me.

Furthermore, now that we have incorporated the include and exclude checks feature, the topological sorter can be created after the set of checks to be run has been worked out, which means it can just take these checks instead of all registered checks. However, I consider this to be not as relevant as an improvement compared to removing `_graph` and `_is_graph_ready` from the class scope of `Registry`.

src/macaron/slsa_analyzer/registry.py
```diff
@@ -498,25 +623,42 @@ def prepare(self) -> bool:
...
+            return False
+
+        if len(checks_to_run) == 0:
+            logger.info("There are no checks to run according to the exclude/include configuration.")
+            return False
+        self.checks_to_run = checks_to_run
+
+        # Store the check tree as dictionary to be used in the HTML report.
+        if not self.check_tree:
+            self.check_tree.update(self._get_check_tree_as_dict())
```
nathanwn, line 658: Hmm, why is it `update` here instead of reassigning `self.check_tree` with the newly created set?

src/macaron/slsa_analyzer/registry.py
nicallen: I don't think the distinction really matters, you can turn any forest into a tree by introducing a special root node whose children are the root nodes of the trees of the forest, which is what is done here (the `check_tree` field of `Registry` is the root node in this case, and does not itself represent a check).

nathanwn: Hmm, yeah with that interpretation calling it `CheckTree` is fine.

nicallen:
> I'm not sure if this is in-scope of this PR, but I am seeing some issues with the attributes `_graph` and `_is_graph_ready` in the `Registry` class...

I agree that the `Registry` class is doing too much and has overly convoluted statefulness going on. As I see it, there are 3 main phases of the `Registry` class state, and different operations are incorrect to invoke during each, making it confusing to use since it is not obvious when something is safe to call or what invariants hold at what times.

The 3 phases are: when checks are being registered (the check graph is mutable, checks referenced may not yet exist, not safe to invoke operations that need a stable, complete graph), after check registration is finished, but topological sorting has not been done, and finally after topological sorting is done (not safe to repeat topological sorting).

An improvement would be to refactor these phases into 3 separate classes, maybe something like `CheckRegistry` which only allows adding checks, then an operation that validates and finalises the graph, returning a `CheckGraph`, which has the operations for traversing the complete graph (and doesn't have the ability to add further checks), and when you actually need to run the checks, a `CheckSchedule` can be constructed which takes the `CheckGraph` and computes the topologically-sorted run order.

src/macaron/slsa_analyzer/registry.py
tromai: Hmmm I am not sure why I use `update` here. I will use assignment instead.

tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/35907f6a57a48f373293eab1046f6939125d9eb8

tromai:
Thanks @nicallen @nathanwn for the feedback regarding the current state of our `Registry` class. I agree with both comments about the limitation of  it. The issues everyone described there are what I encountered (somewhat) during the design and implementation of this feature. I will create a issue for this improvements of registering and scheduling the checks.

src/macaron/slsa_analyzer/registry.py
```diff
@@ -331,7 +344,111 @@ def _validate_check_relationship(relationship: Any) -> bool:
...
+
+        Returns
+        -------
+        set[str]
+            The set of ids for all parent checks.
+        """
+        check = self._all_checks_mapping.get(check_id)
+        if not check:
+            # It won't happen as we have validated the existence of check_id in registry.prepare().
+            return set()
```
nathanwn, line 363: The `dict.get` method allows passing in a default value in case the key does not exist.
```suggestion
        check = self._all_checks_mapping.get(check_id, set())
```

nathanwn: You may be doing this to include the comment that "this branch won't happen". I would argue the comment is kind of irrelevant to what this function is doing because whatever the state of the `self._all_checks_mapping` attribute this function is still correct regardless.

tromai: I agree, I previously put this here before the comments in [here](https://github.com/oracle/macaron/pull/254#discussion_r1487025498) were made. I will address it as you suggested.

nathanwn: Ah, nevermind about the suggestion about using `dict.get`. I think I misunderstood this function earlier.

tromai: All good! I guess this method shouldn't change anything then.

nathanwn (approved):
LGTM. Thanks for working on this PR.

nicallen (approved):

src/macaron/graph/graph.py
behnazh-w, line 1: I'm not sure why you have decided to create this module. Right now it only contains one function to compute reachable nodes, which happens to be using DFS. But depending on the graph, different traversals might be needed. I'm worried that this function is going to be used only once for the check dependencies. So, why not to make it a method of the `Registry` class?

src/macaron/config/defaults.ini
```diff
@@ -475,3 +475,18 @@ hostname = registry.npmjs.org
...
 attestation_endpoint = -/npm/v1/attestations
 request_timeout = 20
+
+# Configuration options for selecting the checks to run.
+# Both the exclude and include are defined as lists of string:
+#   - The exclude list is used to specify the checks that will not run.
+#   - The include list is used to specify the checks will run. Checks are not specified by "include" will be excluded.
+# The final list of checks to run will always be "include" minus "exclude".
+# information on the effect of include/exclude to the check dependencies.
+# Each element of a list is either the full ID of a check (e.g mcn_build_as_code_1) or a glob pattern
```
behnazh-w, line 485: It is not clear how the user can provide the list of checks. I think one example would be helpful.

src/macaron/output_reporter/templates/base_template.html
```diff
@@ -683,6 +690,13 @@
                 document.getElementById("check_report_content").classList.toggle("hidden");
             });
 
+            // Add a listener toggle the display of run checks.
```
behnazh-w, line 693:
```suggestion
           // Add a listener to toggle the display of run checks.
```

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -72,6 +72,12 @@ def __init__(self, output_path: str, build_log_path: str) -> None:
             logger.error("Cannot start the analysis. Exiting ...")
             sys.exit(1)
 
+        logger.info(
+            "The following checks are excluded from user configuration: %s",
```
behnazh-w, line 76:
```suggestion
            "The following checks are excluded based on the user configuration: %s",
```

src/macaron/slsa_analyzer/registry.py
```diff
@@ -331,7 +344,111 @@ def _validate_check_relationship(relationship: Any) -> bool:
...
+        ----------
+        check_id : str
+            The check id we want to obtain the children.
+
+        Returns
+        -------
+        set[str]
+            The set of ids for all children checks.
+        """
+        # If this check is not defined in the check relationships mapping, it mean that it
```
behnazh-w, line 380:
```suggestion
        # If this check is not defined in the check relationships mapping, it means that it
```

src/macaron/slsa_analyzer/registry.py
```diff
@@ -23,6 +25,17 @@
 logger: logging.Logger = logging.getLogger(__name__)
 
 
+class CheckRegistryError(MacaronError):
```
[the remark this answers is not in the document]
tromai: In the latest version, this has been removed. Therefore I think I can resolve this conversation.

tests/e2e/expected_results/slsa-verifier/slsa-verifier_provenance_checks_excluded.json
```diff
@@ -0,0 +1,198 @@
...
+        "excluded_checks": [
+            "mcn_provenance_level_three_1",
+            "mcn_provenance_expectation_1",
+            "mcn_provenance_available_1"
+        ]
+    },
+    "target": {
+        "info": {
+            "full_name": "slsa-framework/slsa-verifier",
+            "local_cloned_path": "../../../../output/git_repos/github_com/slsa-framework/slsa-verifier",
```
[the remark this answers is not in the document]
tromai: This issue has been addressed in the latest version. I have had a look at the expected JSON files and they seems to don't have this sort of relative paths anymore.

tests/e2e/expected_results/slsa-verifier/slsa-verifier_provenance_checks_excluded.json
```diff
@@ -0,0 +1,198 @@
...
+            },
+            {
+                "check_id": "mcn_build_service_1",
+                "num_deps_pass": 0
+            }
+        ],
+        "dep_status": []
+    },
+    "policies_passed": [],
+    "policies_failed": []
```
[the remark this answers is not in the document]
tromai: I have had a look and in the latest version, the expected files don't have these lines anymore.

src/macaron/graph/graph.py
tromai: Thanks for the question!
According to this [discussion](https://github.com/oracle/macaron/pull/254#discussion_r1487025498), I have factored out this generic logic for DFS graph traversal. It is implemented so that it could work on all directed acrylic graphs given that we provide the `get_successors` function for that graph. Within the Registry class, this function is used twice, each with different `get_successors` functions, hence different "graph" (the nodes are the same but the relations between each node are different). 
- https://github.com/oracle/macaron/blob/35907f6a57a48f373293eab1046f6939125d9eb8/src/macaron/slsa_analyzer/registry.py#L425-L428
- https://github.com/oracle/macaron/blob/35907f6a57a48f373293eab1046f6939125d9eb8/src/macaron/slsa_analyzer/registry.py#L438-L441
This function is generic enough (its implementation doesn't have to know about the relation between nodes, e.g. the check relationships), I think it doesn't need to belong to the Registry class. 
In additions, I don't think it would cause us any issue if this function is only used within the Registry class. It's a generic function so essentially it can be used at any other location in the future and putting it in a separated module can prepare us for that scenario. Unit testing would be a bit clearer too as we don't put the unit tests in the same place as the registry unit tests.

src/macaron/config/defaults.ini
```diff
@@ -475,3 +475,18 @@ hostname = registry.npmjs.org
 # The attestation REST API.
 attestation_endpoint = -/npm/v1/attestations
 request_timeout = 20
+
+# Configuration options for selecting the checks to run.
+# Both the exclude and include are defined as lists of string:
+#   - The exclude list is used to specify the checks that will not run.
+#   - The include list is used to specify the checks will run. Checks are not specified by "include" will be excluded.
+# The final list of checks to run will always be "include" minus "exclude".
+# information on the effect of include/exclude to the check dependencies.
```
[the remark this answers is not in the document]
tromai: Oops I think I leave it here by accident. I will remove it.

src/macaron/graph/graph.py
behnazh-w: I'm not sure if this function is generic enough to justify creating a separate module for graphs. What if I need to get the reachable nodes in a graph but  the traversal should be BFS because the order matters? I can't use this function anymore and should implement a new `get_reachable_nodes` function.

tromai: What if I need to get the reachable nodes in a graph but the traversal should be BFS because the order matters?
I think in this case, we are still solving a graph problem, but with a different requirement on the traversal order. Therefore, I believe the original `get_reachable_nodes` function would still be suitable to be in a dedicated graph module as it works with all directed acyclic graphs. 
I agree that if we need a different traversal order, we would need to implement a new function to do that. But that function would be perfectly fit within this graph module if it's made generic enough. I can rename the original `get_reachable_nodes` function to `get_reachable_nodes_dfs` if it helps making the intention clearer.

scripts/dev_scripts/integration_tests.sh
```diff
@@ -577,6 +577,17 @@ fi
 rm -rf "$SOURCE_REPO"
 rm -rf "$TARGET_REPO"
 
+echo -e "\n----------------------------------------------------------------------------------"
+echo "Running the analysis with all checks excluded should returns an error code."
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/7c94792b6406387beec45d59fcbc2002323e8491

src/macaron/config/defaults.ini
```diff
@@ -475,3 +475,18 @@ hostname = registry.npmjs.org
 # The attestation REST API.
 attestation_endpoint = -/npm/v1/attestations
 request_timeout = 20
+
+# Configuration options for selecting the checks to run.
+# Both the exclude and include are defined as lists of string:
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/7c94792b6406387beec45d59fcbc2002323e8491

src/macaron/config/defaults.ini
```diff
@@ -475,3 +475,18 @@ hostname = registry.npmjs.org
...
+#   - The exclude list is used to specify the checks that will not run.
+#   - The include list is used to specify the checks will run. Checks are not specified by "include" will be excluded.
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/7c94792b6406387beec45d59fcbc2002323e8491

src/macaron/config/defaults.ini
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/7c94792b6406387beec45d59fcbc2002323e8491

src/macaron/output_reporter/templates/base_template.html
tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/7c94792b6406387beec45d59fcbc2002323e8491

src/macaron/slsa_analyzer/analyzer.py
tromai: Fixed in https://github.com/oracle/macaron/pull/254/commits/7c94792b6406387beec45d59fcbc2002323e8491

src/macaron/graph/graph.py
behnazh-w: I still don't find it necessary to create a separate `graph` module for implementing a DFS traversal. Often times the graph representations can be different depending on the problem. So, I'm not sure if this method will get called anywhere else. How about adding this module when we find a good use case?

nathanwn: Looking at the signature of this function, I think it is fine as is. Neither the signature of the function, the docstring, or the corresponding unit tests implies that the result must follow a particular traversal order. Therefore, the consumer of this interface should not rely on the order of nodes returned from this function.

If we rename it to `get_reachable_nodes_dfs`, meaning it does imply a particular order, then we will need to mention that in the docstring and reflect that in the unit tests, which we don't at the moment because the order of traversal is not part of the initial motivation for the creation of this function. We also need to define what exactly "DFS order" means because the order of DFS traversal is not unique without some specific constraints (e.g. the adjacency list representation of the graph must be traversed in some particular order).

I think if we want to traverse the graph in whatever particular order, then it is very likely that we will need to reimplement something new and test it to make sure it traverses in the order that we want. However, I don't see that new function invalidating the purpose of this function, nor the creation of that new traversal function to be non-desirable.

About whether separating this function to a new module. I don't have a strong opinion on this, but I think it is probably best if we go with Behnaz's suggestion of not separating this to a new module until we have another use case for it which concretely proves it effectiveness as a generic implementation.

src/macaron/config/defaults.ini
tromai: I have added a few examples in https://github.com/oracle/macaron/pull/254/commits/5efae2c587d19d1e172a7eb749417f62eeb0ae96

src/macaron/graph/graph.py
tromai: Thanks everyone for the feedback. I have move this function back into the Registry class.

src/macaron/graph/graph.py
```diff
@@ -0,0 +1,50 @@
...
+"""This module contains generic methods for graph operations."""
+
+
+from collections.abc import Callable, Iterable
+from typing import TypeVar
+
+T = TypeVar("T")
+
+
+def get_transitive_closure(
```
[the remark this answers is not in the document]
tromai: Addressed in https://github.com/oracle/macaron/pull/254/commits/9347251dc789dbe18710d39e72173fcca98f5b89

behnazh-w (approved):

PR #260, opened 2023-05-30
feat: add problog rule inference to build as code check

sophie-bates:
## Priority (v1)
- [ ] Refactor build as code `run_check` method into smaller subchecks.
- [ ] Use ProbLog predicates to receive subcheck values from within prolog string.
- [ ] Write ProbLog rules to evaluate the check's overall confidence score.
- [ ] Update BuildAsCodeTable to receive the information.
- [ ] Use overall confidence score to make a decision about check passing.
- [ ] Update check_result to store confidence score.

## Later
- [ ] Update tests. 
- [ ] Refine refactoring - can things be done more efficiently/effectively?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,7 @@ dependencies = [
     "jinja2 >=3.1.2,<4.0.0",
     "SQLAlchemy >=2.0.0,<3.0.0",
     "defusedxml >=0.7.1,<1.0.0",
+    "problog >=2.2.4,<3.0.0"
 ]
 keywords = []
 # https://pypi.org/classifiers/


src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -130,164 +86,110 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
...
+                )
+
+                # TODO: query each of the methods, and take the values from the one with the highest confidence.
+                confidence_score = 0.0
+                result = get_evaluatable().create_from(prolog_string).evaluate()
+                for key, value in result.items():
+                    print(key, value)
+                    if str(key) == "build_as_code_check":
+                        confidence_score = float(value)
+                results = vars(build_as_code_subchecks.build_as_code_subcheck_results)
```
tromai, line 128: It's okay to use this built in function for now. But we should keep in mind that this function would raise TypeError when `__dict__` attribute could not be found within the target object.

sophie-bates: Thanks for your comment Nhan, I'm actually about to push a change that will mean I'm no longer using this anyway so hopefully it's all good.

src/macaron/slsa_analyzer/checks/build_as_code_subchecks.py
```diff
@@ -0,0 +1,246 @@
...
+
+    # store analyze context
+    def __init__(self, ctx: AnalyzeContext, ci_info: CIInfo) -> None:
+        self.ctx = ctx
+        self.build_tool: BaseBuildTool = ctx.dynamic_data["build_spec"].get("tool")  # type: ignore
+        self.ci_services = ctx.dynamic_data["ci_services"]
+        self.check_results: dict = {}  # Update this with each check.
+        self.ci_info = ci_info
+        self.ci_service = ci_info["service"]
+        self.failed_check = 0.0
```
tromai, line 79: I am a bit confused by this attribute `failed_check`. Could you document what this attribute is used for and what is the value that it contains?
From a first look, I was expecting a `boolean` type instead of a float.

src/macaron/slsa_analyzer/checks/build_as_code_subchecks.py
```diff
@@ -0,0 +1,246 @@
...
+
+        # TODO: Before each check is run, check whether a certainty result already exists in self.check_results
+        # to avoid re-running unecessarily.
+
+    def ci_parsed(self) -> float:
+        """Check whether parsing is supported for this CI service's CI config files."""
+        check_certainty = 1.0
+        # If this check has already been run on this repo, return certainty.
+
+        justification: list[str | dict[str, str]] = ["The CI workflow files for this CI service are parsed."]
```
tromai, line 91: We could define the justification inside the if branch so that we don't need to do it when `self.ci_info["bash_commands"]` is not available.

src/macaron/slsa_analyzer/checks/build_as_code_subchecks.py
sophie-bates: Yes, it's the value that should be returned if the sub check fails.

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -130,164 +86,98 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
...
-                            return CheckResultType.PASSED
+                        ]
+
+                # TODO: compile all justifications
+                # check_result["justification"].append()
+
+                # TODO: Investigate using proofs
+
+                # Check whether the confidence score is greater than the minimum threshold for this check.
+                if confidence_score >= self.confidence_score_threshold:
```
behnazh-w, line 178: At this line, if a CI service has a `confidence_score` larger than the threshold, but smaller than other CI services, wouldn't the check return the result for the smaller `confidence_score`?

sophie-bates: True. I can update this implementation to store all of the required information (i.e. deploy commands) for each `ci_service` and then populate the `BuilasAsCodeTable` and `AnalyzeContext` with whichever `ci_service` has the highest `confidence_score`?

behnazh-w: Sure, however I don't understand why `AnalyzeContext` needs to be updated?

sophie-bates: I just meant where the `ci_info` information is updated: https://github.com/oracle-samples/macaron/blob/add-problog-rule-inference-for-build-as-code/src/macaron/slsa_analyzer/checks/build_as_code_check.py#L145-L159.

behnazh-w: Oh, that's the inferred provenance representation. We don't update `ci_info` itself.

PR #263, opened 2023-05-30
feat: use database to store and retrieve found repos

benmss:
This adds the third part of the repo finding extension. Making remote calls to retrieve POMs is a costly operation that should be avoided when possible. By adding found repositories to the Macaron database, it can first be polled before any remote calls are needed.


src/macaron/dependency_analyzer/dependency_resolver.py
```diff
@@ -173,16 +178,29 @@ def add_latest_version(
...
+        artifact = f"{item['group']}:{item['name']}"
         if f"{item['group']}:{item['name']}" in defaults.get_list("repofinder.java", "artifact_ignore_list"):
-            logger.debug("Skipping GAV: %s", gav)
+            logger.debug("Skipping artifact: %s", artifact)
             return
 
+        if db_man and defaults.getboolean("repofinder.java", "use_database"):
+            # Perform database lookup
+            query = sqlalchemy.text(
+                "SELECT remote_path FROM _repository WHERE namespace = :group and name = :artifact"
```
behnazh-w, line 194: Instead of raw queries, we should use SQLAlchemy's [ORM queries](https://docs.sqlalchemy.org/en/20/orm/session_basics.html#querying), which performs proper validation.

behnazh-w:
We've decided to review this PR once the new data model [PR](https://github.com/oracle/macaron/pull/305) is merged.

PR #265, opened 2023-05-30
feat(release): generate SLSA provenance for the Docker image

### LINKED ISSUE #264: Generate SLSA provenance for the Docker image

behnazh-w:
Closes #264

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -106,6 +106,8 @@ jobs:
     outputs:
       release-tag: ${{ steps.release-info.outputs.release-tag }}
       release-url: ${{ steps.release-info.outputs.release-url }}
+      image-name: ${{ steps.push-docker.outputs.image-name }}
+      image-digest: ${{ steps.push-docker.outputs.image-digest }}
     runs-on: ubuntu-latest
     permissions:
       contents: write # To publish release notes.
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -153,9 +155,18 @@ jobs:
 
     # Push the Docker image as part of the release pipeline.
     - name: Push the Docker image
+      id: push-docker
       env:
         RELEASE_TAG: ${{ steps.release-info.outputs.release-tag }}
-      run: make push-docker
+      run: |
+        make push-docker
+        IMAGE_AND_DIGEST=$(docker inspect --format='{{index .RepoDigests 0}}' "$IMAGE_NAME")
```
tromai, line 163: When I try to build the Docker image (to create an image called `ghcr.io/oracle-samples/macaron
`) and run this script manually, it cannot resolve. This is the error that I have:
```
docker inspect --format='{{index .RepoDigests 0}}' ghcr.io/oracle-samples/macaron

template parsing error: template: :1:2: executing "" at <index .RepoDigests 0>: error calling index: reflect: slice index out of range
```
However, if I run this script with an image that currently exists in ghcr (i.e `ghcr.io/oracle-samples/macaron-base
`, it would work normally.
It is correct to understand that `make push-docker` will push the image to ghcr previously and this script will work fine in that case?

behnazh-w: > It is correct to understand that `make push-docker` will push the image to ghcr previously and this script will work fine in that case?

Yes, that's right. Certain information, such as `RepoDigests` is not finalized until the image is pushed. That's why you get an empty field before pushing the image.

tromai (approved):

Issue #250, opened 2023-05-29
Using the show-prelude option in the verify-policy command still require the path to the policy file 


tromai:
**Description**
In the `verify-policy` action, there is the `--show-prelude` flag for dumping out the policy preclude.
In this method, it indicates that showing the preclude does not require any user policy file - https://github.com/oracle-samples/macaron/blob/907eedf543131a2dd8d406bfa3ad54b3df8fae9a/src/macaron/policy_engine/policy_engine.py#L159 .
However, when we try to use the `--show-prelude` flag, this error happens:
```
usage: macaron verify-policy [-h] -d DATABASE -f FILE [-s]
macaron verify-policy: error: the following arguments are required: -f/--file
```

PR #270, opened 2023-06-01
chore: remove running the policy engine when --show-prelude is provided

tromai:
**Description**
I fixed issue #250 by assigning both `--file` and `--show-prelude` flags in the `verify-policy` action command to a mutual exclusive group.
This will effective only allow the user to use either `--file` or `--show-prelude` at one time.
I also refactored the code to generate the prelude from a database to a different method rather than putting it inside the same method used for running the policy engine against the policy file provided in `--file`.


behnazh-w:
As discussed, the scope for this PR/commit should not be `fix` because we don't want it to appear in the changelog.

tromai:
Thanks @behnazh-w, I have renamed the PR and change the commit message.

behnazh-w (approved):

PR #299, opened 2023-06-08
fix: fixes the result summary for UNKNOWN check results

behnazh-w:


tests/e2e/expected_results/slsa-verifier/slsa-verifier_yaml_invalid.json
```diff
@@ -1686,7 +1686,7 @@
                 "FAILED": 0,
                 "PASSED": 7,
                 "SKIPPED": 0,
-                "UNKNOWN": 0
+                "UNKNOWN": 1
```
tromai, line 1689: I wonder which check was flagged as UNKNOWN in these cases?

tromai: I think it's the Provenance Expectation check when the expectation file is not valid.

tromai (approved):

PR #307, opened 2023-06-13
chore: change repository links to oracle/macaron

behnazh-w:


tromai:
I believe re running the `build_docker_image.yaml` workflow would fix the issue as the `macaron-base` image was not available before? Otherwise everything LGTM

tromai (approved):

behnazh-w:
> I believe re running the `build_docker_image.yaml` workflow would fix the issue as the `macaron-base` image was not available before? Otherwise everything LGTM

We had to also change the digest of the base image for it to work. Separately I pushed the base image from `oracle-samples/macaron` registry back to the registry. So we don't need to update the digest anymore.

PR #308, opened 2023-06-13
chore: update the documentation to make the repo compliant

spavlusieva:


behnazh-w (commented):
Thanks @spavlusieva . Following our branching model, can you please change the target branch to `staging`. We will sync `main` with `staging` in a separate PR.

Also, we follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) both for PR title and commit message, otherwise the GitHub Action checks will fail, for example:
```
chore: delete build_spec.yaml 
```
Should work.

behnazh-w:
@spavlusieva All the commit messages should also follow conventional commits for the check to pass.

PR #316, opened 2023-06-15
feat: add support for cloning GitLab repositories

### LINKED ISSUE #301: Add support for cloning from GitLab
Macaron is currently only supporting cloning from GitHub. We need to add support for cloning from other git services.
This issue is about adding support for cloning from GitLab.

nathanwn:
Closes #301.

# Description

This PR adds support for cloning GitLab repositories in Macaron. Specifically, GitLab repositories can now be cloned as either main targets or dependencies in a `macaron analyze` run.

This feature comes with the following additions/changes:

## `.ini` configuration

* The old `[git]` section in the `.ini` configuration will now be replaced by the new `[git_service.*]`  sections.
* Each `[git_service]` section corresponds to a different git service with a `domain` key. This domain is used to determine if a Git service hosts a repo given the repo's Git URL.
* If the `domain` key is not found in any `[git_service.*]`  section, the command `macaron analyze`  will error right away before any analysis is done.
* Users can configure 2 different GitLab services in the `[git_service.gitlab.*]`  sections: a public one and a private one. The domain of the public GitLab service is `gitlab.com` by default.

## Access Token Environment Variables

An access token for the public GitLab service is not required when it is enabled. However, an access token for the private GitLab service is always required when it is enabled.

We accept two environment variables storing the access tokens that correspond to the two GitLab services:
  * `MCN_PUBLIC_GITLAB_TOKEN`  for the public GitLab service.
  * `MCN_PRIVATE_GITLAB_TOKEN`  for the private GitLab service.

The `MCN_PRIVATE_GITLAB_TOKEN` environment variable must be set when the `[git_service.gitlab.private]` section is enabled in the `.ini` config. If this is not the case, the command `macaron analyze`  will error right away before any analysis is done.

### DIFF
--- a/src/macaron/__main__.py
+++ b/src/macaron/__main__.py
@@ -143,7 +143,7 @@ def perform_action(action_args: argparse.Namespace) -> None:
                 for git_service in GIT_SERVICES:
                     git_service.load_defaults()
             except ConfigurationError as error:
-                logger.error(str(error))
+                logger.error(error)
                 sys.exit(os.EX_USAGE)
 
             analyze_slsa_levels_single(action_args)
--- a/src/macaron/__main__.py
+++ b/src/macaron/__main__.py
@@ -16,10 +16,12 @@ import macaron
 from macaron.config.defaults import create_defaults, load_defaults
 from macaron.config.global_config import global_config
 from macaron.config.target_config import TARGET_CONFIG_SCHEMA
+from macaron.errors import ConfigurationError
 from macaron.output_reporter.reporter import HTMLReporter, JSONReporter, PolicyReporter
 from macaron.parsers.yaml.loader import YamlLoader
 from macaron.policy_engine.policy_engine import run_policy_engine, show_prelude
 from macaron.slsa_analyzer.analyzer import Analyzer
+from macaron.slsa_analyzer.git_service import GIT_SERVICES
 
 logger: logging.Logger = logging.getLogger(__name__)
 
@@ -134,6 +136,16 @@ def perform_action(action_args: argparse.Namespace) -> None:
                 logger.error("GitHub access token not set.")
                 sys.exit(os.EX_USAGE)
             global_config.gh_token = gh_token
+
+            # TODO: Here we should try to statically analyze the config before
+            # actually running the analysis.
+            try:
+                for git_service in GIT_SERVICES:
+                    git_service.load_defaults()
+            except ConfigurationError as error:
+                logger.error(str(error))
+                sys.exit(os.EX_USAGE)
+
             analyze_slsa_levels_single(action_args)
         case _:
             logger.error("Macaron does not support command option %s.", action_args.action)
--- a/src/macaron/errors.py
+++ b/src/macaron/errors.py
@@ -22,3 +22,7 @@ class CUEExpectationError(MacaronError):
 
 class CUERuntimeError(MacaronError):
     """Happens when there are errors in CUE expectation validation."""
+
+
+class ConfigurationError(MacaronError):
+    """Happens when there is an error in the configuration (.ini) file."""
--- a/src/macaron/errors.py
+++ b/src/macaron/errors.py
@@ -26,3 +26,7 @@ class CUERuntimeError(MacaronError):
 
 class ConfigurationError(MacaronError):
     """Happens when there is an error in the configuration (.ini) file."""
+
+
+class CloneError(MacaronError):
+    """Happens when cannot clone a git repository."""
--- a/src/macaron/slsa_analyzer/git_service/__init__.py
+++ b/src/macaron/slsa_analyzer/git_service/__init__.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2022 - 2022, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """The git_service package contains the supported git services for Macaron."""
@@ -6,8 +6,8 @@
 from .base_git_service import BaseGitService
 from .bitbucket import BitBucket
 from .github import GitHub
-from .gitlab import GitLab
+from .gitlab import PrivateGitLab, PublicGitLab
 
 # The list of supported git services. The order of the list determines the order
 # in which each git service is checked against the target repository.
-GIT_SERVICES: list[BaseGitService] = [GitHub(), GitLab(), BitBucket()]
+GIT_SERVICES: list[BaseGitService] = [GitHub(), PublicGitLab(), PrivateGitLab(), BitBucket()]
--- a/src/macaron/slsa_analyzer/git_service/base_git_service.py
+++ b/src/macaron/slsa_analyzer/git_service/base_git_service.py
@@ -6,7 +6,7 @@
 from abc import abstractmethod
 
 from macaron.config.defaults import defaults
-from macaron.errors import ConfigurationError
+from macaron.errors import CloneError, ConfigurationError
 from macaron.slsa_analyzer import git_url
 
 
@@ -109,6 +109,25 @@ class BaseGitService:
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def clone_repo(self, clone_dir: str, url: str) -> None:
+        """Clone a repository.
+
+        Parameters
+        ----------
+        clone_dir: str
+            The name of the directory to clone into.
+            This is equivalent to the <directory> argument of ``git clone``.
+        url : str
+            The url to the repository.
+
+        Raises
+        ------
+        CloneError
+            If there is an error cloning the repo.
+        """
+        raise NotImplementedError()
+
 
 class NoneGitService(BaseGitService):
     """This class can be used to initialize an empty git service."""
@@ -153,3 +172,16 @@ class NoneGitService(BaseGitService):
             True if the repo can be cloned, else False.
         """
         return False
+
+    def clone_repo(self, _clone_dir: str, url: str) -> None:
+        """Clone a repo.
+
+        In this particular case, since this class represents a ``None`` git service,
+        we do nothing but raise a ``CloneError``.
+
+        Raises
+        ------
+        CloneError
+            Always raise, since this method should not be used to clone any repository.
+        """
+        raise CloneError(f"Internal error encountered when cloning the repo '{url}'.")
--- a/src/macaron/slsa_analyzer/git_service/base_git_service.py
+++ b/src/macaron/slsa_analyzer/git_service/base_git_service.py
@@ -7,6 +7,7 @@ from abc import abstractmethod
 
 from macaron.config.defaults import defaults
 from macaron.errors import ConfigurationError
+from macaron.slsa_analyzer import git_url
 
 
 class BaseGitService:
@@ -67,7 +68,10 @@ class BaseGitService:
         return domain
 
     def is_detected(self, url: str) -> bool:
-        """Return True if the remote repo is using this git service.
+        """Check if the remote repo at the given ``url`` is hosted on this git service.
+
+        This check is done by checking the URL of the repo against the domain of this
+        git service.
 
         Parameters
         ----------
@@ -77,9 +81,17 @@ class BaseGitService:
         Returns
         -------
         bool
-            True if this git service is detected else False.
+            True if the repo is indeed hosted on this git service.
         """
-        raise NotImplementedError
+        if self.domain is None:
+            return False
+        return (
+            git_url.parse_remote_url(
+                url,
+                allowed_git_service_domains=[self.domain],
+            )
+            is not None
+        )
 
     @abstractmethod
     def can_clone_remote_repo(self, url: str) -> bool:
--- a/src/macaron/slsa_analyzer/git_service/base_git_service.py
+++ b/src/macaron/slsa_analyzer/git_service/base_git_service.py
@@ -32,7 +32,7 @@ class BaseGitService:
     def load_domain(self, section_name: str) -> str | None:
         """Load the domain of the git service from the ini configuration section ``section_name``.
 
-        The section may or may not be available in the configuration. In both case,
+        The section may or may not be available in the configuration. In both cases,
         the method should not raise ``ConfigurationError``.
 
         Meanwhile, if the section is present but there is a schema violation (e.g. a key such as
--- a/src/macaron/slsa_analyzer/git_service/base_git_service.py
+++ b/src/macaron/slsa_analyzer/git_service/base_git_service.py
@@ -94,22 +94,6 @@ class BaseGitService:
         )
 
     @abstractmethod
-    def can_clone_remote_repo(self, url: str) -> bool:
-        """Return True if the remote repository can be cloned.
-
-        Parameters
-        ----------
-        url : str
-            The remote url.
-
-        Returns
-        -------
-        bool
-            True if the repo can be cloned, else False.
-        """
-        raise NotImplementedError
-
-    @abstractmethod
     def clone_repo(self, clone_dir: str, url: str) -> None:
         """Clone a repository.
 
@@ -155,21 +139,6 @@ class NoneGitService(BaseGitService):
         -------
         bool
             True if this git service is detected else False.
-        """
-        return False
-
-    def can_clone_remote_repo(self, url: str) -> bool:
-        """Return True if the remote repository can be cloned.
-
-        Parameters
-        ----------
-        url : str
-            The remote url.
-
-        Returns
-        -------
-        bool
-            True if the repo can be cloned, else False.
         """
         return False
 
--- a/src/macaron/slsa_analyzer/git_service/base_git_service.py
+++ b/src/macaron/slsa_analyzer/git_service/base_git_service.py
@@ -1,9 +1,12 @@
-# Copyright (c) 2022 - 2022, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """This module contains the BaseGitService class to be inherited by a git service."""
 
 from abc import abstractmethod
+
+from macaron.config.defaults import defaults
+from macaron.errors import ConfigurationError
 
 
 class BaseGitService:
@@ -18,13 +21,51 @@ class BaseGitService:
             The name of the git service.
         """
         self.name = name
+        self.domain: str | None = None
 
     @abstractmethod
     def load_defaults(self) -> None:
-        """Load the default values from defaults.ini."""
+        """Load the values for this git service from the ini configuration."""
         raise NotImplementedError
 
-    @abstractmethod
+    def load_domain(self, section_name: str) -> str | None:
+        """Load the domain of the git service from the ini configuration section ``section_name``.
+
+        The section may or may not be available in the configuration. In both case,
+        the method should not raise ``ConfigurationError``.
+
+        Meanwhile, if the section is present but there is a schema violation (e.g. a key such as
+        ``domain`` is missing), this method will raise a ``ConfigurationError``.
+
+        Parameters
+        ----------
+        section_name : str
+            The name of the git service section in the ini configuration file.
+
+        Returns
+        -------
+        str | None
+            The domain. This can be ``None`` if the git section is not found in
+            the ini configuration file, meaning the user does not enable the
+            corresponding git service.
+
+        Raises
+        ------
+        ConfigurationError
+            If there is a schema violation in the git service section.
+        """
+        if not defaults.has_section(section_name):
+            # We do not raise ConfigurationError here because it is not compulsory
+            # to have all available git services in the ini config.
+            return None
+        section = defaults[section_name]
+        domain = section.get("domain")
+        if not domain:
+            raise ConfigurationError(
+                f'The "domain" key is missing in section [{section_name}] of the .ini configuration file.'
+            )
+        return domain
+
     def is_detected(self, url: str) -> bool:
         """Return True if the remote repo is using this git service.
 
@@ -65,7 +106,11 @@ class NoneGitService(BaseGitService):
         super().__init__("")
 
     def load_defaults(self) -> None:
-        """Load the default values from defaults.ini."""
+        """Load the values for this git service from the ini configuration.
+
+        In this particular case, since this class represents a ``None`` git service,
+        we do nothing.
+        """
 
     def is_detected(self, url: str) -> bool:
         """Return True if the remote repo is using this git service.
--- a/src/macaron/slsa_analyzer/git_service/base_git_service.py
+++ b/src/macaron/slsa_analyzer/git_service/base_git_service.py
@@ -46,7 +46,7 @@ class BaseGitService:
         Returns
         -------
         str | None
-            The domain. This can be ``None`` if the git section is not found in
+            The domain. This can be ``None`` if the git service section is not found in
             the ini configuration file, meaning the user does not enable the
             corresponding git service.
 
--- a/src/macaron/slsa_analyzer/git_service/gitlab.py
+++ b/src/macaron/slsa_analyzer/git_service/gitlab.py
@@ -22,7 +22,6 @@ import os
 from abc import abstractmethod
 
 from macaron.errors import ConfigurationError
-from macaron.slsa_analyzer import git_url
 from macaron.slsa_analyzer.git_service.base_git_service import BaseGitService
 
 
@@ -54,24 +53,7 @@ class GitLab(BaseGitService):
         """
         return False
 
-    def is_detected(self, url: str) -> bool:
-        """Return True if the remote repo is using this git service.
 
-        Parameters
-        ----------
-        url : str
-            The url of the remote repo.
-
-        Returns
-        -------
-        bool
-            True if this git service is detected else False.
-        """
-        parsed_url = git_url.parse_remote_url(url)
-        if not parsed_url or self.name not in parsed_url.netloc:
-            return False
-
-        return True
 
 
 class PrivateGitLab(GitLab):
--- a/src/macaron/slsa_analyzer/git_service/gitlab.py
+++ b/src/macaron/slsa_analyzer/git_service/gitlab.py
@@ -21,6 +21,7 @@ is in the ``[git_service.gitlab.private]`` section.
 import logging
 import os
 from abc import abstractmethod
+from urllib.parse import ParseResult, urlunparse
 
 from macaron.errors import CloneError, ConfigurationError
 from macaron.slsa_analyzer import git_url
@@ -75,11 +76,23 @@ class GitLab(BaseGitService):
                 f"Cannot clone the repo '{url}' due to the URL format being invalid or not supported by Macaron."
             )
 
+        # Construct clone URL from ``urlparse`` result, with or without an access token.
+        # https://docs.gitlab.com/ee/gitlab-basics/start-using-git.html#clone-using-a-token
         if self.access_token:
-            # https://docs.gitlab.com/ee/gitlab-basics/start-using-git.html#clone-using-a-token
-            clone_url = f"https://oauth2:{self.access_token}@{self.domain}/{url_parse_result.path}"
+            clone_url_netloc = f"oauth2:{self.access_token}@{self.domain}"
         else:
-            clone_url = f"https://{self.domain}/{url_parse_result.path}"
+            clone_url_netloc = self.domain
+
+        clone_url = urlunparse(
+            ParseResult(
+                scheme=url_parse_result.scheme,
+                netloc=clone_url_netloc,
+                path=url_parse_result.path,
+                params="",
+                query="",
+                fragment="",
+            )
+        )
 
         return clone_url
 
--- a/src/macaron/slsa_analyzer/git_service/gitlab.py
+++ b/src/macaron/slsa_analyzer/git_service/gitlab.py
@@ -39,21 +39,6 @@ class GitLab(BaseGitService):
         """Load the .ini configuration."""
         raise NotImplementedError()
 
-    def can_clone_remote_repo(self, url: str) -> bool:
-        """Return True if the remote repository can be cloned.
-
-        Parameters
-        ----------
-        url : str
-            The remote url.
-
-        Returns
-        -------
-        bool
-            True if the repo can be cloned, else False.
-        """
-        return False
-
     def construct_clone_url(self, url: str) -> str:
         """Construct a clone URL for GitLab, with or without access token.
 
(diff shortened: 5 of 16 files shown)


src/macaron/slsa_analyzer/git_service/__init__.py
```diff
@@ -1,13 +1,115 @@
...
+        git_service_section = config[section_name]
+
+        # Domain is compulsory in every ``git_service.*`` section.
+        domain = git_service_section.get("domain")
+        if not domain:
+            raise ConfigurationError(
+                f'The "domain" key is missing in section [{section_name}] of the .ini configuration file.'
+            )
+
+        if section_name.startswith("git_service.github"):
```
behnazh-w, line 59: What's the reason to move loading defaults here and using `if`-`else` to handle each service separately?  The intention to have a [`load_defaults()` function](https://github.com/oracle/macaron/pull/316/files#diff-833baa746fec95ad48a6fdccb975226a06a0668dd98a214103275861689de193L23) to be implemented per Git service was to avoid such an `if-else` implementation.

nathanwn: The way I do it is: I create one object of the `GitLab` class for each GitLab service (public or private). Each `GitLab` object currently does not know which section of the config it should look into (either `[git_service.gitlab.public]` or `[git_service.gitlab.private]`) to populate itself with the domain and access token.

behnazh-w: You can set the values like [here](https://github.com/oracle/macaron/blob/2f658bde1f185a0d265ee05f8f31eb02bfd7fd96/src/macaron/slsa_analyzer/ci_service/github_actions.py#L85) as part of the `load_defaults` function. You can also have two subclasses for the public and private versions of  the `GitLab` for handling the tokens differently.

nathanwn: There is another reason for this `load_git_services` function, which is configuration validation (you can see what I do [here](https://github.com/oracle/macaron/blob/6754bb2e9b026a13f97f070f53f9c63ad33f8942/src/macaron/__main__.py#L140)). The way I think about it, this validation should be done before any git service object is created, not the other way around.

behnazh-w: The validation should happen earlier [here](https://github.com/oracle/macaron/blob/2f658bde1f185a0d265ee05f8f31eb02bfd7fd96/src/macaron/__main__.py#L278). I will address that in a separate PR as it is not directly related to this feature.

nathanwn: > You can also have two subclasses for the public and private versions of GitLab for handling the tokens differently.
I don't like the idea of introducing another level of abstraction here.
In addition, doing it like this means we must always "speculatively" create 1 public GitLab service object and 1 private GitLab service object. In reality, neither can be necessary at all because a user does not enable either of them in their config.
Also by having this `load_defaults`, we are somehow introducing different states to the git service objects, i.e. after being created with the constructor, the git service object is not yet ready to be used as it also has to call this `load_defaults` method to get ready. 
Another issue is: what if the configuration is invalid, e.g. there's no domain in the section? Do we throw the object away?

nathanwn: > The validation of the ini file happens [here](https://github.com/oracle/macaron/blob/c1b5435961e2b61f5377f0cc8a441c0fff28aa55/src/macaron/__main__.py#L340) already as part of the load_defaults function.

By validation, I mean also checking the semantic of the configuration file, not just checking whether we can load it in or not.

nathanwn: Previously, the `load_defaults` method, although not really doing anything, getting called over and over again [here](https://github.com/oracle/macaron/blob/102853cfeaf318391b8d66d2e6302b2ce4e5c359/src/macaron/slsa_analyzer/analyzer.py#L580). The call stack here is `get_git_service <- _prepare_repo <- run_single <- run`. `run_single` is called multiple times, not just for the main repo but also for its dependency.
This is not great. Optimally, loading config should only be done once when the object is created, or right after the fact.
I feel like the fact that we are introducing states into the object is part of the reason why things like this can happen. We either need to be extra careful, or we don't have the "created but config not loaded" state in the first place.

behnazh-w: Two points:

1. If the domain provided for a Git service in the ini file is not valid, Macaron run should not proceed and exit early.
2. The current design in general assumes certain objects exist, so that we can iterate through them without constantly checking if they are `None` or not. Please follow this design.

src/macaron/slsa_analyzer/git_service/__init__.py
```diff
@@ -1,13 +1,115 @@
...
+__all__ = [
+    "BaseGitService",
+    "BitBucket",
+    "GitHub",
+    "GitLab",
+    "NoneGitService",
+]
+
+
+def load_git_services(config: ConfigParser) -> list[BaseGitService]:
```
tromai, line 25: From my understanding, this method and `load_git_service_domains` could be called multiple times and it will try to read the os.env value and the `.ini` config value all over again the next time we call it right? 
For example, `load_git_services` is called 1 time before the analysis start to see if there is any error. After that, it is called against in `Analyzer.get_git_service()`. Does this mean that the first one and the second one would result in different results if there are changes to the env variables we read :thinking: ?

src/macaron/slsa_analyzer/git_service/__init__.py
tromai: I think the reason why we need to do it differently this time is because:
- The initialization of the base git services depend on the combination of values we have in the `.ini` config file especially in the case of GitLab private and public. Therefore, we need to read the config values first before initializing the BaseGitService instances accordingly. If we were doing it using the original method, I thing it would be more complicated given the time limitation (e.g we need to check if we are not creating 2 GitLab instances public and private at the same time as it doesn't make sense, how would we return if during the analysis, the GitLab instance when try to load the default values would recognise that the environmental variable is not set correctly). 
- I think the reason why I put `GIT_SERVICES: list[BaseGitService] = [GitHub(), GitLab(), BitBucket()]` originally is because the original Git services did not load the values from defaults.ini so it worked at that time because we đidn't need to prepare those class instances with any values from users.
- I just recognize that because `get_git_service` method is called again and again [here](https://github.com/oracle/macaron/blob/102853cfeaf318391b8d66d2e6302b2ce4e5c359/src/macaron/slsa_analyzer/analyzer.py#L580), we could potentially not read the same original values (i.e if the env var changed during the analysis , etc.)

behnazh-w: @tromai 
>we need to check if we are not creating 2 GitLab instances public and private at the same time as it doesn't make sense

We can have two public and private instances at the same time. Assume a case where a main target is private but the dependencies are using the public GitLab service.

tromai: > 1. If the domain provided for a Git service is not valid, Macaron run should not proceed and exit early.
I wonder if you mean that the `load_defaults` method in each BaseGitService should detect if there are errors in the `.ini` config? Hmmm I think there is a potential issue with this approach. For example, if we are creating both PublicGitLab and PrivateGitLab instances beforehand. During the analysis we run `load_defaults` for PublicGitLab first, at this stage of execution, if the config for the PublicGitLab is invalid, should we exit or we only exit after checking if PrivateGitLab config exist/valid because exiting. This could introduce more complexity to the implementation I think. :thinking: 

> 2. The current design in general assumes certain objects exist, so that we can iterate through them without constantly checking if they are None or not. Please follow this design.
Could we store the list of BaseGitService returned in the Analyzer and iterate through them instead ? :thinking:

tromai: > We can have two public and private instances at the same time. Assume a case where a main target is private but the dependencies are using the public GitLab service.

Ah I think I get this part now. I think we could either:
- Have a class for PublicGitLab and PrivateGitLab or
- Have one GitLab class but two instances created for each of them (I personally think this would be better as the only difference is the token value and the domain value (I assume cloning, etc for them is the same), it's not necessary for us to create two sub class given that they are not that much different).

behnazh-w: >I think the reason why I put GIT_SERVICES: list[BaseGitService] = [GitHub(), GitLab(), BitBucket()] originally is because the original Git services did not load the values from defaults.ini so it worked at that time because we đidn't need to prepare those class instances with any values from users.

The values were loaded [here](https://github.com/oracle/macaron/blob/c1b5435961e2b61f5377f0cc8a441c0fff28aa55/src/macaron/slsa_analyzer/analyzer.py#L580) though were loaded again and again before cloning a repo (which was not necessary).

nathanwn: > The values were loaded [here](https://github.com/oracle/macaron/blob/c1b5435961e2b61f5377f0cc8a441c0fff28aa55/src/macaron/slsa_analyzer/analyzer.py#L580) though were loaded again and again before cloning a repo (which was not necessary).

I can revert the `GIT_SERVICES` variable and load the config into it. This ensures that we do not repeatedly call this method over again.

src/macaron/errors.py
```diff
@@ -22,3 +22,11 @@ class CUEExpectationError(MacaronError):
...
 class CUERuntimeError(MacaronError):
     """Happens when there are errors in CUE expectation validation."""
+
+
+class ConfigurationError(MacaronError):
+    """Happens when there is an error in the configuration (.ini) file."""
+
+
+class CloneError(MacaronError):
+    """Happens when cannot clone a git repository (.ini) file."""
```
tromai, line 32: I think `(.ini) file` should be removed?

nathanwn: Thanks. Fixed in 8d642231b291181f84e666783d8a26de4061a59c.

src/macaron/slsa_analyzer/git_service/__init__.py
nathanwn: In the new version of my PR, environment variables are only read once at the beginning, right before the analysis starts.

src/macaron/slsa_analyzer/git_service/gitlab.py
```diff
@@ -13,40 +32,128 @@ class GitLab(BaseGitService):
...
+        ------
+        CloneError
+            If there is an error parsing the URL.
         """
-        return False
+        class_name = type(self).__name__
+        if not self.domain:
+            # This should not happen.
+            raise CloneError(
+                f"Internal error: The repository '{url} should not be cloned from the git service '{class_name}'."
```
behnazh-w, line 65:
```suggestion
                f"The repository '{url} should not be cloned from the git service '{class_name}'."
```

behnazh-w, line 65: Instead of printing `'{class_name}`  here, please add a debug log, which will include the line info.

src/macaron/slsa_analyzer/git_service/gitlab.py
```diff
@@ -13,40 +32,128 @@ class GitLab(BaseGitService):
...
+            raise CloneError(
+                f"Cannot clone the repo '{url}' from the git service '{class_name}'"
+                "the URL format is invalid or not supported by Macaron."
+            )
+
+        if self.access_token:
+            # https://docs.gitlab.com/ee/gitlab-basics/start-using-git.html#clone-using-a-token
+            clone_url = f"https://oauth2:{self.access_token}@{self.domain}/{url_parse_result.path}"
+        else:
+            clone_url = f"https://{self.domain}/{url_parse_result.path}"
```
behnazh-w, line 82: Please use `urllib.parse.ParseResult` to construct the URL and catch exceptions.

src/macaron/slsa_analyzer/git_service/gitlab.py
```diff
@@ -13,40 +32,128 @@ class GitLab(BaseGitService):
...
-        -------
-        bool
-            True if this git service is detected else False.
+        Raises
+        ------
+        CloneError
+            If there is an error cloning the repository.
+        """
+        clone_url = self.construct_clone_url(url)
+        git_url.clone_remote_repo(clone_dir, clone_url)
```
behnazh-w, line 106: Do we need a separate method here? Otherwise you could call `git_url.clone_remote_repo(clone_dir, clone_url)` from `construct_clone_url` method.

tests/slsa_analyzer/test_git_url.py
```diff
@@ -122,36 +116,28 @@ def test_get_remote_vcs_url() -> None:
...
+@patch("macaron.slsa_analyzer.git_url.get_allowed_git_service_domains")
+def test_get_remote_vcs_url_with_invalid_allowed_domains(
+    get_allowed_domains_fn: MagicMock,
+) -> None:
     """Test the vcs URL validator method without any allowed git hosts."""
-    original_allowed_hosts = defaults.get("git", "allowed_hosts")
-
-    # Test running with no git hosts defined in defaults.ini
-    defaults.remove_option("git", "allowed_hosts")
+    get_allowed_domains_fn.return_value = []
```
behnazh-w, line 124: Instead of mocking this function, wouldn't it be better to use a real `defaults` object and set invalid values? That way we can make sure that `git_url.get_allowed_git_service_domains` also works as expected.

behnazh-w:
@nathanwn I know you have opened a [ticket](https://github.com/oracle/macaron/pull/316) for integration tests. But please add at least one test for a public GitLab repository in this PR.

src/macaron/slsa_analyzer/git_service/gitlab.py
nathanwn: The main reason for me to have this `construct_clone_url` as a pure method is for testing purposes. I wanted to test the resulting URL and nothing else, as you can see [here](https://github.com/oracle/macaron/pull/316/commits/e2a1c32028d35a77f28a7919ea65c6b821d8eec2#diff-47b7349044e9e3c72fbf12384013033a706a2638cc4b24ae7f6f161c5a91eb6e).

tests/slsa_analyzer/test_git_url.py
nathanwn: Hmm, I can probably create a separate test for `git_url.get_allowed_git_service_domains`?
Modifying the global `defaults` object looks quite ugly to me, and I'll have to make sure I revert it correctly on top of that 😄.
I also thought about mocking the `defaults` object itself, but it is also quite ugly.

src/macaron/__main__.py
```diff
@@ -134,6 +136,16 @@ def perform_action(action_args: argparse.Namespace) -> None:
...
                 sys.exit(os.EX_USAGE)
             global_config.gh_token = gh_token
+
+            # TODO: Here we should try to statically analyze the config before
+            # actually running the analysis.
+            try:
+                for git_service in GIT_SERVICES:
+                    git_service.load_defaults()
+            except ConfigurationError as error:
+                logger.error(str(error))
```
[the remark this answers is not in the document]
nathanwn: This is unnecessary. I fixed in c96e6b7.

src/macaron/slsa_analyzer/git_service/base_git_service.py
```diff
@@ -18,15 +22,56 @@ def __init__(self, name: str) -> None:
...
     def load_defaults(self) -> None:
-        """Load the default values from defaults.ini."""
+        """Load the values for this git service from the ini configuration."""
         raise NotImplementedError
 
-    @abstractmethod
+    def load_domain(self, section_name: str) -> str | None:
+        """Load the domain of the git service from the ini configuration section ``section_name``.
+
+        The section may or may not be available in the configuration. In both case,
```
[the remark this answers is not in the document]
nathanwn: Fixed in b42a9e3.

tests/slsa_analyzer/test_git_url.py
behnazh-w: You can also read read a `defaults.ini` that has the test values and by merge the configurations using `defaults.load_defaults( "my_defaults.ini"))`.

src/macaron/slsa_analyzer/git_service/gitlab.py
```diff
@@ -13,40 +32,128 @@ class GitLab(BaseGitService):
...
+        """
+        try:
+            self.domain = self.load_domain(section_name="git_service.gitlab.private")
+        except ConfigurationError as error:
+            raise error
+
+        if not self.domain:
+            return
+
+        access_token_env_var = "MCN_PRIVATE_GITLAB_TOKEN"  # nosec B105
```
tromai, line 131: What error you are disabling using `# nosec B105`?

src/macaron/slsa_analyzer/git_service/gitlab.py
nathanwn: Fixed in f5a4bd09b58cbfb534527f626ad8fe1ba94c5aa8.

nathanwn: Fixed in f5a4bd09b58cbfb534527f626ad8fe1ba94c5aa8.

src/macaron/slsa_analyzer/git_service/gitlab.py
nathanwn: It's [this bandit error](https://bandit.readthedocs.io/en/latest/plugins/b105_hardcoded_password_string.html).

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -331,17 +339,18 @@ def clone_remote_repo(clone_dir: str, url: str) -> Repo | None:
 
     # The Repo.clone_from method handles creating intermediate dirs.
     try:
-        logger.info("Cloning the repo %s to %s", url, clone_dir)
```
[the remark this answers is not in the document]
nathanwn: Good catch. Fixed in 4712b134569a97e7e46faa9401d2be15d1905ec3.

nathanwn:
> @nathanwn I know you have opened a [ticket](https://github.com/oracle/macaron/pull/316) for integration tests. But please add at least one test for a public GitLab repository in this PR.

Integration test added in dbb274898e6ccde6f58df66b815b4552c31dc51f.

src/macaron/slsa_analyzer/git_service/base_git_service.py
```diff
@@ -18,15 +22,56 @@ def __init__(self, name: str) -> None:
...
+
+        Parameters
+        ----------
+        section_name : str
+            The name of the git service section in the ini configuration file.
+
+        Returns
+        -------
+        str | None
+            The domain. This can be ``None`` if the git section is not found in
```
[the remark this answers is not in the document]
nathanwn: Fixed in 4e41752.

src/macaron/slsa_analyzer/git_service/__init__.py
tromai: I think this conversation can be resolved as we have changed the implementation s.

src/macaron/slsa_analyzer/git_service/gitlab.py
nathanwn: Fixed in 185f5870f73b628816467d0aa3628fdc7351cbd4, but I'm not sure which exceptions you are referring to. The documentation for [`urlunparse`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlunparse) does not mention any errors.

behnazh-w: Sorry I wasn't clear. I think `ValueError` should be the right exception based on the documentations.

tests/slsa_analyzer/test_git_url.py
nathanwn: I added a couple of tests for `get_allowed_git_service_domains` in 2e607da3aa4f5bc3714940469c9e19bea566b14b.

src/macaron/slsa_analyzer/git_service/gitlab.py
nathanwn: If I'm not mistaken, only `urlparse` raises `ValueError` while `urlunparse` does not.

There is a place where I didn't handle `urlparse` properly. Fixed in 407eebb0d8d9514a3d1c0a9959695b1bf9416930.

src/macaron/slsa_analyzer/git_service/__init__.py
nathanwn: I think we agreed to leave this to a later refactor PR.

tromai (approved):
LGTM

tests/slsa_analyzer/test_git_url.py
```diff
@@ -122,36 +118,120 @@ def test_get_remote_vcs_url() -> None:
...
+            {"gitlab.com", "internal.gitlab.org"},
+        ),
+    ],
+)
+def test_get_allowed_git_service_domains(
+    config_input: str,
+    expected_allowed_domain_set: set[str],
+) -> None:
+    """Test the get allowed git service domains function."""
+    config = configparser.ConfigParser()
```
behnazh-w, line 151: Why not to use the `defaults` module directly which already has a function to merge? That way we also test `defaults` itself.

tests/slsa_analyzer/test_git_url.py
```diff
@@ -122,36 +118,120 @@ def test_get_remote_vcs_url() -> None:
...
+    ],
+)
+def test_get_allowed_git_service_domains_with_override(
+    default_config_input: str,
+    override_config_input: str,
+    expected_allowed_domain_set: set[str],
+    tmp_path: Path,
+) -> None:
+    """Test the get allowed git service domains function, in multi-config files scenario."""
+    default_filepath = tmp_path / "default.ini"
```
behnazh-w, line 200: Please use `os.path.join` here and for all the path constructions.

tests/slsa_analyzer/test_git_url.py
```diff
@@ -122,36 +118,120 @@ def test_get_remote_vcs_url() -> None:
...
+            {"github.com", "gitlab.com", "internal.gitlab.org"},
+            id="Private GitLab in user config",
+        ),
+    ],
+)
+def test_get_allowed_git_service_domains_with_override(
+    default_config_input: str,
+    override_config_input: str,
+    expected_allowed_domain_set: set[str],
+    tmp_path: Path,
```
behnazh-w, line 197: Where is `tmp_path` set? I couldn't find any fixture for it.

nathanwn: It's a [pytest pre-defined fixture](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html). It creates temporary directories in `/tmp` that are unique to each invocation of `pytest`.

tests/slsa_analyzer/test_git_url.py
tromai: I wonder why couldn't we use the slash operator here as it is supported by pathlib - https://docs.python.org/3/library/pathlib.html#operators. The `tmp_path` is of type `pathlib.Path` so I think it's okay to use.

tests/slsa_analyzer/test_git_url.py
tromai: I don't think it's a good idea to test the `defaults` module here as we are testing the function in `git_url` which mean we should isolate it to make the test deterministic. Because the method accept a `config.Parser` type, I think it's fair to test it like this.

tests/slsa_analyzer/test_git_url.py
behnazh-w: Nice, I didn't know about it :slightly_smiling_face:

tests/slsa_analyzer/test_git_url.py
behnazh-w: And why is it a good idea to rewrite and mock `defaults.load_defaults()` functionality that already exists?

tests/slsa_analyzer/test_git_url.py
behnazh-w: It's to be consistent with the rest of the codebase and to be more explicit by using a function signature.

tests/slsa_analyzer/test_git_url.py
nathanwn: @behnazh-w The main reason is I didn't want to modify the `defaults` object. Reverting it to the original state is quite cumbersome. [This](https://github.com/oracle/macaron/blob/bbf3ef3c256056b9b6b43788e8e650b6a8d4feb7/tests/slsa_analyzer/test_git_url.py#L148) was the way we do it earlier, which was not nice. The best way to do it is probably having a similar `load_defaults` call to [the one in the `setup_test` fixture](https://github.com/oracle/macaron/blob/bbf3ef3c256056b9b6b43788e8e650b6a8d4feb7/tests/conftest.py#L68). 
I assume the function to merge you are referring to is the [`load_defaults`](https://github.com/oracle/macaron/blob/c1b5435961e2b61f5377f0cc8a441c0fff28aa55/src/macaron/config/defaults.py#L103) function here. However, then again I will need to create different `.ini` files for this test, which results in the drawback of not being able to see the `.ini` input with the test itself together in one single file.
If you insist I can try to write the test this way.

nathanwn: > And why is it a good idea to rewrite and mock defaults.load_defaults() functionality that already exists?
This follows my same train of thought of not wanting to modify the global `defaults` object unless I really have to. I was assuming we already have tests for `load_defaults()` so these git-related tests should only test git-related logic.

tromai: Hmm I don't think this part is mocking `defaults.load_defaults()` as the function `get_allowed_git_service_domains` is a standalone "helper" function which accepts a `config.Parser` type so we want to isolate and test its own logic separately. Therefore, we are testing so that this function works well with any instances for `config.Parser`. 
If we are using `defaults` modules to load and test this function, if we change the `defaults` module in the future, we then need to update this too, which doesn't sound quite right to me as the standalone function `get_allowed_git_service_domains` shouldn't be affected by any changes outside of its own logic. :thinking:

behnazh-w: You don't need to modify any object. You can write the file as you're already doing in the other test [here](https://github.com/oracle/macaron/pull/316/files#diff-2b822b2946b03718e4f6c457cad5b087cfb7ca54934975d925b6b3414125e47dR203) but use `defaults.load_defaults()` for merging instead.

behnazh-w: >If we are using defaults modules to load and test this function, if we change the defaults module in the future, we then need to update this too, which doesn't sound quite right to me as the standalone function get_allowed_git_service_domains shouldn't be affected by any changes outside of its own logic.

That's why we should use `defaults` module and not re-implement its logic. If the logic of `load_defaults` changes in future, this implementation has to change too.

nathanwn: You can see [here in the `load_defaults()` function](https://github.com/oracle/macaron/blob/bbf3ef3c256056b9b6b43788e8e650b6a8d4feb7/src/macaron/config/defaults.py#L125) that it is indeed modifying the global `defaults` object. If it did return a merged `ConfigParser` object, then I could have used it for the test.

behnazh-w: >You can see [here in the load_defaults() function](https://github.com/oracle/macaron/blob/bbf3ef3c256056b9b6b43788e8e650b6a8d4feb7/src/macaron/config/defaults.py#L125) that it is indeed modifying the global defaults object. If it did return a merged ConfigParser object, then I could have used it for the test.

Yes, and that's the logic of `load_defaults`. What I meant was to avoid modifying the `defaults` object like this as it was happening before:
```
defaults["git"]["allowed_hosts"] = "invalid host"
```

nathanwn: Let me try to articulate everything again to make sure we are on the same page and I don't misunderstand anything.

What I mean by "modifying the `defaults` object" is doing anything that changes the state of this object. This includes not only directly modifying it directly as if it is a dictionary, such as the one you just mentioned

```
defaults["git"]["allowed_hosts"] = "invalid host"
```

but also calling any functions that modify it, such as the `load_defaults` function. The reason I don't like changing the state of the `defaults` object is that other tests rely on this object as well. If I don't do my due diligence to revert `defaults` to the original state, then there can be [test pollution](https://github.com/asottile/detect-test-pollution#what-is-test-pollution) affecting tests running after the current one.

Let's say for whatever reason I have to use `load_defaults` in one of my tests. The question now is: how do I ensure that there is no test pollution? I see two ways of doing this:

* Using the dictionary syntax, similar to the following (which I assume you are against?)

```
defaults["git"]["allowed_hosts"] = "invalid host"
```

* Calling `load_defaults` again to re-load the original config.

I was not a big fan of either of these alternatives for a couple of reasons that I mentioned throughout.

behnazh-w: >is that other tests rely on this object as well

Each test function has its own scope in Python and the imported `defaults` object reloads every time a test function runs. I'm not sure I follow how a test pollution can happen in this case. It would help if you give a concrete example to reproduce this using `load_defaults`.

nathanwn: Ok. I think I understand it better now. We have this [`setup_test`](https://github.com/oracle/macaron/blob/bbf3ef3c256056b9b6b43788e8e650b6a8d4feb7/tests/conftest.py#L50) fixture which makes sure the global `defaults` object is reloaded before every test. Therefore I didn't have to worry about test pollution in the first place. I was not aware of this fixture before.

nathanwn: I have modified the tests in 9425dfee3028b3a3fd26963a246d9b09d99d66d9. Let me know if there is anything else you want to address.

tests/slsa_analyzer/test_git_url.py
nathanwn: Should be fixed in 9425dfee3028b3a3fd26963a246d9b09d99d66d9.

tests/slsa_analyzer/test_git_url.py
```diff
@@ -122,36 +118,110 @@ def test_get_remote_vcs_url() -> None:
...
+def test_get_remote_vcs_url_with_user_defined_allowed_domains(tmp_path: Path) -> None:
+    """Test the vcs URL validator method with user-defined allowed domains."""
+    url = "https://internal.gitlab.org/org/name"
+    assert git_url.get_remote_vcs_url(url) == ""
+
+    user_config_path = os.path.join(tmp_path, "config.ini")
+    with open(user_config_path, "w", encoding="utf-8") as user_config_file:
+        user_config_file.write(
+            """
+                               [git_service.gitlab.private]
```
behnazh-w, line 205: Not sure, why `black` hasn't fixed the indentation here?

tests/slsa_analyzer/test_git_url.py
behnazh-w: Thanks, looks good. I just added one comment.

tests/slsa_analyzer/test_git_url.py
nathanwn: Hmm, this is actually the way `black` fixes it. I think initially I typed

```
        user_config_file.write("""
                               [git_service.gitlab.private]
                               ...
                               """)
```
I guess I could have helped `black` a bit here. Fixed in 965282fca5bf228b12b154755cf15f3ec3acfa2a.

behnazh-w (approved):

PR #319, opened 2023-06-19
chore(deps-dev): add colored package to resolve issue in syrupy

behnazh-w:
A breaking change in colored package is causing syrupy to fail.

As a temporary solution, we have to add colored as a test dependency and avoid using version 1.5.0 until syrupy releases a fix.


behnazh-w:
Syrupy has released a fix version.

See: https://github.com/tophat/syrupy/issues/758

PR #329, opened 2023-06-21
chore: catch AttributeError exception while processing SBOM

behnazh-w:
Sometimes SBOMs miss certain elements, such as `metadata.component`, which can throw `AttributeError` exception when `get` is called on the element if it is `None`, e.g., 
```python
bom.get("metadata").get("component").get("bom-ref")
```
On the other hand, because we use the `get` method, `KeyError` is never raised. See the documentation [here](https://docs.python.org/3/library/stdtypes.html#dict.get).

### DIFF
--- a/src/macaron/dependency_analyzer/cyclonedx.py
+++ b/src/macaron/dependency_analyzer/cyclonedx.py
@@ -70,7 +70,7 @@ def get_root_component(root_bom_path: Path) -> Optional[dict | None]:
     try:
         root_bom = deserialize_bom_json(root_bom_path)
         return root_bom.get("metadata").get("component")  # type: ignore
-    except (CycloneDXParserError, KeyError) as error:
+    except (CycloneDXParserError, AttributeError) as error:
         logger.error(error)
 
     return None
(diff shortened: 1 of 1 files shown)


src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -70,7 +70,7 @@ def get_root_component(root_bom_path: Path) -> Optional[dict | None]:
     try:
         root_bom = deserialize_bom_json(root_bom_path)
         return root_bom.get("metadata").get("component")  # type: ignore
-    except (CycloneDXParserError, KeyError) as error:
+    except (CycloneDXParserError, AttributeError) as error:
```
tromai, line 73: I wonder if we should isolate the try-catch block for AttributeError on the line `return root_bom.get("metadata").get("component")` only? To me, `AttributeError` seems like a very popular Error that could happen because of other reasons too so we should only limit it on the line we are expecting it to happen. This could prevent us from accidentally swallow the `AttributeError` not raised by the usages of chainning `get()` on dictionary.
This comment applies to the other 2 changes as well.

behnazh-w: This is on purpose, because the SBOMs  can be missing various elements and the blocks are quite isolated for  `AttributeError`s.

behnazh-w: See commit [877d877](https://github.com/oracle/macaron/pull/329/commits/877d87732d1c47c8fa248b2b066d76aae5712160).

tromai (approved):

PR #305, opened 2023-06-13
feat!: introduce a new data model and software components based on PURL

behnazh-w:
This PR changes the data model in Macaron to analyze at software component level using [PURL](https://github.com/package-url/purl-spec) as the identifier instead of repository level analysis. In this new data model, it is allowed to miss a repository associated with a component, closing this [issue](https://github.com/oracle/macaron/issues/187).

In addition, it closes this [issue](https://github.com/oracle/macaron/issues/115) by changing the database transactions to guarantee atomicity per analysis, i.e., if anything goes wrong and the analysis fails unexpectedly, no changes will be added to the database.

In this first PR the input is still a repository, and a PURL is constructed from the repository provided as input. In a subsequent PR, we will add [support for PURL identifier to be provided as a CLI input](https://github.com/oracle/macaron/issues/375).

Here is the new ERD for Macaron:
image

PR #349, opened 2023-06-29
feat: add support for JFrog Artifactory and witness provenances produced on GitLab CI

nathanwn:
This pull request adds the following features to Macaron:

1. Support downloading artifacts and provenances from a JFrog Maven package registry.
2. A new check for provenances produced by [witness](https://github.com/testifysec/witness).

Note that these will only work for repositories built with Gradle for now. Support for Maven will be added later.

# JFrog Maven package registry

A JFrog Maven package registry is a [JFrog Artifactory instance that acts as a Maven repository](https://jfrog.com/help/r/jfrog-artifactory-documentation/maven-repository). In SLSA terms, we call it a JFrog Maven **package registry** from here on out.

We are making the assumption that artifacts and provenances can be hosted on a JFrog Maven package registry. This pull request allows Macaron to discover artifacts and provenances on such a registry for analysis.

Another assumption is that an artifact and a corresponding provenance are hosted next to each other on the JFrog Maven package registry, i.e. if the artifact is hosted at `https://some/url/of/groupid/artifactid/version/*.jar`, then the corresponding provenance is hosted at `https://some/url/of/groupid/artifactid/version/*.intoto.jsonl`.

For Macaron to be aware of a JFrog Maven package registry, the user needs to provide the following section in the `.ini` config file:

```ini
[package_registry.jfrog.maven]
# In this example, the Maven package registry can be accessed at https://internal.registry.org/maven-repo.
domain = internal.registry.org
repo = maven-repo
download_timeout = 120
```

For Macaron to recognize a repository to possibly publish artifacts onto some JFrog Maven package registry, the repository must have Gradle as its build tool. Support for other build tools such as Maven will be added later.

# The `mcn_provenance_witness_level_one_1` check

This new `mcn_provenance_witness_level_one_1` check is added as a witness-specific replacement for `mcn_provenance_level_three_1`, which currently only supports SLSA provenances generated by [`slsa-github-generator`](https://github.com/slsa-framework/slsa-github-generator).

In this PR, this check only verifies that the digest of each jar file could be found somewhere in the corresponding witness provenance.

For this check to be effective, the following section must be provided in the `.ini` config:

```ini
# Witness provenance.
[provenance.witness]
# The allowed values of the `predicateType` field (data type: list).
predicate_types =
    https://witness.testifysec.com/attestation-collection/v0.1
artifact_extensions =
    jar
```

The following diagram shows a brief summary of the check's logical flow.

jfrogmaven drawio


src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
```diff
@@ -0,0 +1,302 @@
...
+        # JFrog REST API documentaiton: https://jfrog.com/help/r/jfrog-rest-apis/artifacts-storage.
+        url = f"{self.repo_url}/api/storage/{self.repo_key}/{group_path}/{artifact_id}/{version}"
+
+        try:
+            response = requests.get(
+                url=url,
+                timeout=defaults.getint("requests", "timeout", fallback=10),
+            )
+        except requests.exceptions.RequestException as error:
+            logger.error(
```
behnazh-w, line 240: Change to `debug` level log.

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
```diff
@@ -0,0 +1,302 @@
...
+                "Failed to fetch the downloadable assets of Java package %s:%s: %s", group_id, artifact_id, error
+            )
+            return []
+
+        if response.status_code == 200:
+            json_payload = json.loads(response.text)
+        else:
+            logger.error(
+                "Failed to fetch the downloadable assets of Java package %s:%s: got response with status code %d.",
```
behnazh-w, line 249: Same here.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -67,6 +75,246 @@ def __init__(self) -> None:
...
+        with tempfile.TemporaryDirectory() as temp_path:
+            # Note: In certain cases, Macaron can find the same provenance file in
+            # multiple different places on a package registry.
+            #
+            # We may consider de-duplicating this file, so that we do not run the same
+            # steps on the same file multiple times.
+
+            # Download the provenance
+            for prov_asset in provenance_assets:
+                if not jfrog_maven_registry.download_asset(
```
behnazh-w, line 175: Before downloading the asset, we should check for the size, which is returned by the API  and only download if it's smaller than the allowed size provided in `defaults.ini` file (check the GitHub release implementation).

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -67,6 +75,246 @@ def __init__(self) -> None:
...
+                # Store the provenance assets for other checks.
+                ci_info["provenance_assets"].extend(provenance_assets)
+
+                # Download the provenance assets and load the provenance payloads.
+                self.download_provenances_from_github_actions_ci_service(
+                    ci_info,
+                )
+
+                return [
+                    ProvenanceAvailableTable(
```
behnazh-w, line 263: This line adds a row to `ProvenanceAvailableTable`, which should instead be added in the `run_check` method after deciding if the check is passed or not. Please just return the necessary information here.

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -52,3 +54,46 @@ def load(cls, path: str) -> Any:
             raise SLSAProvenanceError(
                 f"Cannot decode the message content of the SLSA attestation - {error.reason}"
             ) from error
+
+
+def load_provenance(filepath: str) -> Any:
```
behnazh-w, line 59: This method seems to duplicate `ProvPayloadLoader.load` method?

src/macaron/slsa_analyzer/checks/provenance_available_check.py
tromai: I wonder if by only create an instance of `ProvenanceAvailableTable` class, the data is automatically added into the database using SQLAlchemy?

nathanwn: I think the result rows are not added to the database until [here](https://github.com/oracle/macaron/blob/fcc9b0a4017862ad0cf69e3571913df4d44f8418/src/macaron/slsa_analyzer/database_store.py#L62), isn't it? In these `find_provenance_assets_*` methods, I just "prepare" the rows if there are provenances found and let the `run_check` method decides whether to write these rows into `check_result["result_tables"]`, which will then will be written into the database later on after the check finishes.
A good thing about making these `find_provenance_assets_*` methods return table rows is that it makes use of a readily-defined datatype of table rows. Previously in the implementation of this check, after provenances are discovered on CI services, they are stored inside an intermediate dictionary which is then converted to a list of table rows type right away, and I feel like this dictionary did not serve any obvious purpose. I can keep using this pattern and make these two methods `find_provenance_assets_*` return 2 lists of dictionaries instead of 2 lists of table rows, but again I'm not sure I see a reason for using dictionaries over an already-defined class.

src/macaron/slsa_analyzer/package_registry/package_registry.py
```diff
@@ -0,0 +1,32 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    @abstractmethod
+    def load_defaults(self) -> None:
+        """Load the .ini configuration for the current package registry."""
+
+    @abstractmethod
+    def is_detected(self, build_tool: BaseBuildTool) -> bool:
+        """Detect if the package registry is compatible with the given build tool.
```
tromai, line 26: The doc string is a bit ambiguous for me (it says `compatible` but the method name is `is_detected`).

nathanwn: The naming of the `is_detected` method here is to be consistent with the methods `is_detected` in other classes (e.g. git services, build tools). I should have written the docstring more carefully though. The docstring is fixed in d0a8ecaf372998213fa8fd69bcb80b7c5ab4b79a.

src/macaron/config/defaults.ini
```diff
@@ -337,3 +337,17 @@ provenance_extensions =
 max_download_size = 70000000
 # This is the timeout (in seconds) to run the SLSA verifier.
 timeout = 120
+
+# Witness provenance.
+# [provenance.witness]
+# # The allowed values of the `predicateType` field (data type: list).
```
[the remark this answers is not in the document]
nathanwn: I added some details in a14aa91c333140b9a03140903718270fd8c7ec06.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
behnazh-w: When `ProvenanceAvailableTable` is instantiated, a [transient](https://docs.sqlalchemy.org/en/20/orm/session_state_management.html#quickie-intro-to-object-states) instance is created by SQLAlchemy, which will be persisted later when it is added to the session. Creating such transient instances should be done with care and it's best to be done where the check status is determined to be passed in our case to avoid accidental addition of rows to the table. 

I do agree though that the intermediate dictionary can be removed if not necessary. But it's best to address that as part of the new data model [PR](https://github.com/oracle/macaron/pull/305).

src/macaron/slsa_analyzer/build_tool/gradle.py
```diff
@@ -135,3 +136,42 @@ def get_dep_analyzer(self, repo_path: str) -> CycloneDxGradle:
...
 
         raise DependencyAnalyzerError(f"Unsupported SBOM generator for Gradle: {tool_name}.")
+
+    def get_group_id(self, project_path: str) -> str | None:
+        """Get the group id of a Java repository.
+
+        Parameters
+        ----------
+        project_path : str
+            Path to the repository.
```
tromai, line 146:
```suggestion
            Path to the Gradle project.
```

src/macaron/slsa_analyzer/build_tool/gradle.py
```diff
@@ -135,3 +136,42 @@ def get_dep_analyzer(self, repo_path: str) -> CycloneDxGradle:
             )
 
         raise DependencyAnalyzerError(f"Unsupported SBOM generator for Gradle: {tool_name}.")
+
+    def get_group_id(self, project_path: str) -> str | None:
+        """Get the group id of a Java repository.
```
tromai, line 141:
```suggestion
        """Get the group id of a Java Gradle repository.
```

src/macaron/slsa_analyzer/analyze_context.py
```diff
@@ -134,7 +138,7 @@ def provenances(self) -> dict:
         """
         try:
             ci_services = self.dynamic_data["ci_services"]
-            result = {}
+            result = {}  # key: ci service's name; values: list of provenance payloads
```
nicallen, line 141: This comment should be moved to the docstring, since it describes the format of the return value.

src/macaron/slsa_analyzer/checks/provenance_witness_l1_check.py
```diff
@@ -0,0 +1,253 @@
...
+                            check_result["justification"].append(
+                                "Failed to locate the hash digest of the jar file of Java package "
+                                f"{corresponding_provenance_asset.group_id}:"
+                                f"{corresponding_provenance_asset.artifact_id}:"
+                                f"{corresponding_provenance_asset.version} "
+                                f"in the provenance {provenance_url}."
+                            )
+                            return CheckResultType.FAILED
+
+        check_result["result_tables"].append(ProvenanceWitnessL1Table())
```
behnazh-w, line 249: We need to add some evidence to `check_result["justification"]` when the check passes, like the list of artifacts that have been checked.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -67,6 +75,246 @@ def __init__(self) -> None:
...
+                    continue
+
+            # Load the provenances to a dictionary in the package registry data entry.
+            # In this dictionary, each key is a URL to the provenance asset and
+            # each value is the corresponding provenance JSON payload.
+            for prov_asset in provenance_assets:
+                # Read the provenance.
+                provenance_filepath = os.path.join(temp_path, prov_asset.name)
+                try:
+                    data_entry.provenances[prov_asset.url] = load_provenance(
```
behnazh-w, line 192: The loaded provenances should be added to the [provenances](https://github.com/oracle/macaron/blob/6cf27845c7f891cd168a94b0abd4094b15cc4a6f/src/macaron/slsa_analyzer/analyze_context.py#L132-L146)  property of `analyze_context` to be picked up by the output HTML and JSON files.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -67,6 +75,246 @@ def __init__(self) -> None:
...
+                                artifact_id=artifact_id,
+                                version=latest_version,
+                                extensions=provenance_extensions,
+                            )
+                        )
+
+                    if not provenance_assets:
+                        continue
+
+                    logger.info("Found the following provenance assets:")
```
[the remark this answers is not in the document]
nathanwn: I think we already discussed and agreed that listing the provenances found is helpful.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -67,6 +75,246 @@ def __init__(self) -> None:
...
+                    for provenance_asset in provenance_assets:
+                        logger.info("* %s", provenance_asset.url)
```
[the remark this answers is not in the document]
nathanwn: I think we already discussed and agreed that listing the provenances found is helpful. (Same comment with [this one](https://github.com/oracle/macaron/pull/349#discussion_r1275747792))

src/macaron/slsa_analyzer/build_tool/gradle.py
```diff
@@ -135,3 +136,42 @@ def get_dep_analyzer(self, repo_path: str) -> CycloneDxGradle:
...
+            gradlew = os.path.join(global_config.resources_path, "gradlew")
+        try:
+            result = subprocess.run(  # nosec B603
+                [gradlew, "properties"],
+                capture_output=True,
+                cwd=project_path,
+                check=False,
+            )
+        except (subprocess.CalledProcessError, OSError) as error:
+            logger.debug("Could not capture the group id of the repo at %s", project_path)
```
nathanwn, line 167: @behnazh-w You asked in [this comment](https://github.com/oracle/macaron/pull/349#discussion_r1257570447)
> This can be a debug level log?

It is fixed now.

src/macaron/slsa_analyzer/build_tool/gradle.py
```diff
@@ -135,3 +136,42 @@ def get_dep_analyzer(self, repo_path: str) -> CycloneDxGradle:
...
+            gradlew = os.path.join(global_config.resources_path, "gradlew")
+        try:
+            result = subprocess.run(  # nosec B603
+                [gradlew, "properties"],
+                capture_output=True,
+                cwd=project_path,
+                check=False,
+            )
+        except (subprocess.CalledProcessError, OSError) as error:
+            logger.info("Could not capture the group id of the repo at %s", project_path)
```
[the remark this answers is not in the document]
nathanwn: See my answer [here](https://github.com/oracle/macaron/pull/349/commits/bee5806b89c28840dcf33037fbfea2bbefeae84b#r1275758159).

src/macaron/slsa_analyzer/build_tool/gradle.py
nathanwn: Fixed in 008474b757db40a13e18340c63c6e22095e2054c.

src/macaron/slsa_analyzer/build_tool/gradle.py
nathanwn: Fixed in 008474b757db40a13e18340c63c6e22095e2054c.

src/macaron/slsa_analyzer/provenance/loader.py
nathanwn: I didn't want to use the original `ProvPayloadLoader.load` methods for a couple of reasons. 
* The error messages of this method mention SLSA provenances but I was loading witness provenances rather than SLSA provenances.
* I also wanted the return type to be `JsonType` instead of `Any`, but I don't want to change the return type of the `ProvPayloadLoader.load` method since it is being used in the provenance level 3 check. Fixing the return type will result in type errors that I don't want to fix in this PR.
If you agree then I can create an issue to refactor these two methods of loading provenances into one.

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
```diff
@@ -0,0 +1,309 @@
...
+        URL to where the Maven asset is hosted.
+    group_id: str
+        The group id.
+    artifact_id: str
+        The artifact id.
+    version: str
+        The version of the Maven asset.
+    repo_url: str
+        The repo URL of the JFrog Maven registry.
+    repo_key: str
```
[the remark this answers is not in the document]
nathanwn: In the newer version:
* `repo_url -> domain`
* `repo_key -> repo`

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
```diff
@@ -0,0 +1,309 @@
...
+from macaron.config.defaults import defaults
+from macaron.errors import ConfigurationError
+from macaron.slsa_analyzer.build_tool.base_build_tool import BaseBuildTool
+from macaron.slsa_analyzer.package_registry.package_registry import PackageRegistry
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class JFrogMavenAsset(NamedTuple):
+    """An asset (file) hosted at a given URL.
```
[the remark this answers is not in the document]
nathanwn: Fixed in ed304b5b6fbd26019c623f954a325e92bd25d18c.

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
```diff
@@ -0,0 +1,309 @@
...
+    def download_asset(self, url: str, dest: str) -> bool:
+        """Download an asset of a Java package release.
+
+        Parameters
+        ----------
+        url : str
+            The URL to the asset on the package registry.
+        dest : str
+            The local destination where the asset is downloaded to.
+        """
```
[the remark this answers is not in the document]
nathanwn: Fixed in the current version.

src/macaron/slsa_analyzer/package_registry/package_registry.py
```diff
@@ -0,0 +1,36 @@
...
+
+        The detection here is based on the repo's detected build tool.
+        If the package registry is compatible with the given build tool, it can be a
+        possible place where the artifacts produced from the repo are published.
+
+        Parameters
+        ----------
+        build_tool : BaseBuildTool
+            A detected build tool of the repository under analysis.
+        """
```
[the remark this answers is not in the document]
nathanwn: Fixed in the current version.

src/macaron/slsa_analyzer/analyze_context.py
nathanwn: This has been fixed in the current version.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
nathanwn: I think creating transient objects is OK as long as I make sure not to add them to the database.
Anyhow, in the current version, I have changed the return type of the two `find_provenance_assets_*` methods.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -67,6 +75,246 @@ def __init__(self) -> None:
...
+        Returns
+        -------
+        list[ProvenanceAvailableTable]
+            A list of provenance assets found on one of the package registry.
+            This is an empty list if there is no provenance assets found.
+        """
+        for package_registry_data_entry in package_registry_data_entries:
+            match package_registry_data_entry:
+                case PackageRegistryData(
+                    build_tool=Gradle() as gradle,
```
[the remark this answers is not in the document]
nathanwn: I added issue #394.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -67,6 +75,246 @@ def __init__(self) -> None:
...
+            # steps on the same file multiple times.
+
+            # Download the provenance
+            for prov_asset in provenance_assets:
+                if not jfrog_maven_registry.download_asset(
+                    prov_asset.url,
+                    os.path.join(temp_path, prov_asset.name),
+                ):
+                    logger.info(
+                        "Could not download the provenance %s. Skip verifying...",
```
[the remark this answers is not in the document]
nathanwn: This has been fixed in the current version.

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
nathanwn: This has been fixed in the current version.

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
nathanwn: This has been fixed in the current version.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
nathanwn: This download size check has been added in commit fc4f0c8d48fc19a7e81738fa6b9730c9f3045f16.

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
```diff
@@ -0,0 +1,309 @@
...
+    name: str
+    group_id: str
+    artifact_id: str
+    version: str
+    repo_url: str
+    repo_key: str
+
+    @property
+    def url(self) -> str:
+        """Get the URL to the asset on the Maven package registry."""
```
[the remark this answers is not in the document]
nathanwn: I changed the docstring in 3c6bd35622df3031b09fdaf1d0288542810a9749.

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
```diff
@@ -0,0 +1,309 @@
...
+        group_path = self.group_id.replace(".", "/")
+        return f"{self.repo_url}/{self.repo_key}/{group_path}/{self.artifact_id}/{self.version}/{self.name}"
```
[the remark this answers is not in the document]
nathanwn: In the new version, I construct URLs using the [`urlunsplit`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlunsplit) function. Hence, the URLs are most likely valid and there should not be any issues, for example, with trailing whitespaces.

src/macaron/slsa_analyzer/ci_service/github_actions.py
```diff
@@ -96,7 +98,7 @@ def load_defaults(self) -> None:
                 self, "max_workflow_persist", defaults.getint("ci.github_actions", "max_workflow_persist", fallback=90)
             )
 
-    def is_detected(self, repo_path: str) -> bool:
+    def is_detected(self, repo_path: str, git_service: BaseGitService | None = None) -> bool:
```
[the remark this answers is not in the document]
nathanwn: This has been fixed in the current version.

src/macaron/slsa_analyzer/checks/provenance_witness_l1_check.py
nathanwn: The list of successfully verified artifacts has been added to the justification in the current version.

src/macaron/slsa_analyzer/build_tool/gradle.py
```diff
@@ -135,3 +136,42 @@ def get_dep_analyzer(self, repo_path: str) -> CycloneDxGradle:
...
+
+        Returns
+        -------
+        str | None
+            The group id, if exists.
+        """
+        # Use the gradlew that comes with the repository first, then use Macaron's
+        # built-in gradlew as a fallback option.
+        if os.path.isfile(os.path.join(project_path, "gradlew")):
+            gradlew = "./gradlew"
```
behnazh-w, line 156: What if `gradlew` is not executable?

src/macaron/slsa_analyzer/checks/provenance_available_check.py
nathanwn: Fixed in 5709a7f58bbab08f3a32d44c8b26090f033a3696.

src/macaron/slsa_analyzer/build_tool/gradle.py
```diff
@@ -135,3 +136,42 @@ def get_dep_analyzer(self, repo_path: str) -> CycloneDxGradle:
             )
 
         raise DependencyAnalyzerError(f"Unsupported SBOM generator for Gradle: {tool_name}.")
+
+    def get_group_id(self, project_path: str) -> str | None:
+        """Get the group id of a Gradle repository.
+
+        Parameters
+        ----------
+        project_path : str
```
behnazh-w, line 145: Instead of passing `project_path`, you should use `BaseBuildTool.get_build_dirs` function that contains all the build directories.

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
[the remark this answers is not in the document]
tromai: Thanks. I will have a look at the new version and resolve this conversation after that.

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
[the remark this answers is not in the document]
tromai: Thanks. I will have a look at the new version and resolve this conversation after that.

src/macaron/slsa_analyzer/build_tool/gradle.py
nathanwn: If the `gradlew` file is not executable here, then do you want `subprocess.run` to raise an error or do you want to fall back to Macaron's built-in `gradlew`? I am assuming the latter.

src/macaron/slsa_analyzer/build_tool/gradle.py
nathanwn: Here I don't necessarily want all directories, but only one single top-level directory where I can execute `./gradlew properties` and get a single group id value.

behnazh-w: First, some projects can have multiple sub-projects, each with their own built tool. Second, you need to run `gradlew` where `gradle.properties` is located, not at the top-level directory necessarily.

Can you also add unit tests for this function? Please follow the other Gradle tests that already use mock repositories.

nathanwn: I have checked and `gradle.properties` seems to be an optional file for Gradle, and it may or may not exist in a Gradle project.
After consulting the Gradle doc, this is what I found:
* [This section](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#declare_properties_in_gradle_properties_file) says "In Gradle, properties can be defined in the build script (i.e. `build.gradle`) , in a `gradle.properties` file or as parameters on the command line."
* The command I am using is `./gradlew properties | grep group`, which inspects the group property of the project.
  * According to the Gradle documentation [here](https://docs.gradle.org/current/dsl/org.gradle.api.Project.html#org.gradle.api.Project:group), the `group` property is under the `Project` interface, and
  * according to the doc [here](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html), "There is a one-to-one relationship between a Project and a ["build.gradle"](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#DEFAULT_BUILD_FILE) file.".

All of this means: the directories where the command `./gradlew properties` is run should be those containing a `build.gradle` file.

Here I see two solutions:
* Using the `BaseBuildTool.get_build_dirs` to get these directories. However, this `get_build_dir` method depends on the configuration in `.ini`, and it may not be, resulting in the command failing.
* Not using the `get_build_dirs` method, but scan the whole repo for the `build.gradle` files specifically.

I think the second solution is more reliable since it is more closely tied to the semantics of Gradle, whereas the first solution depends on the config.

nathanwn: I have checked and `gradle.properties` seems to be an optional file for Gradle, and it may or may not exist in a Gradle project.
After consulting the Gradle doc, this is what I found:
* [This section](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#declare_properties_in_gradle_properties_file) says "In Gradle, properties can be defined in the build script (i.e. `build.gradle`) , in a `gradle.properties` file or as parameters on the command line."
* The command I am using is `./gradlew properties | grep group`, which inspects the group property of the project.
  * According to the Gradle documentation [here](https://docs.gradle.org/current/dsl/org.gradle.api.Project.html#org.gradle.api.Project:group), the `group` property is under the `Project` interface, and
  * according to the doc [here](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html), "There is a one-to-one relationship between a Project and a ["build.gradle"](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#DEFAULT_BUILD_FILE) file.".

All of this means: the directories where the command `./gradlew properties` is run should be those containing a `build.gradle` file.

Here I see two solutions:
* Using the `BaseBuildTool.get_build_dirs` to get these directories. However, this `get_build_dir` method depends on the configuration in `.ini`, and it may give us directories that result in the command failing.
* Not using the `get_build_dirs` method, but scan the whole repo for the `build.gradle` files specifically.

I think the second solution is more reliable since it is more closely tied to the semantics of Gradle, whereas the first solution depends on the config. Let me know what you think.

behnazh-w: >Not using the get_build_dirs method, but scan the whole repo for the build.gradle files specifically.

There is no need to repeat doing what `BaseBuildTool.get_build_dirs` already does, which is looking for `build.gradle` by default. This also makes this function consistent with the build tool detection.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -93,42 +364,45 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
...
+            fallback=["intoto.jsonl"],
+        )
 
-                    # Store the provenance assets for other checks.
-                    ci_info["provenance_assets"].extend(assets)
+        # We look for the provenances in the package registries first, then CI services.
+        # (Note the short-circuit evaluation with OR.)
+        try:
+            provenance_assets = self.find_provenance_assets_on_package_registries(
+                repo_full_name=ctx.component.repository.full_name,
```
behnazh-w, line 377: I think you need to pass the file system path to the clone directory here?
```suggestion
                repo_full_name=ctx.component.repository.fs_path,
```

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
```diff
@@ -0,0 +1,819 @@
...
+    def construct_maven_path(
+        self,
+        group_id: str,
+        artifact_id: str | None = None,
+        version: str | None = None,
+        asset_name: str | None = None,
+    ) -> str:
+        """Construct a path to a folder or file on the registry, assuming Maven layout.
+
+        This is simply done by replacing all dots (``.``) with forward-slashes (``/``).
```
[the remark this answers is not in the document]
nathanwn: For the standard Maven registry layout, this is only applicable to group IDs.

src/macaron/slsa_analyzer/specs/package_registry_data.py
behnazh-w, line 1: To be consistent with `ci_spec.py` and `build_spec.py`, how about renaming this module to `package_registry_spec.py`?

Same comment for `PackageRegistryData`, which can be renamed to `PackageRegistryInfo` to be consistent with `CIInfo` and `BuildInfo`.

src/macaron/slsa_analyzer/specs/package_registry_data.py
```diff
@@ -0,0 +1,46 @@
...
+
+    package_registry : PackageRegistry
+        The package registry matched against the repository. This is dependent on the build tool detected.
+
+    latest_version : str | None
+        The latest version of the artifact found on the registry.
+
+    provenance_assets : list[dict]
+        Release assets for SLSA provenances, e.g., asset for attestation.intoto.jsonl.
+        Each entry of the list is a dictionary with two keys: ``"name"`` - the name of the
```
tromai, line 32: This is probably a nitpicking, if we are using double back-tick here, do we really need to add the double quote inside too? The same goes for 
```
``"url"``
```

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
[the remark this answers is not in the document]
tromai: I think it would be good if we note this in the docs string, saying that for this method (which support Maven layout, we only do this for group ID). What do you about it?

src/macaron/slsa_analyzer/specs/package_registry_data.py
```diff
@@ -0,0 +1,46 @@
...
+    build_tool : BaseBuildTool
+        The build tool matched against the repository.
+
+    package_registry : PackageRegistry
+        The package registry matched against the repository. This is dependent on the build tool detected.
+
+    latest_version : str | None
+        The latest version of the artifact found on the registry.
+
+    provenance_assets : list[dict]
```
tromai, line 30: Should this be `list[Asset]` ?

src/macaron/slsa_analyzer/specs/package_registry_data.py
```diff
@@ -0,0 +1,46 @@
...
+        Release assets for SLSA provenances, e.g., asset for attestation.intoto.jsonl.
+        Each entry of the list is a dictionary with two keys: ``"name"`` - the name of the
+        provenance file, and ``"url"`` - the URL where the provenance can be retrieved.
+
+    provenances : dict[str, dict]
```
tromai, line 35:
```suggestion
    provenances : dict[str, dict[str, JsonType]]
```

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -78,6 +74,281 @@ def __init__(self) -> None:
...
+            The full name of the repo, in the format of ``owner/repo_name``.
+        package_registry_data_entries : list[PackageRegistryData]
+            A list of package registry data entries.
+        provenance_extensions : list[str]
+            A list of provenance extensions. Assets with these extensions are assumed
+            to be provenances.
+
+        Returns
+        -------
+        Sequence[Asset]
```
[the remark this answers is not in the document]
nathanwn: There are a couple of reasons:
* The indirect reason is I want to return objects of type `Asset` instead of `JFrogMavenAsset` from this method. This return type of `Sequence[Asset]` is also consistent with the method `find_provenance_assets_on_ci_services`. The `run_check` method - the caller of these `find_provenance_assets_*` methods - does not need to know of the `JFrogMavenAsset` type or whatever concrete asset type they return. The `Asset` interface is enough.
* The direct reason for `Sequence[Asset]` instead of `list[Asset]` is Python's type variance rule for generic container types. What this method really returns is a `list[JFrogMavenAsset]`. `JFrogMavenAsset` is a subtype of `Asset`. However, because `list` (like other mutable containers in Python) is invariant, a `list[JFrogMavenAsset]` is not necessarily a `list[Asset]`. Meanwhile, `Sequence` is covariant, and therefore a `Sequence[JFrogMavenAsset]` is also a `Sequence[Asset]`.

[This page in the mypy doc](https://mypy.readthedocs.io/en/stable/generics.html#variance-of-generic-types) gives a good example of why a multiple containers like `list` should be invariant. Here is the example they use:
```python
class Circle(Shape):
    # The rotate method is only defined on Circle, not on Shape
    def rotate(self): ...

def add_one(things: list[Shape]) -> None:
    things.append(Shape())

my_circles: list[Circle] = []
add_one(my_circles)     # This may appear safe, but...
my_circles[-1].rotate()  # ...this will fail, since my_circles[0] is now a Shape, not a Circle
```
In this example, if the function `add_one` is annotated as `add_one(things: Sequence[Shape])`, then the method call `things.append` would be prohibited by the type checker, preventing the issue in the first place. `Sequence` is a supertype of both `list` and `tuple`, therefore it cannot be mutated because it will violate the fact that the container can be a `tuple`. While mutable containers like `list` are invariant, immutable containers like `tuple` are covariant, and as a result `Sequence` is also covariant.

Sorry for the wall of text, but hopefully the explanation clarifies my intention. 😄

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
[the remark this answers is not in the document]
nathanwn: This is a mistake. I think this method originally only took group id.
See commit b6275cda1a70aa2484935226057cc021ea786038.

src/macaron/slsa_analyzer/build_tool/gradle.py
nathanwn: This has been fixed in 8a9f82be1bf53574798a0a9f3de8ff9ac843c484.

src/macaron/slsa_analyzer/build_tool/gradle.py
nathanwn: This has been fixed in 8a9f82be1bf53574798a0a9f3de8ff9ac843c484.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
nathanwn: Fixed in 8a9f82be1bf53574798a0a9f3de8ff9ac843c484.

src/macaron/slsa_analyzer/build_tool/gradle.py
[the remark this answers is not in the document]
nathanwn: Fixed in 8a9f82be1bf53574798a0a9f3de8ff9ac843c484.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -78,6 +74,281 @@ def __init__(self) -> None:
...
+            A sequence of provenance assets found on one of the package registry.
+            This sequence is empty if there is no provenance assets found.
```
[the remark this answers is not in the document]
nathanwn: Fixed in 4de00af09bc56409bfcb278d76b4507487bb59bb.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -78,6 +74,281 @@ def __init__(self) -> None:
...
+        package_registry_data_entries : list[PackageRegistryData]
+            A list of package registry data entries.
+        provenance_extensions : list[str]
+            A list of provenance extensions. Assets with these extensions are assumed
+            to be provenances.
+
+        Returns
+        -------
+        Sequence[Asset]
+            A sequence of provenance assets found on one of the package registry.
```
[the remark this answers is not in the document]
nathanwn: Fixed in 78a4c3bc37cc7b86da65b93dac713af3c75b8557.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -78,6 +74,281 @@ def __init__(self) -> None:
...
+                        "slsa.verifier",
+                        "max_download_size",
+                        fallback=1000000,
+                    )
+
+                    for provenance_asset in provenance_assets:
+                        if provenance_asset.size_in_bytes > max_valid_provenance_size:
+                            msg = (
+                                f"The provenance asset {provenance_asset.name} exceeds the "
+                                f"max valid file size of {max_valid_provenance_size} (bytes)."
```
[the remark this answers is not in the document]
nathanwn: Fixed in 74c860cd990b108563de90dfea7e2c60d81439f3.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -78,6 +74,281 @@ def __init__(self) -> None:
...
+                            )
+                            logger.info(msg)
```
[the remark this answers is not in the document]
nathanwn: Fixed in 74c860cd990b108563de90dfea7e2c60d81439f3.

src/macaron/slsa_analyzer/specs/package_registry_data.py
nathanwn: This is now obsolete due to 93beee3fdd0cbbb0f3fc90b77339dcfea22950f7.

src/macaron/slsa_analyzer/specs/package_registry_data.py
nathanwn: This is now obsolete due to 93beee3fdd0cbbb0f3fc90b77339dcfea22950f7.

src/macaron/slsa_analyzer/specs/package_registry_data.py
nathanwn: This is now obsolete due to 93beee3fdd0cbbb0f3fc90b77339dcfea22950f7.

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
```diff
@@ -0,0 +1,819 @@
...
+
+    size_in_bytes: int
+    sha256_digest: str
+    download_uri: str
+
+
+class JFrogMavenRegistry(PackageRegistry):
+    """A JFrog package registry that acts as a Maven registry."""
+
+    def __init__(
```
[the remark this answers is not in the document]
nathanwn: Docstring added in 6fbea9029fe69d04b2ee338242b4b0ecf98b20ae.

src/macaron/slsa_analyzer/specs/package_registry_data.py
nathanwn: See 4e6bd58c3326c2e704a4000b73633b30608ff41e.

behnazh-w: Looks like the module name is still inconsistent. Can you please rename the module to `package_registry_spec.py`?

nathanwn: I renamed the module in 558ccbc1eb521f186274954ab73182cd4e00e103.

tests/slsa_analyzer/checks/test_provenance_available_check.py
```diff
@@ -35,17 +35,14 @@ def __init__(self, profile: dict):
         super().__init__(profile)
         self.release = {
             "assets": [
-                {"name": "attestation.intoto.jsonl", "url": "URL", "size": "10"},
-                {"name": "artifact.txt", "url": "URL", "size": "10"},
+                {"name": "attestation.intoto.jsonl", "url": "URL", "size": 10},
```
nathanwn, line 38: The `"size"` field here is supposed to be a number, not a string. I have checked the [GitHub docs](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release) and tried requesting the API myself.

tromai (approved):

src/macaron/slsa_analyzer/provenance/loader.py
behnazh-w: I think the `SLSAProvenanceError` is not quite accurate in the original implementation. Based on the [in-toto spec](https://github.com/in-toto/attestation/blob/main/spec/v1/envelope.md), it would make more sense to use `IntotoAttestationError` for in-toto attestations instead and remove `load_provenance`.

tests/e2e/expected_results/micronaut-core/micronaut-core.json
```diff
@@ -1523,8 +833,20 @@
...
+                {
+                    "check_id": "mcn_provenance_witness_level_one_1",
+                    "check_description": "Check whether the target has a level-1 witness provenance.",
+                    "slsa_requirements": [
+                        "Provenance - Available - SLSA Level 1",
+                        "Provenance content - Identifies build instructions - SLSA Level 1",
+                        "Provenance content - Identifies artifacts - SLSA Level 1",
+                        "Provenance content - Identifies builder - SLSA Level 1"
                     ],
+                    "justification": [],
```
behnazh-w, line 849: Why is the `mcn_provenance_witness_level_one_1` passing for `micronaut-core`? :thinking:

tests/e2e/expected_results/urllib3/urllib3.json
```diff
@@ -352,10 +352,22 @@
...
+                    "check_id": "mcn_provenance_witness_level_one_1",
+                    "check_description": "Check whether the target has a level-1 witness provenance.",
+                    "slsa_requirements": [
+                        "Provenance - Available - SLSA Level 1",
+                        "Provenance content - Identifies build instructions - SLSA Level 1",
+                        "Provenance content - Identifies artifacts - SLSA Level 1",
+                        "Provenance content - Identifies builder - SLSA Level 1"
+                    ],
+                    "justification": [],
+                    "result_type": "PASSED"
```
behnazh-w, line 369: Same comment here.

tests/e2e/expected_results/urllib3/urllib3_cue_invalid.json
```diff
@@ -356,6 +356,18 @@
...
+                    "check_id": "mcn_provenance_witness_level_one_1",
+                    "check_description": "Check whether the target has a level-1 witness provenance.",
+                    "slsa_requirements": [
+                        "Provenance - Available - SLSA Level 1",
+                        "Provenance content - Identifies build instructions - SLSA Level 1",
+                        "Provenance content - Identifies artifacts - SLSA Level 1",
+                        "Provenance content - Identifies builder - SLSA Level 1"
+                    ],
+                    "justification": [],
+                    "result_type": "PASSED"
```
behnazh-w, line 369: Same comment here.

tests/e2e/expected_results/slsa-verifier/slsa-verifier_cue_PASS.json
```diff
@@ -1768,6 +1768,18 @@
...
+                    "check_id": "mcn_provenance_witness_level_one_1",
+                    "check_description": "Check whether the target has a level-1 witness provenance.",
+                    "slsa_requirements": [
+                        "Provenance - Available - SLSA Level 1",
+                        "Provenance content - Identifies build instructions - SLSA Level 1",
+                        "Provenance content - Identifies artifacts - SLSA Level 1",
+                        "Provenance content - Identifies builder - SLSA Level 1"
+                    ],
+                    "justification": [],
+                    "result_type": "PASSED"
```
behnazh-w, line 1781: Same comment here.

src/macaron/slsa_analyzer/ci_service/base_ci_service.py
```diff
@@ -59,14 +60,19 @@ def get_workflows(self, repo_path: str) -> list:
...
+        self, repo_path: str, git_service: BaseGitService | None = None  # pylint: disable=unused-argument
+    ) -> bool:
         """Return True if this CI service is used in the target repo.
 
         Parameters
         ----------
         repo_path : str
             The path to the target repo.
 
+        git_service : BaseGitService
```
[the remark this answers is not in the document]
nathanwn: This is an abstract method in the abstract class `BaseCIService`. `git_service` is being used in the concrete class `GitHubActions`.

src/macaron/config/defaults.ini
```diff
@@ -337,3 +337,20 @@ provenance_extensions =
...
+# https://github.com/in-toto/attestation/tree/main/spec/v0.1.0#statement
+predicate_types =
+    https://witness.testifysec.com/attestation-collection/v0.1
+artifact_extensions =
+    jar
+
+# Package registries.
+# [package_registry.jfrog.maven]
+# In this example, the Maven repo can be accessed at `https://internal.registry.org/repo-name`.
+# domain = internal.registry.org
```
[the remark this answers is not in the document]
nathanwn: This should also be fixed for git services, all of which are using the term "domain" in the config. I created issue #416.

src/macaron/slsa_analyzer/asset/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+        """Get the size of the asset in bytes."""
+
+    def download(self, dest: str) -> bool:
+        """Download the asset.
+
+        Parameters
+        ----------
+        dest : str
+            The local destination where the asset is downloaded to.
+            Note that this must includes the file name.
```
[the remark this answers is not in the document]
nathanwn: Fixed in 5511ed3ab4d9581a0b565645d1e99658172cb20b.

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
```diff
@@ -0,0 +1,860 @@
...
+        if version:
+            path = "/".join([path, version])
+        if asset_name:
+            path = "/".join([path, asset_name])
+        return path
+
+    def fetch_artifact_ids(self, group_id: str) -> list[str]:
+        """Get all artifact ids under a group id.
+
+        This is done by fetching all children folder under the group folder on the registry.
```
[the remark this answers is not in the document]
nathanwn: Fixed in 5511ed3ab4d9581a0b565645d1e99658172cb20b.

src/macaron/slsa_analyzer/asset/__init__.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""This module defines asset classes.
```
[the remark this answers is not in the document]
nathanwn: Fixed in 7b292b15336136960a429eb2e55093bf53bf7696.

tests/e2e/expected_results/micronaut-core/micronaut-core.json
nathanwn: Previously, I did not return FAILED if there is no witness provenance discovered.
Fixed in d234c2736a370d13320b0b7338ba461224f2a31b.

tests/e2e/expected_results/urllib3/urllib3.json
nathanwn: Fixed in d234c2736a370d13320b0b7338ba461224f2a31b.

tests/e2e/expected_results/urllib3/urllib3_cue_invalid.json
nathanwn: Fixed in d234c2736a370d13320b0b7338ba461224f2a31b.

tests/e2e/expected_results/slsa-verifier/slsa-verifier_cue_PASS.json
nathanwn: Fixed in d234c2736a370d13320b0b7338ba461224f2a31b.

src/macaron/slsa_analyzer/provenance/witness.py
```diff
@@ -0,0 +1,210 @@
...
+    if predicates is None or not isinstance(predicates, dict):
+        return None
+    attestations = predicates.get("attestations", [])
+    if attestations is None or not isinstance(attestations, list):
+        return None
+    for attestation_entry in attestations:
+        if not isinstance(attestation_entry, dict):
+            return None
+        attestation_type = attestation_entry.get("type")
+        if attestation_type != "https://witness.dev/attestations/gitlab/v0.1":
```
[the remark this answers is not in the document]
nathanwn: I have fixed this function [here](https://github.com/oracle/macaron/pull/349/commits/e519b9ab432f4a86b1824731d3d399682b816170#diff-a958e75ec77484c721a410cf555fbe0aeb9bcc6027761c557ebbe0a6d12cbf45R118), using the Strategy pattern on Attestor classes.

src/macaron/config/defaults.ini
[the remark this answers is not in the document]
nathanwn: Fixed in 07ad8f5882a5e0dc6d10c4c8b0660dbefdd85f72.

src/macaron/slsa_analyzer/provenance/loader.py
nathanwn: I have refactored loading and validating provenances in e519b9ab432f4a86b1824731d3d399682b816170.

src/macaron/slsa_analyzer/provenance/intoto/v01/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+    Specification: https://github.com/in-toto/attestation/tree/main/spec/v0.1.0#statement.
+    """
+
+    name: str
+    digest: dict[str, str]
+
+
+def validate_intoto_statement(payload: dict[str, JsonType]) -> TypeGuard[InTotoStatement]:
+    """Validate the statement of an in-toto attestation.
+
```
behnazh-w, line 39: Same comment regarding adding TODO and link this [issue](https://github.com/oracle/macaron/issues/426).

src/macaron/slsa_analyzer/provenance/intoto/v01/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+    Parameters
+    ----------
+    payload : dict[str, JsonType]
+        The JSON statement after being base64-decoded.
+
+    Returns
+    -------
+    TypeGuard[InTotoStatement]
+        ``True`` if the subject element is valid, in which case its type is narrowed to an
+        ``InTotoStatement``; ``False`` otherwise.
```
behnazh-w, line 51: Shouldn't this be the following?
```suggestion
        ``True`` if the attestation statement is valid, in which case its type is narrowed to an
        ``InTotoStatement``; ``False`` otherwise.
```

src/macaron/slsa_analyzer/checks/provenance_witness_l1_check.py
```diff
@@ -0,0 +1,186 @@
...
+
+                        if failure_justification:
+                            check_result["justification"].extend(failure_justification)
+                            return CheckResultType.FAILED
+
+                        verified_artifact_assets.extend(artifact_assets)
+
+        # If Macaron cannot discover any witness provenance, we "fail" the check.
+        # Here, there status ``FAILED`` means: Macaron fails to discover any witness provenance.
+        # This is consistent with how Souffle works: facts in Souffle usually represent things that exist.
```
nicallen, line 174: I think the main point here as to why the check should fail if there are no witness provenances is because the property that the check is intended to capture is `"the project produces verifiable witness provenances"`. While it is true that the reason for this is related to the check result's intended use in policies, I don't think the reference to Souffle itself as the policy language is really relevant, it's more about what is useful for the policies we want to write. Having this property defined in this way is simply more useful than an alternative such as `"all witness provenances produced by the project can be verified"` (which would permit vacuous truth and therefore pass if there are none), because for example you would want to be able to write a policy that requires a verifiable provenance of any kind, which is easily specified as `"the project produces a verifiable witness provenance" OR " the project produces a verifiable SLSA provenance" OR ...`.

src/macaron/slsa_analyzer/provenance/provenance.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @property
+    def asset(self) -> AssetLocator:
+        """Get the asset."""
+
+    @property
+    def payload(self) -> InTotoPayload:
+        """Get the JSON payload of the provenance, in in-toto format.
+
+        This payload must be a JSON object at the top-level, hence the return type.
```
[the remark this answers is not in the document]
nathanwn: See 7ceb8b6dbb19acc504aa70b2efe97242c918cce9.

src/macaron/slsa_analyzer/provenance/intoto/v01/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+    predicate: dict[str, JsonType] | None
+
+
+class InTotoSubject(TypedDict):
+    """An in-toto subject.
+
+    Specification: https://github.com/in-toto/attestation/tree/main/spec/v0.1.0#statement.
+    """
+
+    name: str
```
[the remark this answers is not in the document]
nathanwn: See 7ceb8b6dbb19acc504aa70b2efe97242c918cce9.

src/macaron/slsa_analyzer/provenance/intoto/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+StatementT = TypeVar("StatementT", bound=Mapping)
+"""Type of an in-toto statement.
+
+This is currently either a v0.1 statement or v1 statement.
+"""
+
+
+@dataclass(frozen=True)
+class InTotoPayload(Generic[StatementT]):
+    """The payload of an in-toto provenance."""
```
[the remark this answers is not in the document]
nathanwn: See 7ceb8b6dbb19acc504aa70b2efe97242c918cce9.

src/macaron/slsa_analyzer/provenance/intoto/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+from macaron.util import JsonType
+
+StatementT = TypeVar("StatementT", bound=Mapping)
+"""Type of an in-toto statement.
+
+This is currently either a v0.1 statement or v1 statement.
+"""
+
+
+@dataclass(frozen=True)
```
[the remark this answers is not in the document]
nathanwn: See 72854d6e849828a984d524602c6c5a72639f36f2.

src/macaron/slsa_analyzer/provenance/intoto/v01/__init__.py
nathanwn: See d2b1f228d8c38248bf71cb1e7788011bf09c6e6c.

src/macaron/slsa_analyzer/provenance/intoto/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+
+class InTotoV1Payload(InTotoPayload[v1.InTotoStatement]):
+    """The provenance payload following in-toto v1 schema.
+
+    Specification: https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md.
+    """
+
+
+def validate_intoto_payload(payload: dict[str, JsonType]) -> InTotoPayload:
+    """Validate the schema of an in-toto provenance payload.
```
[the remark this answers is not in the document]
nathanwn: See d2b1f228d8c38248bf71cb1e7788011bf09c6e6c.

src/macaron/slsa_analyzer/provenance/provenance.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""This module defines asset classes.
```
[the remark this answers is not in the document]
nathanwn: See 3a6ea89.

src/macaron/slsa_analyzer/provenance/intoto/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+    -------
+    InTotoPayload
+        The validated in-toto payload.
+
+    Raises
+    ------
+    ValidateInTotoPayloadError
+        When there is an error validating the payload.
+    """
+    _type = payload.get("_type")
```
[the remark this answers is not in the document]
nathanwn: I think I used the leading underscore here not for avoiding conflict with the Python keyword `type`, but to keep the same name with the field `_type` in the payload. This is more of a styling preference I think.
See d3e3ace.

src/macaron/slsa_analyzer/provenance/intoto/v01/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+    TypeGuard[InTotoStatement]
+        ``True`` if the subject element is valid, in which case its type is narrowed to an
+        ``InTotoStatement``; ``False`` otherwise.
+
+    Raises
+    ------
+    ValidateInTotoPayloadError
+        When the payload does not follow the expected schema.
+    """
+    _type = payload.get("_type")
```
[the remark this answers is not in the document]
nathanwn: See d3e3ace.

src/macaron/slsa_analyzer/checks/provenance_witness_l1_check.py
nathanwn: Thanks for your comment. I think this is a very nice way to express the intent of the check:
> `"the project produces verifiable witness provenances"`

See 65a8347.

nicallen (approved):

src/macaron/slsa_analyzer/provenance/intoto/v01/__init__.py
nathanwn: Thanks. This must be a copy-paste error. Fixed in 1a6d6ca.

behnazh-w (approved):

PR #379, opened 2023-07-23
chore: upgrade to python 3.11.4

### LINKED ISSUE #376: Upgrade to Python 3.11.4

tromai:
Closes #376

### DIFF
--- a/docker/Dockerfile.base
+++ b/docker/Dockerfile.base
@@ -21,11 +21,6 @@ ENV HOME="/home/macaron" \
     LANG="C.UTF-8" \
     # The GPG key to verify the source tar ball following the instructions in https://www.python.org/downloads/
     PYTHON_SOURCE_GPG="A035C8C19219BA821ECEA86B64E628F8D684696D" \
-    # https://github.com/docker-library/python/blob/f568f56f28fab0fe87b34db777e2c2861cef002b/3.11/slim-buster/Dockerfile#L133
-    PYTHON_PIP_VERSION=22.3.1 \
-    PYTHON_SETUPTOOLS_VERSION=65.5.1 \
-    PYTHON_GET_PIP_URL="https://github.com/pypa/get-pip/raw/d5cb0afaf23b8520f1bbcfed521017b4a95f5c01/public/get-pip.py" \
-    PYTHON_GET_PIP_SHA256="394be00f13fa1b9aaa47e911bdb59a09c3b2986472130f30aa0bfaf7f3980637" \
     # Setting Java related environment variables.
     JAVA_HOME="/usr/lib/jvm/jdk-17-oracle-x64" \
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.base
```diff
@@ -5,7 +5,7 @@
 # to build and run the Docker image.
 # This image is based on the container-registry.oracle.com/os/oraclelinux:9-slim image and contains the following
 # components:
-#   Python3.11.2 (with pip 22.3.1 and setuptools 65.5.1) compiled and installed from source.
+#   Python3.11.4 (with pip 22.3.1 and setuptools 65.5.1) compiled and installed from source.
```
behnazh-w, line 8: Please upgrade `pip` and `setuptools` to the latest versions too.

tromai: Actually, because we have decided back then that we don't want to install pip and setuptools separately. We have opted to install pip together with the Python installation. I will remove this note on pip and setuptools.
After having a look [here](https://github.com/oracle/macaron/blob/d6f1cc60a5eb1c26165f530afc73ea735e800e38/docker/Dockerfile.base#L24C5-L28C10) I have noticed that these env variables are not used anywhere (they are used back when we still install pip and setuptools separately from the Python installation - following the method [here](https://github.com/docker-library/python/blob/f568f56f28fab0fe87b34db777e2c2861cef002b/3.11/slim-buster/Dockerfile#L133)).
Therefore, I have also removed them and make sure that pip and setuptools are upgraded to the latest version in `Dockerfile.final` - see https://github.com/oracle/macaron/pull/379/commits/3f68c4f73bf717874e45fff416370a0dd84c0053) and https://github.com/oracle/macaron/pull/379/commits/3fffadeb341f13e82fb8325d52a67bd2abb6d2db

behnazh-w: I was actually confused because we had indeed decided to install the latest version of `pip` and `setuptools` for the corresponding Python version. Thanks for the clarification.

docker/Dockerfile.final
```diff
@@ -37,7 +37,7 @@ COPY --chown=macaron:macaron $WHEEL_PATH $HOME/dist/
 RUN : \
     && python3 -m venv $HOME/.venv \
     && . .venv/bin/activate \
-    && pip install --no-compile --no-cache-dir --upgrade pip \
+    && pip install --no-compile --no-cache-dir --upgrade pip setuptools \
```
behnazh-w, line 40: We don't need to install `setuptools` separately here.
```bash
python -m pip install --upgrade pip
```
already installs it:
```bash
$pip list
Package    Version
---------- -------
pip        23.2.1
setuptools 65.5.0
```

tromai: I think we need to run it here with `--upgrade` to update it to the latest version though :thinking:

behnazh-w: That's a good point :+1:

behnazh-w (approved):

Issue #378, opened 2023-07-23
Simplify the RFC3339DateTime custom type implementation


behnazh-w:

jenstroeger:
@behnazh-w I’m working on a PR for this issue 🤓

PR #381, opened 2023-07-24
fix(orm): serialize datetime object’s timezone instead of always coercing to UTC when persisting to the SQLite db

jenstroeger:
Fixes https://github.com/oracle/macaron/issues/378

The existing code doesn’t ensure that the serialized object is a `datetime` and it always coerces the timezone to UTC; the docstring seems to indicate that serialized `datetime` objects are stored as “[naive](https://docs.python.org/3/library/datetime.html#aware-and-naive-objects)” strings but that’s not the case because `Z` is appended to the serialized string which indicated [Zulu](https://en.wikipedia.org/wiki/ISO_8601#Coordinated_Universal_Time_(UTC)) time (GMT).

This change preserves the timezone of the serialized `datetime` objects, and adds UTC for naive `datetime` objects before serializing to string. Always deserializes to the original timezone.

### DIFF
--- a/src/macaron/database/rfc3339_datetime.py
+++ b/src/macaron/database/rfc3339_datetime.py
@@ -4,7 +4,7 @@
 """This module implements SQLAlchemy type for converting date format to RFC3339 string representation."""
 
 import datetime
-from typing import Any, Optional
+from typing import Any
 
 from sqlalchemy import String, TypeDecorator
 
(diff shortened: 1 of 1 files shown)


src/macaron/database/rfc3339_datetime.py
```diff
@@ -14,34 +14,33 @@ class RFC3339DateTime(TypeDecorator):  # pylint: disable=W0223
...
-            result = value.isoformat(sep="T", timespec="seconds") + "Z"
-        return result
-
-    def process_result_value(self, value: Optional[Any], dialect: Any) -> None | datetime.datetime:
+        if value is None:
+            return None
+        if not isinstance(value, datetime.datetime):
+            raise TypeError("RFC3339DateTime type expects a datetime object")
+        if not value.tzinfo:
+            value = value.astimezone(datetime.timezone.utc)  # Consider coercing to host timezone.
```
behnazh-w, line 35: Can you please make this comment a TODO?

src/macaron/database/rfc3339_datetime.py
```diff
@@ -14,34 +14,33 @@ class RFC3339DateTime(TypeDecorator):  # pylint: disable=W0223
...
     """
 
     # It is stored in the database as a string
     impl = String
     cache_ok = True
 
-    def process_bind_param(self, value: Optional[Any], dialect: Any) -> None | str:
-        """Process when storing.
+    def process_bind_param(self, value: None | Any, dialect: Any) -> None | str:
+        """Process when storing. If the provided ``datetime`` is a naive datetime object then UTC is added.
```
behnazh-w, line 25:
```suggestion
        """Process when storing. If the timezone of the serialized ``datetime`` object is provided, this function preserves it. Otherwise, if the provided ``datetime`` is a naive ``datetime`` object then UTC is added.
```

behnazh-w:
@jenstroeger Thanks a lot for your contribution. I left some minor comments. Also, looks like the OCA check is not passing. Can you please try again and see if any information is missing?

src/macaron/database/rfc3339_datetime.py
jenstroeger: Do you want me to make the change here, or in a separate PR?

behnazh-w: I prefer a separate PR.

src/macaron/database/rfc3339_datetime.py
jenstroeger: I made the change locally but this makes for a long line, see new PR https://github.com/oracle/macaron/pull/384. Waiting for that to be merged first.

behnazh-w: We used to address such length issues by line breaks. Thanks for the improvement.

jenstroeger: Line-break won’t work because:
```
src/macaron/database/rfc3339_datetime.py:25:1: D205 1 blank line required between summary line and description
        """Process when storing. If the timezone of the serialized ``datetime`` object is provided, this function
        preserves it. Otherwise, if the provided ``datetime`` is a naive ``datetime`` object then UTC is added.

        value: None | datetime.datetime
            The value being stored
        """
^
src/macaron/database/rfc3339_datetime.py:25:1: D400 First line should end with a period
        """Process when storing. If the timezone of the serialized ``datetime`` object is provided, this function
        preserves it. Otherwise, if the provided ``datetime`` is a naive ``datetime`` object then UTC is added.

        value: None | datetime.datetime
            The value being stored
        """
^
```

behnazh-w: I meant using a line break like this:
```
        """Process when storing datetime objects in SQLite.

        If the timezone of the serialized ``datetime`` object is provided, this function
        preserves it. Otherwise, if the provided ``datetime`` is a naive ``datetime`` object then UTC is added.

        value: None | datetime.datetime
            The value being stored
        """
```

behnazh-w (approved):

PR #383, opened 2023-07-25
build: improve Makefile setup on macOS

behnazh-w:


Makefile
```diff
@@ -134,9 +141,15 @@ souffle:
                 sudo apt update; \
                 sudo apt install souffle; \
                 ;; \
+			"Darwin") \
```
tromai, line 144: I checked on my machine and it seems that this block for `"Darwin"` is indented using tabs instead of spaces.

behnazh-w: See commit [1cdef4e](https://github.com/oracle/macaron/pull/383/commits/1cdef4e8fde94b7fa027149483caf90ea9441a80)

src/macaron/config/defaults.py
```diff
@@ -163,9 +163,6 @@ def create_defaults(output_path: str, cwd_path: str) -> bool:
             os.path.relpath(os.path.join(output_path, "defaults.ini"), cwd_path),
         )
         return True
-    except shutil.Error as error:
+    except (OSError, shutil.Error) as error:
```
tromai, line 166: Should we put a small comment above this line saying why we added the OSError here?

behnazh-w: See commit [dc4a44d](https://github.com/oracle/macaron/pull/383/commits/dc4a44dc54e47d3e09ac886fad768f327802da57)

tromai (approved):

PR #384, opened 2023-07-26
ci: make line-length checks a bit more tolerant using Bugbear only

jenstroeger:
Currently `pylint` and `flake8` check line lenghts in different ways, so this change disables `pylint` and `flake8`’s default checkers and uses Bugbear only (which is a little more lenient anyway, see [here](https://github.com/PyCQA/flake8-bugbear#list-of-warnings)).


behnazh-w:
@jenstroeger thanks for this improvement. Can you please change the title to
```
ci: make line-length checks a bit more tolerant using Bugbear only
```

We don't need to communicate this commit as part of the changelog to the users.

behnazh-w (approved):

PR #388, opened 2023-07-28
feat: enable repo finder to support more languages via Open Source Insights

benmss:
The repo finder is refactored to:
- be usable from anywhere within Macaron
- accept a PURL string as input
- support more languages via Open Source Insights (deps.dev)

Added languages include: Python, NodeJS, .Net, Rust.
Note that these additions won't do anything until dependency analysers for these languages are added.

Use of Open Source Insights can be disabled via a configuration value.


src/macaron/repo_finder/repo_finder_dd.py
```diff
@@ -0,0 +1,203 @@
...
+
+import requests
+
+from macaron.repo_finder.repo_finder_base import BaseRepoFinder
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class RepoFinderDD(BaseRepoFinder):
+    """This class is used to find repositories using Google's Open Source Insights tool (deps.dev)."""
```
tromai, line 18: Do you think that putting the acronym "DD" after `deps.dev` would be a bit clearer. When first looking into the module, I was wondering for some time on what "DD" means. :sweat_smile:

src/macaron/repo_finder/repo_finder_java.py
```diff
@@ -0,0 +1,323 @@
+# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
```
tromai, line 1: I believe the implementation of methods inside `JavaRepoFinder` didn't change much from the previous state and you only move it into a class?

benmss: This is true.

src/macaron/repo_finder/repo_finder_base.py
```diff
@@ -0,0 +1,87 @@
...
+            The URL for the GET request.
+
+        Returns
+        -------
+        str :
+            The retrieved file data or an empty string.
+        """
+
+    @abstractmethod
+    def read_metadata(self, metadata: str) -> tuple[Iterator[str], int]:
```
nicallen, line 74: A tuple of iterator and size is an odd return type, since there can be iterators whose size is unknowable, so if the interface demands that the implementation must concretely enumerate the size anyway, why not just return a list?

benmss: That's a good point. I'll change it to a list.

nicallen (approved):

src/macaron/repo_finder/repo_finder_dd.py
```diff
@@ -0,0 +1,203 @@
...
+        url : str
+            The URL for the GET request.
+
+        Returns
+        -------
+        str :
+            The retrieved file data or an empty string.
+        """
+        try:
+            result = session.get(url)
```
behnazh-w, line 133: To improve handling errors for requests, let's use the [send_get_http](https://github.com/oracle/macaron/blob/fa2622b588546888041cbf3a0c07a67841a98080/src/macaron/util.py#L21) function. As discussed, we will revisit retrying requests when there is a timeout [separately](https://github.com/oracle/macaron/issues/398).

benmss: I've changed this and the other similar calls of the Repo Finder. I chose to use send_get_http_raw as it fits with the class design better, and the functionality seems to be the same.

src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -86,25 +84,17 @@ def create_urls(self, group: str, artifact: str, version: str) -> list[str]:
         list[str]
             The list of created URLs.
         """
-        package_name = f"{group}:{artifact}@{version}"
```
tromai: Why is this variable removed here? Is it not used anywhere and was left here by accident?

src/macaron/__main__.py
```diff
@@ -118,6 +119,31 @@ def verify_policy(verify_policy_args: argparse.Namespace) -> int:
...
+    urls = []
+    # Without version
+    urls.append(repo_finder.create_urls("", "packageurl-python", ""))
+    # With version
+    urls.append(repo_finder.create_urls("", "packageurl-python", "0.11.1"))
+    for url in urls:
+        logger.debug("Verifying: %s", url[0])
+        metadata = repo_finder.retrieve_metadata(url[0])
+        if not metadata:
+            return 1
```
tromai, line 140: For this method. Instead of returning integer as exit code. Let's use the return code constant provided in the `os` library [here](https://docs.python.org/3/library/os.html#os.EX_OK).
For this method. I think for 0 could be replaced by `os.EX_OK` and 1 could be replaced by `os.EX_UNAVAILABLE` ? Please feel free to suggest another constant if it's more suitable.

src/macaron/__main__.py
```diff
@@ -118,6 +119,31 @@ def verify_policy(verify_policy_args: argparse.Namespace) -> int:
     return os.EX_USAGE
 
 
+def verify_repo_finder() -> int:
```
tromai, line 122: Is this verify functionary available for the user or do we only have it for the integration tests only? If it's intended for the user, I think we need to update the Sphinx documentation too (I'm happy to have it done in another PR if that's the case).

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -836,6 +841,52 @@ def perform_checks(self, analyze_ctx: AnalyzeContext) -> dict[str, CheckResult]:
...
+        self, dependencies: dict[str, DependencyInfo], context: AnalyzeContext, description: str
+    ) -> None:
+        """Utilise the Repo Finder to resolve more dependencies to repositories."""
+        purl = PackageURL.from_string(context.component.purl)
+        build_tools = context.dynamic_data["build_spec"]["tools"]
+        if not purl.type or not build_tools:
+            logger.debug("No PURL type found for: %s", purl)
+        else:
+            # Get actual PURL type based on current build tool
+            build_tool = build_tools[0]
```
tromai, line 854: What is the reason we select the first element in `build_tools` to resolve the dependencies. Would it prevent us from acquiring the correct dependencies if a component have more than one detected build tool?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -836,6 +841,52 @@ def perform_checks(self, analyze_ctx: AnalyzeContext) -> dict[str, CheckResult]:
...
+            if len(build_tools) > 1:
+                logger.debug(
+                    "More than one build tool found for %s, assuming type based on class: %s",
+                    description,
+                    build_tool.__class__,
+                )
+            updated_type = ""
+
+            if isinstance(build_tool, Gradle):
```
tromai, line 863: I can't seem to figure out the reason why we need to "reconstruct" the purl of the software component here instead of using it directly?

src/macaron/repo_finder/repo_finder.py
```diff
@@ -0,0 +1,53 @@
...
+    try:
+        purl = PackageURL.from_string(purl_string)
+    except ValueError as error:
+        logger.debug("Invalid PURL: %s, %s", purl_string, error)
+        return
+
+    # Find matching Repo Finder
+    repo_finder: BaseRepoFinder
+    if purl.type == "maven":
+        repo_finder = JavaRepoFinder()
```
tromai, line 42: Does this mean that for software component that uses Maven as the build tool, we always use the JavaRepoFinder (which looks at pom.xml files, etc.) but not the deps dev method?
This behavior is also applied to the dependencies rathered in the `_resolve_more_dependencies` method within `analyzer.py`, isn't it?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -836,6 +841,52 @@ def perform_checks(self, analyze_ctx: AnalyzeContext) -> dict[str, CheckResult]:
...
+                logger.debug("Unsupported PURL type: %s", build_tool.__class__)
+
+            for key, item in dependencies.items():
+                if item["available"] != SCMStatus.MISSING_SCM:
+                    continue
+                dep_namespace, dep_name = key.split(":")
+                purl_string = f"pkg:{updated_type}/{dep_namespace}/{dep_name}"
+                if item["version"] != "unspecified":
+                    purl_string = purl_string + "@" + item["version"]
+                urls = find_repo(purl_string)
```
tromai, line 881: From the behavior that I mentioned in this [discussion](https://github.com/oracle/macaron/pull/388/files#r1286611150):
 
- Let's say that `build_tool` of the main software component here is `Maven`.
- Before we call `_resolve_more_dependencies`, the `JavaRepoFinder` has been used to resolve the repo paths of all dependencies within `dependencies`. 
- Let's say there is a dependency `item` called A:B:C. The `JavaRepoFinder` couldn't resolve A:B:C's repo path before. Therefore, `item["available"] == SCMStatus.MISSING_SCM`.
- On line 878, are new PURL string is created for A:B:C (i.e `pkg:maven/A/B`). 
- On line 881, this new PURL string is provided to the `find_repo` method. The `find_repo` would not use deps dev to resolve `pkg:maven/A/B`'s repo path but will use the `JavaRepoFinder` to resolve it again. Would that make any different as `JavaRepoFinder` has been used before to resolve this dependency but failed?

src/macaron/repo_finder/repo_finder_deps_dev.py
benmss: It was only used for debug output relating to a failed GET request. That is now handled as part of the send_get_http_raw function call, so there was no need to keep it here.

src/macaron/__main__.py
benmss: I don't think there is much value in making it available to the user. It has a very specific use currently, and extending it to be more useful for users might not be a good use of time (in my opinion).

src/macaron/slsa_analyzer/analyzer.py
benmss: The purl here is of the main target, meaning its type is github or similar. A new purl has to be created for the dependency using its package type.

src/macaron/slsa_analyzer/analyzer.py
benmss: You are right, there would be an issue with multiple conflicting build tools. One solution I see for this is to add the name of the tool each dependency was found by. Perhaps in the `note` field of `DependencyInfo`?

src/macaron/slsa_analyzer/analyzer.py
benmss: The repo finder is only called through `_resolve_more_dependencies`.

src/macaron/repo_finder/repo_finder.py
benmss: This is true. But now I will run some more tests on their relative performance.

src/macaron/__main__.py
tromai: Soungs good to me. 

The only concern is that with this design, this `verify-repo-finder` command will be shown when the user run `macaron --help`.
```
usage: macaron [-h] [-V] [-v] [-o OUTPUT_DIR] [-dp DEFAULTS_PATH] [-lr LOCAL_REPOS_PATH]
               {analyze,dump-defaults,verify-policy,verify-repo-finder} ...

positional arguments:
  {analyze,dump-defaults,verify-policy,verify-repo-finder}
                        Run macaron <action> --help for help

...
```
But I guess it won't pose any problem (at least for me). Let's keep it this way for now.

src/macaron/slsa_analyzer/analyzer.py
tromai: Oh I see. I forgot that the first round of the dependency resolution is through the SBOM generator, not the repo finder. It makes sense now.

src/macaron/slsa_analyzer/analyzer.py
tromai: Right. So we use the main's package type (e.g `Maven`) to resolve all of its dependencies by "casting" the dependencies PURL into that particular package type.

src/macaron/__main__.py
behnazh-w: Since we have `verify-policy` as well, to avoid confusion we can change `verify-repo-finder` to `test-repo-finder` to indicate this is just for testing purposes. What do you think?

benmss: That works for me.

src/macaron/slsa_analyzer/analyzer.py
behnazh-w: I think by refactoring the repo finder differently we can avoid this issue. Instead of moving `_resolve_more_dependencies` here, we can add it as an abstract method to `BaseBuildTool`, which can be implemented per build type. Note how we get the [dependency analyzer](https://github.com/oracle/macaron/blob/66f138bd882507afbe8cf854e997b47b97029086/src/macaron/slsa_analyzer/analyzer.py#L275) per build tool. We can subsequently call `resolve_more_dependencies` per build tool afterwards.

src/macaron/__main__.py
tromai: Sounds good.

src/macaron/dependency_analyzer/dependency_resolver.py
```diff
@@ -31,9 +29,13 @@ class DependencyTools(str, Enum):
 class DependencyInfo(TypedDict):
     """The information of a resolved dependency."""
 
-    version: str
-    group: str
+    scheme: str
```
behnazh-w, line 32: Instead of duplicating the PURL spec here, wouldn't it be better to store the PURL string directly, which will be parsed using the `packageurl` module?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -1047,6 +977,34 @@ def perform_checks(self, analyze_ctx: AnalyzeContext) -> dict[str, CheckResult]:
 
         return results
 
+    def _resolve_more_dependencies(self, dependencies: dict[str, DependencyInfo]) -> None:
```
behnazh-w, line 980: This function and `resolve_dependencies` can be moved to the `dependency_resolver` module to keep the dependency resolution code in one place.

src/macaron/dependency_analyzer/dependency_resolver.py
benmss: If we store the PURL string as well as the old information (in their fields), we have duplicates. If we don't keep the old information, we have to create the PURL from the string every time we want to extract any information from it. Perhaps storing the actual PURL would make the most sense?

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -171,9 +171,13 @@ def convert_components_to_artifacts(
             # According to PEP-0589 all keys must be present in a TypedDict.
             # See https://peps.python.org/pep-0589/#totality
             item = DependencyInfo(
-                version=component.get("version") or "",
-                group=component.get("group") or "",
+                scheme="pkg",
+                type_="maven",
```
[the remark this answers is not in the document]
benmss: The class isn't exactly language agnostic when it uses `groupId:artifactId` as the artifact keys, but I agree otherwise.

src/macaron/__main__.py
nathanwn: Hi Ben. What do you think about moving your `test_repo_finder()` function from `src/macaron/__main__.py` into a new Python script e.g. in `tests/e2e/repo_finder.py`, and instead of running `macaron test-repo-finder`, we can run the `python tests/e2e/repo_finder.py` script in our integration tests?

The `tests/e2e/repo_finder.py` script can be run in `integration_test.sh` similar to how the `tests/e2e/compare_e2e_result.py` is being run.

I prefer this to adding the `test-repo-finder` command to Macaron, as it may cause confusion for a user who relies on the CLI help. At the same time, `test_repo_finder()` is only for testing and not actually being used in Macaron, and hence I don't think it should stay in `src`.

src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -0,0 +1,196 @@
...
+from requests.exceptions import ReadTimeout
+
+from macaron.repo_finder.repo_finder_base import BaseRepoFinder
+from macaron.util import send_get_http_raw
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class RepoFinderDepsDev(BaseRepoFinder):
+    """This class is used to find repositories using Google's Open Source Insights A.K.A. deps.dev (DD)."""
```
nathanwn, line 19: I think you have renamed everything from `DD` to `DepsDev`.
```suggestion
    """This class is used to find repositories using Google's Open Source Insights A.K.A. deps.dev."""
```

src/macaron/__main__.py
benmss: I think that's a good idea.

docs/source/pages/using.rst
```diff
@@ -203,30 +203,38 @@ This feature is enabled by default. To disable, or configure its behaviour in ot
 
 See :ref:`dump-defaults <action_dump_defaults>`, the CLI command to dump the default configurations in ``defaults.ini``. After making changes, see :ref:`analyze <analyze-action-cli>` CLI command for the option to pass the modified ``defaults.ini`` file.
 
-Within the configuration file under the ``repofinder.java`` header, five options exist: ``find_repos``, ``artifact_repositories``, ``repo_pom_paths``, ``find_parents``, ``artifact_ignore_list``. These options behave as follows:
+Within the configuration file under the ``repofinder.java`` header, three options exist: ``artifact_repositories``, ``repo_pom_paths``, ``find_parents``. These options behave as follows:
```
nathanwn, line 206: I think the correct term to call `[[repofinder.java]]` is "section" rather than "header" (according to the [documentation of the `configparser` library](https://docs.python.org/3/library/configparser.html#quick-start)). 
```suggestion
Within the configuration file under the ``[[repofinder.java]]`` section,  three options exist: ``artifact_repositories``, ``repo_pom_paths``, ``find_parents``. These options behave as follows:
```

docs/source/pages/using.rst
```diff
@@ -203,30 +203,38 @@ This feature is enabled by default. To disable, or configure its behaviour in ot
...
 
-- ``find_repos`` (Values: True or False) - Enables or disables the Repository Finding feature.
 - ``artifact_repositories`` (Values: List of URLs) - Determines the remote artifact repositories to attempt to retrieve dependency information from.
 - ``repo_pom_paths`` (Values: List of POM tags) - Determines where to search for repository information in the POM files. E.g. scm.url.
 - ``find_parents`` (Values: True or False) - When enabled, the Repository Finding feature will also search for repository URLs in parents POM files of the current dependency.
-- ``artifact_ignore_list`` (Values: List of GAs) - The Repository Finding feature will skip any artifact in this list. Format is "GroupId":"ArtifactId". E.g. org.apache.maven:maven
+
+Under the related header ``repofinder``, two more options exist: ``find_repos``, and ``use_open_source_insights``:
```
nathanwn, line 212:
```suggestion
Under the related section ``[[repofinder]]``, two more options exist: ``find_repos``, and ``use_open_source_insights``:
```

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -536,9 +536,14 @@ def get_remote_vcs_url(url: str, clean_up: bool = True) -> str:
     str
         The remote url to the repo or empty if the url is invalid.
     """
+    if len(url) == 1:
```
nathanwn, line 539: I wonder in which case this happens and why you need to check for that case here.

tests/slsa_analyzer/test_analyzer.py
```diff
@@ -42,49 +42,6 @@ def test_resolve_local_path(self) -> None:
         assert Analyzer._resolve_local_path(self.PARENT_DIR, "././././") == self.PARENT_DIR
 
 
-@pytest.mark.parametrize(
```
nathanwn, line 45: I saw that you moved the `to_repo_path` function from `slsa_analyzer/analyzer.py` to `repo_finder/repo_finder.py`.
I assume that means this test should also be moved and not removed? I don't think I find this test being moved anywhere so I assume it is being removed.

benmss: Good catch. I am fairly sure I moved it, but now I can't find it. I'll add it back.

docs/source/pages/using.rst
benmss: It is essentially the same thing. The documentation you linked calls them "section headers", so either is fine really.

src/macaron/slsa_analyzer/git_url.py
benmss: You are right, I will investigate it more.

docs/source/pages/using.rst
benmss: I suppose if we want to be completely correct, more of the sentence would need to be adjusted as "under the ..." does not fit with "section".
Options are then:

1. "... under the <name> header..."
2. "...within the <name> section..."

src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -0,0 +1,196 @@
...
+                package_name = name.lower().replace("_", "-")
+            case "npm":
+                if namespace:
+                    package_name = f"%40{namespace}%2F{name}"
+                else:
+                    package_name = name
+            case "nuget" | "cargo":
+                package_name = name
+            case "maven":
+                package_name = f"{namespace}%3A{name}"
```
nathanwn, line 190: Can you link to the section(s) in `deps.dev` documentation where these formats are documented?

tests/repo_finder/repo_finder_dd/test_repo_finder_dd.py
```diff
@@ -0,0 +1,16 @@
...
+from macaron.repo_finder.repo_finder_deps_dev import RepoFinderDepsDev
+
+
+def test_repo_finder_dd() -> None:
+    """Test the functions of the repo finder."""
+    repo_finder = RepoFinderDepsDev("pypi")
+    assert (
+        repo_finder.create_type_specific_url("", "packageurl-python")
+        == "https://api.deps.dev/v3alpha/systems/pypi/packages/packageurl-python"
+    )
```
nathanwn, line 16: I saw that you are supporting some other package types in `create_type_specific_url` (`maven`, `cargo`, `nuget`, etc.), but you are only testing `pypi` here. Do you think it's worth testing other package types as well?

src/macaron/dependency_analyzer/dependency_resolver.py
nathanwn: I agree with the option of storing the actual PURL object.

docs/source/pages/using.rst
nathanwn: Hmm, that's very fair. Thanks, Ben. I think you can keep it as is.

src/macaron/dependency_analyzer/cyclonedx.py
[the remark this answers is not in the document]
nathanwn: @benmss Are you referring to [this line](https://github.com/oracle/macaron/pull/388/files/57d13c9d300bbe896b8c315670ef5accfc3e2333#diff-b2e47510ad053dfc5cd207b480a61ff1dce1c947acbc1b89b3a7a0e0b1ae84b6L170)?
```
            key = f"{component.get('group')}:{component.get('name')}"
```
From this particular line, it looks to me like this function seems to only support Java previously because a component that is not a Java package does not have the `group` attribute.


Suggestion: To make this function actually language agnostic, maybe instead of reading attributes such as `'group'` or `'name'` from CycloneDx SBOM Components, we should read the `'purl'` attribute instead ([example](https://github.com/oracle/macaron/blob/32f2f6d01763f867b7092faccc5e29b0970c0802/tests/dependency_analyzer/cyclonedx/resources/apache_maven_root_sbom.json#L98)). @benmss I think you should consider refactoring this function in this direction of utilizing the `'purl'` attribute and remove reads from attributes that do not seem language agnostic (e.g. `'group'`). Let me know what you think.

Additional info I found:
* [CycloneDX SBOM version 1.4 schema in protobuf](https://github.com/CycloneDX/specification/blob/cc15c8550e87dfa9ee051e6795e74a3c6c641264/schema/bom-1.4.proto#L101) (I assume Macaron is supporting version 1.4 because the example SBOMs in our test resources all follows version 1.4).

src/macaron/repo_finder/repo_finder_base.py
```diff
@@ -0,0 +1,83 @@
+# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""This module contains the base class for the repo finders."""
+
+from abc import ABC, abstractmethod
+from collections.abc import Iterator
+
+
+class BaseRepoFinder(ABC):
```
nathanwn, line 10: These questions are more about the interface of this class rather than the implementation.

1. I see that the `find_repo` method has the signature of

```
def find_repo(self, group: str, artifact: str, version: str)
```

which, at first glance, looks to me like it is only designed for Java. In the theme of making things language-agnostic, do you think we can change the signature of this method to make it look more language-agnostic? For example, instead of taking GAV, can the function take something like PURL as param?

2. I'm not sure if the methods `create_urls`, `retrieve_metadata`, and `read_metadata` should be part of the common interface of repo finders, for a couple of reasons:
  * In general, I don't really see why these methods should be used polymorphically outside of the repo finder packages, they seems more like internal implementation to me. They do not seem to be useful from the perspective of the consumer of the interface, e.g. if I want to find the repo of some software component, why do I need to care about creating URLs and retrieving different kinds of data? It's more likely that I just want to provide an identifier (i.e. PURL) and get a repo.
  * I also feel like these abstract methods are creating unnecessary abstractions that make the concrete implementation of each repo finder type more confusing. As an example, it's not clear to me why we should have the very board concept of `metadata` while we can have `retrieve_json_payload`/`read_json_payload` for `RepoFinderDepsDev`, and `retrieve_pom`/`read_pom` for `JavaRepoFinder` instead. These kinds of abstractions, imo, make the implementation looks kind of awkward and harder to read. For example, the `BaseRepoFinder::create_urls` abstract method returns a list of URLs, but the concrete method `RepoFinderDepsDev::create_urls` repo finder only either return 0 or 1 URL, and it is not clear why a list should the return type instead of `str | None`. In another case, it's not too clear to me what the actual return values of the `JavaRepoFinder::read_metadata` and `RepoFinderDepsDev::read_metadata` have in common, while the former is from pom contents and the latter is from request payloads.

src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -0,0 +1,196 @@
...
+
+from requests.exceptions import ReadTimeout
+
+from macaron.repo_finder.repo_finder_base import BaseRepoFinder
+from macaron.util import send_get_http_raw
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class RepoFinderDepsDev(BaseRepoFinder):
```
nathanwn, line 18: Nit: I think this class should be named `DepsDevRepoFinder` to be consistent with the other classes: `BaseRepoFinder` and `JavaRepoFinder`.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -1047,6 +977,34 @@ def perform_checks(self, analyze_ctx: AnalyzeContext) -> dict[str, CheckResult]:
...
+        """Utilise the Repo Finder to resolve the repositories of more dependencies."""
+        for item in dependencies.values():
+            if item.get("available") != SCMStatus.MISSING_SCM:
+                continue
+
+            name = item.get("name") or ""
```
nathanwn, line 986: This part is probably related to [this discussion](https://github.com/oracle/macaron/pull/388/files/57d13c9d300bbe896b8c315670ef5accfc3e2333#r1312331740): if we store PURL in `DependencyInfo`, we wouldn't need to reconstruct PURLs here.

src/macaron/dependency_analyzer/cyclonedx.py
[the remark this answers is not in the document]
behnazh-w: Using `purl` could be a good candidate but because it is optional, the refactoring might not be simple. I have added an issue to refactor this function separately to support other languages: https://github.com/oracle/macaron/issues/464

@benmss you can add a TODO linking this issue but for now, it would be better to avoid fixing the `type`.

src/macaron/dependency_analyzer/dependency_resolver.py
behnazh-w: Storing the `packageurl` object sounds good to me.

src/macaron/dependency_analyzer/cyclonedx.py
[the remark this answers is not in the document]
benmss: As we agreed to retrieve the dependency type directly from the SBOM, the options appear to be:

1. Try to get the type from the optional PURL
2. Fix the type as "maven"

src/macaron/repo_finder/repo_finder_java.py
```diff
@@ -49,12 +46,20 @@ def find_repo(self, group: str, artifact: str, version: str) -> Iterator[str]:
         # - Try to extract SCM metadata and return URLs
         # - Try to extract parent information and change current artifact to it
         # - Repeat
-        group = group.replace(".", "/")
+        group = (purl.namespace or "").replace(".", "/")
```
nathanwn, line 49: It may be better to do the `replace(".", "/")` in the `_create_urls` method, instead of doing it twice in this method (here, and in the `while` loop).

tests/e2e/repo_finder/repo_finder.py
```diff
@@ -20,21 +20,32 @@
...
 def test_repo_finder() -> int:
-    """Test the functionality of the remote API calls used by the repo finder.
-
-    Functionality relating to Java artifacts is not verified for two reasons:
-    - It is extremely unlikely that Maven central will change its API or cease operation in the near future.
-    - Other similar repositories to Maven central (internal Artifactory, etc.) can be provided by the user instead.
-    """
-    defaults.add_section("repofinder")
+    """Test the functionality of the remote API calls used by the repo finder."""
+    if not defaults.has_section("repofinder.java"):
```
nathanwn, line 24: Suggestion: Instead of manipulating the `defaults` object here, I think it is better to provide a custom user config in plain `ini` format and let the `defaults` object read the `ini` string instead. This is [an example of a test](https://github.com/oracle/macaron/blob/staging/tests/slsa_analyzer/package_registry/test_jfrog_maven_registry.py#L31C1-L47C1) following this approach.
Advantages of this approach include:
* It better simulates how user config is loaded into Macaron.
* The `ini` string is probably more readable than Python code that manipulates the `defaults` object.

nathanwn: P/S: I just realized this is not a `pytest` test, but an integration script test. Sorry about that. In that case, you can't utilize the `tmp_path: Path` pytest fixture like the example I sent. I see two alternatives:
* If you still want to create an `ini` string in code, write it to a file, and load that file, you can instead use the `tempfile.TemporaryDirectory()` context manager as an alternative for the `tmp_path` pytest fixture ([documentation](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory), and [example in Macaron](https://github.com/oracle/macaron/blob/32f2f6d01763f867b7092faccc5e29b0970c0802/src/macaron/slsa_analyzer/checks/provenance_available_check.py#L223)).
* You can also create a separate `ini` file and load that instead. The downside to this approach is that to read this test, one needs to open two files.

src/macaron/repo_finder/repo_finder.py
```diff
@@ -0,0 +1,182 @@
...
+    vcs_set = {get_remote_vcs_url(value) for value in urls if get_remote_vcs_url(value) != ""}
+
+    # To avoid non-deterministic results we sort the URLs.
+    vcs_list = sorted(vcs_set)
+
+    if len(vcs_list) < 1:
+        return ""
+
+    # Report the first valid URL.
+    return vcs_list.pop()
```
nathanwn, line 182: `list.pop()` returns the last element in `list`, not the first one ([Python doc](https://docs.python.org/3/tutorial/datastructures.html#more-on-lists)).
I suppose this is probably what you want?
```suggestion
    return vcs_list[0]
```

benmss: This function was moved from `dependency_resolver.py` and has remained unchanged.

tests/e2e/repo_finder/repo_finder.py
benmss: If more people want to weigh in on this, I'll possibly change it. Otherwise, I think it's fine as is.

src/macaron/repo_finder/repo_finder_java.py
```diff
@@ -0,0 +1,327 @@
...
+        ----------
+        purl : PackageURL
+            The PURL of an artifact.
+
+        Yields
+        ------
+        Iterator[Iterator[str]] :
+            The URLs found for the passed artifact.
+        """
+        yield from iter(self._find_repo(purl))  # type: ignore[misc]
```
nathanwn, line 43: Shouldn't this just be:
```suggestion
        yield from self._find_repo(purl)
```
And the return type would just be `Iterator[str]`? The return type of `Iterator[Iterator[str]]` seems quite odd and I don't see how it is necessary.
I also have the similar comment for the `DepsDevRepoFinder::find_repo()` method.

src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -0,0 +1,196 @@
...
+        if not json_data:
+            logger.debug("Failed to retrieve json data for: %s", purl)
+            return
+
+        urls = self._read_json(json_data)
+        if not urls:
+            logger.debug("Failed to extract repository URLs from json data: %s", purl)
+            return
+
+        yield iter(urls)  # type: ignore[misc]
```
nathanwn, line 55: I think this should just be
```suggestion
        yield from urls
```
Since `urls` is already a generator, we use `yield from` instead of `yield` ([delegating to a subgenerator](https://docs.python.org/3/whatsnew/3.3.html#pep-380)).

src/macaron/repo_finder/repo_finder.py
nathanwn: Oh, I see. Up to you to decide whether to change it. 👍🏼

src/macaron/repo_finder/repo_finder_java.py
benmss: The reason it is that way is due to how finding repositories for Java works. Retrieving a POM and analysing it potentially returns multiple URLs, as a list that is iterable. At that stage the list needs to be iterated in search of valid URLs, and if none are found, the search continues with the next parent POM. Using nested iterators with `yield from` allow this process to continue from where it left off, thereby retrieving the next iterable list, and so on.
While deps.dev does not need this, other future finders possibly might. For example, one that checks multiple sources of knowledge in sequence.
Sadly, `type: ignore[misc]` is needed because `mypy` cannot handle this situation at all. The hints it gives are wrong.

src/macaron/repo_finder/repo_finder.py
```diff
@@ -0,0 +1,182 @@
...
+        return ""
+
+    # Call Repo Finder and return first valid URL
+    logger.debug("Analyzing %s with Repo Finder: %s", purl.to_string(), repo_finder.__class__)
+    for urls in repo_finder.find_repo(purl):
+        url = find_valid_url(urls)
+        if url:
+            return url
+
+    return ""
```
nathanwn, line 85: Related to [this comment](https://github.com/oracle/macaron/pull/388/files#r1316858675) about the return type of the `BaseRepoFinder::find_repo` method: if the return type is just `Iterator[str]`, then I think here you can simplify this quite a bit:
```suggestion
    return find_valid_url(repo_finder.find_repo(purl))
```

src/macaron/repo_finder/repo_finder_java.py
nathanwn: Hmm, right. I think I see what you are trying to do.
> Retrieving a POM and analysing it potentially returns multiple URLs, as a list that is iterable. At that stage the list needs to be iterated in search of valid URLs, and if none are found, the search continues with the next parent POM.

Having said that, shouldn't this logic be internal to only the `JavaRepoFinder`? I don't quite see the point of iterating through multiple calls of `repo_finder.find_repo` until a valid result is found, as opposed to getting a valid result from a single call. Interacting until a valid URL is found, imo, should happen within the `JavaRepoFinder` class, not outside of it.

src/macaron/dependency_analyzer/dependency_resolver.py
```diff
@@ -307,6 +261,121 @@ def tool_valid(tool: str) -> bool:
             return False
         return True
 
+    @staticmethod
+    def resolve_dependencies(main_ctx: Any, sbom_path: str) -> dict[str, DependencyInfo]:
```
nathanwn, line 265: I wonder why `main_ctx` is type-annotated as `Any` instead of `AnalyzeContext` here.

behnazh-w: @nathanwn This question is not relevant to this PR because @benmss has not authored this function and just moved it to [dependency_resolver.py](https://github.com/oracle/macaron/pull/388/files/724b3993ef12517afc94f457e4998a1667160b90#diff-c95724433501b4a68d8ffe2e35305bff56f2c26c45571f9250641511a8d48269) to keep dependency resolution functions in the same place.

nathanwn: Got it.

tromai: Hmm I don't see why the question could be irrelevant even though the change belongs to this PR. 
I guess the concerned part is whether we want to address it in this PR or not. I guess because this function wasn't authored by @benmss , it's okay for us to address in another PR (if applicable). However, I don't think there is any issue if we want to know more about the code in the first place. 
To me, this `Any` type annotation is a bit confusing and I would raise the same question if I were reviewing this PR.

behnazh-w: @tromai a review tip: when we review a PR we need to understand the context and why a change has happened. If a question needs to be asked outside the context of a PR, we can explicitly mention that, ask the relevant contributor, and follow up separately. For example in this case, @benmss was not the author of the code in the first place, so asking him about this code is not useful. But this question can be asked from the actual author (by mentioning that person) and follow up if needed separately.

src/macaron/repo_finder/repo_finder.py
benmss: @behnazh-w @tromai Was the original intent of this to return the first element as the comment suggests, or the last as the code performs?

behnazh-w: The issue is in my comment, which should be the last element not first.

src/macaron/repo_finder/repo_finder_java.py
benmss: I may have got carried away with the idea of iterators inside iterators. The simpler solution _is_ to move the URL validation function (again) so it can be used inside the Repo Finders.

src/macaron/util.py
```diff
@@ -254,3 +256,28 @@ def get_if_exists(doc: JsonType, path: list[str | int]) -> JsonType | None:
         else:
             return None
     return doc
+
+
+def find_valid_url(urls: Iterable[str]) -> str:
```
nathanwn, line 261: I guess you need to move this function out of `repo_finder.py` due to cyclic imports. 

Also now that I look at it, this function seems to be about URLs of git repositories, and thus it may go into the `macaron.slsa_analyzer.git_url` module, and named `find_valid_git_url`/`find_valid_vcs_url`? Not sure how helpful it is in terms of organizing the code. I don't have too strong of an opinion about this.

nathanwn (approved):

src/macaron/util.py
behnazh-w: This function shouldn't be moved to the `util` module. In fact, we have been trying to remove functions from this module over time. Otherwise `util` quickly becomes a bag of things.

This function is selecting a repository URL when there are multiple of them. When we didn't have a generic repo finder, multiple URLs would be collected while analyzing an SBOM only. Now that we allow `PURL` identifier as input, it will also be meaningful for the main target. So, it's better to be part of the repo finder module.

tromai (approved):

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -527,20 +527,31 @@ def to_analysis_target(config: Configuration, available_domains: list[str]) -> A
...
             If the PURL provided from the user is invalid.
         """
-        # Due to the current design of Configuration class. purl, repo_path, branch and digest are initialized
-        # as empty strings and we assumed that they are always set with input values as non-empty strings.
+        # Due to the current design of Configuration class, repo_path, branch and digest are initialized
+        # as empty strings, and we assumed that they are always set with input values as non-empty strings.
         # Therefore, their true types are ``str``, and an empty string indicates that the input value is not provided.
-        purl_input_str: str = config.get_value("purl")
+        # The purl might be a PackageURL type, or a string, and therefore requires extra care.
+        parsed_purl: PackageURL | None
```
behnazh-w, line 534: The comment says the ` parsed_purl` can be `PackageURL` or a string. I think the comment needs to be adjusted because it can also be `None` and the type should also allow `str`?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -527,20 +527,31 @@ def to_analysis_target(config: Configuration, available_domains: list[str]) -> A
...
+        if config.get_value("purl") is None or config.get_value("purl") == "":
+            parsed_purl = None
+        elif isinstance(config.get_value("purl"), str):
+            try:
+                parsed_purl = PackageURL.from_string(config.get_value("purl"))
+            except InvalidPURLError as error:
```
behnazh-w, line 540: Looking at the [exceptions](https://github.com/package-url/packageurl-python/blob/0d3336804ce6dac59975c8c170d4015149442a93/src/packageurl/__init__.py#L484) raised by `PackageURL.from_string`, we should catch `ValueError` instead of `InvalidPURLError`, which is a Macaron exception.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -527,20 +527,31 @@ def to_analysis_target(config: Configuration, available_domains: list[str]) -> A
...
+                raise PURLNotFoundError("Invalid input PURL.") from error
```
behnazh-w, line 541: I think `InvalidPURLError` would be more suitable to be raised. Also, it would be helpful to include the `config.get_value("purl"),` in the error.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -527,20 +527,31 @@ def to_analysis_target(config: Configuration, available_domains: list[str]) -> A
...
+        else:
```
behnazh-w, line 542: Can we check that the type is `PackageURL` and in the `else` branch raise an exception?

tests/slsa_analyzer/test_analyzer.py
```diff
@@ -112,5 +111,5 @@ def test_resolve_analysis_target(
 )
 def test_resolve_analysis_target_invalid_purl(config: Configuration) -> None:
     """Test the resolve analysis target method with invalid inputs."""
-    with pytest.raises(InvalidPURLError):
+    with pytest.raises(ValueError):  # noqa: PT011
```
behnazh-w, line 114: If you fix the exceptions, this line shouldn't change.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -724,33 +561,24 @@ def to_analysis_target(config: Configuration, available_domains: list[str]) -> A
...
-                    parsed_purl = PackageURL.from_string(purl_input_str)
-                except ValueError as error:
-                    raise InvalidPURLError(f"The package url {purl_input_str} is not valid.") from error
-
-                converted_repo_path = Analyzer.to_repo_path(parsed_purl, available_domains)
-                # TODO: If ``converted_repo_path`` is None, this means that the PURL given by the user if not pointing
-                # to a git repository (e.g ``pkg:maven/apache/maven@1.0.0``). Resolving the repository
-                # path from such PURl string will be handled in https://github.com/oracle/macaron/pull/388.
+                repo = ""
+                converted_repo_path = repo_finder.to_repo_path(parsed_purl, available_domains)  # type: ignore[arg-type]
```
behnazh-w, line 567: Looks like mypy fails to support `match case` properly. Instead of ignoring the type, can you please check that `parsed_purl` is not `None` and add a comment that this assertion is just to help mypy resolve the type correctly, even though `parsed_purl` cannot be `None` here?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -724,33 +561,24 @@ def to_analysis_target(config: Configuration, available_domains: list[str]) -> A
...
+                if converted_repo_path is None:
+                    # Try to find repo from PURL
+                    repo = repo_finder.find_repo(parsed_purl)  # type: ignore[arg-type]
```
behnazh-w, line 570: Same comment here.

src/macaron/output_reporter/reporter.py
```diff
@@ -120,6 +120,9 @@ def generate(self, target_dir: str, report: Report | dict) -> None:
         try:
             dep_file_name = os.path.join(target_dir, "dependencies.json")
             serialized_configs = list(report.get_serialized_configs())
+            for report_dict in serialized_configs:
+                # Serialize PackageURL objects as strings
+                report_dict["purl"] = str(report_dict.get("purl"))
```
behnazh-w, line 125: Instead of modifying the `Configuration` instance here, it's better to make sure it is always stored as a `str` when it is instantiated. So, I suggest to remove this change and instead store it as a string `"purl": str(value.get("purl")),` here:
https://github.com/oracle/macaron/blob/0f426982837708539fb64b76001b823b79596b0e/src/macaron/dependency_analyzer/dependency_resolver.py#L266

src/macaron/dependency_analyzer/dependency_resolver.py
```diff
@@ -165,60 +164,14 @@ def add_latest_version(
         else:
             try:
                 if (
-                    (latest_version := latest_value.get("version"))
-                    and (item_version := item.get("version"))
+                    (latest_version := latest_value.get("purl").version)  # type: ignore[union-attr]
+                    and (item_version := item.get("purl").version)  # type: ignore[union-attr]
```
behnazh-w, line 168: Instead of ignoring the types, can you add a check to make sure `latest_value.get("purl")` and  `item.get("purl")` are not `None`?

benmss: I'm not sure I can. All of these fail to make mypy happy:
```
if (latest_value.get("purl") is not None and item.get("purl") is not None):
    if (
        (latest_version := latest_value.get("purl").version)
        and (item_version := item.get("purl").version)
        and version.Version(latest_version) < version.Version(item_version)
    ):
        latest_deps[key] = item
```
        
```
if (
    latest_value.get("purl") is not None
    and item.get("purl") is not None
    and (latest_version := latest_value.get("purl").version)
    and (item_version := item.get("purl").version)
    and version.Version(latest_version) < version.Version(item_version)
):
    latest_deps[key] = item
```

```
if latest_value.get("purl") is not None and item.get("purl") is not None:
    latest_value_purl: PackageURL = latest_value.get("purl")
    item_purl: PackageURL = item.get("purl")
    if (
        (latest_version := latest_value_purl.version)
        and (item_version := item_purl.version)
        and version.Version(latest_version) < version.Version(item_version)
    ):
        latest_deps[key] = item
```

benmss: It seems this works:
```
latest_value_purl = latest_value.get("purl")
item_purl = item.get("purl")
    if (
        latest_value_purl is not None
        and item_purl is not None
        and (latest_version := latest_value_purl.version)
        and (item_version := item_purl.version)
        and version.Version(latest_version) < version.Version(item_version)
    ):
        latest_deps[key] = item
```

src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -0,0 +1,198 @@
...
+
+        if version:
+            return [f"{base_url}/versions/{version}"]
+
+        # Find the latest version.
+        try:
+            response = send_get_http_raw(base_url, {})
+        except ReadTimeout:
+            logger.debug("Failed to retrieve version (timeout): %s:%s", namespace, name)
+            return []
```
behnazh-w, line 94: Catching this exception shouldn't be necessary anymore because this [PR](https://github.com/oracle/macaron/pull/475) [adds exception handling](https://github.com/oracle/macaron/blob/0f426982837708539fb64b76001b823b79596b0e/src/macaron/util.py#L83) in `send_get_http_raw`.

src/macaron/repo_finder/repo_finder.py
```diff
@@ -0,0 +1,150 @@
...
+Artifact PURLs
+--------------
+For artifact PURLs, the PURL type determines how the repositories are searched for.
+Currently, for Maven PURLs, SCM meta data is retrieved from the matching POM retrieved from Maven Central (or
+other configured location).
+
+For Python, .NET, Rust, and NodeJS type PURLs, Google's Open Source Insights API is used to find the meta data.
+
+In either case, any repository links are extracted from the meta data, then checked for validity via
+``utils::find_valid_url`` which accepts URLs that point to a Github repository or similar.
```
behnazh-w, line 22:
```suggestion
``repo_validator::find_valid_repository_url`` which accepts URLs that point to a Github repository or similar.
```

src/macaron/dependency_analyzer/dependency_resolver.py
```diff
@@ -309,6 +267,121 @@ def tool_valid(tool: str) -> bool:
...
+
+                # We collect the generated SBOM as a best effort, even if the build exits with errors.
+                # TODO: add improvements to help the SBOM build succeed as much as possible.
+                deps_resolved |= dep_analyzer.collect_dependencies(str(working_dir))
+
+            logger.info("Stored dependency resolver log for %s to %s.", dep_analyzer.tool_name, log_path)
+
+        # Use repo finder to find more repositories to analyze.
+        if defaults.getboolean("repofinder", "find_repos"):
+            DependencyAnalyzer._resolve_more_dependencies(deps_resolved)
```
behnazh-w, line 366: The repo finder seems to be called only when an SBOM is not available. This function returns [earlier](https://github.com/oracle/macaron/pull/388/files#diff-c95724433501b4a68d8ffe2e35305bff56f2c26c45571f9250641511a8d48269R292) if an SBOM is available and does not try to resolve more dependencies.

benmss:
Blocked by #479

behnazh-w (approved):

PR #397, opened 2023-08-04
fix(orm): use the host’s timezone when persisting datetime objects without a timezone, instead of forcing them to UTC

jenstroeger:


behnazh-w (approved):
Thanks for the contribution!

PR #400, opened 2023-08-04
chore: return the runner back to runner queue on check exception

tromai:
Similar to https://github.com/oracle/macaron/pull/216 . But in https://github.com/oracle/macaron/pull/216 I missed one case, which is addressed in this PR.


tromai:
Force pushing at https://github.com/oracle/macaron/commit/32c59834148acdf3deb3210b1e5e3a60be919a69 due to me pushing an irrelevant commit to this branch :facepalm:

tromai:
Close as #650 has completely removed the multi-thread bits.

PR #408, opened 2023-08-10
docs: adjust the Makefile target `docs` to include untracked files

nathanwn:
In #370, I forgot that `git ls-files` does not include untracked files by default. This means newly created source files that have not been staged using `git add` are not included in the dependencies of `make docs`.

This PR is to fix the issue.

We use these flags with `git ls-files`:
* [`-c|--cached`](https://git-scm.com/docs/git-ls-files#Documentation/git-ls-files.txt---cached) to include tracked files (Note: this flag is on by default but is off if the flags below are enabled).
* [`-o|--others`](https://git-scm.com/docs/git-ls-files#Documentation/git-ls-files.txt---others) to include untracked files.
* [`--exclude-standard`](https://git-scm.com/docs/git-ls-files#Documentation/git-ls-files.txt---exclude-standard) to prevent listing files excluded by `.gitignore`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -293,7 +293,7 @@ dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-build-epoch.txt:
 	echo $(SOURCE_DATE_EPOCH) > dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-build-epoch.txt
 
 # Build the HTML documentation from the package's source.
-DOCS_SOURCE := $(shell git ls-files docs/source)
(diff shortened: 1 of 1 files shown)


behnazh-w:
Is this PR ready for review?

nathanwn:
@behnazh-w As I discussed, I think it is best that we remove caching for the `docs` target altogether, as it seems like we don't have a reliable caching mechanism while building the docs is relatively fast already.

nathanwn:
Closing this. The fix is in #411.

PR #409, opened 2023-08-13
feat: add docker build detection

timyarkov:
This PR adds support for the detection of Dockerfiles, so as to cover scenarios where Docker gets used as a build tool, e.g. when Macaron gets built using its Dockerfile.

- Docker presence is detected by finding files either named `Dockerfile` or in the formats `*.Dockerfile` or `Dockerfile.*` to cover different naming conventions of dockerfiles, e.g. `dev.Dockerfile` or like Macaron's own `Dockerfile.base` and `Dockerfile.final`. This is defined in `defaults.ini` under `[builder.docker]`
- The supported build command keyword is `build`, and supported deploy command keyword is `push`, defined in `defaults.ini` under `[builder.docker]`
- For CI deploy commands the Github action [docker/build-push-action@v4](https://github.com/docker/build-push-action) is supported, defined in `defaults.ini` under `[builder.docker.ci.deploy]`

To do from here:

- Implement dependency resolution for dockerfiles, as of now this is purely detection and does not include anything regarding SBOMs
- Further testing; as of now it has a single test for detection of docker using https://github.com/timyarkov/docker_test, however it would be good to expand tests to cover scenarios like multiple build tools alongside docker, more naming variations, etc.

### DIFF
--- a/src/macaron/config/defaults.ini
+++ b/src/macaron/config/defaults.ini
@@ -300,6 +300,20 @@ deploy_arg =
 [builder.poetry.ci.deploy]
 github_actions = pypa/gh-action-pypi-publish
 
+# This is the spec for trusted Docker build tool usages.
+[builder.docker]
+entry_conf =
+# Also account for if there's multiple dockerfiles, e.g. test.Dockerfile
+build_configs =
+    Dockerfile
+    Dockerfile.*
+    *.Dockerfile
+builder = docker
+build_arg = build
+deploy_arg = push
+[builder.docker.ci.deploy]
+github_actions = docker/build-push-action@v4
+
 # This is the spec for GitHub Actions CI.
 [ci.github_actions]
 entry_conf =
--- a/src/macaron/config/defaults.ini
+++ b/src/macaron/config/defaults.ini
@@ -308,11 +308,15 @@ build_configs =
     Dockerfile
     Dockerfile.*
     *.Dockerfile
-builder = docker
-build_arg = build
-deploy_arg = push
+builder =
+    docker
+build_arg =
+    build
+deploy_arg =
+    push
 [builder.docker.ci.deploy]
-github_actions = docker/build-push-action@v4
+github_actions =
+    docker/build-push-action
 
 # This is the spec for GitHub Actions CI.
 [ci.github_actions]
--- a/src/macaron/slsa_analyzer/build_tool/docker.py
+++ b/src/macaron/slsa_analyzer/build_tool/docker.py
@@ -1,0 +1,92 @@
+# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""This module contains the Docker class which inherits BaseBuildTool.
+
+This module is used to work with repositories that use Docker as a build tool.
+"""
+
+from macaron.config.defaults import defaults
+from macaron.dependency_analyzer.dependency_resolver import NoneDependencyAnalyzer
+from macaron.slsa_analyzer.build_tool.base_build_tool import BaseBuildTool, file_exists
+
+
+class Docker(BaseBuildTool):
+    """This class contains the information of Docker when used as a build tool."""
+
+    def __init__(self) -> None:
+        """Initialize instance."""
+        super().__init__(name="docker")
+
+    def load_defaults(self) -> None:
+        """Load the default values from defaults.ini."""
+        if "builder.docker" in defaults:
+            for item in defaults["builder.docker"]:
+                if hasattr(self, item):
+                    setattr(self, item, defaults.get_list("builder.docker", item))
+
+        if "builder.docker.ci.deploy" in defaults:
+            for item in defaults["builder.docker.ci.deploy"]:
+                if item in self.ci_deploy_kws:
+                    self.ci_deploy_kws[item] = defaults.get_list("builder.docker.ci.deploy", item)
+
+    def is_detected(self, repo_path: str) -> bool:
+        """Return True if this build tool is used in the target repo.
+
+        Parameters
+        ----------
+        repo_path : str
+            The path to the target repo.
+
+        Returns
+        -------
+        bool
+            True if this build tool is detected, else False.
+        """
+        for file in self.build_configs:
+            if file_exists(repo_path, file):
+                return True
+
+        return False
+
+    def prepare_config_files(self, wrapper_path: str, build_dir: str) -> bool:
+        """Make necessary preparations for using this build tool.
+
+        Parameters
+        ----------
+        wrapper_path : str
+            The path where all necessary wrapper files are located.
+        build_dir : str
+            The path of the build dir. This is where all files are copied to.
+
+        Returns
+        -------
+        bool
+            True if succeed else False.
+        """
+        # TODO: Future dependency analysis may require some preprocessing, e.g.
+        #       saving images to tar files. Need to investigate when implementing
+        #       and work with this method accordingly.
+
+        return False
+
+    def get_dep_analyzer(self, repo_path: str) -> NoneDependencyAnalyzer:
+        """Create a DependencyAnalyzer for the Docker build tool. Currently unimplemented.
+
+        Parameters
+        ----------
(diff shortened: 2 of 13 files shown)


go.mod
timyarkov, line 1: Dunno where the go.mod and go.sum got changed, I tried checking it out from staging while I was having issues committing my stuff but it seems between then and me finally getting it working it changed again. Does it look fine or should I make sure its reset to what it is in staging?

tromai:
Please rebase on staging as this PR should address the issue with our CI pipeline - https://github.com/oracle/macaron/pull/410

tromai:
@timyarkov It would be good if we have a description of the feature's behaviors in the PR's description whenever a new feature is developed. It could really help us with reviewing the code changes.
For example: https://github.com/oracle/macaron/pull/67 and https://github.com/oracle/macaron/pull/316. 
Note that it doesn't need a full detail explanation of the implementation but a short description of how the features work as the user's perspective (any configuration needed, what sort of inputs does it accept, etc.).

timyarkov:
> @timyarkov It would be good if we have a description of the feature's behaviors in the PR's description whenever a new feature is developed. It could really help us with reviewing the code changes. For example: #67, #316 and #254 (minus the implementation section). Note that it doesn't need a full detail explanation of the implementation but a short description of how the features work as the user's perspective:
> 
> * any configuration needed
> * what sort of inputs does it accept
> * what is the problem the feature is going to solve

My bad, have updated it with more details, hopefully that is enough?

tromai:
> > @timyarkov It would be good if we have a description of the feature's behaviors in the PR's description whenever a new feature is developed. It could really help us with reviewing the code changes. For example: #67, #316 and #254 (minus the implementation section). Note that it doesn't need a full detail explanation of the implementation but a short description of how the features work as the user's perspective:
> > 
> > * any configuration needed
> > * what sort of inputs does it accept
> > * what is the problem the feature is going to solve
> 
> My bad, have updated it with more details, hopefully that is enough?

No worries. That's great. Thanks for that.

src/macaron/config/defaults.ini
```diff
@@ -295,6 +295,20 @@ deploy_arg =
 [builder.poetry.ci.deploy]
 github_actions = pypa/gh-action-pypi-publish
 
+# This is the spec for trusted Docker build tool usages.
+[builder.docker]
+entry_conf =
+# Also account for if there's multiple dockerfiles, e.g. test.Dockerfile
+build_configs =
+    Dockerfile
+    Dockerfile.*
```
tromai, line 304: Hmm, the [`file_exists`](https://github.com/oracle/macaron/blob/staging/src/macaron/slsa_analyzer/build_tool/base_build_tool.py) method is not supposed to expand the wildcard pattern from file names (at least that is what I have been assuming for a long time). 
I guess it will expand the wild card pattern because at [these lines](https://github.com/oracle/macaron/blob/ae2c859ba13451bff35dc7986a029a8151006ca0/src/macaron/slsa_analyzer/build_tool/base_build_tool.py#L35C1-L36C57) we are concatenating the `file_name` variable (which is the direct input from `defaults.ini`) to the call of `glob.iglob`. 
@behnazh-w Is this behavior expected ? If yes, I guess we should improve the documentation in the `files_exist` method to notice about this behavior.

src/macaron/slsa_analyzer/build_tool/docker.py
```diff
@@ -0,0 +1,92 @@
...
+        Returns
+        -------
+        bool
+            True if succeed else False.
+        """
+        # TODO: Future dependency analysis may require some preprocessing, e.g.
+        #       saving images to tar files. Need to investigate when implementing
+        #       and work with this method accordingly.
+
+        return True
```
tromai, line 71: By returning True here, do you mean that we never need to prepare anything to use this build tool, or there isn't any implementation yet for it.
If it's the latter case, I guess we should return False here :thinking:

tromai:
Please rebase on staging. Hopefully this [PR](https://github.com/oracle/macaron/pull/419) would prevent this issue from happening again.

src/macaron/config/defaults.ini
```diff
@@ -295,6 +295,20 @@ deploy_arg =
...
+    *.Dockerfile
```
tromai, line 305: Thinking more about this pattern. Have you ever encountered such usage in any of the Github projects? Because the usages like `dev.Dockerfile` or `test.Dockerfile` are very unusual to me. I haven't seen them any where. 
If there is an existing project that use this pattern, it's good to share the link here so that we could discuss too.

timyarkov: I chose to include those since while during research I saw somewhere that those would be considered valid names, as such I thought it would be good to include for the sake of completeness. Searching github does show some repos that use this convention though:
- `dev.Dockerfile` in https://github.com/logspace-ai/langflow
- Didn't find one with `test.Dockerfile` specifically from a quick search but did see a few along the lines of `latest.Dockerfile`/similar, e.g. https://github.com/cube-js/cube/blob/c8ef602495edfd6e6c8b898bb9a61bb5be6602ae/packages/cubejs-docker/latest.Dockerfile#L4 among others

tromai: I think it's also interesting to see how each of those Dockerfile are used and whether it's actually used **build** the final package or not. 
- In the first case - `dev.Dockerfile in https://github.com/logspace-ai/langflow`. I have a quick look and it seems that this is only used for development purposes and doesn't build the final production image. Let's say, a repository has `dev.Dockerfile` for development purposes only and there is not other files that match these patterns, Macaron still consider that repository has Dockerfile build tool (in this case that is a False Positive). Could we do anything to prevent such cases :thinking: ? 
- As for the second case, that is a use case for building the production image.

timyarkov: Hm good point, don't know if its a bit much but maybe seeing what gets referenced in pipelines could be an indicator for whether to count a dockerfile or not? Could mean that repos without pipelines could potentially get missed out on, but at least it would be reduced false positives?

src/macaron/config/defaults.ini
behnazh-w: Yes, this behavior is expected. Based on the [`glob.iglob`](https://docs.python.org/3/library/glob.html#glob.iglob), which follows [`glob.glob`](https://docs.python.org/3/library/glob.html#glob.glob):
>pathname can contain shell-style wildcards

We could add a note to [`file_exists`](https://github.com/oracle/macaron/blob/staging/src/macaron/slsa_analyzer/build_tool/base_build_tool.py) but that's expected of `glob.iglob` to handle wildcards anyway.

src/macaron/config/defaults.ini
behnazh-w: At this point we can detect candidate build tools. The `build_as_code` check then looks for commands executing the build tool in shell scripts that are reachable from GitHub Actions workflows. I think at that time we are in a better position to report whether the build tool is actually used for build and deployment.

behnazh-w (commented):
Please add unit tests following the examples for build tools [here](https://github.com/oracle/macaron/tree/0f56f663c9262d52d96dd905915c8c08de3b68b9/tests/slsa_analyzer/build_tool).

timyarkov:
> Please add unit tests following the examples for build tools [here](https://github.com/oracle/macaron/tree/0f56f663c9262d52d96dd905915c8c08de3b68b9/tests/slsa_analyzer/build_tool).

Added in latest commit https://github.com/oracle/macaron/pull/409/commits/33d4595a70dc3413d23f7cf324ad86371f900902

src/macaron/config/defaults.ini
```diff
@@ -295,6 +295,20 @@ deploy_arg =
...
+builder = docker
+build_arg = build
+deploy_arg = push
+[builder.docker.ci.deploy]
+github_actions = docker/build-push-action@v4
```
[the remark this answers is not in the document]
behnazh-w: Why is the version for this GitHub Action specified here?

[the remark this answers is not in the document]
timyarkov: It was the one docker's documentation used, just looked though that it looks like the codebase as its setup can accept any version? In that case I can fix this up

behnazh-w (approved):

tromai (approved):

PR #411, opened 2023-08-14
docs: always clean and build docs in Makefile docs target

behnazh-w:
It turns out that keeping track of source files under docs/ is not trivial. So, this commit changes the docs target in Makefile by always removing the `docs/_build/` and building the docs.


Makefile
```diff
@@ -293,12 +293,9 @@ dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-build-epoch.txt:
 	echo $(SOURCE_DATE_EPOCH) > dist/$(PACKAGE_NAME)-$(PACKAGE_VERSION)-build-epoch.txt
 
 # Build the HTML documentation from the package's source.
-DOCS_SOURCE := $(shell git ls-files docs/source)
 .PHONY: docs
-docs: docs/_build/.built-on
-docs/_build/.built-on: $(DOCS_SOURCE)
+docs: docs-clean
```
nathanwn, line 297: Do we need to run the `docs-clean` target every time we run `make docs`?
I think Sphinx has some caching mechanism out-of-the-box and doing this will completely invalidate Sphinx's own cache every time.
There is also, to my knowledge, no side-effect of running `make docs` repeatedly, so we probably do not need to run `docs-clean` all the time.

behnazh-w: If a source file is deleted, there will be a lingering HTML file that Sphinx does not remove. That's why the  `docs-clean` is added to have the correct outputs, giving up on Sphinx's caching mechanism as well.

nathanwn: I see. Thanks for clarifying.

nathanwn (approved):

PR #421, opened 2023-08-21
feat!: implement two-person reviewed check

Yao-Wen-Chang:
Add the new check called "two-person reviewed".
Add new parameter (params) to def send_get_http_raw(url: str, headers: dict, params: dict | None) -> Response | None .
Add two methods in api_client.py (list_pull_requests, get_a_review).
Write a test function for the check.


behnazh-w:
Closing as it duplicates #425

PR #424, opened 2023-08-21

Yao-Wen-Chang:
SLSA Level 4 requirement contains two-person reviewed.
This check guarantees each PR includes at least one authenticated reviewer.
And, this check would scan through the whole PR IDs, then summarize the pass rate.


behnazh-w:
Closing as it duplicates #425

PR #425, opened 2023-08-21

Yao-Wen-Chang:
SLSA Level 4 requirement contains two-person reviewed.
This check guarantees each PR includes at least one authenticated reviewer.
And, this check would scan through the whole PR IDs, then summarize the pass rate.


Yao-Wen-Chang:
[two-person review requirement](https://slsa.dev/spec/v0.1/requirements)
In the above link, the author mentioned: 
**The following combinations are acceptable:**
- Uploader and reviewer are two different trusted persons.
- Two different reviewers are trusted persons.

Does this mean only one authenticated reviewer is acceptable, since the Uploader is the contributor other than the maintainer?

PR #432, opened 2023-08-23
fix: update justifications to be complete for multi build tool projects

timyarkov:
Right now while projects with multiple build tools are analysed properly, the justifications in the report don't reflect all build tools used due to logic that short-circuits the `build_as_code` and `build_service` checks upon a single tool passing. This is a problem when testing for if certain build tools get picked up in scenarios of multiple other build tools being present, since there is a chance it likely won't get reflected in the report, thus making it difficult to properly test such cases.

- `build_as_code_check.py` and `build_service_check.py` now perform their checks for every single build tool, regardless of if there's a pass or not (still preserving previous one pass = check passed logic however)
- Results for analysis of https://github.com/timyarkov/multibuild_test updated to reflect changes

### DIFF
--- a/src/macaron/slsa_analyzer/checks/build_as_code_check.py
+++ b/src/macaron/slsa_analyzer/checks/build_as_code_check.py
@@ -366,19 +366,22 @@ class BuildAsCodeCheck(BaseCheck):
         ci_services = ctx.dynamic_data["ci_services"]
 
         # Check if "build as code" holds for each build tool.
+        overall_res = CheckResultType.FAILED
+
         for tool in build_tools:
             res = self._check_build_tool(tool, ctx, ci_services, check_result)
 
             if res == CheckResultType.PASSED:
-                # Since the check passing is contingent on at least one passing,
-                # short-circuit if we do get a pass
+                # The check passing is contingent on at least one passing, if
+                # one passes treat whole check as passing. We do still need to
+                # run the others for justifications though to report multiple
+                # build tool usage.
                 # TODO: When more sophisticated build tool detection is
                 # implemented, consider whether this should be one fail = whole
                 # check fails instead
-                return CheckResultType.PASSED
+                overall_res = CheckResultType.PASSED
 
-        # No passes, so overall fail
-        return CheckResultType.FAILED
+        return overall_res
 
 
 registry.register(BuildAsCodeCheck())
--- a/src/macaron/slsa_analyzer/checks/build_as_code_check.py
+++ b/src/macaron/slsa_analyzer/checks/build_as_code_check.py
@@ -127,6 +127,21 @@ class BuildAsCodeCheck(BaseCheck):
                         return str(com)
         return ""
 
+    def _add_to_result_table(self, facts: BuildAsCodeFacts, check_result: CheckResult) -> None:
+        """Add to the check result result tables, creating the list if it doesn't yet exist.
+
+        Parameters
+        ----------
+        facts : BuildAsCodeFacts
+            Facts to add to the result tables.
+        check_result : CheckResult
+            The object containing result data of a check.
+        """
+        if "result_tables" in check_result.keys():
+            check_result["result_tables"].append(facts)
+        else:
+            check_result["result_tables"] = [facts]
+
     def _check_build_tool(
         self, build_tool: BaseBuildTool, ctx: AnalyzeContext, ci_services: list[CIInfo], check_result: CheckResult
     ) -> CheckResultType | None:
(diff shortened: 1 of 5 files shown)


src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -366,19 +366,22 @@ def run_check(self, ctx: AnalyzeContext, check_result: CheckResult) -> CheckResu
         ci_services = ctx.dynamic_data["ci_services"]
 
         # Check if "build as code" holds for each build tool.
+        overall_res = CheckResultType.FAILED
+
         for tool in build_tools:
             res = self._check_build_tool(tool, ctx, ci_services, check_result)
```
tromai, line 372: I think there might be issues if we run `self._check_build_tool` multiple times because at this [line](https://github.com/oracle/macaron/blob/68a54ac5573460e8d61fcb9662baa81c783c0c09/src/macaron/slsa_analyzer/checks/build_as_code_check.py#L224), we are creating a new list with the corresponding `BuildAsCodeFacts` table (which will be stored into the database). 
The list was created this way because we have been assuming to stop on a passing BuildAsCode run for a single build tool. 
Now if we run it multiple times, this list will be overwritten each time :thinking: , whereas the expected behavior is that each time we run `self._check_build_tool` with a new build tool, a different `BuildAsCodeFacts` instance should be appended to the list. Can you have a look to see if this is the case in terms of the values stored into the database? The `BuildAsCodeFacts` instance will be stored as the table called `_build_as_code_check`.

timyarkov: Thanks for the catch, far as I can see it was .append() in my original PR (but also there was still an equals in one spot either because I missed it or some other reason?), but it changed to an equals at some point. In any case I've updated it now so it should hopefully be all good. I've also made a couple tests for this in my latest commit, do they look alright?

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -127,6 +127,21 @@ def _has_deploy_command(self, commands: list[list[str]], build_tool: BaseBuildTo
...
+        """Add to the check result result tables, creating the list if it doesn't yet exist.
+
+        Parameters
+        ----------
+        facts : BuildAsCodeFacts
+            Facts to add to the result tables.
+        check_result : CheckResult
+            The object containing result data of a check.
+        """
+        if "result_tables" in check_result.keys():
```
tromai, line 140: If I understand correctly, the `result_tables` field in the `CheckResult` instance passed to the check has already been initialized to an empty list (see [here](https://github.com/oracle/macaron/blob/386c085d9ad30dd9eb9b35aef0c19b91718e02f7/src/macaron/slsa_analyzer/checks/base_check.py#L81)). 
Then wouldn't we need to check again to see if `result_tables` is defined. So I guess you could just append new `BuildAsCodeFacts` instances to it directly. @behnaz Please let me know if my understanding is correct.

tests/e2e/expected_results/multibuild_test/multibuild_test.json
```diff
@@ -211,19 +212,19 @@
                 "num_deps_pass": 0
             },
             {
-                "check_id": "mcn_provenance_available_1",
+                "check_id": "mcn_provenance_witness_level_one_1",
```
tromai, line 215: Are these new changes because of the latest rebase on `staging`? That's what I thought, but just want to make sure.

src/macaron/slsa_analyzer/checks/build_as_code_check.py
timyarkov: I did try just .append, but there somehow were scenarios where I'd be getting a `KeyError` for `result_tables`, so I thought it would be best to just do it like this to be safe. If its an issue though I can try digging into why that happened

tests/slsa_analyzer/checks/test_build_as_code_check.py
```diff
@@ -287,3 +287,52 @@ def test_travis_ci_deploy(
     gradle_deploy.dynamic_data["ci_services"] = [ci_info]
 
     assert check.run_check(gradle_deploy, check_result) == expect_result
+
+
+@pytest.mark.parametrize(
```
tromai, line 292: This is related to my comment [here](https://github.com/oracle/macaron/pull/432/files#r1306875239). We could decide whether we need this test after that comment has been agreed upon. I believe what you are trying to test in this method is whether the checks work fine if the `result_tables` fields are defined/not defined.

tests/e2e/expected_results/multibuild_test/multibuild_test.json
timyarkov: The changes to the justifications were to reflect the new non-short-circuiting behaviour, this in particular yeah I believe its from the latest on staging (had to add similar to my other PR too)

tests/slsa_analyzer/checks/test_build_as_code_check.py
timyarkov: Agreed, if its a core issue that needs fixing I can update the tests accordingly

src/macaron/slsa_analyzer/checks/build_as_code_check.py
timyarkov: Managed to reproduce the KeyError, I missed though that it was happening in `test_travis_ci_deploy` and not the code itself. Latest commit should fix that + reverts to just `.append()`s

tromai (approved):
LGTM. Please wait for @behnazh-w 's feedback before merging.

behnazh-w (approved):
Thanks for your contribution @timyarkov!

PR #436, opened 2023-08-27
docs: fix a typo and missing docstrings in `jfrog_maven_registry` module

behnazh-w:


src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
```diff
@@ -132,6 +117,24 @@ def __init__(
...
+        ----------
+        hostname : str
+            The hostname of the JFrog instance.
+        repo : str
+            The Artifactory repository with Maven layout on the JFrog instance.
+        request_timeout : int
+            The timeout (in seconds) for regular requests made to the package registry.
+        download_timeout : int
+            The timeout (in seconds) for downloading files from the package registry.
+        enabled : bool
```
nathanwn, line 133: I think it should be this instead (as these are the types of constructor params, not the types of instance variables): 

```suggestion
        hostname : str | None
            The hostname of the JFrog instance.
        repo : str | None
            The Artifactory repository with Maven layout on the JFrog instance.
        request_timeout : int | None
            The timeout (in seconds) for regular requests made to the package registry.
        download_timeout : int | None
            The timeout (in seconds) for downloading files from the package registry.
        enabled : bool | None
```

nathanwn (approved):
There is only a comment about type annotations of the `JFrogMavenRegistry` constructor.
Other than that, looks good to me. Thanks, Behnaz.

src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py
behnazh-w: See commit [33e0321](https://github.com/oracle/macaron/pull/436/commits/33e03210f8109f1c5992ff0035926c4b73309ad4)

PR #446, opened 2023-08-29
fix: add timeout to Gradle Group ID detection

behnazh-w:
This PR sets a timeout on the `subprocess.run()` call that runs `gradlew properties` to detect the group ID of a Gradle project. The `timout` is read from `build_timeout` in `defaults.ini` for the detected build tool.

### DIFF
--- a/src/macaron/config/defaults.ini
+++ b/src/macaron/config/defaults.ini
@@ -118,7 +118,6 @@ deploy_arg =
     org.sonatype.plugins:nexus-staging-maven-plugin:deploy-staged-repository
     # See https://help.sonatype.com/repomanager3/integrations/nexus-repository-maven-plugin.
     nxrm3:staging-deploy
-
 build_log = Apache Maven
 wrapper_files =
     .mvn/wrapper/maven-wrapper.jar
@@ -242,6 +241,10 @@ jenkins =
     gradle-git-publish
     gitPublishPush
 
+[builder.gradle.runtime]
+# This is the timeout (in seconds) to run the build tool.
+build_timeout = 600
+
 # This is the spec for trusted Pip packaging tools.
 [builder.pip]
 entry_conf =
@@ -271,6 +274,7 @@ build_arg =
 deploy_arg =
     publish
     upload
+
 [builder.pip.ci.deploy]
 github_actions = pypa/gh-action-pypi-publish
 
@@ -292,6 +296,7 @@ build_arg =
     build
 deploy_arg =
     publish
+
 [builder.poetry.ci.deploy]
 github_actions = pypa/gh-action-pypi-publish
 
--- a/src/macaron/slsa_analyzer/build_tool/base_build_tool.py
+++ b/src/macaron/slsa_analyzer/build_tool/base_build_tool.py
@@ -43,8 +43,11 @@ def file_exists(path: str, file_name: str) -> bool:
 
 
 @dataclass
-class RuntimeConfig:
-    """The class for build tool runtime configuration read from `defaults.ini`."""
+class RuntimeOptions:
+    """The class for build tool runtime configurations read from `defaults.ini`.
+
+    Note that Macaron uses the options in this class to "run" a build tool.
+    """
 
     #: The timeout used for running the build tool commands.
     build_timeout: float = 600
@@ -88,7 +91,7 @@ class BaseBuildTool(ABC):
         }
         self.build_log: list[str] = []
         self.wrapper_files: list[str] = []
-        self.runtime_config = RuntimeConfig()
+        self.runtime_options = RuntimeOptions()
 
     def __str__(self) -> str:
         return self.name
--- a/src/macaron/slsa_analyzer/build_tool/base_build_tool.py
+++ b/src/macaron/slsa_analyzer/build_tool/base_build_tool.py
@@ -8,6 +8,7 @@ import logging
 import os
 from abc import ABC, abstractmethod
 from collections.abc import Iterable
+from dataclasses import dataclass
 from pathlib import Path
 
 from macaron.dependency_analyzer import DependencyAnalyzer
@@ -39,6 +40,14 @@ def file_exists(path: str, file_name: str) -> bool:
         return True
     except StopIteration:
         return False
+
+
+@dataclass
+class RuntimeConfig:
+    """The class for build tool runtime configuration read from `defaults.ini`."""
+
+    #: The timeout used for running the build tool commands.
+    build_timeout: float = 600
 
 
 class BaseBuildTool(ABC):
@@ -79,6 +88,7 @@ class BaseBuildTool(ABC):
         }
         self.build_log: list[str] = []
         self.wrapper_files: list[str] = []
+        self.runtime_config = RuntimeConfig()
 
     def __str__(self) -> str:
         return self.name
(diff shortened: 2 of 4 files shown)


src/macaron/slsa_analyzer/build_tool/base_build_tool.py
```diff
@@ -41,6 +42,14 @@ def file_exists(path: str, file_name: str) -> bool:
         return False
 
 
+@dataclass
+class RuntimeConfig:
+    """The class for build tool runtime configuration read from `defaults.ini`."""
+
+    #: The timeout used for running the build tool commands.
+    build_timeout: float = 600
```
tromai, line 50: Could we just set an optional instance variable called `run_timeout` for `BaseBuildTool` sub-classes? Is this class going to store any other fields for each BuildTool?
Do you think that naming it `RunTimeConfig` would be confusing for developers as this class is entirely different from the "defaults.ini" config or the global config for example.

src/macaron/slsa_analyzer/build_tool/gradle.py
```diff
@@ -217,13 +230,17 @@ def get_group_id(self, gradle_exec: str, project_path: str) -> str | None:
             The group id of the project, if exists.
         """
         try:
+            logger.debug(
```
tromai, line 233: Is this meant to be an `INFO` level log ?

src/macaron/config/defaults.ini
```diff
@@ -242,6 +241,10 @@ jenkins =
     gradle-git-publish
     gitPublishPush
 
+[builder.gradle.runtime]
+# This is the timeout (in seconds) to run the build tool.
+build_timeout = 600
```
nathanwn, line 246: I am wondering why we need to create this new `[builder.gradle.runtime]` section for the `build_timeout` config, instead of adding it to the `[build.gradle]` section we already have.

tests/slsa_analyzer/build_tool/test_gradle.py
```diff
@@ -86,3 +87,32 @@ def test_get_group_ids_separate_projects(tmp_path: Path, gradle_tool: Gradle) ->
         "io.micronaut.foo",
         "io.micronaut.bar",
     }
+
+
+@pytest.mark.parametrize(
+    ("timeout", "expected"),
+    [("0", set()), ("invalid", {"io.micronaut"})],
```
nathanwn, line 94: Nit: I personally prefer using trailing comma here to have each test case on a separate line.
```suggestion
    [
        ("0", set()),
        ("invalid", {"io.micronaut"}),
    ],
```

src/macaron/slsa_analyzer/build_tool/base_build_tool.py
behnazh-w: > Could we just set an optional instance variable called `run_timeout` for `BaseBuildTool` sub-classes? Is this class going to store any other fields for each BuildTool?

I think we need to have a section in the `.ini` file for runtime settings explicitly. I expect more options to be added in future, which should be adjusted per build tool.

behnazh-w: > Do you think that naming it `RunTimeConfig` would be confusing for developers as this class is entirely different from the "defaults.ini" config or the global config for example.

Agreed. How about `RuntimeOptions`?

src/macaron/config/defaults.ini
behnazh-w: The reason is that `[build.gradle]` section contains the build tool specification we use in our checks to analyze a build system. However, I have added `[builder.gradle.runtime]` to keep the options that are actually used by Macaron to run the build tools separate.

tests/slsa_analyzer/build_tool/test_gradle.py
behnazh-w: Interesting. I was expecting `black` to pick that up. I guess it's not supporting the pytest annotations?

nathanwn: I think `black` doesn't usually add trailing commas automatically by itself, except for certain cases (e.g. line is too long).

src/macaron/slsa_analyzer/build_tool/base_build_tool.py
behnazh-w: See commit [fd880ec](https://github.com/oracle/macaron/pull/446/commits/fd880ec99dc041209783be5fc8980d7f0163966c)

src/macaron/slsa_analyzer/build_tool/gradle.py
behnazh-w: See commit [fd880ec](https://github.com/oracle/macaron/pull/446/commits/fd880ec99dc041209783be5fc8980d7f0163966c)

tests/slsa_analyzer/build_tool/test_gradle.py
behnazh-w: See commit [fd880ec](https://github.com/oracle/macaron/pull/446/commits/fd880ec99dc041209783be5fc8980d7f0163966c)

nathanwn: Uh, I don't think it is properly formatted now. 😅 The whole decorator collapsed onto one line.

src/macaron/config/defaults.ini
nathanwn: Do you mean these settings are only used in the checks?
```
[builder.gradle]
entry_conf =
    settings.gradle
    settings.gradle.kts
build_configs =
    build.gradle
    build.gradle.kts
builder =
    gradle
    gradlew
build_arg =
deploy_arg =
    artifactoryPublish
    publish
    publishToSonatype
    gradle-git-publish
    gitPublishPush
build_log = gradle
wrapper_files =
    gradle/wrapper/gradle-wrapper.jar
    gradle/wrapper/gradle-wrapper.properties
    gradlew
```

tests/slsa_analyzer/build_tool/test_gradle.py
behnazh-w: Blame `black`!

src/macaron/config/defaults.ini
nathanwn: I am still a bit unclear about the distinguishment between `[build.gradle]` and `[builder.gradle.runtime]`. I saw you mentioned to @tromai that you expect some other things to be added to `[builder.gradle.runtime]`. What would be some examples?

behnazh-w: > I am still a bit unclear about the distinguishment between `[build.gradle]` and `[builder.gradle.runtime]`.

It might be helpful if you take a look at the `build_as_code` check as an example to see how it uses the build tool specs to determine a deployment pipeline.

behnazh-w: > Do you mean these settings are only used in the checks?

Yes

behnazh-w: > I saw you mentioned to @tromai that you expect some other things to be added to `[builder.gradle.runtime]`. What would be some examples?

For example, we might need to enable/disable certain errors, look for certain outputs, etc.

nathanwn: Ok, I think I see what you are trying to do.

nathanwn (approved):

tromai (approved):

PR #448, opened 2023-08-30
fix: always pull latest docker image in run_macaron.sh

### LINKED ISSUE #438: The `latest` tag used in `run_macaron.sh` does not pull the latest image
Running Macaron using `run_macaron.sh`, if an image with `latest` tag already exists on the host, the latest image will not be pulled from the registry even when there is a new release.

nicallen:
Closes #438

By default, always pull latest version of docker image, but allow this behaviour to be overidden by setting the DOCKER_PULL env var.

### DIFF
--- a/.github/workflows/_build_docker.yaml
+++ b/.github/workflows/_build_docker.yaml
@@ -58,5 +58,6 @@ jobs:
       env:
         # This environment variable will be picked up by run_macaron.sh.
         MACARON_IMAGE_TAG: test
+        DOCKER_PULL: never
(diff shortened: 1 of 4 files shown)


tromai (approved):

behnazh-w (commented):
Can you please add a note to the [installation](https://github.com/oracle/macaron/blob/438-run-macaron-doesnt-pull-latest-image/docs/source/pages/installation.rst) documentation as well?

behnazh-w (approved):

PR #450, opened 2023-08-30
chore: update SECURITY.md

spavlusieva:


behnazh-w:
@spavlusieva The check is failing because of an issue that was fixed earlier today. Please rebase your branch on `staging` and it should be resolved.

behnazh-w:
The change in this PR is already merged via #456 so that it appears in the upcoming release.

PR #454, opened 2023-08-31
refactor!: remove --config-path CLI option

### LINKED ISSUE #417: Discussion on how to handle --config-path option of the analyze command after pull-request 401
As part of https://github.com/oracle/macaron/pull/401, the `--config-path` CLI option of Macaron has been agreed that it is not important for Macaron _at the moment_ because of the following reasons:
- In our Sphinx documentation, we don't have any examples for using this `--config-path` option.
- The content of this config file is a duplication of others CLI options we have to specify the main component (e.g. `repo-path` or `branch` or `digest`). The adding of `-purl` in https://github.com/oracle/macaron/pull/401, we also decide that's not worth adding `-purl` option to the config file because it would lead to potential conflicting values between the input yaml config file and other options.

Therefore, it's decided to find a way to potentially remove the involvement of `--config-path` with the existing design of Macaron. There are multiple options, each with its own pros/cons. This ticket is created for the discussion on what is the best option to proceed.

1. Remove the `--config-path` option completely. Because this feature is not likely to be known and used by the users of Macaron (as far as we know), we don't need to go through a "Deprecated" process for it.
- Pros: remove any entangle with the current development. This make sure that this feature won't be in our way. It's reasonable to do this as we are still < v1.0 so it's possible for these changes.
- Cons: what if there is need for this feature in the future, especially that the `--config-path` has support for pinning the dependencies version of the main software component that Macaron analyzes.

2. Leave the `--config-path` as it is and ignore it as we move on with the development (**need to make sure that its existence won't be conflicted with any existing or future CLI options**).
- Pros: We could still keep this feature for future usage
- Cons: It could get in our way when we change the interface of Macaron.

3. Move the `--config-path` to a separated option/command as a "light-weight" SBOM option where the user could define the specific dependencies by themselves.
- However, I think that this could still require a lot of effort to move and maintain it without bringing any major benefits (as far as it is now).

When making any changes to this option, these are the aspects that we need to care about:
- The integration tests stills uses this options
- Remove the documentation for [this option in our Sphinx documentation](https://oracle.github.io/macaron/pages/cli_usage/action_analyze.html#cmdoption-c)

Please feel free to give your opinions and suggestions about this.

tromai:
Closes #417 .


scripts/dev_scripts/integration_tests.sh
```diff
@@ -123,109 +118,45 @@ $RUN_MACARON analyze -rp https://github.com/apache/maven -b master -d 6767f2500f
...
 
 echo -e "\n----------------------------------------------------------------------------------"
-echo "micronaut-projects/micronaut-core: Check the e2e output JSON file with config and no dependency analyzing."
+echo "micronaut-projects/micronaut-core: Check the e2e output JSON file with no dependency analyzing."
 echo -e "----------------------------------------------------------------------------------\n"
-JSON_RESULT_DIR=$WORKSPACE/output/reports/github_com/micronaut-projects/micronaut-core/
-JSON_EXPECT_DIR=$WORKSPACE/tests/e2e/expected_results/micronaut-core
-
-declare -a COMPARE_FILES=(
-    "micronaut-core.json"
```
behnazh-w, line 156: Instead of removing this test case, how about using a custom SBOM to analyze selected dependencies?

scripts/dev_scripts/integration_tests.sh
```diff
@@ -123,109 +118,45 @@ $RUN_MACARON analyze -rp https://github.com/apache/maven -b master -d 6767f2500f
...
+echo "apache/maven: Check the e2e output JSON file with no dependency analyzing."
 echo -e "----------------------------------------------------------------------------------\n"
-JSON_RESULT_DIR=$WORKSPACE/output/reports/github_com/apache/maven
-JSON_EXPECT_DIR=$WORKSPACE/tests/e2e/expected_results/maven
-
-declare -a COMPARE_FILES=(
-    "maven.json"
-    "guava.json"
-    "mockito.json"
-)
```
behnazh-w, line 192: Same comment as above.

tromai:
This PR will be closed because we will not proceed with removing the input yaml file completely. See https://github.com/oracle/macaron/issues/417#issuecomment-1707472105

PR #466, opened 2023-09-10
fix: encode PURL qualifiers as a normalized string

behnazh-w:
This PR sets ``encode=True`` to encode qualifiers of a PURL string as a normalized string while converting it to a dictionary because SQLite doesn't support ``dict`` type.

It also adds exception handling for deserializing a PURL string while initializing a Component instance.

### DIFF
--- a/src/macaron/database/table_definitions.py
+++ b/src/macaron/database/table_definitions.py
@@ -24,7 +24,7 @@ from sqlalchemy.orm import Mapped, mapped_column, relationship
 
 from macaron.database.database_manager import ORMBase
 from macaron.database.rfc3339_datetime import RFC3339DateTime
-from macaron.errors import CUEExpectationError, CUERuntimeError
+from macaron.errors import CUEExpectationError, CUERuntimeError, InvalidPURLError
 from macaron.slsa_analyzer.provenance.expectations.cue import cue_validator
 from macaron.slsa_analyzer.provenance.expectations.expectation import Expectation
 from macaron.slsa_analyzer.slsa_req import ReqName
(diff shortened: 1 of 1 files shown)


src/macaron/database/table_definitions.py
```diff
@@ -172,9 +172,21 @@ def __init__(self, purl: str, analysis: Analysis, repository: "Repository | None
...
-        purl_parts = PackageURL.from_string(purl)
-        purl_kwargs = purl_parts.to_dict()
+        try:
+            purl_parts = PackageURL.from_string(purl)
+        except ValueError as error:
+            raise InvalidPURLError(f"The package url {purl} is not valid.") from error
+
+        # We set ``encode=True`` to encode qualifiers as a normalized string because SQLite doesn't support ``dict`` type.
+        # TODO: Explore the ``dbm`` or ``shelve`` packages to support dict type, which are part of the Python standard library.
+        purl_kwargs = purl_parts.to_dict(encode=True)
```
tromai, line 188: Can you specify that the `qualifiers` here is the `qualifiers` component of the PackageURL? 
I wonder what is the behavior of SQLite if we don't set `encode=True` in this place? Would it raise warnings when storing the `Component` instances into the database or it raise an exception?

behnazh-w: >Can you specify that the qualifiers here is the qualifiers component of the PackageURL?

If you mean the documentation here is not enough, I would disagree. ` purl_parts` refers to the parts of a PURL and I'm documenting why `encode=True` is passed as argument to `to_dict` function. There are various references to the PURL spec in the implementation and the reader should read them if not familiar.

In general it's not necessary to fully document a third-party concept when the references are provided.

behnazh-w: >I wonder what is the behavior of SQLite if we don't set encode=True in this place? Would it raise warnings when storing the Component instances into the database or it raise an exception?

Because the `qualifiers` will be of `dict` type, it throws an exception while storing the value.

tromai: > In general it's not necessary to fully document a third-party concept when the references are provided

I didn't mean that we should document third-party concept. I think it's hard for us (developers) to understand where is the `qualifiers` coming from. Is the `qualifiers` something related to the PURL or something else. In additions, when I first read this line, I don't know right away in which concrete case the `qualifiers` will be displayed as a dictionary if `encode=False`. Perhaps we could adjust it slightly to something like this:
```
We set ``encode=True`` so that the ``purl_parts``'s qualifiers will returned as a normalized string instead of a dictionary. This is because these values are used to initialized the instances of `Component`, which will be stored into the database , and SQLite doesn't support ``dict`` type. See https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst for more information on the ``qualifiers`` component of a PURL.
```

behnazh-w: You shouldn't see this change in isolation. By the time a developer reaches this line of code, they have already been introduced to PURL via an earlier reference to https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst and should know about `qualifiers`. I don't think it's needed to keep repeating this reference every time we do something with a PURL string.

tromai (approved):

PR #467, opened 2023-09-11
fix: check if repository is available in provenance available check

behnazh-w:
If a repository is not available for an artifact/analysis target identified by a PURL string, the  `mcn_provenance_available_1` check throws an exception. 

This PR fixes this bug by checking if the repository is available before running the check.


nathanwn (approved):
As we have discussed, I wonder if this issue can be prevented with static type checking. We should revisit this in the future to see if there is a solution that works well with both SQLAlchemy and mypy.

behnazh-w:
> As we have discussed, I wonder if this issue can be prevented with static type checking. We should revisit this in the future to see if there is a solution that works well with both SQLAlchemy and mypy.

Yes, it would be nice to enforce type checking on repositories, which can be `None` in our data model. But that requires some investigation.

PR #471, opened 2023-09-12
Add artifact match pipeline check

behnazh-w:
This PR adds a new check, `mcn_infer_artifact_pipeline_1` to detect a potential pipeline from which an artifact is published.

When a verifiable provenance is found for an artifact, the result of this check can be discarded. Otherwise, we check whether a CI workflow run has automatically published the artifact.

We use several heuristics in this check:
* The workflow run should have started before the artifact is published.
* The workflow step that calls a deploy command should have run successfully.
* The workflow step that calls a deploy command should have started before the artifact is published.

Note: due to a limitation, we cannot specify the provenance checks as parents of this check because a check cannot have more than one parent in the current design. It would be good to skip this with a success result if the relevant provenance checks pass in the future.


src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
```diff
@@ -0,0 +1,172 @@
...
+        ----------
+        group_id : str
+            The group id of the artifact.
+        artifact_id: str
+            The artifact id of the artifact.
+
+        Returns
+        -------
+        datetime
+            The artifact publish timestamp as a timezone-aware datetime object.
```
tromai, line 123: I don't quite understand the phase `timezone-aware`. Does it means that the returned `datetime` object will be different depending on the timezone of the local machine? 
I think because we always return `datetime.fromtimestamp(timestamp / 1000, tz=timezone.utc)` which mean that the timestamp is always converted to UTC timezone, it's not going to change depending on the timezone of the local machine?

src/macaron/slsa_analyzer/git_service/api_client.py
```diff
@@ -264,6 +264,79 @@ def get_workflow_runs(self, full_name: str, branch_name: str, created_after: str
...
+
+        Returns
+        -------
+        dict
+            The json query result or an empty dict if failed.
+
+        Examples
+        --------
+        The following call to this method will perform a query to
+        ``https://api.github/com/repos/{full_name}/
```
tromai, line 288:
```suggestion
        ``https://api.github.com/repos/{full_name}/
```

src/macaron/slsa_analyzer/git_service/api_client.py
```diff
@@ -264,6 +264,79 @@ def get_workflow_runs(self, full_name: str, branch_name: str, created_after: str
...
+            The json query result or an empty dict if failed.
+
+        Examples
+        --------
+        The following call to this method will perform a query to
+        ``https://api.github/com/repos/owner/repo/actions/runs?created=2022-11-05T20:38:40..2022-11-05T20:38:58``
+
+        >>> gh_client.get_workflow_run_for_date_time_range(
+            full_name="owner/repo",
+            created=datetime-range
```
tromai, line 328:
```suggestion
            created="2022-11-05T20:38:40..2022-11-05T20:38:58"
```

src/macaron/slsa_analyzer/ci_service/github_actions.py
```diff
@@ -410,6 +412,116 @@ def has_latest_run_passed(
...
+        logger.debug(
+            "Getting the latest workflow run of %s at %s within time range %s",
+            workflow,
+            str(date_time),
+            str(time_range),
+        )
+
+        html_urls: set[str] = set()
+        try:
+            datetime_from = date_time - timedelta(seconds=time_range)
```
tromai, line 458: Do we need to make sure that `date_time` here is in the correct timezone (UTC) ?

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
```diff
@@ -0,0 +1,172 @@
...
+        """
+        query_params = [f"q=g:{group_id}", f"a:{artifact_id}"]
+        if version:
+            query_params.append(f"v:{version}")
+        url = urlunsplit(
+            SplitResult(
+                scheme="https",
+                netloc=self.hostname,
+                path=f"/{self.search_endpoint}",
+                query="&".join(["+AND+".join(query_params), "core=gav", "rows=1", "wt=json"]),
```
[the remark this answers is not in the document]
behnazh-w: The endpoint doesn't seem to accept the encodings: https://search.maven.org/solrsearch/select?q=g%3Acom.google.guava%2BAND%2Ba%3Aguava%2BAND%2Bv%3A32.1.2-jre&core=gav&rows=1&wt=json

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
```diff
@@ -0,0 +1,172 @@
...
+                )
+
+            # We only consider the first ``docs`` element.
+            timestamp = res_obj.get("response").get("docs")[0].get("timestamp")
+            if not timestamp:
+                raise InvalidHTTPResponseError(f"The timestamp is missing in the response returned by {url}.")
+
+            logger.debug("Found timestamp: %s.", timestamp)
+
+            # The timestamp published in Maven Central is in milliseconds and needs to be divided by 1000.
```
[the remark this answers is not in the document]
behnazh-w: Unfortunately, I couldn't find any documentation about that :-(

src/macaron/slsa_analyzer/git_service/api_client.py
```diff
@@ -264,6 +264,79 @@ def get_workflow_runs(self, full_name: str, branch_name: str, created_after: str
...
+        )
+        """
+        logger.debug("Query GitHub to get run jobs for %s with run ID %s", full_name, run_id)
+
+        url = f"{GhAPIClient._REPO_END_POINT}/{full_name}/actions/runs/{run_id}/jobs"
+        response_data = send_get_http(url, self.headers)
+
+        return response_data
+
+    def get_workflow_run_for_date_time_range(self, full_name: str, datetime_range: str) -> dict:
```
tromai, line 303: Instead of providing the `datetime_range` as a string, should we provide the range as `datetime` objects (which must have the same timezone of UTC) and then these `datetime` objects will be converted into the string format for the f query?

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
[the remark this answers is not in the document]
tromai: Hmmm that's quite unfortunate. I can also see that the documentation for the Search API of Central Repository ([here](https://central.sonatype.org/search/rest-api-guide/) quite limited.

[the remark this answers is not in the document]
tromai: We could add a small comment saying that we observed and came up with this logic. But it's not officially documented anywhere. But it's also fine as is.

src/macaron/slsa_analyzer/git_service/api_client.py
behnazh-w: I'm not sure what you mean. GitHub API accepts specific query formats, such as : `<start-datetime>..<end-datetime>` or simply  `<datetime>`. It is the responsibility of the caller of this function i.e., `ci_service.workflow_run_in_date_time_range()` to construct the final query value.

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
behnazh-w: The documentation here might be helpful to understand  “aware” and “naive” date time objects: https://docs.python.org/3/library/datetime.html#aware-and-naive-objects

src/macaron/slsa_analyzer/ci_service/github_actions.py
behnazh-w: It is the responsibility of the caller to provide the correct timezone.

src/macaron/slsa_analyzer/git_service/api_client.py
behnazh-w: See commit [7199992](https://github.com/oracle/macaron/pull/471/commits/71999921ecebde4ca7448ed69f704b746cf5729c)

src/macaron/slsa_analyzer/git_service/api_client.py
behnazh-w: See commit [7199992](https://github.com/oracle/macaron/pull/471/commits/71999921ecebde4ca7448ed69f704b746cf5729c)

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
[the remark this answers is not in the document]
behnazh-w: See commit [7199992](https://github.com/oracle/macaron/pull/471/commits/71999921ecebde4ca7448ed69f704b746cf5729c)

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
```diff
@@ -0,0 +1,172 @@
...
+
+        If version is not provided, the timestamp of the latest version will be returned.
+
+        Parameters
+        ----------
+        group_id : str
+            The group id of the artifact.
+        artifact_id: str
+            The artifact id of the artifact.
+
```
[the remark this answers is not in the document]
behnazh-w: See commit [7199992](https://github.com/oracle/macaron/pull/471/commits/71999921ecebde4ca7448ed69f704b746cf5729c)

src/macaron/slsa_analyzer/ci_service/github_actions.py
```diff
@@ -410,6 +412,116 @@ def has_latest_run_passed(
...
+            return html_urls
+
+        if not run_data:
+            logger.debug("Unable to find any run data for the workflow %s", workflow)
+            return html_urls
+
+        logger.debug("Checking workflow run of %s.", workflow)
+        try:
+            # iterate through the responses in reversed order to add the run closest to the date_time timestamp first.
+            for item in reversed(run_data["workflow_runs"]):
```
tromai, line 480: From my understanding, the list of workflow runs when returned by the REST API will be in the order of **decreasing** created date. This will mean that the latest workflow run will be at the first position in `run_data["workflow_runs"]`, etc. 
However, from the comment here
```
# iterate through the responses in reversed order to add the run closest to the date_time timestamp first.
```
If we are checking them in the reversed order, the workflow runs closest to `date_time-time_range` will be encountered first. I think there is an inconsistent between the comment and the implementation.

docs/source/pages/supported_technologies/index.rst
```diff
@@ -41,6 +41,9 @@ Package Registries
    * - `JFrog Artifactory <https://jfrog.com/artifactory>`_
      - Only projects built with Gradle and publishing to a JFrog Artifactory repo following `Maven layout <https://maven.apache.org/repository/layout.html>`_
      - :doc:`page </pages/supported_technologies/jfrog>`
+   * - `Maven Central Artifactory <https://central.sonatype.com>`_
+     - This registry is searched for artifacts built with Gradle or Maven.
```
tromai, line 45: I think putting it this way might make the reader misunderstand that the Maven Central Artifactory only supports artifacts built with Gradle and Maven.
Perhaps we could say:
```
Only projects built with Gradle or Maven and published to the Maven Central Artifactory.
```

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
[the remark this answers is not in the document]
tromai: Hmm. I don't understand the reason why it behaves this way. Perhaps we could revisit it again later.

src/macaron/slsa_analyzer/ci_service/github_actions.py
behnazh-w: >If we are checking them in the reversed order, the workflow runs closest to date_time-time_range will be encountered first. 

This workflow has happened earlier and appears as the last element in the `run_data["workflow_runs"]`. We want to prioritize it and show it as the first candidate in the justification, so we need to iterate in a reversed order.

docs/source/pages/supported_technologies/index.rst
behnazh-w: See commit [1aabb11](https://github.com/oracle/macaron/pull/471/commits/1aabb111ad2d55425a391d884238aaa9d4f2a8a3)

src/macaron/slsa_analyzer/ci_service/github_actions.py
tromai: If we are prioritizing the earliest workflow run in the range of `date_time - time_range` to `date_time` (given that `date_time - time_range` is earlier than `date_time`) then should the comment here be:
```
# iterate through the responses in reversed order to add the run closest to the ``date_time - time_range`` timestamp first.
```

tromai (approved):

src/macaron/slsa_analyzer/ci_service/github_actions.py
behnazh-w: See commit [a656eb7](https://github.com/oracle/macaron/pull/471/commits/a656eb719141173d47f473ea1fc1607cab610525)

PR #480, opened 2023-09-21
test: improve expectation regex value to better match usage in urllib3

### LINKED ISSUE #479: Urllib3 provenance expectation problem
The expectation file urllib3_PASS.cue can fail unexpectedly due to the urllib3 repository sometimes using version branches/tags that start with a `v`. E.g. `v2.0.5`. This can be solved by adding an optional token to the file's uri regular expression string.

benmss:
Closes #479 
Regex value in expectation file is adjusted slightly so as to account for the urllib3 version branch name / tag sometimes starting with a `v`.


behnazh-w:
The scope of the PR and commit title can change to `ci:` or `test:` because it is not necessary to communicate this fix in the release changelog.

behnazh-w (approved):

PR #481, opened 2023-09-21
test: adjust flaky test which relies on filesystem permissions

nathanwn:
I am removing a test that fails in environments where the user running the test actually has write access to the `/` directory.
This may happen in, for example, a test environment inside a container where the user is `root`, which is quite common in CI systems (e.g. [GitLab Runner with docker executor](https://docs.gitlab.com/runner/executors/docker.html), or [Jenkins docker pipelines](https://www.jenkins.io/doc/book/pipeline/docker/)).

### DIFF
--- a/tests/config/test_defaults.py
+++ b/tests/config/test_defaults.py
@@ -31,6 +31,15 @@ def test_create_defaults() -> None:
     assert create_defaults(output_dir, global_config.macaron_path) is True
 
 
+@pytest.mark.xfail(
+    os.geteuid() == 0,
+    reason="Only effective for non-root users",
+)
+def test_create_defaults_without_permission() -> None:
+    """Test dumping default config in cases where the user does not have write permission to the output location."""
+    assert create_defaults(output_path="/", cwd_path="/") is False
+
+
 @pytest.mark.parametrize(
     ("section", "item", "delimiter", "strip", "duplicated_ok", "expect"),
     [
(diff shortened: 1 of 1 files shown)


tromai (approved):

tests/config/test_defaults.py
```diff
@@ -29,7 +29,6 @@ def test_create_defaults() -> None:
     """Test dumping the default values."""
     output_dir = os.path.dirname(os.path.abspath(__file__))
     assert create_defaults(output_dir, global_config.macaron_path) is True
-    assert create_defaults("/", "/") is False
```
behnazh-w, line 32: This particular test has been helpful in finding bugs related to permission issues in the past. Instead of removing it, please replace it by a testcase that raises `OSError` [here](https://github.com/oracle/macaron/blob/0f426982837708539fb64b76001b823b79596b0e/src/macaron/config/defaults.py#L166).

nathanwn: Do you have any suggestions for improving this test?
One solution I can think of now is to detect if the current user running tests is `root`, and disable the test if it is the case.

nathanwn: I fixed the test in 268839891f7c4719bbe1339038a4652d90ee38b2.
I set it to expect to fail in case the user is `root`.

behnazh-w: The adjustment to the test looks good.

behnazh-w (approved):

PR #483, opened 2023-09-22
feat: add download timeout config

### LINKED ISSUE #392: Add download timeout setting to ini config
Currently, we do not have a separate setting for download timeout in the ini config - the download timeout is the same as the request timeout.

```ini
[requests]
# The default timeout in seconds for 'requests' API calls.
timeout = 10
```

Although this default 10-second timeout for requests is sensible, it is most likely too low for downloads and has caused issues when downloading artifacts of the size ~ hundreds of MB.

We need a separate setting for download timeout.

nathanwn:
Closes #392.

I am adding this because I encountered this issue again today in [this job](https://github.com/oracle/macaron/actions/runs/6269369980/job/17025702477#step:8:1282).

I am not sure if the PR type should be `feat` though. Let me know what you think.

### DIFF
--- a/src/macaron/config/defaults.ini
+++ b/src/macaron/config/defaults.ini
@@ -13,6 +13,10 @@ timeout = 5
 # The default timeout in seconds for 'requests' API calls.
 timeout = 10
 
+[downloads]
+# The default timeout in seconds for downloading assets.
+timeout = 120
+
 # This is the database to store Macaron's results.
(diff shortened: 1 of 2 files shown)


tromai:
It's okay for me too to not use `feat` here.

tromai (approved):

behnazh-w (approved):

PR #491, opened 2023-09-28
fix: do not use git set-branches if the target branch is not currently available in the repository

### LINKED ISSUE #486: Repository checkout fails in some cases
If we run the `analyze` command on a remote path at a specific branch that does not exist, and then run it again just using the remote path with no branch and commit provided as input (which should fall back to the main branch and latest commit), checkout fails.

To reproduce this bug run:
```shell
macaron analyze -rp https://github.com/behnazh-w/example-maven-app -b master
```

Then run
```shell
macaron analyze -rp https://github.com/behnazh-w/example-maven-app
```

tromai:
Closes #486 

## Bug description

This bug happened because of the way branch check out is handling in this [method](https://github.com/oracle/macaron/blob/71accbf81e849947ca238ebfa3a9a6ef88fd01f6/src/macaron/slsa_analyzer/git_url.py#L65).

At the moment, the process Macaron in which Macaron tries to checkout the correct branch and commit for the analysis is as follows:
- The branch to checkout is either the value provided by the user (via `--branch` in the CLI) or the default branch of the repository ([implementation](https://github.com/oracle/macaron/blob/71accbf81e849947ca238ebfa3a9a6ef88fd01f6/src/macaron/slsa_analyzer/git_url.py#L251)):
  - We are using `git rev-parse --abbrev-ref origin/HEAD` to obtain the value `origin/<default_branch_name>` ([references](https://stackoverflow.com/questions/28666357/git-how-to-get-default-branch)).

- Macaron try to check out that branch (implementation [here](https://github.com/oracle/macaron/blob/71accbf81e849947ca238ebfa3a9a6ef88fd01f6/src/macaron/slsa_analyzer/git_url.py#L109-L141))..
  - The way we check if a given branch exists in the repository:
    - `res_branch in [os.path.basename(ref.name) for ref in org_remote.refs]`
    - The `refs` in `org_remote` instance would return `git.RemoteReference` instances (according to gitpython [documentation](https://gitpython.readthedocs.io/en/stable/reference.html?highlight=Reference#git.remote.Remote.refs)).
    - The documentation for git remote references can be seen [here](https://git-scm.com/book/en/v2/Git-Internals-Git-References).
    - Therefore, each `git.RemoteReference` can be considered a reference to the latest commit of a remote branch (**from the last time we communicate with remote**). Therefore, `org_remote.refs` will return an iterator of remote references that this local repository currently have (which mean that it could be missing any new remote branches, etc.).
    - I have manually check in a shallow clone, a call to `org_remote.refs` does not return all available remote branches. It only contains: 1. `refs/remotes/origin/<branch_name>` which point to the branch which we shallow clone (if no branch is provided e.g. `git clone --depth=1 <url>`), this will be `refs/remotes/origin/<default_branch>`. 2. if it's a shallow clone of the default branch, there will be also a `refs/remotes/origin/HEAD`.

- If Macaron cannot find the target branch from `org_remote.refs`, Macaron will try to setup a local branch that track that branch from origin remote (implementation [here](https://github.com/oracle/macaron/blob/71accbf81e849947ca238ebfa3a9a6ef88fd01f6/src/macaron/slsa_analyzer/git_url.py#L133-L138)). 
  - The full command is: `git remote set-branches --add origin '<target_branch>'` ([references](https://stackoverflow.com/questions/23708231/git-shallow-clone-clone-depth-misses-remote-branches.)) What this command do is that it setup the branch list to track a branch `<target_branch>` from the remote. The `--add` will make sure that we only add the new tracked branch into the list of tracked branches. The next time we fetch, git will also try to fetch this branch from the remote repository. The reason I did it this way is to make sure that *we only need to fetch the branch that we are interested in when working with shallow clones*.

  - This is the command that causes the bug in https://github.com/oracle/macaron/issues/486:
    - This command this will update the `.git/config` file in the target repository with `<target_branch>`. Therefore, if the branch doesn't exist, subsequent fetching/pulling will always fail because git will try to ask the remote for this non-exist branch and return error everytime.
    - The second issue: there is a fetching operation happened [here](https://github.com/oracle/macaron/blob/71accbf81e849947ca238ebfa3a9a6ef88fd01f6/src/macaron/slsa_analyzer/git_url.py#L137) without checking if `offline_mode` is set or not.
    - The third issue is that we don't have a good way to revert the effect of set-branches (https://stackoverflow.com/questions/47726087/undo-set-branches-on-git-remote). Overall, it's not a good idea to perform git operations that alter `.git/config`.

## Solution.
To resolve this bug, I simplify the logic for checkout branch and commit for a target repository with the following goals:
- Remove any conditional logic for handling shallow-clone repositories:
  - The usage of `git remote set-branches --add origin '<target_branch>'`
  - Unshallow when trying to pull from latest (note [here](https://github.com/oracle/macaron/pull/491/files#r1340879300)).
- Make sure that the only online operations included in the `check_out_repo_target` method are:
  - `git fetch`: happen before checkout branch.
  - `git pull --ff-only` before checkout a specific commit.

If the checkout operation fails for either branch or commit, we return errors.

### DIFF
--- a/src/macaron/slsa_analyzer/git_url.py
+++ b/src/macaron/slsa_analyzer/git_url.py
@@ -109,7 +109,7 @@ def check_out_repo_target(git_obj: Git, branch_name: str = "", digest: str = "",
 
     if not offline_mode:
         # Fetch from remote by running ``git fetch`` inside the target repository.
-        # We don't specify any remote name (e.g. origin) is because we want git to resolve the default fetching
+        # We don't specify any remote name (e.g. origin) because we want git to resolve the default fetching
         # target by itself.
         # For example, the user runs Macaron on a local repository where the remote is set to have name "foo_origin"
         # instead.
@@ -117,7 +117,7 @@ def check_out_repo_target(git_obj: Git, branch_name: str = "", digest: str = "",
         try:
             git_obj.repo.git.fetch()
         except GitCommandError as error:
-            logger.error("Cannot perform fetching on this repository. Error: %s", error)
+            logger.error("Unable to fetch from the remote repository. Error: %s", error)
             return False
 
     try:
@@ -183,7 +183,7 @@ def get_default_branch(git_obj: Git) -> str:
     branch of the remote repository and it's usually set when the repository is first cloned with
     ``git clone <url>``.
     Therefore, this method will fail to obtain the default branch name if ``origin/HEAD`` is not
-    available. An example of this case is when a repository is shallow-cloned with a non-default the branch
+    available. An example of this case is when a repository is shallow-cloned from a non-default branch
     (e.g. ``git clone --depth=1 <url> -b some_branch``).
 
     Parameters
--- a/src/macaron/slsa_analyzer/git_url.py
+++ b/src/macaron/slsa_analyzer/git_url.py
@@ -13,7 +13,6 @@ from configparser import ConfigParser
 
 from git import GitCommandError
 from git.objects import Commit
-from git.remote import Remote
 from git.repo import Repo
 from pydriller.git import Git
 
@@ -74,10 +73,12 @@ def check_out_repo_target(git_obj: Git, branch_name: str = "", digest: str = "",
     If ``digest`` is provided, this method will checkout that specific commit. If ``digest``
     cannot be found in the current branch, this method will pull (fast-forward only) from remote.
 
-    This method supports repositories which are cloned (full or shallow) from existing remote repositories.
+    This method supports repositories which are cloned from existing remote repositories.
     Other scenarios are not covered (e.g. a newly initiated repository).
 
-    If ``offline_mode`` is set, this method will not pull from remote while checking out the branch or commit.
+    This method will not try to un-shallow an existing shallow-clone repository.
+
+    If ``offline_mode`` is set, this method will not pull/fetch from remote while checking out the branch or commit.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


src/macaron/slsa_analyzer/git_url.py
```diff
@@ -191,67 +175,16 @@ def check_out_repo_target(git_obj: Git, branch_name: str = "", digest: str = "",
     return True
 
 
-def commit_exists(git_obj: Git, digest: str) -> bool:
```
tromai, line 194: The reason I removed this method because it's just a try-except of the `is_ancestor` method. The try-except block is not needed for this command because this method doesn't raise `GitCommandError` (see implementation of that method [here](https://github.com/gitpython-developers/GitPython/blob/7d4f6c68c657586f27125f25b15f02c5a3d7f35c/git/repo/base.py#L699C4-L712C20)). In additions, because this method returns a bool value already, we don't need to cast it.

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -191,67 +175,16 @@ def check_out_repo_target(git_obj: Git, branch_name: str = "", digest: str = "",
...
-    try:
-        return bool(git_obj.repo.is_ancestor(digest, "HEAD"))
-    except GitCommandError as error:
-        # The exception could be raised because the digest does not exist in the history of the branch
-        # or the digest is not a valid digest.
-        logger.debug(error)
-        return False
-
-
-def pull_latest_changes(git_obj: Git) -> bool:
```
tromai, line 217: This method is doing 2 things:
- If the target repository is a shallow clone repository, it will try to unshallow it. 
- Pull the latest changes by running `git pull -f --ff-only`.

There are a few issues I can see with this method:
- The `-f` flag is not needed (references of this flag [here](https://git-scm.com/docs/git-pull#Documentation/git-pull.txt--f)).
- The first task (unshallow a repository) was historically put there because we want Macaron support running against shallow clone repositories. However, I think that we shouldn't do this anymore:
  - If the user provide a shallow repository for Macaron to analyze as a local repo, Macaron will not try to fetch/pull any latest changes in the `check_out_repo_target`.
  - Macaron at the moment **always fully clones** repositories in `<output>/git_repos/...`. The only case where a shallow repository exists `<output>/git_repos/...` is when the user perform a shallow clone of a repository inside that directory (which they shouldn't do unless they know what they are doing). Even the user tries to do that, Macaron would still work (running `git fetch` on a shallow repository won't raise any error except not being able to fetch the whole history, etc.).

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -191,67 +175,16 @@ def check_out_repo_target(git_obj: Git, branch_name: str = "", digest: str = "",
...
-
-    This function does not perform any online operation.
+    """Return the default branch name of the target repository.
+
+    This function does not perform any online operation. It depends on the existence of the remote
+    reference ``origin/HEAD`` in the git repository. This remote reference will point to the default
+    branch of the remote repository and it's usually set when the repository is first cloned with
+    ``git clone <url>``.
+    Therefore, this method will fail to obtain the default branch name if ``origin/HEAD`` is not
+    available. An example of this case is when a repository is shallow-cloned with a non-default the branch
```
behnazh-w, line 186:
```suggestion
    available. An example of this case is when a repository is shallow-cloned from a non-default branch
```

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -106,76 +107,59 @@ def check_out_repo_target(git_obj: Git, branch_name: str = "", digest: str = "",
...
-    #   - no digest is provided.
-    #   - or a commit digest is provided but it does not exist in the current local branch.
-    if not offline_mode and (not digest or not commit_exists(git_obj, digest)):
-        logger.info("Pulling the latest changes of branch %s fast-forward only.", res_branch)
-        if not pull_latest_changes(git_obj):
-            logger.error(
-                "Cannot pull the latest changes for branch %s.",
-                res_branch,
-            )
+            logger.error("Cannot perform fetching on this repository. Error: %s", error)
```
behnazh-w, line 120:
```suggestion
           logger.error("Unable to fetch from the remote repository. Error: %s", error)
```

src/macaron/slsa_analyzer/git_url.py
behnazh-w: How would removing `git_obj.repo.git.fetch("--unshallow")` affect our new upcoming feature that supports shallow clones when an artifact is mapped to a commit?

src/macaron/slsa_analyzer/git_url.py
tromai: Fixed in https://github.com/oracle/macaron/pull/491/commits/7b35e849e44cc4bdfd4b23499db6c16430b55270

src/macaron/slsa_analyzer/git_url.py
tromai: Fixed in https://github.com/oracle/macaron/pull/491/commits/7b35e849e44cc4bdfd4b23499db6c16430b55270

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -106,76 +107,59 @@ def check_out_repo_target(git_obj: Git, branch_name: str = "", digest: str = "",
...
-            "Find branch %s in the remote's refs of the local repository. The branch is checked out directly.",
-            res_branch,
-        )
-        git_obj.repo.git.checkout(res_branch)
-    else:
-        # This might happen because the git repository is a shallow clone.
-        logger.info("Cannot find branch %s. Try to fetch the branch from remote.", res_branch)
+    if not offline_mode:
+        # Fetch from remote by running ``git fetch`` inside the target repository.
+        # We don't specify any remote name (e.g. origin) is because we want git to resolve the default fetching
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle/macaron/pull/491/commits/7b35e849e44cc4bdfd4b23499db6c16430b55270

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -106,76 +107,59 @@ def check_out_repo_target(git_obj: Git, branch_name: str = "", digest: str = "",
...
+        # (which point to the tip of the branch).
+        # If the commit we want to analyze is same as HEAD, that commit is still considered as the ancestor of HEAD.
+        # The ``is_ancestor`` method runs ``git merge-base`` behind the scence.
+        # For more information on computing the ancestor status of two commits: https://git-scm.com/docs/git-merge-base.
+        if not digest or not git_obj.repo.is_ancestor(digest, "HEAD"):
+            logger.info("Pulling the latest changes of branch %s fast-forward only.", res_branch)
             try:
-                git_obj.checkout(digest)
+                # Pull the latest changes on the current branch fast-forward only.
+                git_obj.repo.git.pull("--ff-only")
```
tromai, line 148: Thinking more about it. We don't need to pull before checking out a commit. If there is a newer commit from remote that has been fetched to the local repository, we could checkout that commit directly without pulling. 
In theory, this could save us from downloading the uneccessary trees or blobs of commits that we don't care about. However, I don't have enough evidence to validate it yet. I will continue working it.

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -74,10 +73,12 @@ def check_out_repo_target(git_obj: Git, branch_name: str = "", digest: str = "",
     If ``digest`` is provided, this method will checkout that specific commit. If ``digest``
     cannot be found in the current branch, this method will pull (fast-forward only) from remote.
 
-    This method supports repositories which are cloned (full or shallow) from existing remote repositories.
+    This method supports repositories which are cloned from existing remote repositories.
     Other scenarios are not covered (e.g. a newly initiated repository).
 
-    If ``offline_mode`` is set, this method will not pull from remote while checking out the branch or commit.
+    This method will not try to un-shallow an existing shallow-clone repository.
```
behnazh-w, line 79: Not sure why you want to convey here?

src/macaron/slsa_analyzer/git_url.py
behnazh-w: Is there any update for my question?

src/macaron/slsa_analyzer/git_url.py
tromai: Hm right, because I haven't decided to remove support for un-shallow repositories, I probably shouldn't mention it too. I will remove this line as it's vague and confusing.

src/macaron/slsa_analyzer/git_url.py
behnazh-w: As discussed offline, removing this should not affect `treeless` cloning, which will be added in this [PR](https://github.com/oracle/macaron/pull/389).

src/macaron/slsa_analyzer/git_url.py
tromai: Removed in https://github.com/oracle/macaron/pull/491/commits/54ed439f235d5e724550aad4e6c9670b18d2aad7

behnazh-w (approved):

PR #504, opened 2023-10-05
feat: support gzipped provenance files

nicallen:


tromai (approved):

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -33,8 +35,12 @@ def load_provenance_file(filepath: str) -> dict[str, JsonType]:
         If there is an error loading the provenance JSON payload.
     """
     try:
-        with open(filepath, encoding="utf-8") as file:
-            provenance = json.load(file)
+        try:
+            with gzip.open(filepath, mode="rt", encoding="utf-8") as file:
+                provenance = json.load(file)
```
behnazh-w, line 40: If this line raises an `OSError`, we try to open the file normally after catching `gzip.BadGzipFile`, which inherits `OSError`. The second read could potentially raise an  `OSError` again but we won't catch it. We can catch `OSError` at this [line](https://github.com/oracle/macaron/pull/504/files#diff-83d632a57b7fff707ceba30cfe00bb82d195071cf8e93172c1fcd90c372d1410L38), which seems to be missing anyway.


Also, from the [documentation](https://docs.python.org/3/library/gzip.html#gzip.BadGzipFile), [zlib.error](https://docs.python.org/3/library/zlib.html#zlib.error) can also be raised for invalid gzip files, if I read it correctly? I guess we should catch [zlib.error](https://docs.python.org/3/library/zlib.html#zlib.error) as well to be on the safe side?

nicallen: I've changed it to catch OSError (from either the gzip.open or the regular open) and raise LoadIntotoAttestationError in that case.

I've also changed it to catch EOFError and zlib.error when reading the gzip file and fall back to trying to read the file as json in that case. The alternative would be to raise LoadIntotoAttestationError in those cases, by presuming that these are only thrown for corrupt gzip files that couldn't possible be valid json, but the documentation isn't clear exactly in what circumstances each may be thrown so it is probably safer the way I have done it (tradeoff is between potentially rejecting valid non-gzipped json files vs potentially silently suppressing corrupt-file errors).

PR #508, opened 2023-10-10
feat: map artifacts to commits via repo tags

benmss:
This feature allows artifacts with a specified version to be mapped to the repository commit they were created from.
The mapping is achieved through use of repository tags, where these tags match the specified version sufficiently.
Tag to version comparisons are performed using regular expression.

Tags are expected to be of the format (with allowances for extra characters):
`[<artifact_name>-]<version>`

`artifact_name` is an optional prefix that matches the name of the related artifact, e.g. "commons-io".

`<version>` is a string of some number of alphanumeric characters that can be (multi-) separated by any number of non-numeric characters. 

Examples of accepted tag versions:

- `v1.2.3`
- `1_2_3`
- `1.2.3-DEV`
- `1.2.3.RELEASE`
- `r1rv23`
- `1.2.3.v123123`

Further changes to this PR are in progress:

- [x] Support tags whose suffix does not match their related version
- [x] Do not fallback to latest commit if the commit finder fails
- [x] Use default branch for commits with no associated branch (for now)
- [x] Use purls with versions instead of commits in integration tests
- [x] Allow for purls that represent repositories not just artefacts
- [x] Extend test file to include purls and comments
- [x] Use Hypothesis to test commit finder input

### DIFF
--- a/src/macaron/config/defaults.ini
+++ b/src/macaron/config/defaults.ini
@@ -51,6 +51,7 @@ recursive = False
 [repofinder]
 find_repos = True
 use_open_source_insights = True
+find_commits = True
 
 [repofinder.java]
 # The list of maven-like repositories to attempt to retrieve artifact POMs from.
--- a/src/macaron/config/defaults.ini
+++ b/src/macaron/config/defaults.ini
@@ -51,7 +51,6 @@ recursive = False
 [repofinder]
 find_repos = True
 use_open_source_insights = True
-find_commits = True
 
 [repofinder.java]
 # The list of maven-like repositories to attempt to retrieve artifact POMs from.
--- a/src/macaron/repo_finder/commit_finder.py
+++ b/src/macaron/repo_finder/commit_finder.py
@@ -196,21 +196,22 @@ def get_commit_from_version(git_obj: Git, name: str, version: str) -> tuple[str,
     logger.debug("Searching for commit of artifact version using tags: %s@%s", name, version)
 
     # Only consider tags that have a commit.
-    valid_tags = []
+    valid_tags = {}
     for tag in git_obj.repo.tags:
         commit = _get_tag_commit(tag)
         if not commit:
             logger.debug("No commit found for tag: %s", tag)
             continue
 
-        valid_tags.append(tag)
+        tag_name = str(tag)
+        valid_tags[tag_name] = tag
 
     if not valid_tags:
         logger.debug("No tags with commits found for %s", name)
         return "", ""
 
     # Match tags.
-    matched_tags = _match_tags(valid_tags, name, version)
+    matched_tags = match_tags(list(valid_tags.keys()), name, version)
 
     if not matched_tags:
         logger.debug("No tags matched for %s", name)
@@ -221,8 +222,11 @@ def get_commit_from_version(git_obj: Git, name: str, version: str) -> tuple[str,
         logger.debug("Best match: %s", matched_tags[0])
         logger.debug("Up to 5 others: %s", matched_tags[1:6])
 
-    tag = matched_tags[0]
-    tag_name = str(tag)
+    tag_name = matched_tags[0]
+    tag = valid_tags[tag_name]
+    if not tag:
+        # Tag names are taken from valid_tags and should always exist within it.
+        logger.debug("Missing tag name from tag dict: %s not in %s", tag_name, valid_tags.keys())
 
     branch_name = _get_branch_of_commit(git_obj.get_commit_from_tag(tag_name))
     if not branch_name:
@@ -240,7 +244,7 @@ def get_commit_from_version(git_obj: Git, name: str, version: str) -> tuple[str,
     return branch_name, tag.commit.hexsha
 
 
-def _build_version_pattern(version: str) -> tuple[Pattern, list[str], bool]:
+def _build_version_pattern(version: str) -> tuple[Pattern | None, list[str], bool]:
     """Build a version pattern to match the passed version string.
 
     Parameters
@@ -250,14 +254,19 @@ def _build_version_pattern(version: str) -> tuple[Pattern, list[str], bool]:
 
     Returns
     -------
-    tuple[Pattern, list[str], bool]
+    tuple[Pattern | None, list[str], bool]
         The tuple of the regex pattern that will match the version, the list of version parts that were extracted, and
         whether the version string has a non-numeric suffix.
 
     """
     # The version is split on non-alphanumeric characters to separate the version parts from the non-version parts.
     # e.g. 1.2.3-DEV -> [1, 2, 3, DEV]
-    split = split_pattern.split(version)
+    try:
+        split = split_pattern.split(version)
+    except TypeError as error:
+        logger.debug("Failed to perform regex split on version string %s -- %s", version, error)
+        return None, [], False
+
     logger.debug("Split version: %s", split)
     if not split:
         # If the version string contains no separators use it as is.
@@ -314,15 +323,20 @@ def _build_version_pattern(version: str) -> tuple[Pattern, list[str], bool]:
         f"^(?:(?:{PREFIX_WITH_SEPARATOR})|(?:{PREFIX}{PREFIX_SEPARATOR}))(?P<version>"
         f"{this_version_pattern}){SUFFIX_SEPARATOR}{SUFFIX}$"
     )
-    return re.compile(this_version_pattern, flags=re.IGNORECASE), parts, has_non_numeric_suffix
+    try:
+        return re.compile(this_version_pattern, flags=re.IGNORECASE), parts, has_non_numeric_suffix
+    except Exception as error:  # pylint: disable=broad-exception-caught
+        # The regex library uses an internal error that cannot be used here to satisfy pylint.
+        logger.debug("Error while compiling version regex: %s", error)
+        return None, [], False
 
 
-def _match_tags(tag_list: list[TagReference], artifact_name: str, artifact_version: str) -> list[TagReference]:
+def match_tags(tag_list: list[str], artifact_name: str, artifact_version: str) -> list[str]:
     """Return items of the passed tag list that match the passed artifact name and version.
 
     Parameters
     ----------
-    tag_list: list[TagReference]
+    tag_list: list[str]
         The list of tags to check.
     artifact_name: str
         The name of the artifact to match.
@@ -331,17 +345,18 @@ def _match_tags(tag_list: list[TagReference], artifact_name: str, artifact_versi
 
     Returns
     -------
-    list[TagReference]
+    list[str]
         The list of tags that matched the pattern.
     """
     # Create the pattern for the passed version.
     pattern, parts, has_non_numeric_suffix = _build_version_pattern(artifact_version)
+    if not pattern:
+        return []
 
     # Match the tags.
     matched_tags = []
     for tag in tag_list:
-        tag_name = str(tag)
-        match = pattern.match(tag_name)
+        match = pattern.match(tag)
         if not match:
             continue
         # Tags are append with their match information for possible further evaluation.
--- a/src/macaron/repo_finder/commit_finder.py
+++ b/src/macaron/repo_finder/commit_finder.py
@@ -14,7 +14,8 @@ logger: logging.Logger = logging.getLogger(__name__)
 
 ALPHANUMERIC = "0-9a-z"
 PREFIX = "(?:.*)"
-INFIX = "[^0-9]{1,3}"  # 1 to 3 non-numeric characters
+# Version token separator: 1 to 3 alphabetic characters OR 1 to 3 non-alphanumeric characters.
+INFIX = "([a-z]{1,3}|[^0-9a-z]{1,3})"
 split_pattern = re.compile(f"[^{ALPHANUMERIC}]+", flags=re.IGNORECASE)
 validation_pattern = re.compile(f"[{ALPHANUMERIC}]+", flags=re.IGNORECASE)
 
--- a/src/macaron/repo_finder/commit_finder.py
+++ b/src/macaron/repo_finder/commit_finder.py
@@ -7,8 +7,13 @@ import re
 from re import Pattern
 
 from git import TagReference
+from gitdb.exc import BadName
 from packageurl import PackageURL
-from pydriller import Git
+from pydriller import Commit, Git
+
+from macaron.repo_finder import repo_finder_deps_dev
+from macaron.repo_finder.repo_finder import to_domain_from_known_purl_types
+from macaron.slsa_analyzer.git_service import GIT_SERVICES
 
 logger: logging.Logger = logging.getLogger(__name__)
 
@@ -75,10 +80,12 @@ numeric_only_pattern = re.compile("[0-9]+")
 versioned_string = re.compile("[a-z]+[0-9]+", flags=re.IGNORECASE)  # e.g. RC1, M5, etc.
 
 
-def get_commit_from_version(git_obj: Git, purl: PackageURL) -> tuple[str, str]:
-    """Try to find the matching commit in a repository of a given version via tags.
+def find_commit(git_obj: Git, purl: PackageURL) -> tuple[str, str]:
+    """Try to find the commit matching the passed PURL.
 
-    The version of the passed PackageURL is used to match with the tags in the target repository.
+    The PURL may have be a repository type, e.g. GitHub, in which case the commit might be in its version part.
+    Otherwise, the PURL should be a package manager type, e.g. Maven, in which case the commit must be found from
+    the artifact version.
 
     Parameters
     ----------
@@ -92,10 +99,85 @@ def get_commit_from_version(git_obj: Git, purl: PackageURL) -> tuple[str, str]:
     tuple[str, str]
         The branch name and digest as a tuple.
     """
-    if purl.version is None:
-        logger.debug("Missing version for artifact: %s", purl.name)
+    version = purl.version
+    if version is None:
+        logger.debug("Missing version for analysis target: %s", purl.name)
         return "", ""
-    logger.debug("Searching for commit of artifact version using tags: %s@%s", purl.name, purl.version)
+
+    available_domains = [git_service.hostname for git_service in GIT_SERVICES if git_service.hostname]
+    domain = to_domain_from_known_purl_types(purl.type) or (purl.type if purl.type in available_domains else None)
+    if domain:
+        # PURL is a repository type.
+        return get_commit_from_purl(git_obj, version)
+    if purl.type in repo_finder_deps_dev.SUPPORTED_TYPES:
+        # PURL is a package manager type.
+        return get_commit_from_version(git_obj, purl.name or "", version)
+
+    logger.debug("Type of PURL is not supported for commit finding: %s", purl.type)
+    return "", ""
+
+
+def get_commit_from_purl(git_obj: Git, version: str) -> tuple[str, str]:
+    """Try to find the commit in the PURL or from the tag in the PURL.
+
+    Parameters
+    ----------
+    git_obj: Git
+        The repository.
+    version: str
+        The version of the analysis target.
+
+    Returns
+    -------
+    tuple[str, str]
+        The branch name and digest as a tuple.
+    """
+    # A commit hash is 40 characters in length, but commits are often referenced using only some of those.
+    commit: Commit | None = None
+    if 7 <= len(version) <= 40:
+        try:
+            commit = git_obj.get_commit(version)
+        except BadName as error:
+            logger.debug("Failed to retrieve commit: %s", error)
+
+    if not commit:
+        # Treat the 'commit' as a tag.
+        try:
+            commit = git_obj.get_commit_from_tag(version)
+        except IndexError as error:
+            logger.debug("Failed to retrieve commit: %s", error)
+
+    if not commit:
+        return "", ""
+
+    branch_name = _get_branch_of_commit(commit)
+    if not branch_name:
+        logger.debug("No valid branch found for commit: %s", commit.hash)
+        return "", ""
+
+    return branch_name, commit.hash
+
+
+def get_commit_from_version(git_obj: Git, name: str, version: str) -> tuple[str, str]:
+    """Try to find the matching commit in a repository of a given version via tags.
+
+    The version of the passed PackageURL is used to match with the tags in the target repository.
+
+    Parameters
+    ----------
+    git_obj: Git
+        The repository.
+    name: str
+        The name of the analysis target.
+    version: str
+        The version of the analysis target.
+
+    Returns
+    -------
+    tuple[str, str]
+        The branch name and digest as a tuple.
+    """
+    logger.debug("Searching for commit of artifact version using tags: %s@%s", name, version)
 
     # Only consider tags that have a commit.
     valid_tags = []
@@ -110,37 +192,25 @@ def get_commit_from_version(git_obj: Git, purl: PackageURL) -> tuple[str, str]:
         valid_tags.append(tag)
 
     if not valid_tags:
-        logger.debug("No tags with commits found for %s", str(purl))
+        logger.debug("No tags with commits found for %s", name)
         return "", ""
 
     # Match tags.
-    matched_tags = _match_tags(valid_tags, purl.name, purl.version)
+    matched_tags = _match_tags(valid_tags, name, version)
 
     if not matched_tags:
-        logger.debug("No tags matched for %s", str(purl))
+        logger.debug("No tags matched for %s", name)
         return "", ""
 
     if len(matched_tags) > 1:
-        logger.debug("Tags found for %s: %s", str(purl), len(matched_tags))
+        logger.debug("Tags found for %s: %s", name, len(matched_tags))
         logger.debug("Best match: %s", matched_tags[0])
         logger.debug("Up to 5 others: %s", matched_tags[1:6])
 
     tag = matched_tags[0]
     tag_name = str(tag)
-    branches = git_obj.get_commit_from_tag(tag_name).branches
-    logger.debug("Branches: %s", branches)
 
-    if not branches:
-        logger.debug("No branch associated with tag (commit): %s (%s)", tag_name, tag.commit.hexsha)
-        return "", ""
-
-    branch_name = ""
-    for branch in branches:
-        # Ensure the detached head branch is not picked up.
-        if "(HEAD detached at" not in branch:
-            branch_name = branch
-            break
-
+    branch_name = _get_branch_of_commit(git_obj.get_commit_from_tag(tag_name))
     if not branch_name:
         logger.debug("No valid branch associated with tag (commit): %s (%s)", tag_name, tag.commit.hexsha)
         return "", ""
@@ -150,8 +220,8 @@ def get_commit_from_version(git_obj: Git, purl: PackageURL) -> tuple[str, str]:
         tag,
         tag.commit.hexsha,
         branch_name,
-        purl.name,
-        purl.version,
+        name,
+        version,
     )
     return branch_name, tag.commit.hexsha
 
@@ -370,3 +440,21 @@ def _count_parts_in_tag(tag_version: str, tag_suffix: str, version_parts: list[s
                 count = count - 1
 
     return count
+
+
+def _get_branch_of_commit(commit: Commit) -> str:
+    """Get the branch of the passed commit as a string or return None."""
+    branches = commit.branches
+
+    if not branches:
+        logger.debug("No branch associated with commit: %s", commit.hash)
+        return ""
+
+    branch_name = ""
+    for branch in branches:
+        # Ensure the detached head branch is not picked up.
+        if "(HEAD detached at" not in branch:
+            branch_name = branch
+            break
+
+    return branch_name
--- a/src/macaron/repo_finder/commit_finder.py
+++ b/src/macaron/repo_finder/commit_finder.py
@@ -1,0 +1,160 @@
+# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""This module contains the logic for matching PackageURL versions to repository commits via the tags they contain."""
+import logging
+import re
+from re import Pattern
+
+from git import TagReference
+from packageurl import PackageURL
+from pydriller import Git
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+ALPHANUMERIC = "0-9a-z"
+PREFIX = "(?:.*)"
+INFIX = "[^0-9]*"
+split_pattern = re.compile(f"[^{ALPHANUMERIC}]", flags=re.IGNORECASE)
+validation_pattern = re.compile(f"[{ALPHANUMERIC}]+", flags=re.IGNORECASE)
+
+
+def get_commit_from_version(git_obj: Git, purl: PackageURL) -> tuple[str, str]:
+    """Try to find the matching commit in a repository of a given version via tags.
+
+    The version of the passed PackageURL is used to match with the tags in the target repository.
+
+    Parameters
+    ----------
+    git_obj: Git
+        The repository.
+    purl: PackageURL | None
+        The PURL of the artifact.
+
+    Returns
+    -------
+    tuple[str, str]
+        The branch name and digest as a tuple.
+    """
+    if purl.version is None:
+        logger.debug("Missing version for artifact: %s", purl.name)
+        return "", ""
+    logger.debug("Searching for commit of artifact version using tags: %s@%s", purl.name, purl.version)
+    # Create version based regex from the PackageURL for matching against tags.
+    # The version is split on non-alphanumeric characters to separate the version parts from the non-version parts.
+    # e.g. 1.2.3-DEV -> [1, 2, 3, DEV]
+    split = split_pattern.split(purl.version)
+    logger.debug("Split version: %s", split)
+    if not split:
+        split = [purl.version]
+    this_version_pattern = ""
+    for part in split:
+        # Validate the split part by checking it is only comprised of alphanumeric characters.
+        valid = validation_pattern.match(part)
+        if not valid:
+            continue
+        if this_version_pattern:
+            # To maximise chances of matching tags, the regex between the version parts allows for zero or more
+            # non-numeric characters.
+            this_version_pattern = this_version_pattern + INFIX
+        this_version_pattern = this_version_pattern + str(part)
+    # Prepend the optional prefix, add a named capture group for the version, and enforce end of string analysis.
+    this_version_pattern = PREFIX + "(?P<version>" + this_version_pattern + ")" + "$"
+    logger.debug("Created pattern: %s", this_version_pattern)
+    target_version_pattern = re.compile(this_version_pattern, flags=re.IGNORECASE)
+
+    # A pattern for very weakly confirming the presence of a version.
+    has_version_pattern = re.compile(".*[0-9].*", flags=re.IGNORECASE)
+
+    # Tags are examined as followed:
+    # - Any without a corresponding commit are discarded.
+    # - If any tag matches the has_name_pattern, only tags that match it will be examined.
+    # - If no tag matches the has_name_pattern, all tags will be examined.
+    named_tags = []
+    other_tags = []
+    for tag in git_obj.repo.tags:
+        try:
+            if not tag.commit:
+                raise ValueError
+        except ValueError:
+            logger.debug("No commit found for tag: %s", tag)
+            continue
+
+        tag_name = str(tag)
+
+        if has_version_pattern.match(tag_name):
+            named_tags.append(tag)
+        else:
+            other_tags.append(tag)
+
+    # Match tags.
+    if named_tags:
+        matched_tags = _match_tags(named_tags, target_version_pattern)
+    else:
+        matched_tags = _match_tags(other_tags, target_version_pattern)
+
+    # Report matched tags.
+    if not matched_tags:
+        logger.debug("No tags found for %s", str(purl))
+    else:
+        logger.debug("Tags found for %s: %s", str(purl), len(matched_tags))
+
+    if len(matched_tags) > 1:
+        # TODO decide how to handle multiple matching tags, and if it is possible
+        logger.debug("Found multiple tags for %s: %s", str(purl), len(matched_tags))
+
+    for tag in matched_tags:
+        tag_name = str(tag)
+        branches = git_obj.get_commit_from_tag(tag_name).branches
+
+        logger.debug("Branches: %s", branches)
+
+        if not branches:
+            continue
+
+        branch_name = ""
+        for branch in branches:
+            # Ensure the detached head branch is not picked up.
+            if "(HEAD detached at" not in branch:
+                branch_name = branch
+                break
+
+        if not branch_name:
+            continue
+
+        logger.debug(
+            "Found tag %s with commit %s of branch %s for artifact version %s@%s",
+            tag,
+            tag.commit.hexsha,
+            branch_name,
+            purl.name,
+            purl.version,
+        )
+        return branch_name, tag.commit.hexsha
+
+    logger.debug("Could not find tagged commit for artifact version: %s@%s", purl.name, purl.version)
+    return "", ""
+
+
+def _match_tags(tag_list: list[TagReference], pattern: Pattern) -> list[TagReference]:
+    """Return items of the passed tag list that match the passed pattern.
+
+    Parameters
+    ----------
+    tag_list: list[TagReference]
+        The list of tags to check.
+    pattern: Pattern
+        The pattern to match against.
+
+    Returns
+    -------
+    The list of tags that matched the pattern.
+    """
+    matched_tags = []
+    for tag in tag_list:
+        tag_name = str(tag)
+        match = pattern.match(tag_name)
+        if not match:
+            continue
+        matched_tags.append(tag)
+    return matched_tags
--- a/src/macaron/repo_finder/commit_finder.py
+++ b/src/macaron/repo_finder/commit_finder.py
@@ -281,7 +281,7 @@ def _build_version_pattern(version: str) -> tuple[Pattern, list[str], bool]:
             # - 3.1.test.2 -> 'test' and '2' become optional.
             has_non_numeric_suffix = True
 
-        if has_trailing_zero or has_non_numeric_suffix:
+        if count == len(split) - 1 and has_trailing_zero or has_non_numeric_suffix:
             # This part will be made optional in the regex, hence the grouping bracket.
             this_version_pattern = this_version_pattern + "("
 
@@ -293,7 +293,7 @@ def _build_version_pattern(version: str) -> tuple[Pattern, list[str], bool]:
         # Add the current part to the pattern.
         this_version_pattern = this_version_pattern + part
 
-        if has_trailing_zero or has_non_numeric_suffix:
+        if count == len(split) - 1 and has_trailing_zero or has_non_numeric_suffix:
             # Complete the optional capture group.
             this_version_pattern = this_version_pattern + ")?"
 
@@ -441,12 +441,11 @@ def _count_parts_in_tag(tag_version: str, tag_suffix: str, version_parts: list[s
                         versioned_string_match = True
                     else:
                         count = count + 1
+
+        if tag_suffix != last_part:
+            count = count + 1
         else:
-            # The tag suffix is a single part: reduce the count if it matches the version.
-            if tag_suffix != last_part:
-                count = count + 1
-            else:
-                count = count - 1
+            count = count - 1
 
     return count
 
--- a/src/macaron/repo_finder/commit_finder.py
+++ b/src/macaron/repo_finder/commit_finder.py
@@ -12,14 +12,67 @@ from pydriller import Git
 
 logger: logging.Logger = logging.getLogger(__name__)
 
-ALPHANUMERIC = "0-9a-z"
-# Any number of irrelevant prefix characters followed by 1 to 2 prefix-version separators (optional).
-PREFIX = "(?:.*[-_/vr]{1,2})?"
-# Version token separator: 1 to 3 alphabetic characters OR 1 to 3 non-alphanumeric characters.
-INFIX = "([a-z]{1,3}|[^0-9a-z]{1,3})"
-split_pattern = re.compile(f"[^{ALPHANUMERIC}]+", flags=re.IGNORECASE)
-validation_pattern = re.compile(f"[{ALPHANUMERIC}]+", flags=re.IGNORECASE)
-alphabetic_pattern = re.compile(".*[a-z].*", flags=re.IGNORECASE)
+# An optional named capture group "prefix" that accepts one of the following:
+# - A string of any characters starting with an alphabetic character, ending with one of:
+#   - One alphabetic character and one or more numbers.
+#   - One number and one alphabetic character.
+#   - Two alphabetic characters.
+# - OR
+# - Two alphabetic characters.
+# E.g.
+# - name_prefix
+# - prefix-a444
+# - vm
+# - name-prefix-j5u
+PREFIX = "^(?P<prefix>(?:[a-z].*(?:[a-z][0-9]+|[0-9][a-z]|[a-z]{2}))|[a-z]{2})?"
+
+# An optional named capture group "prefix_sep" that accepts one of:
+# - A 'v', 'r', or 'c' character that is not preceded by a non-alphanumeric character.
+# ('c' is probably a typo as it was found in only one example tag, but accepting any single alphabetic character
+# would also most likely be fine.)
+# - A non-alphanumeric character followed by 'v', 'r', or 'c'.
+# - A non-alphanumeric character.
+# Then optionally ending with one non-alphanumeric character.
+# E.g.
+# - _v-
+# - r_
+# - c
+# - .
+PREFIX_SEPARATOR = "(?P<prefix_sep>(?:(?:(?<![0-9a-z])[vrc])|(?:[^0-9a-z][vrc])|[^0-9a-z])(?:[^0-9a-z])?)?"
+
+# Together, the prefix and prefix separator exist to separate the prefix from version part of a tag, while ensuring that
+# the prefix is free from non-prefix characters (the separator). Note that the prefix is expected to be at least two
+# characters in length to prevent overlap with separators and confusion with versions; the prefix separator is at most
+# three characters; and a negative lookback passes if there are no preceding characters.
+
+# The infix exists between parts of the version string. The most recent design resulted in use of a back reference to
+# ensure non-suffix version parts were separated by the same separator, e.g. 1.2.3 but not 1.2-3. However, one edge
+# case required this to be partially reverted, as it required 1.2-3 to be accepted, while another edge case where
+# additional zeros need to be added after a version to pad its length, e.g. 1.2 becomes 1.2.0.0, still requires it.
+
+# The infix accepts either:
+# - One to three alphabetic characters.
+# - One to three non-alphanumeric characters.
+# Note: The upper limit of three could be reduced to two based on current data.
+INFIX_3 = "([a-z]{1,3}|[^0-9a-z]{1,3})"
+INFIX_1 = f"(?P<sep>{INFIX_3})"  # A named capture group of INFIX_3.
+INFIX_2 = "(?P=sep)"  # A back reference to INFIX_1.
+
+# The suffix separator exists for much the same purpose as the prefix separator: splitting the suffix into the actual
+# suffix, and the characters that join it to the version.
+# It optionally accepts:
+# One to two non-alphanumeric characters that are followed by a non-numeric character (positive lookahead).
+SUFFIX_SEPARATOR = "(?P<suffix_sep>[^0-9a-z]{1,2}(?=[^0-9]))?"
+
+# The suffix optionally accepts:
+# A string that starts with an alphabetic character, and continues for one or more characters of any kind.
+SUFFIX = "(?P<suffix>[a-z].*)?"
+
+split_pattern = re.compile("[^0-9a-z]", flags=re.IGNORECASE)
+validation_pattern = re.compile("[0-9a-z]+", flags=re.IGNORECASE)
+alphabetic_only_pattern = re.compile("[a-z]+", flags=re.IGNORECASE)
+numeric_only_pattern = re.compile("[0-9]+")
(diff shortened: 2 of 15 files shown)


benmss:
The regex might still change as I will be performing another iteration of tests to check its correctness against the dataset I have.

src/macaron/repo_finder/repo_finder.py
```diff
@@ -148,3 +162,104 @@ def to_repo_path(purl: PackageURL, available_domains: list[str]) -> str | None:
...
+        tag_name = str(tag)
+        adjusted_tag_name = tag_name
+        contains_name = False
+        if purl.name.lower() in tag_name.lower():
+            adjusted_tag_name = re.sub(purl.name, "", adjusted_tag_name, re.IGNORECASE)
+            contains_name = True
+
+        match = tag_pattern.match(adjusted_tag_name)
+
+        if not require_name_match and match and match.group("version") and contains_name:
```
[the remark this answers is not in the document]
benmss: `require_name_match` is true when at least one of the tags contains `purl.name`. 
`contains_name` is true when the current tag contains `purl.name`. 
This check is for tags with identifying prefixes, but not all repositories use them (or need to). If a relevant identifying prefix is found, it becomes a requirement for a proper match. If none is found, a version tag can be considered a match regardless of extra prefixes.

src/macaron/repo_finder/repo_finder.py
```diff
@@ -148,3 +162,104 @@ def to_repo_path(purl: PackageURL, available_domains: list[str]) -> str | None:
             fragment="",
         )
     )
+
+
+def get_commit_from_version_tag(git_obj: Git, purl: PackageURL) -> tuple[str, str]:
```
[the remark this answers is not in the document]
benmss: It is the version from the tag and the PackageURL. How about `get_commit_from_version`?

src/macaron/repo_finder/repo_finder.py
```diff
@@ -148,3 +162,104 @@ def to_repo_path(purl: PackageURL, available_domains: list[str]) -> str | None:
...
+    -------
+    tuple[str, str]
+        The branch name and digest as a tuple.
+    """
+    # Try to resolve the branch and digest from the version
+    logger.debug("Searching for commit of artifact version using tags: %s@%s", purl.name, purl.version)
+    matched_tags = []
+    # Iterate over tags, keeping any that match the regex version pattern, contain the purl.name, and/or match the
+    # specific version.
+    # If any tags contain the purl.name of the artifact and a valid version, tags that only contain the version
```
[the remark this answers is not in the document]
benmss: There is nothing missing here. I will try and make it clearer though.

tests/repo_finder/test_repo_finder.py
```diff
@@ -69,3 +74,45 @@ def test_resolve_analysis_target(
...
+    if init_repo:
+        tags = [
+            "test-name-v1.0.1-A",
+            "v1.0.1-B",
+            "v1.0.3+test",
+            "1.0.5",
+            "50.0",
+            "78A",
+        ]
+        files = [path.joinpath(".git", "description")]
```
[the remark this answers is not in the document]
benmss: Not sure I see the issue given the repository only exists for testing purposes. I'm not against changing it, but further elaboration as to why it should be changed would be appreciated.

tests/repo_finder/test_repo_finder.py
```diff
@@ -69,3 +74,45 @@ def test_resolve_analysis_target(
...
+        "1.0.1-B",
+        "1.0.3+test",
+        "1.0.5",
+        "50.0",
+        "78A",
+    ]
+    purl_name = "test-name"
+    for count, value in enumerate(versions):
+        _test_tag(git_obj, PackageURL(type="maven", name=purl_name, version=value), str(count))
+        purl_name = purl_name + "-" + str(count)
```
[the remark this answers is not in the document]
benmss: The only real requirement is that all iterations after the first are different, but you are right in that I most likely did not intend for the string to grow as the above demonstrates. I will change it.

tests/repo_finder/test_repo_finder.py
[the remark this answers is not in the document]
tromai: Hmm right, I'm just curious if there is any particular reason why we need to do it this way and whether committing `.git/description` would have any side effects that we should know of (that can cause the test to fail unexpectedly). If not, I don't see why we need to change it.

src/macaron/repo_finder/repo_finder.py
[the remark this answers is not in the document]
tromai: > This check is for tags with identifying prefixes, but not all repositories use them (or need to)

Hmm right, then I think this is essentially an edge case as not all repositories use them. I strongly believe we should add some comments to describe this one and/or an example to explain it.

src/macaron/repo_finder/repo_finder.py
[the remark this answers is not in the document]
tromai: I guess my question with this comment is:
- What is going to happen to `tags contain the purl.name of the artifact and a valid version`, are they going to be ignored?

tests/repo_finder/test_repo_finder.py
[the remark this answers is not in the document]
benmss: That makes sense but I've decided to skip committing the file altogether as only the commit, message, and tag are actually required for the test.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # A pattern for very weakly confirming the presence of a version.
+    has_version_pattern = re.compile(".*[0-9].*", flags=re.IGNORECASE)
+
+    # Tags are examined as followed:
+    # - Any without a corresponding commit are discarded.
+    # - If any tag matches the has_name_pattern, only tags that match it will be examined.
+    # - If no tag matches the has_name_pattern, all tags will be examined.
+    named_tags = []
+    other_tags = []
```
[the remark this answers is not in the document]
nathanwn: I think this is not entirely true. At least for mypy, when a generic list is initialized like this, the type of the list is inferred from the first `.append` to the list. You can try running the following examples through mypy.

**Example 1**: mypy cannot infer the type of an empty list just from initialization.
```python
from typing_extensions import reveal_type
a = []
reveal_type(a)
```
`mypy t.py` shows:
```
t.py:2: error: Need type annotation for "a" (hint: "a: List[<type>] = ...")  [var-annotated]
t.py:3: note: Revealed type is "builtins.list[Any]"
Found 1 error in 1 file (checked 1 source file)
```


**Example 2**: mypy can infer the type of the list from the first `.append`.
```python
from typing_extensions import reveal_type
a = []
a.append(1)
reveal_type(a)
```
`mypy t.py` shows:
```
t.py:4: note: Revealed type is "builtins.list[builtins.int]"
Success: no issues found in 1 source file
```

**Example 3**: mypy infers the type of the list from the first `.append` and type-checks the subsequent `.append` accordingly.
```python
from typing_extensions import reveal_type
a = []
a.append(1)
a.append("hello")
reveal_type(a)
```
`mpyp t.py` shows:
```
t.py:4: error: Argument 1 to "append" of "list" has incompatible type "str"; expected "int"  [arg-type]
t.py:5: note: Revealed type is "builtins.list[builtins.int]"
Found 1 error in 1 file (checked 1 source file)
```

This is somewhat unrelated, but I have seen this pattern of type inference for containers from the first push/append in another language, [Rust](https://doc.rust-lang.org/rust-by-example/types/inference.html).

In this particular case, my personal opinion is it is fine whether these lists are type-annotated or not. You can make the argument for both sides. This is not about typing issues but more about code readability and developer ergonomics.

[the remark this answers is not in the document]
tromai: Hmm that's true. I think that there might be some issues with my mypy settings in my editor. It didn't raise any errors in example 3 so I thought It couldn't resolve the types from the usages of `append`. This explanation makes sense, Thanks @nathanwn. 
@benmss I guess it's up to you regarding the final decision, I'm also fine with both options as they are not typing issues.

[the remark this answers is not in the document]
behnazh-w: Nice conversation. Since these are local variables, the name of the variables can be enough to improve readability. See a similar discussion on usage of `var` in Java: https://openjdk.org/projects/amber/guides/lvti-faq

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,160 @@
...
+    tag_list: list[TagReference]
+        The list of tags to check.
+    pattern: Pattern
+        The pattern to match against.
+
+    Returns
+    -------
+    The list of tags that matched the pattern.
+    """
+    matched_tags = []
```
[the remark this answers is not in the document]
tromai: See https://github.com/oracle/macaron/pull/508#discussion_r1362938860. This fix is optional.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,160 @@
...
+    # The version is split on non-alphanumeric characters to separate the version parts from the non-version parts.
+    # e.g. 1.2.3-DEV -> [1, 2, 3, DEV]
+    split = split_pattern.split(purl.version)
+    logger.debug("Split version: %s", split)
+    if not split:
+        split = [purl.version]
+    this_version_pattern = ""
+    for part in split:
+        # Validate the split part by checking it is only comprised of alphanumeric characters.
+        valid = validation_pattern.match(part)
```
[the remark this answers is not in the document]
benmss: There was a slight error in the regex that I've now fixed. It is supposed to split on any number of separators, so your example above would produce `['1', '2', '3', 'A']`.
This extra validation was added more for protection against shenanigans, as the version string is essentially user input, and parts of it will be directly used in a new regex pattern. I'm not sure whether that kind of injection matters, or if it's not already too late at that point, but I thought making an attempt was still worthwhile.

src/macaron/repo_finder/commit_finder.py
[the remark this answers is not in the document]
benmss: Interesting for sure. I've added the type as I think it does help readability in this case.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -40,49 +40,27 @@ def get_commit_from_version(git_obj: Git, purl: PackageURL) -> tuple[str, str]:
...
-        this_version_pattern = this_version_pattern + str(part)
-    # Prepend the optional prefix, add a named capture group for the version, and enforce end of string analysis.
-    this_version_pattern = PREFIX + "(?P<version>" + this_version_pattern + ")" + "$"
-    logger.debug("Created pattern: %s", this_version_pattern)
-    target_version_pattern = re.compile(this_version_pattern, flags=re.IGNORECASE)
 
-    # A pattern for very weakly confirming the presence of a version.
-    has_version_pattern = re.compile(".*[0-9].*", flags=re.IGNORECASE)
+    target_version_pattern = _build_version_pattern(purl.version)
+    has_name_pattern = re.compile(f".*{purl.name}.*[0-9].*", flags=re.IGNORECASE)
```
tromai, line 45: Would there be any case where there would be non-alphanumeric characters in `purl.name` that could change the semantic of this regex string? If yes, I think we should escape those or use a [raw string](https://docs.python.org/3/library/re.html#raw-string-notation) instead.

tromai (approved):

behnazh-w (commented):
Looks like for `pkg:maven/com.google.guava/guava@32.1.2-jre?type=jar` the commit finder maps the artifact to (adde55ee1f9bc7fb053183b5465f413547c198fe) which is the latest commit on `master`.

Can you please create tasks for the following items in the PR description?

1. We should cover cases where artifact versions have suffixes that do not match tags.
2. If a tag is not found, as we had discussed before, we shouldn't fall back to the latest commit. Instead, the repository should  be None to fit it the current data model.
3. Since we have a commit mapping feature, we shouldn't need to pin tests in the integration tests to a commit anymore. Just passing the purl should be enough. So, we can probably remove at least some of the -c config.yaml  runs in the integration tests with purl

Here is one dependency from `apache_maven_root_sbom.json` which we use in our integration tests that is also mapped to the latest commit: `pkg:maven/org.junit.jupiter/junit-jupiter-engine@5.8.1?type=jar`

tests/e2e/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env python3
```
nathanwn, line 1: FYI in #520, we have decided to avoid running Python scripts with system Python and instead use the Python interpreter in the virtual env. With that, we opt to avoid using the Python shebang in the scripts to avoid any confusion regarding how the scripts are run.
I saw that in `integration_tests.sh` you run this script with `python <script-name>` which is good as is. However, the shebang here should be removed.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+    # If multiple tags still remain, sort them based on the closest match in terms of individual parts.
+    if len(matched_tags) > 1:
+        matched_tags.sort(
+            key=lambda matched_tag: _count_parts_in_tag(matched_tag["version"], matched_tag["suffix"], parts)
+        )
+
+    return _extract_tags_from_list(matched_tags)
+
+
+def _extract_tags_from_list(matched_tags: list[dict[str, str | TagReference]]) -> list[TagReference]:
```
nathanwn, line 321: From where this function is called, shouldn't the type of the `matched_tags` param just be as follows?
```diff
-def _extract_tags_from_list(matched_tags: list[dict[str, str | TagReference]]) -> list[TagReference]:
+def _extract_tags_from_list(matched_tags: list[dict[str, str]]) -> list[TagReference]:
```
I ran this through mypy and this seems correct.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+
+    Returns
+    -------
+    list[TagReference]
+        The list of tags.
+    """
+    tags: list[TagReference] = []
+    for matched_tag in matched_tags:
+        tag = matched_tag.get("tag")
+        if tag is not None and isinstance(tag, TagReference):
```
nathanwn, line 337: Looking at the way you populate the `matched_tags` list, the `tag` field always exists and always has the type `TagReference`. Therefore, this `not None` and dynamic type check is probably unnecessary, and using `matched_tag["tag"]` is more appropriate than `matched_tag.get("tag")` here.

In addition, I think this private method is probably over-DRYing a bit - it is only called once and can be replaced with a simple list comprehension like so:
```python
tags = [_["tag"] for _ in matched_tags]
```
Therefore, I think you can consider removing this method with a similar list comprehension, which is simpler to read.

src/macaron/config/defaults.ini
```diff
@@ -47,6 +47,7 @@ recursive = False
 [repofinder]
 find_repos = True
 use_open_source_insights = True
+find_commits = True
```
nathanwn, line 50: Can you help me understand why we need another config knob here? 🤔 In which case should we not map versions to commits?

src/macaron/repo_finder/commit_finder.py
benmss: I had planned to take another look at this but forgot. Thanks for the reminder, I'll use the one-liner.

src/macaron/repo_finder/commit_finder.py
benmss: This seems a bit strange to me as one of the items in the dictionary is a `TagReference`. Regardless, this function will be removed.

src/macaron/config/defaults.ini
benmss: Given that nothing will be returned if the commit finder fails, it may no longer be an optional feature. So you may be right in that it is not necessary.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -700,6 +705,13 @@ def _prepare_repo(
             logger.error("Cannot reset the target repository.")
             return None
 
+        # Find the digest and branch if a version has been specified
+        if defaults.getboolean("repofinder", "find_commits") and not digest and purl and purl.version:
```
nathanwn, line 709: Checking if `purl.version` is `None` here is probably not necessary, since you are checking it in the `get_commit_from_version` function anyway.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+def get_commit_from_version(git_obj: Git, purl: PackageURL) -> tuple[str, str]:
+    """Try to find the matching commit in a repository of a given version via tags.
+
+    The version of the passed PackageURL is used to match with the tags in the target repository.
+
+    Parameters
+    ----------
+    git_obj: Git
+        The repository.
+    purl: PackageURL | None
```
nathanwn, line 87:
```suggestion
    purl: PackageURL
```

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+        The PURL of the artifact.
```
nathanwn, line 88: Looking at the context where this function is called, I think this should be the PURL of the "analysis target" rather than the "artifact".
The call chain here is: `add_component -> _prepare_repo -> get_commit_from_version`. I see that the `purl` param of `_prepare_repo` is documented as "PURL of the artifact" and it is probably why you also documented similarly here. However, from the top-level call site, I think "the PURL of the analysis target" is more precise since it makes a less specific assumption of where this PURL comes from.
It also makes sense to fix the docstring of `_prepare_repo` accordingly.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+
+    # Only consider tags that have a commit.
+    valid_tags = []
+    for tag in git_obj.repo.tags:
+        try:
+            if not tag.commit:
+                raise ValueError("The commit object is None")
+        except ValueError:
+            logger.debug("No commit found for tag: %s", tag)
+            continue
```
nathanwn, line 110: This `try-raise-except` combination here is probably not necessary. 😄

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+    -------
+    tuple[str, str]
+        The branch name and digest as a tuple.
+    """
+    if purl.version is None:
+        logger.debug("Missing version for artifact: %s", purl.name)
+        return "", ""
+    logger.debug("Searching for commit of artifact version using tags: %s@%s", purl.name, purl.version)
+
+    target_version_pattern, parts, has_non_numeric_suffix = _build_version_pattern(purl.version)
```
nathanwn, line 100: Because these values are only relevant in the `_match_tags` function, I reckon it is better to move this call of `_build_version_pattern` into `_match_tags` instead of leaving it here.

src/macaron/slsa_analyzer/analyzer.py
benmss: If there is no version we don't want to call the commit finder, because doing so would result in a failed attempt to prepare the repo versus continuing and most likely checking out the latest commit (for a non-versioned PURL),

src/macaron/repo_finder/commit_finder.py
benmss: Sadly it is. @tromai and I discussed it during his review.

nathanwn: Oh, I see what is happening here. `commit` here is a `@property` method and can raise `ValueError`.

I was confused because I thought the `except ValueError` was only catching the `ValueError` raised on the previous line, and was not aware of the error potentially raised by the property method.

To communicate the intention here, maybe something like this is a bit clearer:
```python
try:
    # commit is a property method which can raise ``ValueError``
    commit = tag.commit
except ValueError:
    logger.debug(...)
    continue

if not commit:
    logger.debug(...)
    continue
```
I suppose it's up to you to decide which way is better, but at the very least I think there should be a comment about `tag.commit` potentially raising `ValueError`.

nicallen: I think it would be clearer to extract this out into a function to hide the exception logic:
```
def _get_tag_commit(tag: TagReference) -> Commit | None:
    try:
        return tag.commit
    except ValueError:
        return None
```

src/macaron/slsa_analyzer/analyzer.py
nathanwn: Ah, I see what you mean. The `get_commit_from_version` function should only be called when `purl.version` is present. That makes sense.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+            continue
+        parts.append(part)
+
+        # Detect versions that end with a zero, so the zero can be made optional.
+        has_trailing_zero = len(split) > 2 and count == len(split) - 1 and part == "0"
+        numeric_only = numeric_only_pattern.match(str(part))
+
+        if not has_non_numeric_suffix and not numeric_only:
+            # A non-numeric part enables the flag for treating this and all remaining parts as version suffix parts.
+            has_non_numeric_suffix = True
```
nathanwn, line 201: Can you give some examples of version strings that should return `has_non_numeric_suffix` to be `True`? I am unsure about the meaning of it.
From my testing, it returns `True` for the version string `v0.1.0`, which I am not sure is as intended.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+SUFFIX = "(?P<suffix>[a-z].*)?"
+
+split_pattern = re.compile("[^0-9a-z]", flags=re.IGNORECASE)
+validation_pattern = re.compile("[0-9a-z]+", flags=re.IGNORECASE)
+alphabetic_only_pattern = re.compile("[a-z]+", flags=re.IGNORECASE)
+numeric_only_pattern = re.compile("[0-9]+")
+versioned_string = re.compile("[a-z]+[0-9]+", flags=re.IGNORECASE)  # e.g. RC1, M5, etc.
+
+
+def get_commit_from_version(git_obj: Git, purl: PackageURL) -> tuple[str, str]:
```
nathanwn, line 78: As we have discussed, for GitHub PURLs (as well as other repository PURLs), `purl.version` is not a package version, but rather a commit or a tag (see [github type](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#github) and [bitbucket type](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#bitbucket) sections). Therefore, we also need to consider `purl.type` for the `purl.version -> commit` mapping.

src/macaron/repo_finder/commit_finder.py
benmss: If any part of a version string contains a non-numeric character then this flag will become `True`.
E.g. `1.1A.2` -> `True`, `3.5.RELEASE` -> `True`
Your example above would cause it to be true, but only if you include the `v`. Typically, it is only tags that start with an alphabetic character like that.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+    int
+        The sort value based on the similarity between the tag and version, lower means more similar.
+
+    """
+    count = len(version_parts)
+    # Reduce count for each direct match between version parts and tag version.
+    tag_version_text = tag_version
+    for part in version_parts:
+        if part in tag_version_text:
+            tag_version_text.replace(part, "", 1)
```
nathanwn, line 365: This is likely incorrect, because `.replace` does not mutate the string.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+    # If any of the matches contain a prefix that matches the target artifact name, remove those that don't.
+    named_tags = []
+    for item in matched_tags:
+        prefix: str | None = item["prefix"]
+        if not prefix:
+            continue
+        if "/" in prefix:
+            # Exclude prefix parts that exists before a forward slash, e.g. rel/
+            split = prefix.split("/")
+            prefix = split[len(split) - 1]
```
nathanwn, line 305: Suggestion: this is probably cleaner:
```suggestion
        # Exclude prefix parts that exist before a forward slash, e.g. rel/
        _, _, prefix = prefix.rpartition("/")
```

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+    #  version group.
+    if has_non_numeric_suffix:
+        filtered_tags = []
+        for item in matched_tags:
+            # Discard tags with no suffix or with one that does not match the version.
+            suffix: str | None = item["suffix"]
+            if not suffix:
+                filtered_tags.append(item)
+                continue
+            if suffix == parts[len(parts) - 1]:
```
nathanwn, line 290:
```suggestion
            if suffix == parts[-1]:
```

tests/repo_finder/test_repo_finder.py
```diff
@@ -69,3 +75,38 @@ def test_resolve_analysis_target(
 ) -> None:
     """Test the resolve analysis target method with valid inputs."""
     assert Analyzer.to_analysis_target(config, available_domains) == expect
+
+
```
nathanwn, line 79: I prefer having tests related to the commit finder in a different file, e.g. `test_commit_finder.py`.

tests/repo_finder/test_repo_finder.py
```diff
@@ -69,3 +75,38 @@ def test_resolve_analysis_target(
...
+def test_get_commit_from_version() -> None:
```
nathanwn, line 80: I think the thing you are actually testing in this test method is the mapping from PURLs (or more specifically the version part in PURLs) to tags.
In that case, I think mocking a Git repo is unnecessary. I think we can simplify matters here by not unit-testing directly on `get_commit_from_version(git_obj: Git, purl: PackageURL) ->  [Branch, Commit]`. Instead, we can create a smaller function that `get_commit_from_version` itself calls within, but have a simpler interface, something along the line of:

```
find_matching_tags(tags: Tags, purl: PackageURL) -> [Branch, Commit]
```

and unit test this instead, which is much easier to handle without any complex mocking setup.

src/macaron/repo_finder/commit_finder.py
nathanwn: Hmm, I checked the [PURL type definitions](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst) page and it does seem that most version strings do not have a `v` at the start.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+    has_non_numeric_suffix = False
+    for count, part in enumerate(split):
+        # Validate the split part by checking it is only comprised of alphanumeric characters.
+        valid = validation_pattern.match(part)
+        if not valid:
+            continue
+        parts.append(part)
+
+        # Detect versions that end with a zero, so the zero can be made optional.
+        has_trailing_zero = len(split) > 2 and count == len(split) - 1 and part == "0"
```
tromai, line 196: I think `len(split)` can be computed outside of the for loop.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+        numeric_only = numeric_only_pattern.match(str(part))
```
tromai, line 197: Why does we need to coerce part to a string here while we don't for `valid = validation_patter.match(part)`?

src/macaron/repo_finder/commit_finder.py
nathanwn: Do you mean `has_trailing_zero` can be computed outside of the for loop instead?

benmss: I've moved `has_trailing_zero` out of the loop and simplified it.
`len(split)` is O(1) so there's no real benefit there.

src/macaron/repo_finder/commit_finder.py
benmss: Indeed, it is not needed.

tests/repo_finder/test_repo_finder.py
benmss: I will look into this as part of adding tests with Hypothesis.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+#   - One number and one alphabetic character.
+#   - Two alphabetic characters.
+# - OR
+# - Two alphabetic characters.
+# E.g.
+# - name_prefix
+# - prefix-a444
+# - vm
+# - name-prefix-j5u
+PREFIX = "^(?P<prefix>(?:[a-z].*(?:[a-z][0-9]+|[0-9][a-z]|[a-z]{2}))|[a-z]{2})?"
```
nathanwn, line 27: I think for these examples, it will be nice if you can also include the full version strings, and then point out which part in each version string is considered a prefix.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+            # Complete the optional capture group.
+            this_version_pattern = this_version_pattern + ")?"
+
+        # Extend the pattern with up to three additional zeros to help map mismatches between versions and tags.
+        # E.g. 1.0 to 1.0.0, or 3.2 to 3.2.0, etc.
+        if not has_non_numeric_suffix and count == len(split) - 1 and 1 < numeric_parts < 4:
+            while numeric_parts < 4:
+                # Additional zeros added for this purpose make use of a back reference to the first matched separator.
+                this_version_pattern = this_version_pattern + "(" + INFIX_2 + "0)?"
+                numeric_parts = numeric_parts + 1
```
nathanwn, line 229: This `if` statement only runs when the outer `for` loop reaches the last iteration, and it literally runs at the end of this last iteration. Therefore I think it should be moved out of the `for` loop.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+        this_version_pattern = this_version_pattern + str(part)
+
+        if has_trailing_zero or has_non_numeric_suffix:
+            # Complete the optional capture group.
+            this_version_pattern = this_version_pattern + ")?"
+
+        # Extend the pattern with up to three additional zeros to help map mismatches between versions and tags.
+        # E.g. 1.0 to 1.0.0, or 3.2 to 3.2.0, etc.
+        if not has_non_numeric_suffix and count == len(split) - 1 and 1 < numeric_parts < 4:
+            while numeric_parts < 4:
```
nathanwn, line 226: Can you leave some comments here explaining these conditions: `if 1 < numeric_parts < 4` and `while numeric_parts < 4`?

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,397 @@
...
+# case required this to be partially reverted, as it required 1.2-3 to be accepted, while another edge case where
+# additional zeros need to be added after a version to pad its length, e.g. 1.2 becomes 1.2.0.0, still requires it.
+
+# The infix accepts either:
+# - One to three alphabetic characters.
+# - One to three non-alphanumeric characters.
+# Note: The upper limit of three could be reduced to two based on current data.
+INFIX_3 = "([a-z]{1,3}|[^0-9a-z]{1,3})"
+INFIX_1 = f"(?P<sep>{INFIX_3})"  # A named capture group of INFIX_3.
+INFIX_2 = "(?P=sep)"  # A back reference to INFIX_1.
```
nathanwn, line 59: From your implementation, I think these should not be called "infix", because you also have the notion of "prefix" and "suffix" in your tag-matching implementation, and it sounds like "infix" should be "anything and everything between the prefix and the suffix", while in reality these "INFIX_x" patterns here are for "separators between numeric parts in the actual infix of the version string".

If I understand it right, given this version string `foobar-1-2-3-baz`, these infix patterns will match the `-` between `1` and `2`, and `2` and `3`. I think `INFIX_SEP` or something along the line would be a better name to avoid confusion.

I also want to see some examples of these cases in your comments. Similarly to my previous remark, it would be nice if you could also include the full version strings, and then point out which parts are matched as "infix separators".

benmss: A definition of infix:

> An infix is an affix inserted inside a word stem.

It goes onto say other things but given we are dealing with the creation of regex strings rather than words. I think any addition to the middle parts of the regex would count.

The infix pattern can also match alphabetic characters if they are between the version numbers, e.g. `r1rc2` -> `rc` is an infix.

Given the terminology here, it's probably more accurate to say that the version numbers are the infix separators.

src/macaron/repo_finder/commit_finder.py
nathanwn: It makes sense to leave some comments in the code to describe what a `has_non_numeric_suffix` is (some examples would be nice), and also to describe its effect on the regex pattern being built.

nathanwn: I want to confirm if I understand this right. Let's take one of the examples you gave:
```
1.1A.2
```
Here the regex generated will be "1(...)?", since the `1A` part switches `has_non_numeric_suffix` to `True`. I assume your assumption here is: parts that are not all-numeric are likely insignificant and they can be matched optionally.

nathanwn: FYI the version string `1.1A.2` results in `has_non_numeric_suffix` being `False`.
I added the following `print` statements to figure out what is going on:
```diff
         # Detect versions that end with a zero, so the zero can be made optional.
         has_trailing_zero = len(split) > 2 and count == len(split) - 1 and part == "0"
         numeric_only = numeric_only_pattern.match(str(part))
+        print(f"{part=}")
+        print(f"{numeric_only=}")
```
and called `_build_version_pattern("1.1A.2")`. This printed
```
part='1'
numeric_only=<re.Match object; span=(0, 1), match='1'>
part='1A'
numeric_only=<re.Match object; span=(0, 1), match='1'>
part='2'
numeric_only=<re.Match object; span=(0, 1), match='2'>
```
which is wrong, because `1A` should not be matched as being "numeric only".
I think the issue here is with the `numeric_only_pattern` regex. Currently, it is
```python3
numeric_only_pattern = re.compile("[0-9]+")
```
This pattern still matches against `1A` because it is not exhaustive. What we need here is
```diff
-numeric_only_pattern = re.compile("[0-9]+")
+numeric_only_pattern = re.compile("^[0-9]+$")
```
After this fix, the result is now as expected.
I suspect some other regex patterns also need to be changed, one of which is `validation_pattern`.

src/macaron/repo_finder/commit_finder.py
nathanwn: For me it's not the word infix by itself that confuses, but the fact that it appears alongside "prefix" and "suffix".
Something like "version_number_separator" is probably better IMO.

benmss: I have to disagree. The words "prefix", "suffix", and "infix" are all being used to describe parts of the regex pattern that are being added based on where they are being added, not necessarily on how they relate to the version. The base pattern, or word stem from the definition, is simply the parts from the version, e.g. `12A3`.  All "affixes" are modifications to the base pattern that allow the version to be mapped to a tag.

nathanwn: Alright. I think as long as you leave some examples in the comments to clarify what these regexes should match, I think it is all good for me. Thanks, Ben. 👍🏼

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,480 @@
...
+            count = count + 1
+        else:
+            count = count - 1
+
+    return count
+
+
+def _get_branch_of_commit(commit: Commit) -> str:
+    """Get the branch of the passed commit as a string or return None."""
+    branches = commit.branches
```
tromai, line 455: It looks to that the [`Commit` ](https://gitpython.readthedocs.io/en/stable/reference.html#module-git.objects.commit)object doesn't have the `branches` attribute or method. The only object I think have the `branches` method is [`git.Repo`](https://gitpython.readthedocs.io/en/stable/reference.html#git.repo.base.Repo.branches) :thinking:

nicallen: `Commit` in this case is `pydriller.Commit` (which does have `branches`) , not `git.objects.Commit`. The overloaded name is a bit confusing.

tromai: Righttt, my bad! It's indeed quite confusing.

benmss: It's the `Commit` from `PyDriller` not `GitPytjpm`.

tests/repo_finder/test_commit_finder.py
```diff
@@ -34,14 +25,86 @@ def test_get_commit_from_version() -> None:
...
+    """Retrieve tag matching version and check it is correct."""
+    matched_tags = commit_finder.match_tags(tags, name, version)
+    assert matched_tags
+    assert matched_tags[0] == target_tag
+
+
+input_pattern = re.compile(r"[0-9]{1,3}(\.[0-9a-z]{1,3}){,5}([-+#][a-z0-9].+)?", flags=re.IGNORECASE)
+
+
+@given(hypothesis.strategies.text())
```
tromai, line 48: Is there any reason why we have `hypothesis.strategies.text()` here while other places in this file use the strategy in a different manner? For example on line 71
```python
from hypothesis.strategies import DataObject, data

@given(data())
@settings(max_examples=VERSION_ITERATIONS, deadline=None)
```

tests/repo_finder/test_commit_finder.py
```diff
@@ -34,14 +25,86 @@ def test_get_commit_from_version() -> None:
...
+    # Try creating a PURL from the version, if successful pass the purl.version to commit finder's pattern builder.
+    if not version:
+        return
+    try:
+        purl = PackageURL(name="test", version=version, type="maven")
+        if not purl.version:
+            return
+        commit_finder._build_version_pattern(purl.version)
+        assert True
+    except ValueError as error:
```
tromai, line 62: I wonder where is ValueError raised from? I had a quick look at `commit_finder._build_version_pattern` but I don't see it being raised anywhere :thinking:.
If it's to catch the ValueError raised by `PackageURL` then I think we shoudn't wrap ` commit_finder._build_version_pattern(purl.version)` in this try block too. This is because it would hide the `ValueError` if it's unexpectedly came from `commit_finder._build_version_pattern`, which defeats the goal of this test.

src/macaron/repo_finder/repo_finder.py
```diff
@@ -111,7 +111,7 @@ def to_repo_path(purl: PackageURL, available_domains: list[str]) -> str | None:
     - The pre-defined repository-based PURL type as defined in
     https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst
 
-    - The supported git service domains (e.g. ``github.com``) defined in ``available_domains``.
+    - The supprted git service domains (e.g. ``github.com``) defined in ``available_domains``.
```
tromai, line 114: I think this is a typo.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+MAX_ZERO_DIGIT_EXTENSION = 4
+
+split_pattern = re.compile("[^0-9a-z]", flags=re.IGNORECASE)
+validation_pattern = re.compile("[0-9a-z]+$", flags=re.IGNORECASE)
+alphabetic_only_pattern = re.compile("[a-z]+$", flags=re.IGNORECASE)
+numeric_only_pattern = re.compile("[0-9]+$")
+versioned_string = re.compile("[a-z]+[0-9]+$", flags=re.IGNORECASE)  # e.g. RC1, M5, etc.
+
+
+def find_commit(git_obj: Git, purl: PackageURL) -> tuple[str, str]:
```
behnazh-w, line 104: Can you please add a unit test for this function. If possible, try to pass invalid inputs that trigger error-handling branches.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+        return get_commit_from_purl(git_obj, version)
+    if purl.type in repo_finder_deps_dev.SUPPORTED_TYPES:
+        # PURL is a package manager type.
+        return get_commit_from_version(git_obj, purl.name or "", version)
+
+    logger.debug("Type of PURL is not supported for commit finding: %s", purl.type)
+    return "", ""
+
+
+def get_commit_from_purl(git_obj: Git, version: str) -> tuple[str, str]:
```
behnazh-w, line 141: Same comment regarding adding  a unit test.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+
+    branch_name = _get_branch_of_commit(commit)
+    if not branch_name:
+        logger.debug("No valid branch found for commit: %s", commit.hash)
+        return "", ""
+
+    return branch_name, commit.hash
+
+
+def get_commit_from_version(git_obj: Git, name: str, version: str) -> tuple[str, str]:
```
behnazh-w, line 182: Same comment regarding adding a unit test.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+        whether the version string has a non-numeric suffix.
+
+    """
+    # The version is split on non-alphanumeric characters to separate the version parts from the non-version parts.
+    # e.g. 1.2.3-DEV -> [1, 2, 3, DEV]
+    try:
+        split = split_pattern.split(version)
+    except TypeError as error:
+        logger.debug("Failed to perform regex split on version string %s -- %s", version, error)
+        return None, [], False
```
behnazh-w, line 275: Can you please add a unit test to trigger this error?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -695,6 +699,13 @@ def _prepare_repo(
             logger.error("The target repository does not have any commit.")
             return None
 
+        # Find the digest and branch if a version has been specified
+        if not digest and purl and purl.version:
+            branch_name, digest = find_commit(git_obj, purl)
+            if not (branch_name and digest):
+                logger.error("Could not map purl version to specific commit in repository.")
```
behnazh-w, line 706:
```suggestion
                logger.error("Could not map the input purl string to a specific commit in the corresponding repository.")
```

tests/e2e/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,71 @@
...
+from packageurl import PackageURL
+
+from macaron.repo_finder import commit_finder
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+# Set logging debug level.
+logger.setLevel(logging.DEBUG)
+
+path = Path(__file__).parent.joinpath("resources", "java_tags.json")
```
behnazh-w, line 21:
```suggestion
java_tags_file_path = Path(__file__).parent.joinpath("resources", "java_tags.json")
```

tests/repo_finder/test_commit_finder.py
```diff
@@ -0,0 +1,112 @@
...
+        tag_pattern = tag_pattern + f"({sep}[a-z].*)?"
+        compiled_pattern = re.compile(tag_pattern, flags=re.IGNORECASE)
+        # Generate tags to match the generated version.
+        for _ in range(TAG_ITERATIONS):
+            tag = _data.draw(hypothesis.strategies.from_regex(compiled_pattern, fullmatch=True))
+            # Perform the match.
+            match = pattern.match(tag)
+            assert match
+    except ValueError as error:
+        logger.debug(error)
```
behnazh-w, line 112: This try-except block seems too broad and I'm not sure I understand it. Which statements did you expect to throw a `ValueError`?

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+
+def _get_tag_commit(tag: TagReference) -> Commit | None:
+    """Return the commit of the passed tag.
+
+    This is a standalone function to more clearly handle the potential error raised by accessing the tag's commit
+    property.
+    """
+    try:
+        return tag.commit
+    except ValueError:
```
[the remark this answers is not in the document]
nathanwn: `commit` here is not an attribute but a property method that may raise `ValueError`. We discussed it [here](https://github.com/oracle/macaron/pull/508#discussion_r1379555532).

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+
+    tag_name = matched_tags[0]
+    tag = valid_tags[tag_name]
+    if not tag:
+        # Tag names are taken from valid_tags and should always exist within it.
+        logger.debug("Missing tag name from tag dict: %s not in %s", tag_name, valid_tags.keys())
+
+    branch_name = _get_branch_of_commit(git_obj.get_commit_from_tag(tag_name))
+    if not branch_name:
+        logger.debug("No valid branch associated with tag (commit): %s (%s)", tag_name, tag.commit.hexsha)
```
behnazh-w, line 238: What if `tag.commit.hexsha` raises a `ValueError`?

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+        return "", ""
+
+    logger.debug(
+        "Found tag %s with commit %s of branch %s for artifact version %s@%s",
+        tag,
+        tag.commit.hexsha,
```
behnazh-w, line 244: Same comment regarding catching `ValueError`.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+        branch_name,
+        name,
+        version,
+    )
+    return branch_name, tag.commit.hexsha
```
behnazh-w, line 249: Same comment regarding catching `ValueError`.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+    Returns
+    -------
+    tuple[str, str]
+        The branch name and digest as a tuple.
+    """
+    # A commit hash is 40 characters in length, but commits are often referenced using only some of those.
+    commit: Commit | None = None
+    if 7 <= len(version) <= 40:
+        try:
+            commit = git_obj.get_commit(version)
```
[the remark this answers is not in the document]
benmss: I have added a hex pattern check for treating the version as a commit string, but I have left it out for when it is treated as a tag due to the open set of characters that can be present in that case. However, I have added a regex escape command that should alleviate concerns there.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+        logger.debug("Best match: %s", matched_tags[0])
+        logger.debug("Up to 5 others: %s", matched_tags[1:6])
+
+    tag_name = matched_tags[0]
+    tag = valid_tags[tag_name]
+    if not tag:
+        # Tag names are taken from valid_tags and should always exist within it.
+        logger.debug("Missing tag name from tag dict: %s not in %s", tag_name, valid_tags.keys())
+
+    branch_name = _get_branch_of_commit(git_obj.get_commit_from_tag(tag_name))
```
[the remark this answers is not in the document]
benmss: It seems GitHub [recommends](https://docs.github.com/en/get-started/using-git/dealing-with-special-characters-in-branch-and-tag-names): a-z A-Z 0-9 . - _ /
However, that's probably a bit too restrictive. I have seen '#' used, but also if we want to support repositories maintained by non-English speakers.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+    version = purl.version
+    if version is None:
+        logger.debug("Missing version for analysis target: %s", purl.name)
+        return "", ""
+
+    available_domains = [git_service.hostname for git_service in GIT_SERVICES if git_service.hostname]
+    domain = to_domain_from_known_purl_types(purl.type) or (purl.type if purl.type in available_domains else None)
+    if domain:
+        # PURL is a repository type.
+        return get_commit_from_purl(git_obj, version)
```
nathanwn, line 132: I think this part can be extracted into a function to check if the PURL/PURL type is a repository type, which can be unit-tested in isolation by itself.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+    purl: PackageURL
+        The PURL of the analysis target.
+
+    Returns
+    -------
+    tuple[str, str]
+        The branch name and digest as a tuple.
+    """
+    version = purl.version
+    if version is None:
```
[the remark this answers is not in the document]
nathanwn: From the implementation of the `PackageURL` library ([here](https://github.com/package-url/packageurl-python/blob/0d3336804ce6dac59975c8c170d4015149442a93/src/packageurl/__init__.py#L161)), I don't think `purl.version` could be an empty string because they do some "normalization" on the value of `version` before returning the result from the `PackageURL.from_string` function, which converts any falsy value to `None`.
If we want to be extra careful, then I guess a falsy check `if version` instead of `if version is None`, as Behnaz suggested, may be a good idea.

src/macaron/repo_finder/commit_finder.py
nathanwn, line 141: Shouldn't the name of the function be 
```suggestion
def get_commit_from_purl_version(git_obj: Git, purl_version: str) -> tuple[str, str]:
```
instead? Because a full PURL is not passed in there, but only the version part of a PURL.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+    name: str
+        The name string.
+    version: str
+        The version string.
+
+    Returns
+    -------
+    tuple[Pattern | None, list[str], bool]
+        The tuple of the regex pattern that will match the version, the list of version parts that were extracted, and
+        whether the version string has a non-numeric suffix.
```
nathanwn, line 266: Nit: It may be helpful to explain the fail case, i.e. the case where the first element (the regex) is `None` instead of `Pattern`.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+    ----------
+    git_obj: Git
+        The repository.
+    purl: PackageURL
+        The PURL of the analysis target.
+
+    Returns
+    -------
+    tuple[str, str]
+        The branch name and digest as a tuple.
```
nathanwn, line 121: Nit: It may be helpful to mention the return values in the fail case (empty strings).

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+    """Return items of the passed tag list that match the passed artifact name and version.
+
+    Parameters
+    ----------
+    tag_list: list[str]
+        The list of tags to check.
+    artifact_name: str
+        The name of the artifact to match.
+    artifact_version: str
+        The version of the artifact to match.
```
nathanwn, line 351: The only place I see this `match_tags` function being called is in the `get_commit_from_version` function, where the `name` and `version` params here are documented as "the name/version of the analysis target" rather than "of the artifact". I think it is better to be consistent when naming/documenting these parameters.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,509 @@
...
+        return "", ""
+
+    available_domains = [git_service.hostname for git_service in GIT_SERVICES if git_service.hostname]
+    domain = to_domain_from_known_purl_types(purl.type) or (purl.type if purl.type in available_domains else None)
+    if domain:
+        # PURL is a repository type.
+        return get_commit_from_purl(git_obj, version)
+    if purl.type in repo_finder_deps_dev.SUPPORTED_TYPES:
+        # PURL is a package manager type.
+        return get_commit_from_version(git_obj, purl.name or "", version)
```
nathanwn, line 135: I have some questions about the `purl.name or ""` part on this line:
* Can `purl.name` be falsy? Have you encountered any specific case where `purl.name` here is falsy, and the empty string `""` is passed in instead? From my reading of the PackageURL library [here](https://github.com/package-url/packageurl-python/blob/main/src/packageurl/__init__.py), the name attribute seems to be always a non-empty string.
* What are the reasons for the empty string `""` to be passed in here, and what are the implications of passing the empty string as PURL name into the `get_commit_from_version` function? I am asking because I do not see the empty string case being handled in the `get_commit_from_version` function, as well as the functions called by it.

src/macaron/repo_finder/commit_finder.py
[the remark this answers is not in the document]
behnazh-w: Because `PackageURL.from_string` makes no guarantee that it will not return an empty string as part of the interface, its behavior can change in future. We need to put a guard and protect our code in this case.

src/macaron/repo_finder/commit_finder.py
benmss: Indeed, it seems it is not needed. Thank you for pointing it out.

src/macaron/repo_finder/commit_finder.py
benmss: After more testing I think the only way this can trigger is if the version is `None`. I've added a check to prevent that from happening at this stage.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,549 @@
...
+    version = purl.version
+    if not version:
+        logger.debug("Missing version for analysis target: %s", purl.name)
+        return "", ""
+
+    repo_type = abstract_purl_type(purl)
+    if repo_type == PurlType.REPOSITORY:
+        return extract_commit_from_version(git_obj, version)
+    if repo_type == PurlType.ARTIFACT:
+        return find_commit_from_version_and_name(git_obj, re.escape(purl.name), version)
```
nathanwn, line 145: I don't think the `re.escape` call should be here, because it seems more like an implementation details of `find_commit_from_version_and_name` or some other function lower down the call stack, than an assumption on the interface. From my brief look, I assume the `re.escape` should happen somewhere in `match_tags` function.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -0,0 +1,549 @@
...
+    """
+    version = purl.version
+    if not version:
+        logger.debug("Missing version for analysis target: %s", purl.name)
+        return "", ""
+
+    repo_type = abstract_purl_type(purl)
+    if repo_type == PurlType.REPOSITORY:
+        return extract_commit_from_version(git_obj, version)
+    if repo_type == PurlType.ARTIFACT:
```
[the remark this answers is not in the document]
benmss: I was sort of referring to it that way in some places already, so I think officially changing it is a good idea.

nathanwn (approved):
I think it's an LGTM for now. Thanks for your work.

behnazh-w (approved):

PR #509, opened 2023-10-10
chore: fix type correctness and remove many "type: ignore" directives

nicallen:


src/macaron/database/table_definitions.py
```diff
@@ -287,7 +287,7 @@ class Repository(ORMBase):
     #: The path to the repo on the file system.
     fs_path: Mapped[str] = mapped_column(String, nullable=False)
 
-    def __init__(self, *args, files: list[str] | None = None, **kwargs):  # type: ignore[no-untyped-def]
+    def __init__(self, *args: Any, files: list[str] | None = None, **kwargs: Any):
```
behnazh-w, line 290: I don't think `Any` is better than `# type: ignore` here. What I didn't know at the time I suppressed the type error here was that it's recommended to handle `args` and `kwargs` using `ParamSpec`: https://docs.python.org/3/library/typing.html#typing.ParamSpec

nicallen: It's slightly better in that at least it is requiring the `files` argument to be typed (which it already was).
I hadn't seen `ParamSpec`, but I think it is mainly relevant in expressing forwarding args/kwargs to a passed callable where the function is generic on the callable's signature, which is not the case here, where they are being passed to the superclass __init__ whose types are fixed, so args/kwargs should simply be typed with the correct element type, which in this case cannot be more precise than Any because the superclass signature is `def __init__(self, **kw: Any)`.

src/macaron/util.py
```diff
@@ -254,8 +254,10 @@ def get_if_exists(doc: JsonType, path: list[str | int]) -> JsonType | None:
     """Get a json dict value if it exists."""
     while len(path) > 0:
         this = path.pop(0)
-        if isinstance(doc, (dict, list)) and this in doc:
-            doc = doc[this]  # type: ignore
+        if isinstance(this, str) and isinstance(doc, dict) and this in doc:
+            doc = doc[this]
+        elif isinstance(this, int) and isinstance(doc, list) and this in doc:
```
tromai, line 259: Hmm I think in this branch, we are trying to extract a nth element from `doc` if `doc` is a list. 
I don't think we should run `this in doc` as this means that if `this` is contained in `doc` (a list can contains an int). I guess we should use `this < len(doc)` instead ? :thinking:

nicallen: You are correct, and this is the same logic as previously, which means it was always incorrect for lists. I guess if this function is not currently being used then that wasn't noticed. Regardless, I have fixed it to check if the index is within the range of the list.

src/macaron/database/table_definitions.py
behnazh-w: > which in this case cannot be more precise than Any because the superclass signature is `def __init__(self, **kw: Any)`.

Huh! I had assumed that the `__init__` function in the `DeclarativeBase` class would use type parameters, but maybe that's tricky because `ParamSpec` was introduces in Python 3.10 and SQLAlchemy needs to be compatible with older Python versions, hence using `Any`.

src/macaron/slsa_analyzer/checks/check_result.py
```diff
@@ -37,7 +39,7 @@ class CheckResult(TypedDict):
     justification: list[str | dict[str, str]]
     # human_readable_justification: str
     # result_values: dict[str, str | float | int] | list[dict[str, str | float | int]]
-    result_tables: list[DeclarativeBase | Table]
+    result_tables: list[DeclarativeBase | Table | Expectation]
```
behnazh-w, line 42: `Table` is a left over from an old implementation and can be removed too.

nicallen: Removed

behnazh-w (approved):

PR #512, opened 2023-10-12
fix: resolve podman compatibility issues

nathanwn:
This PR adds support for Podman as an alternative container engine to run the Macaron image.

The current `run_macaron.sh` script is not fully compatible with Podman, due to known cases where Docker and Podman behave differently. Changes in this PR address these cases.

## Volume-mounting non-existing directories on host

Podman completely bans volume-mounting non-existing directories on the host into the container. See https://github.com/containers/podman/issues/6234 for more details.

Meanwhile, mounting a non-existing directory on the host into a container is allowed in Docker. There is a peculiar behavior: the non-existing directory is owned by `root` both inside and outside the container.

**Solution**: Before volume-mounting a directory, we can either (1) create that directory if it does not exist, or (2) error. The choice should be consistent with how the Macaron Python package behaves. 

## UID mapping and Volume mount owner

By default
* `docker run` maps the host user `$UID` to a user with the same `$UID` in the container.
* `podman run` maps the host user `$UID` to the `root` user in the container.

**Solution**: To make sure Podman behaves exactly like docker w.r.t. volume mount owner, we can set the environment variable `PODMAN_USERNS` to `keep-id`. For more details, see https://docs.podman.io/en/v4.4/markdown/options/userns.container.html#userns-mode.

Example (Note that in the following example, you must create the `$PWD/d` directory on host beforehand):

```bash
$ id -u
1000

# Case 1: docker run with volume mount
$ docker run --rm -it -v $PWD/d:/tmp/d localhost/myimage:latest bash -c "ls -ln /tmp"
total 4 drwxrwxr-x 2 1000 1000 4096 Sep 17 16:24 d

# Case 2: podman run with volume mount and `--userns` not set
$ podman run --rm -it -v $PWD/d:/tmp/d localhost/myimage:latest bash -c "ls -ln /tmp"
total 4 drwxrwxr-x 2 0 0 4096 Sep 17 16:24 d

# Case 3: podman run with volume mount and`--userns=keep-id`
$ podman run --rm -it --userns=keep-id -v $PWD/d:/tmp/d localhost/myimage:latest bash -c "ls -ln /tmp"
total 4 drwxrwxr-x 2 1000 1000 4096 Sep 17 16:24 d
```

## Mount option `:Z`

At the moment, when the Macaron container starts up, the UID of the user `macaron` in the container gets changed to match the UID of the user on the host. This is done with the `usermod` command. Consequently, the owner UID of the `/home/macaron` directory in the container gets changed.

There has not been any issue with Docker so far. However, Podman errors in cases where we mount any volume under `/home/macaron` in the container without the `:Z` option.

Here is how to reproduce.

* We have the following `Dockerfile`.

```Dockerfile
FROM <base image>

RUN : \
    && groupadd --gid 43147 macaron \
    && useradd --uid 43147 --create-home --gid 43147 macaron

USER root
```

* Build this image and give it the name `localhost/foobar`:

```bash
$ podman build . -t localhost/foobar
```

* Create a subdirectory `d` in the current working directory and mount it into the container.

```bash
$ podman run --rm -ti -v $PWD/d:/home/macaron/d localhost/foobar bash
```

* Run the following commands in the container:

```
[root@26dab0c2dee7 /]# groupmod --non-unique --gid 1000 macaron
[root@26dab0c2dee7 /]# usermod --non-unique --uid 1000 --gid 1000 macaron
usermod: Failed to change ownership of the home directory[root@26dab0c2dee7 /]#
```

* Now try `docker run` again, this time with the `:Z` mount option. The `usermod` error should be gone.

**Solution**: For all volume mounts under `/home/macaron`, we need to provide the `:Z` mount option to tell Podman that the volume is not shared with any other container, and thus modifying the owner of `/home/macaron` is safe.

### DIFF
--- a/scripts/release_scripts/run_macaron.sh
+++ b/scripts/release_scripts/run_macaron.sh
@@ -43,6 +43,7 @@ if [[ "${BASH_VERSINFO[0]}" -lt "4" ]]; then
     log_warning "Using bash version >=4 is recommended."
 fi
 
+# Determine the Macaron image tag.
 if [[ -z ${MACARON_IMAGE_TAG:-} ]]; then
     MACARON_IMAGE_TAG="latest"
 fi
@@ -104,8 +105,14 @@ function to_absolute_path() {
 # With the `set -e` option turned on, this function exits the script with
 # return code 1 if the directory does not exist.
 function assert_dir_exists() {
-    if [[ ! -d "$1" ]]; then
-        log_err "Directory $1 of argument $2 does not exist."
+    path=$1
+    arg_name=$2
+    if [[ ! -d "$path" ]]; then
+        if [[ -z "${arg_name:-}" ]]; then
+            log_err "Directory $path does not exist."
+        else
+            log_err "Directory $path of argument $arg_name does not exist."
+        fi
         return 1
     fi
 }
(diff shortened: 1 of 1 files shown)


scripts/release_scripts/run_macaron.sh
```diff
@@ -124,6 +131,68 @@ function assert_path_exists() {
...
+# Add a file to the list of volume mounts.
+#
+# Arguments:
+#   $1: The macaron argument from which the file is passed into this script.
+#   $2: The path to the file on the host.
+#   $3: The path to the file inside the container.
+#   $4: Mount option. Note: this MUST be either `ro,Z` for readonly volumes,
+#       or `rw,Z` otherwise.
+# Globals:
+#   mounts: the volume mount is added to this array
```
tromai, line 184: nit: 
```suggestion
#   mounts: the array where volume mount is added.
```

tromai (approved):

scripts/release_scripts/run_macaron.sh
```diff
@@ -124,6 +131,68 @@ function assert_path_exists() {
...
+# Add a directory to the list of volume mounts.
+#
+# Arguments:
+#   $1: The macaron argument from which the directory is passed into this script.
+#   $2: The path to the directory on the host.
+#   $3: The path to the directory inside the container.
+#   $4: Mount option. Note: this MUST be either `ro,Z` for readonly volume mounts,
+#       or `rw,Z` otherwise.
+# Globals:
+#   mounts: the volume mount is added to this array
```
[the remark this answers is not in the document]
nathanwn: Here I didn't really mean to explain what `mounts` is but rather the effect of the function on the global array `mounts`.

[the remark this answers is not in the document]
nathanwn: See 1f00686.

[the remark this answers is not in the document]
tromai: The change looks good. I will close this conversation.

scripts/release_scripts/run_macaron.sh
```diff
@@ -140,6 +147,64 @@ function assert_path_exists() {
...
+#   $1: The directory to create.
+function create_dir_if_not_exists() {
+    dir=$1
+    if [ ! -d "$dir" ]; then
+        mkdir --parents "$dir"
+    fi
+}
+
+# Add a directory to the list of volume mounts stored in the ``mounts`` global variable.
+#
```
behnazh-w, line 161: I think it would be good to rename the function to indicate that it can potentially create a directory.

In addition, please add a note here that this function can have side effect, i.e., if the mounted directory does not exist, it will create it.

scripts/release_scripts/run_macaron.sh
```diff
@@ -140,6 +147,64 @@ function assert_path_exists() {
...
+# Arguments:
+#   $1: The macaron argument from which the directory is passed into this script.
+#   $2: The path to the directory on the host.
+#   $3: The path to the directory inside the container.
+#   $4: Mount option. Note: this MUST be either `ro,Z` for readonly volume mounts,
+#       or `rw,Z` otherwise.
+function mount_dir() {
```
behnazh-w, line 168: It would be good to pass an option to create the directory, which is disabled by default. This would make us explicitly enable directory creation at callsites. Otherwise, it is easy to call this function thinking we are just mounting a directory, but end up creating it on the host by mistake.

nathanwn: In b8babbae18e20bedc7568d62a894227a2d2fcc0e, I create 3 functions for mounting dirs:
* `mount_dir_ro`
* `mount_dir_rw_allow_create`
* `mount_dir_rw_forbid_create`
I think the last 2 are more readable than having an additional param and calling the `mount_dir` function like this:
```
mount_dir_rw [...] 0
mount_dir_rw [...] 1
```
where `0` and `1` are values for whether directory creation is allowed or not.

behnazh-w (approved):

PR #513, opened 2023-10-13
refactor: refactor interface of base check

nicallen:
Refactor BaseCheck.run_check to return all result info as part of return value, rather than passing in and mutating a CheckResult object containing other data irrelevant to run_check.
Refactor SLSA requirement status information in AnalyzeContext to avoid unnecessary replication of generic info about a SLSA requirement, which can be looked up by ReqName when needed instead.

### DIFF
--- a/src/macaron/slsa_analyzer/analyze_context.py
+++ b/src/macaron/slsa_analyzer/analyze_context.py
@@ -180,7 +180,7 @@ class AnalyzeContext:
         # Remove result_tables since we don't have a good json representation for them.
         sorted_on_id = []
         for res in _sorted_on_id:
-            # res is CheckResult(TypedDict)
+            # res is CheckResult
             res_dict: dict = dict(res.get_dict())
             res_dict.pop("result_tables")
             sorted_on_id.append(res_dict)
--- a/src/macaron/slsa_analyzer/analyze_context.py
+++ b/src/macaron/slsa_analyzer/analyze_context.py
@@ -181,7 +181,7 @@ class AnalyzeContext:
         sorted_on_id = []
         for res in _sorted_on_id:
             # res is CheckResult
-            res_dict: dict = dict(res.get_dict())
+            res_dict: dict = dict(res.get_summary())
             res_dict.pop("result_tables")
             sorted_on_id.append(res_dict)
         sorted_results = sorted(sorted_on_id, key=lambda item: item["result_type"], reverse=True)
--- a/src/macaron/slsa_analyzer/analyze_context.py
+++ b/src/macaron/slsa_analyzer/analyze_context.py
@@ -16,7 +16,7 @@ from macaron.slsa_analyzer.git_service import BaseGitService
 from macaron.slsa_analyzer.git_service.base_git_service import NoneGitService
 from macaron.slsa_analyzer.levels import SLSALevels
 from macaron.slsa_analyzer.provenance.expectations.expectation import Expectation
-from macaron.slsa_analyzer.slsa_req import ReqName, SLSAReq, get_requirements_dict
+from macaron.slsa_analyzer.slsa_req import ReqName, SLSAReqStatus, create_requirement_status_dict
 from macaron.slsa_analyzer.specs.build_spec import BuildSpec
 from macaron.slsa_analyzer.specs.ci_spec import CIInfo
 from macaron.slsa_analyzer.specs.package_registry_spec import PackageRegistryInfo
@@ -64,7 +64,7 @@ class AnalyzeContext:
             The output dir.
         """
         self.component = component
-        self.ctx_data: dict[ReqName, SLSAReq] = get_requirements_dict()
+        self.ctx_data: dict[ReqName, SLSAReqStatus] = create_requirement_status_dict()
 
         self.slsa_level = SLSALevels.LEVEL0
         # Indicate whether this repo fully reach a level or
@@ -176,14 +176,14 @@ class AnalyzeContext:
 
     def get_dict(self) -> dict:
         """Return the dictionary representation of the AnalyzeContext instance."""
-        _sorted_on_id = sorted(self.check_results.values(), key=lambda item: item["check_id"])
+        _sorted_on_id = sorted(self.check_results.values(), key=lambda item: item.check.check_id)
         # Remove result_tables since we don't have a good json representation for them.
         sorted_on_id = []
         for res in _sorted_on_id:
             # res is CheckResult(TypedDict)
-            res_copy: dict = dict(res.copy())
-            res_copy.pop("result_tables")
-            sorted_on_id.append(res_copy)
+            res_dict: dict = dict(res.get_dict())
+            res_dict.pop("result_tables")
+            sorted_on_id.append(res_dict)
         sorted_results = sorted(sorted_on_id, key=lambda item: item["result_type"], reverse=True)
         check_summary = {
             result_type.value: len(result_list) for result_type, result_list in self.get_check_summary().items()
@@ -219,7 +219,7 @@ class AnalyzeContext:
         result: dict[CheckResultType, list[CheckResult]] = {result_type: [] for result_type in CheckResultType}
 
         for check_result in self.check_results.values():
-            match check_result["result_type"]:
+            match check_result.result.result_type:
                 case CheckResultType.PASSED:
                     result[CheckResultType.PASSED].append(check_result)
                 case CheckResultType.SKIPPED:
@@ -243,8 +243,8 @@ class AnalyzeContext:
             output = "".join(
                 [
                     output,
-                    f"Check {check_result['check_id']}: {check_result['check_description']}\n",
-                    f"\t{check_result['result_type'].value}\n",
+                    f"Check {check_result.check.check_id}: {check_result.check.check_description}\n",
+                    f"\t{check_result.result.result_type.value}\n",
                 ]
             )
 
--- a/src/macaron/slsa_analyzer/checks/base_check.py
+++ b/src/macaron/slsa_analyzer/checks/base_check.py
@@ -48,16 +48,34 @@ class BaseCheck:
         result_on_skip : CheckResultType
             The status for this check when it's skipped based on another check's result.
         """
-        self.check_info = CheckInfo(
+        self._check_info = CheckInfo(
             check_id=check_id, check_description=description, eval_reqs=eval_reqs if eval_reqs else []
         )
 
         if not depends_on:
-            self.depends_on = []
+            self._depends_on = []
         else:
-            self.depends_on = depends_on
+            self._depends_on = depends_on
 
-        self.result_on_skip = result_on_skip
+        self._result_on_skip = result_on_skip
+
+    @property
+    def check_info(self) -> CheckInfo:
+        """Get the information identifying/describing this check."""
+        return self._check_info
+
+    @property
+    def depends_on(self) -> list[tuple[str, CheckResultType]]:
+        """Get the list of parent checks that this check depends on.
+
+        Each member of the list is a tuple of the parent's id and the status of that parent check.
+        """
+        return self._depends_on
+
+    @property
+    def result_on_skip(self) -> CheckResultType:
+        """Get the status for this check when it's skipped based on another check's result."""
+        return self._result_on_skip
 
     def run(self, target: AnalyzeContext, skipped_info: SkippedInfo | None = None) -> CheckResult:
         """Run the check and return the results.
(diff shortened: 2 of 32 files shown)


src/macaron/slsa_analyzer/checks/base_check.py
```diff
@@ -7,18 +7,22 @@
...
 logger: logging.Logger = logging.getLogger(__name__)
 
 
 class BaseCheck:
     """This abstract class is used to implement Checks in Macaron."""
 
-    # The dictionary that contains the data of all SLSA requirements.
-    SLSA_REQ_DATA = get_requirements_dict()
-
     def __init__(
```
nathanwn, line 26: The docstring of this constructor needs updating.

src/macaron/slsa_analyzer/checks/check_result.py
```diff
@@ -24,25 +29,54 @@ class CheckResultType(str, Enum):
...
     check_id: str
     check_description: str
-    # The string representations of the slsa requirements and their
-    # corresponding slsa level.
-    slsa_requirements: list[str]
+    eval_reqs: list[ReqName]
+
+
+@dataclass
+class CheckResultData:
```
nathanwn, line 42: Adding docstrings for these classes may be a good idea.
You can follow [instructions here](https://oracle.github.io/macaron/pages/developers_guide/style_guide.html#docstrings-for-classes) to write docstrings that are compatible with our Sphinx setup (the second bullet point is for `dataclass` classes).

tests/slsa_analyzer/test_slsa_requirements.py
```diff
@@ -7,41 +7,24 @@
 
 from unittest import TestCase
 
-from macaron.slsa_analyzer.levels import SLSALevels
-from macaron.slsa_analyzer.slsa_req import BUILD_REQ_DESC, Category, SLSAReq, get_requirements_dict
+from macaron.slsa_analyzer.slsa_req import SLSAReqStatus
 
 
 class TestSLSARequirements(TestCase):
```
nathanwn, line 13: It's probably a good idea to also convert this `unittest` class into single test functions since we are moving away from `unittest`.

src/macaron/slsa_analyzer/checks/base_check.py
nicallen: What needs updating? The parameters have not changed, and the docstring still seems correct to me.
Only thing I think we could add would be declarations for the class' fields, since they used to be 1-1 with the constructor parameters but now are not.

nathanwn: Oh, sorry nvm. My mistake. I think it was decided that we don't document class fields in this case. ([Our style guide](https://oracle.github.io/macaron/pages/developers_guide/style_guide.html#docstrings-for-classes))

nathanwn:
You probably need to regenerate the API doc with `make docs-api` when the PR is finalized.

src/macaron/slsa_analyzer/checks/base_check.py
nicallen: In this case the class fields are not private, they are accessed by registry and runner, so according to the style guide I suppose they should be made into properties and should be documented.
They are also accessed by subclasses, and creating additional subclasses of the BaseCheck class is intended to be an extension point for Macaron, so in that sense they are part of the "public interface", and so it seems worthwhile documenting them for that reason alone (though if they were only intended to be accessible to subclasses and nothing else, the style guide doesn't specify whether we should do anything special for such a case of a "protected" field).

nathanwn: That sounds reasonable.

behnazh-w:
@nicallen It would be good to add a [Developer Docs page](https://oracle.github.io/macaron/pages/developers_guide/apidoc/index.html) to describe the check interface. We already have a [`README.md`](https://github.com/oracle/macaron/blob/0f426982837708539fb64b76001b823b79596b0e/src/macaron/slsa_analyzer/checks/README.md) through to help contributors add a new check. To avoid duplication, we can move the content of the `README.md` to the RST page and add a link to the `README.md`?

behnazh-w:
@nicallen I think it makes sense to change the commit scope to `refactor` so that it gets added to the release changelog.

behnazh-w:
> You probably need to regenerate the API doc with `make docs-api` when the PR is finalized.

Sounds good. @nicallen please rebase on `staging` to include [the latest](https://github.com/oracle/macaron/pull/525) docs updates.

src/macaron/slsa_analyzer/checks/check_result.py
```diff
@@ -24,25 +29,54 @@ class CheckResultType(str, Enum):
...
+    justification: Justification
     # human_readable_justification: str
     # result_values: dict[str, str | float | int] | list[dict[str, str | float | int]]
-    result_tables: list[DeclarativeBase | Expectation]
+    result_tables: ResultTables
     # recommendation: str
     result_type: CheckResultType
 
 
+@dataclass
```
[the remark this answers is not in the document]
nicallen: True, better for them to be immutable. Dataclasses can be defined as immutable with `@dataclass(frozen=True)`, so I think it is preferable to do that rather than use `NamedTuple`.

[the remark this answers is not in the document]
behnazh-w: Curious why you prefer `@dataclass(frozen=True)` over `NamedTuple`? One nice thing about `NamedTuple` is that it supports unpacking and the types **must** be specified. That seems particularly useful for `CheckInfo` and `CheckResultData`.

src/macaron/slsa_analyzer/checks/check_result.py
```diff
@@ -24,25 +29,54 @@ class CheckResultType(str, Enum):
...
+        dict
+        """
+        return {
+            "check_id": self.check.check_id,
+            "check_description": self.check.check_description,
+            "slsa_requirements": [str(BUILD_REQ_DESC.get(req)) for req in self.check.eval_reqs],
+            "justification": self.result.justification,
+            "result_tables": self.result.result_tables,
+            "result_type": self.result.result_type,
+        }
```
behnazh-w, line 77:
```suggestion

    def get_dict(self) -> dict:
        """Get a flattened dictionary representation for this CheckResult.
        Returns
        -------
        dict
        """
        return self.check.asdict() | self.result.asdict()
```

tromai: I don't think `asdict(self.check) | asdict(self.result)` would be equivalent to the original implementation because it don't handle this logic `"slsa_requirements": [str(BUILD_REQ_DESC.get(req)) for req in self.check.eval_reqs],` in the original implementation. Are we meant to remove this from the returned dictionary?

nicallen: Yes, that would change the behaviour. The purpose of this method is to produce the output report format representation, which isn't necessarily the same as an `asdict` of all of the fields. Even if it was the same, being explicit about the output-formatted dict being created makes it easier to change or refactor the representation of this class without accidentally changing the user-visible report format.

src/macaron/slsa_analyzer/checks/check_result.py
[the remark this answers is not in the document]
nicallen: I read this (https://peps.python.org/pep-0557/#why-not-just-use-typing-namedtuple) and it seemed that `dataclass` was the more strongly-typed option, since `NamedTuple` falls back to just being a tuple in various cases.
As far as I can tell, you do have to declare all the fields (with types) for a `dataclass(frozen=True)`, since you need to provide types to declare the field, and even if you try and define an explicit constructor to avoid that it won't let you assign self fields.

src/macaron/slsa_analyzer/checks/check_result.py
behnazh-w: That could be an indication of unnecessary data being stored in the results and or with an undesirable type. I would avoid duplicating data and adjust the original interface instead, i.e., the attributes in `CheckInfo` and `CheckResultData`.

src/macaron/slsa_analyzer/checks/check_result.py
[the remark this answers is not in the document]
behnazh-w: Yes, looks like `dataclass` also requires the attribute to be typed (explicitly or trough assignment).  I don't think one is better than the other in terms of typing. See examples below, both allowing a `str` value being assigned to an `int` attribute: :disappointed:  

```bash
>>> class A(NamedTuple):
...     foo: int
>>> a = A(foo="foo")
>>> type(a.foo)
<class 'str'>
```
vs
```bash
>>> @dataclass(frozen=True)
... class B():
...     foo: int
>>> b = B(foo="foo")
>>> type(b.foo)
<class 'str'>
```
So at this point, I guess there is no big difference.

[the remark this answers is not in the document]
nicallen: Yes, the static types are not enforced at runtime, but that's true of all python static typing. Mypy gives a type error in both cases, so it's as type-safe as python can be.

src/macaron/slsa_analyzer/checks/check_result.py
nicallen: The conversion here is from `eval_reqs` which is a `list[ReqName]` (i.e. list of references to the SLSA requirements, referred to via their identifer enum value) to what the output format expects, which is `slsa_requirements`, containing a list of the SLSA requirements as a textual description. Originally, this conversion happened in the constructor for `CheckResult`, however, I removed it from there because repeatedly storing the SLSA req textual description for each one is redundant since with a ReqName you can look up the description at any time, so there's no need for `CheckResult` to denormalise the data in this way. However, the existing output format has the info in that form (whether this is desirable is another question, the it repeats the same strings but makes the report easier to read, and in any case these changes were intended to be a refactoring and I was not proposing any changes to the reporting format), so the conversion had to happen at some point, so I moved it later to `get_dict` which I considered to be part of the "output layer".

In general, I don't think it is necessarily desirable that the internal representation of results must exactly match the format produced for the reports, having an explicit report generation step that is decoupled from the internal representation makes it easier to have alternative reporting formats (for example, the verification summary could end up being an alternative report format).

behnazh-w: Agreed that changing the output format can be out of scope for this PR, but I still prefer to clean the data and avoid duplication as much as possible (maybe in future) unless really necessary.

Can you please rename the function from `get_dict()` to `get_summary()` and adjust the docstring to reflect the intent you discuss here?

src/macaron/slsa_analyzer/checks/check_result.py
nicallen: I've added doc comments for the fields.

tests/slsa_analyzer/test_slsa_requirements.py
nicallen: Done.

src/macaron/slsa_analyzer/checks/check_result.py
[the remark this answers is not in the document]
nicallen: I have changed them to `@dataclass(frozen=True)`.

src/macaron/slsa_analyzer/analyze_context.py
```diff
@@ -176,14 +176,14 @@ def get_slsa_level_table(self) -> SLSALevel:
 
     def get_dict(self) -> dict:
         """Return the dictionary representation of the AnalyzeContext instance."""
-        _sorted_on_id = sorted(self.check_results.values(), key=lambda item: item["check_id"])
+        _sorted_on_id = sorted(self.check_results.values(), key=lambda item: item.check.check_id)
         # Remove result_tables since we don't have a good json representation for them.
         sorted_on_id = []
         for res in _sorted_on_id:
             # res is CheckResult(TypedDict)
```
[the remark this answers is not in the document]
nicallen: Done.

src/macaron/slsa_analyzer/checks/check_result.py
nicallen: Done.

nicallen:
> @nicallen It would be good to add a [Developer Docs page](https://oracle.github.io/macaron/pages/developers_guide/apidoc/index.html) to describe the check interface. We already have a [`README.md`](https://github.com/oracle/macaron/blob/0f426982837708539fb64b76001b823b79596b0e/src/macaron/slsa_analyzer/checks/README.md) though to help contributors add a new check. To avoid duplication, we can move the content of the `README.md` to the RST page and add a link to the `README.md`?

I have created an separate issue for this (#535), I don't think we need to do this immediately as part of this PR.

behnazh-w (approved):

nathanwn (approved):

PR #524, opened 2023-10-19
chore: improve cyclonedx-mvn bom file check to include files with cus…

### LINKED ISSUE #518: Support custom named cyclonedx maven bom output
Some artifacts use a custom name for the SBOM output of cyclonedx-maven via the property `outputName`. This is either specified in the artifact's `pom.xml` or is inherited from a parent.
E.g. `apache/commons-io` with parent of `commons-parent`
```
<configuration>
<outputName>${project.artifactId}-${project.version}-bom</outputName>
</configuration>
```
Thereby creating the file: `commons-lang3-<version>-bom.json` instead of the default `bom.json` as Macaron expects.

Supplying the property with the default value in Macaron does not appear to override the inherited value.
An alternative solution is to accept all `json` files in the expected directory, and rely upon current validation to filter out any bad files.

Artifacts with custom directories in their cyclonedx configuration are not considered in this issue, due to the considerable amount of extra complexity that would add.

benmss:
Allows Macaron to discover `cyclonedx-mvn` SBOM files that have been created with a non-default name.
Does not provide support for custom directories or cases where multiple SBOM files end up in the expected `target` directory, as these would require POM parsing to properly detect.
Closes #518


src/macaron/dependency_analyzer/cyclonedx_mvn.py
```diff
@@ -64,11 +64,20 @@ def collect_dependencies(self, dir_path: str) -> dict[str, DependencyInfo]:
         """
         # Load the top level file separately as it has different content.
         top_path = Path(os.path.join(dir_path, "target", self.file_name))
+        top_path_altered = False
+        if not os.path.exists(top_path):
+            # If the expected bom file does not exist, allow other named .json files instead.
+            possible_paths = glob.glob(os.path.join(dir_path, "target", "*.json"))
+            if possible_paths:
+                top_path = Path(possible_paths[0])
```
tromai, line 72: If we are not going to support the case where multiple BOM files are generated into the `target` directory, should we exit on error here if `possible_paths` have multiple files, instead of picking the first element in `possible_paths`, which could not be the top BOM file we expect?

tromai:
Is it possible to have a unit test for this scenario?

tests/dependency_analyzer/cyclonedx/test_cyclonedx.py
```diff
@@ -106,3 +107,11 @@ def test_multiple_versions(snapshot: dict[str, DependencyInfo]) -> None:
     bom_path = Path(RESOURCES_DIR, "bom_multi_versions.json")
     result = get_deps_from_sbom(bom_path)
     assert snapshot == result
+
+
+def test_custom_sbom_name_with_maven() -> None:
+    """Test reading cyclonedx maven sbom that was created using a custom name."""
+    cyclonedx: CycloneDxMaven = CycloneDxMaven(
+        "", "bom.json", "maven", defaults.get("dependency.resolver", "dep_tool_maven"), "localhost"
```
tromai, line 115: nit: is there any reason why the `repo_path` attribute of `CycloneDxMaven` is `localhost`?

tests/dependency_analyzer/cyclonedx/test_cyclonedx.py
```diff
@@ -106,3 +107,11 @@ def test_multiple_versions(snapshot: dict[str, DependencyInfo]) -> None:
...
+    )
+    assert cyclonedx.collect_dependencies(str(RESOURCES_DIR))
```
tromai, line 117: Should we compare the actual value of the returned dictionary instead of checking whether the dictionary is not empty?

tromai (approved):
This PR looks good to me, only one questions left.

tests/dependency_analyzer/cyclonedx/test_cyclonedx.py
benmss: Not really. It's not used so perhaps an empty string is better.

tests/dependency_analyzer/cyclonedx/test_cyclonedx.py
benmss: We are really only testing that the plugin can find the json file when it has a custom name. Verifying the contents is outside of the scope I'd say.

behnazh-w:
Since the `apache/maven` repo that we analyze in the integration tests also [customizes](https://repo1.maven.org/maven2/org/apache/maven/maven-parent/40/maven-parent-40.pom) the name of the BOM file, it would make sense to change the pinned commit to the latest version and test the change this PR.

src/macaron/dependency_analyzer/cyclonedx_mvn.py
```diff
@@ -64,11 +64,23 @@ def collect_dependencies(self, dir_path: str) -> dict[str, DependencyInfo]:
...
+                    return {}
+                top_path = Path(possible_paths[0])
+                top_path_altered = True
 
         # Collect all the dependency files recursively.
         child_paths = [
             Path(path)
-            for path in glob.glob(os.path.join(dir_path, "**", "target", self.file_name), recursive=True)
+            for path in glob.glob(
+                os.path.join(dir_path, "**", "target", "*.json" if top_path_altered else self.file_name), recursive=True
```
tromai, line 82: What happen if there is multiple `*.json` files in the "children" `taget` directories, given that there exist a custom-named SBOM file in the top path?

behnazh-w (changes requested):
Because the BOM file can have a custom name, we also need to adjust the error message [here](https://github.com/oracle/macaron/blob/512b5aa3dc220a20dd83907d2569ba008aa9e9eb/src/macaron/dependency_analyzer/cyclonedx.py#L50) and avoid printing the default file name to avoid confusion:

```python
raise CycloneDXParserError(f"Unable to locate any BOM files at: {str(file_path.parent)}.")
```

behnazh-w (approved):

src/macaron/dependency_analyzer/cyclonedx_mvn.py
```diff
@@ -64,14 +64,34 @@ def collect_dependencies(self, dir_path: str) -> dict[str, DependencyInfo]:
...
+            )
             if Path(path) != top_path
         ]
 
+        # Ensure recursively found SBOMs are at most one per directory.
+        child_paths_set = set()
```
tromai, line 88: This set contains the parent directories of all children SBOM files. I think we could re-name it to something like `child_sbom_dir_names`

behnazh-w (approved):

src/macaron/dependency_analyzer/cyclonedx_mvn.py
```diff
@@ -52,6 +52,12 @@ def get_cmd(self) -> list:
     def collect_dependencies(self, dir_path: str) -> dict[str, DependencyInfo]:
         """Process the dependency JSON files and collect direct dependencies.
 
+        We allow the dependency JSON files to be accepted as long as there is only one JSON file in the target
+        directory. The name of the file is not considered because projects can be configured to produce a custom named
+        SBOM, which cannot be overridden if included at the parent POM level.
+        The presence of multiple JSON files within a target directory differs too greatly from the expectations of the
+        plugin's output. It is for this reason that an error is thrown in such cases.
```
tromai, line 59:
```suggestion
        plugin's output. It is for this reason that these cases are treated as error and this method will return an empty dictionary.
```
I think it's better to be explicitly on how the error are handled. We don't throw any error in this method, but only return an empty dictionary.

src/macaron/dependency_analyzer/cyclonedx_mvn.py
```diff
@@ -63,24 +69,19 @@ def collect_dependencies(self, dir_path: str) -> dict[str, DependencyInfo]:
...
-        if not os.path.exists(top_path):
-            # If the expected bom file does not exist, allow other named .json files instead.
-            possible_paths = glob.glob(os.path.join(dir_path, "target", "*.json"))
-            if possible_paths:
-                if len(possible_paths) > 1:
-                    logger.debug("Too many JSON SBOM files found. Expected: 1, Found: %s", len(possible_paths))
-                    return {}
-                top_path = Path(possible_paths[0])
-                top_path_altered = True
+        possible_paths = glob.glob(os.path.join(dir_path, "target", "*.json"))
```
behnazh-w, line 72: I preferred the previous version to first check for the expected file name. If that file cannot be found, we can look for other JSON files. This can avoid missing the BOM files when a project uses expected names but also produces other JSON files.

tromai (approved):

PR #528, opened 2023-10-25
fix: fix bash syntax error when running `run_macaron.sh` on MacOS

nathanwn:
The bash pattern matching syntax `@(...)` requires the shopt option `extglob` to be set. Although this option is set by default on newer versions of bash, it is not always the case. One known case is on MacOS, where the default bash version is `bash 3.2`. (even on the latest version of MacOS).

Reproduction: Run the following bash script in the `bash:3.2` docker container.

```
#!/usr/bin/env bash

if [[ "${DOCKER_PULL}" != @(always|missing|never) ]]; then
    echo "DOCKER_PULL must be one of: always, missing, never (default: always)"
    exit 1
fi
```

### DIFF
--- a/docs/source/pages/installation.rst
+++ b/docs/source/pages/installation.rst
@@ -10,7 +10,10 @@ Installation Guide
 -------------
 Prerequisites
 -------------
-- Installations of ``wget`` or ``curl`` and ``bash`` must be available and on the path.
+- Installations of ``wget`` or ``curl`` and ``bash`` must be available on ``PATH``.
+
+  - Macaron has been tested with ``bash 5.1.16(1)-release``.
+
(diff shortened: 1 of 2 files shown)


behnazh-w (commented):
It's good to add `shopt -s extglob` for older versions of `bash`, but  we should add a requirement for the bash version for running `run_macaron.sh`. In general running old bash versions is not a good idea for security reasons, etc. and it's better if people upgrade their bash version. Can you please add a check for the bash version and "recommend" to update it if it's below our supported version?

nathanwn:
> It's good to add `shopt -s extglob` for older versions of `bash`, but we should add a requirement for the bash version for running `run_macaron.sh`. In general running old bash versions is not a good idea for security reasons, etc. and it's better if people upgrade their bash version. Can you please add a check for the bash version and "recommend" to update it if it's below our supported version?

I have added a warning in efccf1d.

nathanwn:
A note about bash version has been added to our installation instructions in cd73e23.

behnazh-w (approved):

tromai (approved):

PR #529, opened 2023-10-26
chore: display the repository URL status of dependencies in html report instead of the analysis status

tromai:
# Description
This Pull request fixes the issues in the HTML report of the main target where the dependencies are all displayed as `AVAILABLE` even though the repository URL is not available for them - see example below:
```shell
macaron analyze -rp https://github.com/apache/maven -sbom test_sbom.json
```
with the content of `test_sbom.json` as follows: [test_sbom.json](https://github.com/oracle/macaron/files/13234713/test_sbom.json)

image

In this example, Macaron cannot find the repository URL for the dependency component `javax.inject:javax.inject`. However, Macaron still tries to run all checks on that component and produce an "empty" HTML report for it. This is part of the design implemented in this [PR](https://github.com/oracle/macaron/pull/165):
- Macaron still run the checks and produce a HTML report for all software component (including both the main target and dependencies if available) even when no repository URL is found for them (consistent with the data model where the [repository is an optional field of a software component](https://github.com/oracle/macaron/blob/1b7740b682030e42f41d319d38af8aa20764643a/src/macaron/database/table_definitions.py#L139)).

Therefore, it's required to correctly display the repository URL status in the `status` column (as shown in the example above). One of the reason why this is happening is because we are using [`SCMStatus`](https://github.com/oracle/macaron/blob/ecc21626209222cb4c40ef678916dd631ddee04e/src/macaron/output_reporter/scm.py#L9) enum for two purposes:
- Show the status of repository URL for a software component (used in DependencyAnalyzer)
- Show the status of running the analysis for a software component (used in Analyzer)

# Implementation
This PR currently introduce a work-around for this issue. @behnazh-w Please let me know if it's preferred to use this work-around at the moment or implement a proper fix for it directly.

### DIFF
--- a/src/macaron/output_reporter/jinja2_extensions.py
+++ b/src/macaron/output_reporter/jinja2_extensions.py
@@ -147,12 +147,12 @@ def j2_filter_get_flatten_dict(data: Any, has_key: bool = False) -> dict | Any:
     return {"0": str(data)}
 
 
-def j2_filter_get_dep_status_color(dep_status: str) -> str:
-    """Return the html class name for the color of the dep status.
+def j2_filter_get_dep_status_color(repo_url_status: str) -> str:
+    """Return the html class name for the color of the dep repo url status.
 
     Parameters
     ----------
-    dep_status : str
+    repo_url_status : str
         The dep status as string.
 
     Returns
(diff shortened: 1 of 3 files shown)


behnazh-w (approved):
Yes, It's good enough for now.

PR #533, opened 2023-10-27
chore: improve the log message in repo finder

behnazh-w:
When the `purl` string's type, which is provided as input is not a repository, which not log it as an error. This PR changes the log level and improves the log message.


tromai:
I think this change has been made by @benmss in this [PR](https://github.com/oracle/macaron/pull/508) :thinking: . Is this because we want to split it into a separate PR for clarity?

behnazh-w:
> I think this change has been made by @benmss in this [PR](https://github.com/oracle/macaron/pull/508) 🤔 . Is this because we want to split it into a separate PR for clarity?

I hadn't noticed the change in that PR, but since it is still in progress, it's better to fix the log sooner.

tromai (approved):

benmss (approved):

PR #539, opened 2023-11-02
docs: add the new build tools and adjust supported technologies

behnazh-w:


docs/source/pages/supported_technologies/index.rst
```diff
@@ -5,27 +5,55 @@
...
+such as GitHub Actions workflows.
+
+* Maven
+* Gradle
+* Pip
+* Poetry
+* npm
+* Yarn
+* Go
+* Docker container
```
tromai, line 22:
```suggestion
* Docker
```
Because I think Docker is a tool while Docker container in this case is not.

docs/source/pages/supported_technologies/index.rst
```diff
@@ -5,27 +5,55 @@
...
 .. list-table::
    :header-rows: 1
 
    * - CI Service
+     - Support
    * - `GitHub Actions <https://github.com/features/actions>`_
-
+     -
+        * Detecting deployment steps by building a call graph for workflows and reachable shell scripts
+        * Support for various GitHub APIs, such as Releases
```
tromai, line 48: I think it's not clear here what are the purposes of supporting various GitHub APIs (what are they used for?, which GitHub API entrypoints are supported).
Perhaps instead of saying that we support various GitHub APIs, we could say that we support verifying SLSA provenances published on GitHub Release instead?

tromai (approved):
LGTM. The PR can be merged once the two questions are addressed.

docs/source/pages/supported_technologies/index.rst
behnazh-w: That's not correct and we do use APIs for purposes other than releases, such as checking workflow run status. I don't want to make this a long description and just indicating that we interact with GitHub is good enough.

behnazh-w:
> LGTM. The PR can be merged once the two questions are addressed.

Please approve again based on the new changes.

docs/source/pages/supported_technologies/index.rst
behnazh-w: See commit [4afc607](https://github.com/oracle/macaron/pull/539/commits/4afc607a121890332d21ab39dac01fca39e2d24f)

tromai (approved):

PR #543, opened 2023-11-03
chore: remove pull latest when a commit is not provided for a software component (WIP)

### LINKED ISSUE #519: Potential optimization in the git operations used in preparing a repository for the analysis
With the decision to switch to partial cloning in this [PR](https://github.com/oracle/macaron/pull/389) to improve the cloning time in Macaron's analysis, I have noticed that there could be potential optimization for the git operations performed on a repository before Macaron analyzes it. 
The git operations in this context includes the git operations performed to checkout the branch and commit of any software component's repository.
The current implementation of such git operations can be found here:
- [reset_git_repo](https://github.com/oracle/macaron/blob/71accbf81e849947ca238ebfa3a9a6ef88fd01f6/src/macaron/slsa_analyzer/git_url.py#L26)
- [check_out_repo_target](https://github.com/oracle/macaron/blob/71accbf81e849947ca238ebfa3a9a6ef88fd01f6/src/macaron/slsa_analyzer/git_url.py#L65)

The potential optimization are as follow:
- https://github.com/oracle/macaron/pull/491/files#r1357738772: Only checkout the commit we want without `git pull` can save us from downloading unnecessary objects because only checking out would still follow the partial cloning strategy. 

The below optimization was initally raised but because https://github.com/oracle/macaron/pull/536 removed stashing and resetting the repository, it is not needed:
- Stashing, Reseting the index and working directory and fetching before checking out the branch lead to a significant time consumed to check out.
  - We should also consider only stashing, reset the index, working directory when necessary ONLY.
  - We could also consider don't run any fetch before checking out a branch if not necessary.
    - For example, if the use want to analyze a specific branch and commit. If that branch and commit already exists, we shouldn't try to run `git fetch`.
    - However, for cases like if the user provide only a branch, or no branch and commit at all, we still need to fetch to analyze the latest commit.
    - To know whether it's worth to do it this way, much investigation effort is still needed.

tromai:
Closes #519 .

Blocking PR:
- https://github.com/oracle/macaron/pull/548

### DIFF
--- a/src/macaron/slsa_analyzer/git_url.py
+++ b/src/macaron/slsa_analyzer/git_url.py
@@ -92,27 +92,6 @@ def check_out_repo_target(git_obj: Git, branch_name: str = "", digest: str = "",
 
     logger.info("Successfully checkout branch %s.", res_branch)
 
-    if not offline_mode:
-        # We only pull the latest changes if one of these scenarios happens:
-        #   - no digest is provided: we need to pull and analyze the latest commit.
-        #   - a commit digest is provided but it does not exist locally: we need to
-        #     pull the latest changes to check if that commit is available.
-        # We want to check if the commit already exist locally first because we want to avoid pulling unecessary
-        # if it does.
-        # We do this by checking if the commit we want to analyze is an ancestor of the commit being referenced by HEAD
-        # (which point to the tip of the branch).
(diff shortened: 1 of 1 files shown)


tromai:
Close because #519 will be addressed in the PR #548

Issue #542, opened 2023-11-03
Add updater for expected results of integration tests


benmss:
Currently any changes that create a legitimate difference in Macaron's results and the expected results used in the integration tests must be handled manually. An automatic method for updating the expected results files would save considerable time moving forwards.

behnazh-w:
Yes please! Looking forward to the PR :+1:

nathanwn:
Can we utilize syrupy for this purpose? It already supports comparing complex JSON structures and overwriting existing results.

behnazh-w:
> Can we utilize syrupy for this purpose? It already supports comparing complex JSON structures and overwriting existing results.

We have our custom scripts to process the JSON outputs. We trim the non-deterministic parts and only check the parts that are crucial. So, I doubt Syrupy being a good solution.

All we need to do really is to update the JSON outputs by copying over the files if an update flag is provided.

nathanwn:
If we have non-deterministic parts in the expected result files, then can't we just "eliminate those non-deterministic parts" from these expected result files? These files can store just the "deterministic subset" of the JSON result that Macaron produces, and not the whole result. This way we may be able to delegate both comparing and updating to surupy and remove the burden of maintaining the compare script.
Here by "eliminate the non-deterministic parts", I mean both (1) literally removing certain things that we don't need to check for and (2) forcing certain things to be deterministic when Macaron outputs them, such as the order of lists in the output. Point (2) also helps to minimize the diff of these expected output files when we commit them, making it easier to inspect and code review.

behnazh-w:
> This way we may be able to delegate both comparing and updating to surupy and remove the burden of maintaining the compare script.

Removing certain values from the output and storing it as intermediate files for testing also needs maintenance and can be even more complex. I don't see the benefit here since we already have comparison scripts.

nathanwn:
Yep. You're right that in that case, we need to cherry-pick things from the current output, and that part requires maintenance. Something I am wondering is whether the compare script or the cherry-picking will require more maintenance long-term.

tromai:
Can this issue be closed as #764 has been merged :thinking: ?

PR #544, opened 2023-11-03
test: add update option to integration script

benmss:
This PR aims to provide a way of automatically updating the expected results used by the integration tests, closing #542.

The initial implementation achieves this by replacing all `python` commands with a variable that can be overridden when the integration test script is passed a specific third argument, thereby performing a copy operation on the two file arguments that would otherwise be compared. At the current time, this avoids performing copy operations on other possible `python` commands by relying on the fact that only comparison commands use three arguments, with arguments 2 and 3 being files.

Usage:

1. Run the integration tests as usual.
2. Verify any differences between Macaron output and expected results are legitimate.
3. Run updater to replace outdated expected files.

Command: 
`make integration-tests-update`

### DIFF
--- a/scripts/dev_scripts/integration_tests.sh
+++ b/scripts/dev_scripts/integration_tests.sh
@@ -13,24 +13,31 @@ TEST_REPO_FINDER=$WORKSPACE/tests/e2e/repo_finder/repo_finder.py
 RUN_MACARON="python -m macaron -o $WORKSPACE/output"
 PYTHON="python "
 RESULT_CODE=0
+UPDATE=0
 
-function _copy() {
-  # To replace all python commands when updating results, rather than comparing them.
-  # Note that this function will always attempt to perform a file copy operation on arg #2 and #3, if there are
-  # sufficient arguments.
-  if [ $# -eq 3 ] ; then
-    # Copy arg 2 to arg 3.
-    cp "$2" "$3"
+# Optional argument for updating the expected results.
+if [ $# -eq 3 ] && [ "$3" == "--update" ] ; then
+  echo "Updating the expected results to match those currently produced by Macaron."
+  UPDATE=1
+fi
+
+function check_or_update_expected_output() {
+  if [ $UPDATE -eq 1 ] ; then
+    # Perform update of expected results by copying over produced output files.
+    # The copy only takes place if sufficient arguments are present.
+    # This function assumes arguments #2 and #3 are files: <actual_result>, <expected_result>.
+    if [ $# -eq 3 ] ; then
+      echo "Copying $2 to $3"
+      cp "$2" "$3"
+    else
+      # Calls with insufficient arguments are ignored to avoid some needless computation during updates.
+      echo "Ignoring $@"
+    fi
+  else
+    # Perform normal operation.
+    python "$@"
   fi
 }
-
-# Optional arguments for updating the expected results.
-if [ $# -ge 3 ] ; then
-  if [ $3 == "--update" ] ; then
-    echo "Updating expected results after running the tests needed to create them."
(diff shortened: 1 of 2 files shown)


scripts/dev_scripts/integration_tests.sh
```diff
@@ -11,8 +11,27 @@ COMPARE_DEPS=$WORKSPACE/tests/dependency_analyzer/compare_dependencies.py
...
+  # Note that this function will always attempt to perform a file copy operation on arg #2 and #3, if there are
+  # sufficient arguments.
+  if [ $# -eq 3 ] ; then
+    # Copy arg 2 to arg 3.
+    cp "$2" "$3"
+  fi
+}
+
+# Optional arguments for updating the expected results.
+if [ $# -ge 3 ] ; then
```
behnazh-w, line 28: How about checking that the number of arguments is 4?

scripts/dev_scripts/integration_tests.sh
```diff
@@ -11,8 +11,27 @@ COMPARE_DEPS=$WORKSPACE/tests/dependency_analyzer/compare_dependencies.py
...
+  if [ $3 == "--update" ] ; then
+    echo "Updating expected results after running the tests needed to create them."
```
behnazh-w, line 30:
```suggestion
    echo "Updating the expected results based on the previous run of integration tests."
```

scripts/dev_scripts/integration_tests.sh
```diff
@@ -11,8 +11,27 @@ COMPARE_DEPS=$WORKSPACE/tests/dependency_analyzer/compare_dependencies.py
 COMPARE_JSON_OUT=$WORKSPACE/tests/e2e/compare_e2e_result.py
 TEST_REPO_FINDER=$WORKSPACE/tests/e2e/repo_finder/repo_finder.py
 RUN_MACARON="python -m macaron -o $WORKSPACE/output"
+PYTHON="python "
```
behnazh-w, line 14: Instead of `PYTHON` I think it's good to use a more descriptive variable name, like `CHECK_EXPECTED_OUTPUT`.

scripts/dev_scripts/integration_tests.sh
```diff
@@ -11,8 +11,27 @@ COMPARE_DEPS=$WORKSPACE/tests/dependency_analyzer/compare_dependencies.py
...
 RESULT_CODE=0
 
+function _copy() {
```
behnazh-w, line 17:
```suggestion
function _update_expected_output() {
```

scripts/dev_scripts/integration_tests.sh
```diff
@@ -11,8 +11,27 @@ COMPARE_DEPS=$WORKSPACE/tests/dependency_analyzer/compare_dependencies.py
...
+  # sufficient arguments.
+  if [ $# -eq 3 ] ; then
+    # Copy arg 2 to arg 3.
+    cp "$2" "$3"
+  fi
+}
+
+# Optional arguments for updating the expected results.
+if [ $# -ge 3 ] ; then
+  if [ $3 == "--update" ] ; then
```
behnazh-w, line 29:
```suggestion
  if [ "$3" == "--update" ] ; then
```

scripts/dev_scripts/integration_tests.sh
benmss: Macaron is still run to produce the output before the expected results are overwritten, so perhaps it should be something like:
"Updating the expected results to match those currently produced by Macaron."

behnazh-w: Yes, right. Sounds good :+1:

tromai (approved):

behnazh-w (approved):

PR #548, opened 2023-11-07
chore: update git fetch command and remove git pull command

### LINKED ISSUE #519: Potential optimization in the git operations used in preparing a repository for the analysis

### LINKED ISSUE #547: Running git fetch only does not guarantee that we have fetched all changes related to tags
# Summary
`git fetch` **will not** fetch the changes of existing tags in the local repository if that tag has been modified in the remote repository. To be able to fetch that modification, we must run `git fetch --force --tags`. 
Note, before git 2.20, running `git fetch --tags` has the same behavior as `git fetch --force --tags` in this scenario ([documentation](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt-ltrefspecgt)).

# Details
From the [documentation](https://git-scm.com/docs/git-fetch), the default behavior of `git fetch` is:
Fetch branches and/or tags (collectively, "refs") from one or more other repositories, along with the objects necessary to complete their histories. 
```
...
By default, any tag that points into the histories being fetched is also fetched; the effect is to fetch tags that point at branches that you are interested in.
...
When no remote is specified, by default the origin remote will be used, unless there's an upstream branch configured for the current branch.
...
```
With that say, in the context of fetching, when you run `git fetch` , it will fetch new references (e.g. branches) and the objects necessary to complete their histories (commit objects). It will only fetch the tags that points to the new histories or any new tags from the remote repository (even in the case that new tags point to an old commit that already exists).
However, it will not update an existing tag in the local repository if that tag has been modified in the remote repository. By "modified", I mean that someone would do this:
```
git tag --delete existing_tag_name
git tag existing_tag_name <another_new_commit>
git push --force --tags  
```
which essentially modify the commit that the tag `existing_tag_name` is pointing to in the remote repository.
In this case, if you run `git fetch` on the local repository, `existing_tag_name` will not be updated to reflect the modification from the remote repository (however, it will not raise any error, and new commits, new branches, new refs are still being fetched as usual).
To make `existing_tag_name` in the local repository point to `another_new_commit` you must run:
`git fetch --force --tags`
Running `git fetch --tags` only: 1. wouldn't update the modified existing tags, 2. still fetch the new commit history from remote, 3. the command will return an error status code:
```
remote: Enumerating objects: 1, done.
remote: Counting objects: 100% (1/1), done.
remote: Total 1 (delta 0), reused 1 (delta 0), pack-reused 0
Receiving objects: 100% (1/1), 851 bytes | 851.00 KiB/s, done.
From github.com:tromai/test-repo
   14800f3..58b0069  main       -> origin/main
 ! [rejected]        v1.0.0     -> v1.0.0  (would clobber existing tag)
```
# Solution
We have decided that for now, it's better to use `git fetch --tags --force` to make sure we analyze the most up-to-date version of the repository.

tromai:
Closes #547 
Closes #519 

This PR does the following changes to the `check_out_repo_target` function.

## Add additional flags to the git fetch command
For more details on why and which flags were used, see #547 
See commit https://github.com/oracle/macaron/pull/548/commits/46cb61d4cef9fcb3fab53ba9dcefab3190194ebf

## Remove unnecessary logging in the `check_out_repo_target` function
See commit https://github.com/oracle/macaron/pull/548/commits/fe81a0ce027226ee7f5d72e18daab59eba872b6e

## Remove git pull, instead check out the latest commit directly
For more details on why, see #519.
This change is incorporated with the new design of `check_out_repo_target`.
 
## Implement a new design for the `check_out_repo_target` function
See commit https://github.com/oracle/macaron/pull/548/commits/ce3dda7c2971e542eed6250deb2e488ae275ed96

### Current supported use cases of "checking out" a specific branch and/or commit in Macaron:
  - No branch and commit hash are given: Macaron will checkout the latest commit of the default branch. (1)
  - No branch and commit hash is given: The CLI options won't allow the user to provide a commit hash without specifying a branch first. (2)
  - Branch and no commit hash given: Macaron will checkout the latest commit of that branch. (3)
  - Branch and commit hash are given: Macaron will checkout the commit with the given hash in that branch. (4)

### The new design will remove the need for checking out a branch, and instead check out the target commit directly.
- In scenario (1): The latest commit of the default branch of a repository can be obtained through `origin/HEAD` -> `git checkout --force origin/HEAD`.
- In scenario (2): We don't support at the moment but this is a totally normal use case. A git commit hash is supposed to be **unique**. If we are only interested in viewing the repository state at a given commit, a commit hash alone is sufficient.
- In scenario (3): To checkout the latest commit of a branch, we can checkout the reference of that branch in remote: `git checkout --force origin/<branch_name>`. Assume that git fetch has updated the local repository to be the up-to-date version of remote, this command will ensure that the reference `origin/<branch_name>` always point to the latest commit of that branch as observed from remote. For more information on why it is, see the Remotes section in this [article](https://git-scm.com/book/en/v2/Git-Internals-Git-References).
- In scenario (4): This use case can be rephrased as: we want to checkout a commit A, we have the information that commit A exists in the history of branch B. Therefore, it's Macaron responsibility to **only** checkout commit A if commit A exists in branch B. With that being said, we still only need to checkout once which is `git checkout --force <commit>` if we have made sure that the commit exists in the user provided branch.

### Why "origin" remote but not other remote?
- With all repositories that Macaron clones into `<output>/git_repos`, Macaron just clones and do nothing else. This mean that the default origin remote always exist.
- If we just pick one origin remote, we can prevent any extra complexity:
  - Given a branch name, which remote we check the branch name against ? what if multiple remotes have the same branch name (totally possible), but that branch in each remote have a different history. This is especially important for use case (4) as the result is not correct anymore if we don't be specific on the remote we check.
  - When we fetch, which remote do we need to fetch from?
  - How do we find the default branch ? What if different remote have different default branch ?

The drawback of pinning the remote named origin is not large. The only issue I can see right now is if we run the analysis on a local repository provided by the user. However, we could put some documentation saying that **we expect the origin remote is the repository we want to analyze**.

### How to find all the branches from origin remote whose history contains a commit hash.
To find all branches that contains a specified commit we could use the `git branch` command as documented [here](https://git-scm.com/docs/git-branch).
Ideally, what we want to run is:
```
git branch -r --list origin/* --contains <commit_hash>
```
The `-r` flag ensures that it only returns the branch from all remotes in the repositories
The `--list origin/*` makes sure to only filter out the branches that are tracked from the origin remote.
The `--contains <commit_hash>` makes sure that only return the branch names that contain the commit hash we are interested in.

For example, let's try that on the Macaron repository:
```shell
$ git clone https://github.com/oracle/macaron
$ cd macaron
$ git rev-parse v0.1.0 # obtain the commit that was tagged with v0.1.0
9217e9319ce4b3e08b4e4eb12c951f6a97b33c6e
$ git branch -r --list origin/* --contains 9217e9319ce4b3e08b4e4eb12c951f6a97b33c6e
...
origin/main
origin/staging
...
```
I have create a function in `git_url.py` to perform the above command and extract the list of branches from the output. With the above example, we will have a list of string: `[..., "origin/main", "origin/staging", ...]` whose elements are the branches that contains the commit we are interested in.


behnazh-w:
Please also address this [issue](https://github.com/oracle/macaron/issues/519) in this PR and make it a more general optimization for git operations.

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -25,22 +25,126 @@
...
+        # can have extra new line at the end, which can be picked up as an empty element in `raw_branch_names`.
+        if len(name.strip()) == 0:
+            continue
+
+        # The ``*`` symbol will appear next to the branch name where HEAD is currently on.
+        # Branches in git cannot have ``*`` in its name so we can safely replace without tampering with its actual name.
+        # https://git-scm.com/docs/git-check-ref-format
+        cleaned_up_branch_name = name.replace("*", "").strip()
+
+        result.append(cleaned_up_branch_name)
```
nathanwn, line 76: Nit: Just some naming suggestions:
`raw_branch_names -> git_branch_output_lines`
`result -> branches`
`name -> line`
`cleaned_up_branch_name` -> `branch`

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -25,22 +25,126 @@
...
+
+    return result
+
+
+def get_branches_for_commit(git_obj: Git, commit: str, remote: str = "origin") -> list[str]:
```
nathanwn, line 81: `get_branches_containing_commit` may be a more descriptive name.

tests/dependency_analyzer/configurations/micronaut_core_config.yaml
```diff
@@ -13,13 +13,13 @@ dependencies:
 - id: slf4j
     # For version 1.7.36
     # https://github.com/qos-ch/slf4j/commit/e9ee55cca93c2bf26f14482a9bdf961c750d2a56
-  branch: v_1.7.36
+  branch: 1.7-maintenance
```
nathanwn, line 16: I wonder why you need to update the branches in these `yaml` input files for the integration tests 🤔.

src/macaron/slsa_analyzer/git_url.py
tromai: Fixed in https://github.com/oracle/macaron/pull/548/commits/30665ed12f2842bb4f1e4c6c48ccd119f1858f60

src/macaron/slsa_analyzer/git_url.py
tromai: Fixed in https://github.com/oracle/macaron/pull/548/commits/30665ed12f2842bb4f1e4c6c48ccd119f1858f60

tests/dependency_analyzer/configurations/micronaut_core_config.yaml
tromai: The reason I made this changes is because `v_1.7.36` is a tag, but not a branch. In the old implementation of the check out logic, the `branch` value here is used in `git checkout v_1.7.36`. This check out is OKAY because we can checkout a tag in git normally -> we missed this error.
However, my intention with this part in the first place is that: Macaron only checkouts the commit if it belong to the branch name provided in `branch`  in the yaml config file.
Therefore, with the new implementation where the commit is explicitly checked if it is contained within `branch`, the value `v_1.7.36` is flagged as error. It has been updated here with the actually branch name `1.7-maintenance` instead.

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -25,22 +25,126 @@
...
+    If ``branch_name`` is not provided and a commit is provided, this function will checkout the commit directly.
 
-    If ``offline_mode`` is set, this method will not pull/fetch from remote while checking out the branch or commit.
+    If both ``branch_name`` and a commit are provided, this function will checkout the commit directly only if that
+    commit exists in the branch origin/<branch_name>. If not, this fucntion will return False.
+
+    For all scenarios:
+    - If the checkout failed (e.g. a branch or a commit doesn't exist), this function will return
+    False.
+    - This function will perform a force checkout, which mean it will reset the index and working tree of the target
```
nathanwn, line 143: Here I think:
* by "index" you probably mean "staging area"
* by "working tree" you probably mean "working directory"

I assume you are referring to the 2 in 3 states of files in git, including "working directory" (where unstaged changes live), "staging area"/"index" (where staged changes live), and "commits". For reference, this is similar to what [this Git documentation page describes](https://git-scm.com/book/en/v2/Getting-Started-What-is-Git%3F).

I think "index" is fine, just a bit less unintuitive than "staging area" so I would personally prefer "staging area", but I don't have a strong opinion about it.
However, "working tree" is probably not a term to use here because it can have a different meaning in the context of Git. It may refer to [Git worktree](https://git-scm.com/docs/git-worktree/2.24.0) instead, which is a different concept and probably not what you are referring to here.

tromai: I agree that using `staging area` is better than `index`. I will update it.

With `working tree`, it's still correct with my intention of referring to the files checked out from the git repository at a commit to the file system:

> The working tree is a single checkout of one version of the project. These files are pulled out of the compressed database in the Git directory and placed on disk for you to use or modify.

In the documentation page you mentioned, `working tree` is used more frequently than `working directory` (even though they should shared the same definition). That's why I thought `working tree` would be a better choice. This is also consistent with the usage of `working tree` in other git command documentation too. For example: 
- https://git-scm.com/docs/git-reset
- This [post](https://stackoverflow.com/questions/39128500/working-tree-vs-working-directory) about `working tree` vs `working directory` in `git status`
  - i.e  They claimed that `working directory` can mean the current directory where a git command is run instead of the root directory of the checked out git repo files.

In the documentation of [Git worktree](https://git-scm.com/docs/git-worktree/2.24.0), I think `working tree` still keeps the same meaning as above because Git worktree allows us to manage different working trees (each working tree is checkout from a branch for example). 

I think what we can do is instead of mentioning the `index/staging are` and `working tree/working directory` in the doc string, we could reference to the documentation of git checkout --force ([here](https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt---force)) which will ensure one source of truth and prevent any further confusion? :thinking:

nathanwn: > In the documentation of [Git worktree](https://git-scm.com/docs/git-worktree/2.24.0), I think working tree still keeps the same meaning as above

After re-reading, I think I agree. I think it was just me mixing up the meaning of "working tree" with the `git worktree` feature 😓. Thanks for verifying, Nhan.

I can see that the meaning of "working tree" should be the same as "working directory", because "trees" in Git refer to "directories". Whether the word "tree" is confusing or not I guess depends on how familiar the reader to the tree concept in Git.

With that, I don't have a strong opinion on whether a "working tree" or "working directory" should be used. I think as long as we are consistent with how we use these terms, then it's fine for me.

tromai: Thanks Nathan. I would prefer going with working tree to be consistent with how it's frequently used in the documentation. As for index/staging area, I would go with staging area. 
However, in this pydoc string, I think it's best if I reference the documentation of the `--force` flag for checkout ([here](https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt---force)) instead of mentioning the terminologies.

tromai: Addressed in https://github.com/oracle/macaron/pull/548/commits/7ee578fd355d5fbaba745e6ac5ada110896bc957

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -25,22 +25,126 @@
 logger: logging.Logger = logging.getLogger(__name__)
 
 
-def check_out_repo_target(git_obj: Git, branch_name: str = "", digest: str = "", offline_mode: bool = False) -> bool:
+def parse_git_branch_output(content: str) -> list[str]:
+    """Return a list of branch name from the output of ``git branch``.
+
+    This function only work for the output format of the first form of ``git branch`` where the branches are listed.
+    See https://git-scm.com/docs/git-branch for definition of the first form of ``git branch``.
```
behnazh-w, line 32: I don't understand this comment. Can you please elaborate?

tests/dependency_analyzer/expected_results/skipdep_micronaut-projects_micronaut-test.json
```diff
@@ -3,7 +3,7 @@
         "id": "slf4j",
         "purl": "",
         "path": "https://github.com/qos-ch/slf4j.git",
-        "branch": "v_1.7.36",
+        "branch": "1.7-maintenance",
```
behnazh-w, line 6: How about leaving the branch empty? With the support added by this PR for checking out commits it should work, right?

tests/slsa_analyzer/test_git_url.py
```diff
@@ -242,3 +242,38 @@ def test_get_remote_vcs_url_with_user_defined_allowed_hostnames(tmp_path: Path)
...
+        ),
+        ("* master\n ", ["master"]),
+        (
+            "* (HEAD detached at origin/master)\n some_other_branch",
+            ["(HEAD detached at origin/master)", "some_other_branch"],
+        ),
+        ("origin/main\n origin/dev", ["origin/main", "origin/dev"]),
+    ],
+)
+def test_parse_git_branch_output(content: str, expected_output: list[str]) -> None:
```
behnazh-w, line 277: Can you please add a hypothesis test as well to make sure parsing random/unexpected values is handled correctly?

tests/dependency_analyzer/expected_results/skipdep_micronaut-projects_micronaut-test.json
tromai: At the moment, providing a commit digest without providing a branch is not supported by Macaron, unfortunately. To be able to do that, we must also update the Analyzer class. 
I can certain update the logic to know allow providing a commit digest to checkout directly and remove the support for the case of `if both ``branch_name`` and a commit are provided, this function will checkout the commit directly only if that commit exists in the branch origin/<branch_name>. If not, this fucntion will return False.` However, I think it's out of scope for this PR. This PR was only intended to improve the git checkout operation, therefore I tried my best to keep the same behavior as before.
I suggest we can totally remove the branch name here in a separate PR. Please let me know what you think about it.

src/macaron/slsa_analyzer/git_url.py
tromai: There is two form of the `git branch` command.
The first form is using `git branch` to list out existing branch names. For example: `git branch --list`:
```
$ git branch --list
branch-1
branch-2
...
```
The second form is using `git branch` to manage existing branch or create new branches. For example: `git branch new_branch`.
What I mentioned in this comment is that this function is intended to used against the stdout output of the first form of `git branch`. I can update the comment to explain what would the "first form" mean.

tests/slsa_analyzer/test_git_url.py
tromai: This `parse_git_branch_output` function is supposed to use with the output of the first form of git branch (listing out existing branches). 
I am not sure `parsing random/unexpected values is handled correctly` would mean in this case. 
Does it mean that this function doesn't raise any Exception given unexpected values in `content` or does it mean we make sure we parse the output correctly even when `content` is still using the same output format as the first form of git branch but the branch names are of random values? 
For both cases, we need to properly come up with the hypothesis properties for the output of `git branch --list` and the properties of each branch name (for example, a branch name cannot have `*`) because the input to this function is not user input, but the output from `git branch --list`. 
My assumption was git is a well defined program, we can trust that the format of its output is consistent, hence I didn't use hypothesis for this test case. 
Can you please let me know what you mean by random/unexpected values, and how these values could be generated using hypothesis? Should they be any random string or should they follow the properties of the output format of git branch ?

docs/source/pages/using.rst
```diff
@@ -297,6 +297,9 @@ Analyzing a locally cloned repository
   Therefore, any uncommitted changes in the repository need to be backed up to prevent loss (these include unstaged changes, staged changes and untracked files).
   However, Macaron will not modify the history of the repository.
 
+.. note::
+  Macaron will always assume that ``origin`` remote exists in the repository and only checkout commits from ``origin`` remote for the analysis.
```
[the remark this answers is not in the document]
tromai: I think the important bit I want to keep here is that Macaron will always check out commits from `origin` remote but no any other remote. I have added it to your recommendation, please let me know what you think
```
We assume that the ``origin`` remote exists in the cloned repository and checkout the relevant commits from `origin` only.
```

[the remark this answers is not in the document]
behnazh-w: Sounds good.

src/macaron/slsa_analyzer/git_url.py
behnazh-w: The goal of this function is to list the branch names. It's confusing and unnecessary to mention first and second form. Please remove the second part of the comment and just keep the following:

```
Return a list of branch names using ``git branch --list``.
```

tests/dependency_analyzer/expected_results/skipdep_micronaut-projects_micronaut-test.json
behnazh-w: No, we shouldn't remove support for the branches from the CLI. The purpose of this PR is a little confusing because on one hand a case is added in `check_out_repo_target` to allow checking out a commit without when a branch is not provided, but we can't actually test it because more changes are needed in the `Analyzer`. Can you please add that change in this PR? The goal is to allow checking out a commit without a branch being provided.

tests/slsa_analyzer/test_git_url.py
behnazh-w: What if an attacker creates a repo with especially crafted branch names? Even if git reports branches "correctly", this parsing function might not handle them as expected. To avoid making the test complicated, let's start simple and just make sure the function doesn't raise exceptions and let Hypothesis produce random text values.

src/macaron/slsa_analyzer/git_url.py
tromai: This function takes in the output (as string) from stdout of running the command `git branch` in first form to parse and return a list of branch names. However, this function doesn't run the actual git command itself. The reason I have separated this logic for parsing the output string is to unit test it. 
Perhaps what we can do here is to rewrite the doc string as follows:
```
def parse_git_branch_output(content: str) -> list[str]:
    """Return a list of branch name from the output of ``git branch`` where existing branches are listed.

    Note that this function assumes that ``content`` has the format of running ``git branch`` with the ``--list`` option or without non-option arguments (e.g. name of a new branch).
    See https://git-scm.com/docs/git-branch for more details on how to use ``git branch`` to list existing branch names.

    Parameters
    ----------
    content : str
        The raw output as string from the ``git branch`` command.

    Returns
    -------
    list[str]
        The list of strings where each string is a branch element from the raw output.

    Examples
    --------
    >>> from pprint import pprint
    >>> content = '''
    ... * (HEAD detached at 7fc81f8)
    ...   master
    ...   remotes/origin/HEAD -> origin/master
    ...   remotes/origin/master
    ...   remotes/origin/v2.dev
    ...   remotes/origin/v3.dev
    ... '''
    >>> pprint(parse_git_branch_output(content))
    ['(HEAD detached at 7fc81f8)',
     'master',
     'remotes/origin/HEAD -> origin/master',
     'remotes/origin/master',
     'remotes/origin/v2.dev',
     'remotes/origin/v3.dev']
    """
```
My intention with the doc string here is to emphasize that there is an assumption on the format of ``content`` passed to this function. Because this function is only used by ``get_branches_containing_commit`` in ``git_url.py`` so it's safe to make such assumption. I believe that putting the assumption explicit would be very beneficial for developers to understand what is expected for ``content`` here. In additions, the tests and the Example section in that doc string would further help with understanding how ``parse_git_branch_output`` behaves. Please let me know what you think.

behnazh-w: Still to avoid confusion let's keep the comment to the point. We can just say:
```
Return the list of branch names from a string that has a format similar to the output of ``git branch --list``.
```

tests/slsa_analyzer/test_git_url.py
tromai: I added a simple hypothesis test case to generate random string to this function. It is added in https://github.com/oracle/macaron/pull/548/commits/11877d55a113ecb3f7d727bee00a5b04241aaa9d.

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -25,22 +25,126 @@
...
+    remote (i.e. origin/<branch_name).
+
+    If ``branch_name`` is not provided and a commit is provided, this function will checkout the commit directly.
+
+    If both ``branch_name`` and a commit are provided, this function will checkout the commit directly only if that
+    commit exists in the branch origin/<branch_name>. If not, this fucntion will return False.
 
-    If ``offline_mode`` is set, this method will not pull/fetch from remote while checking out the branch or commit.
+    For all scenarios:
+    - If the checkout failed (e.g. a branch or a commit doesn't exist), this function will return
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle/macaron/pull/548/commits/c8ddd08465e6f3df53e97632426a33cd61bff87b

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -25,22 +25,126 @@
...
+    try:
+        raw_output: str = git_obj.repo.git.branch(
+            "--remotes",
+            "--list",
+            f"{remote}/*",
+            "--contains",
+            commit,
+        )
+    except GitCommandError:
+        logger.debug("Error while using git branch to look for branches that contains %s.", commit)
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle/macaron/pull/548/commits/c8ddd08465e6f3df53e97632426a33cd61bff87b

src/macaron/slsa_analyzer/git_url.py
tromai: Fixed in https://github.com/oracle/macaron/pull/548/commits/581d6e675fdce5f4746bb383bc331d1c6a76f6a8

tests/dependency_analyzer/expected_results/skipdep_micronaut-projects_micronaut-test.json
tromai: Added in https://github.com/oracle/macaron/pull/548/commits/9db427c1a4889df7e18ff93f005288c20f73d5e3

docs/source/pages/using.rst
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle/macaron/commit/c8ddd08465e6f3df53e97632426a33cd61bff87b

tests/e2e/expected_results/micronaut-test/micronaut-test.json
```diff
@@ -1,14 +1,14 @@
...
+        "timestamps": "2023-12-22 01:32:17",
         "has_passing_check": true
     },
     "target": {
         "info": {
             "full_name": "pkg:github.com/micronaut-projects/micronaut-test@7679d10b4073a3b842b6c56877c35fa8cd10acff",
             "local_cloned_path": "git_repos/github_com/micronaut-projects/micronaut-test",
             "remote_path": "https://github.com/micronaut-projects/micronaut-test",
-            "branch": "master",
+            "branch": null,
```
behnazh-w, line 11: Can you please leave at least one of the tests that uses both branch and commit.

tromai: I think we still have a test with both branch and commit here - https://github.com/oracle/macaron/blob/c9f5a77f0ed7160fe21d34c9e26076e352ca2001/scripts/dev_scripts/integration_tests.sh#L223-L232.
Do you mean having a test case having both branch and commit defined in the yaml config file Behnaz ?

behnazh-w: yes

tromai: Added a new test case in https://github.com/oracle/macaron/pull/548/commits/2f53835a7db412c10191caa22e806bbf54727a68

behnazh-w: I didn't mean to add a new test, but to keep the branch and commit in one of the existing tests. Adding a new test will make the CI runs longer. So, if not necessary I'd like to avoid that.

tromai: We currently have this test case [here](https://github.com/oracle/macaron/blob/2f53835a7db412c10191caa22e806bbf54727a68/scripts/dev_scripts/integration_tests.sh#L302-L319) that use both branch and digest in the configuration file. We could still have it and I can revert the new test case I just added.

behnazh-w (approved):

Issue #550, opened 2023-11-09
Verify SLSA provenances for npm packages


behnazh-w:
We need to verify the discovered `npm` provenances using [the experimental feature in SLSA verifier](https://github.com/slsa-framework/slsa-verifier/releases/tag/v2.4.1).

behnazh-w:
We no longer need to use slsa-verifier for npm packages. See https://github.com/oracle/macaron/pull/747.

Issue #549, opened 2023-11-09
Discover SLSA v1 provenances for npm packages


behnazh-w:
`npm` is producing SLSA v1 provenances, e.g., see [sigstore@2.1.0](https://registry.npmjs.org/-/npm/v1/attestations/sigstore@2.1.0). We need to support these provenances.

PR #551, opened 2023-11-09
feat: find SLSA provenance v0.2 published on npm registry

### LINKED ISSUE #365: Discover SLSA provenances for npm packages
With the recent [feature](https://github.blog/2023-04-19-introducing-npm-package-provenance/) of `npm` publishing provenances when run on GitHub Actions, we need to discover the published provenances on npm registry and verify them using the [SLSA verifier.](https://github.com/slsa-framework/slsa-verifier). [Here is an example provenance](https://registry.npmjs.org/-/npm/v1/attestations/sigstore@1.6.0) produced for sigstore npm package.

behnazh-w:
This PR adds a feature to detect SLSA provenance v0.2 for `npm` packages that are published on [npm registry](https://www.npmjs.com). An example SLSA provenance v0.2 for [sigstore/mock](https://www.npmjs.com/package/@sigstore/mock) package can be found [here](https://registry.npmjs.org/-/npm/v1/attestations/@sigstore/mock@0.1.0).

Some notes:

1. This PR does not add verification of these provenances. The verification will be handled separately: https://github.com/oracle/macaron/issues/550
2. Support for SLSA provenance v1 will be added in future: https://github.com/oracle/macaron/issues/549
3. A test for an `npm` package is added to the integration tests. If making REST Calls to the `npm` registry is not possible, developers can skip that test by setting `NO_NPM` environment variable before running `make integration-test`.

Closes #365


src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -195,7 +207,50 @@ def find_provenance_assets_on_package_registries(
...
+                    try:
+                        with tempfile.TemporaryDirectory() as temp_dir:
+                            download_path = os.path.join(temp_dir, f"{artifact_id}.intoto.jsonl")
+                            if not npm_provenance_asset.download(download_path):
+                                logger.debug("Unable to find an npm provenance for %s@%s", artifact_id, version)
+                                return []
+                            try:
+                                npm_provenance_payload = load_provenance_payload(download_path)
+                            except LoadIntotoAttestationError as loadintotoerror:
+                                logger.error("Error while loading provenance %s", loadintotoerror)
```
nathanwn, line 246: You probably also need to `return []` here, otherwise `npm_provenance_payload` can be unbound when you push it into `npm_info_entry_provenances`. This was reported by `pyright` on my side.
```suggestion
                            except LoadIntotoAttestationError as loadintotoerror:
                                logger.error("Error while loading provenance %s", loadintotoerror)
                                return []
```

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -0,0 +1,265 @@
...
+        """
+        section_name = "package_registry.npm"
+        if not defaults.has_section(section_name):
+            self.enabled = False
+            return
+        section = defaults[section_name]
+
+        if not section.getboolean("enabled", fallback=True):
+            self.enabled = False
+            logger.debug("npm registry is disabled in section [{section_name}] of the .ini configuration file.")
```
nathanwn, line 74: You probably didn't mean to use f-string here.
```suggestion
            logger.debug(
                "npm registry is disabled in section [%s] of the .ini configuration file.",
                section_name,
            )
```

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -0,0 +1,265 @@
...
+        * publish with "https://github.com/npm/attestation/tree/main/specs/publish/v0.1" predicateType
+        * SLSA with "https://slsa.dev/provenance/v0.2" predicateType
+
+        We download the SLSA provenance in this method.
+
+        Parameters
+        ----------
+        url: str
+            The attestation URL.
+        download_path: srt
```
nathanwn, line 138:
```suggestion
        download_path: str
```

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -0,0 +1,265 @@
...
+            The download path for the asset.
+
+        Returns
+        -------
+        bool
+            ``True`` if the asset is downloaded successfully; ``False`` if not.
+        """
```
nathanwn, line 145: I think you need to mention that this method raises `InvalidHTTPResponseError`.

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -0,0 +1,265 @@
...
+            Note that this must include the file name.
+
+        Returns
+        -------
+        bool
+            ``True`` if the asset is downloaded successfully; ``False`` if not.
+        """
+        try:
+            return self.npm_registry.download_attestation_payload(self.url, dest)
+        except InvalidPURLError as error:
```
nathanwn, line 263: Shouldn't this be `InvalidHTTPResponseError` instead?

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def download_attestation_payload(self, url: str, download_path: str) -> bool:
+        """Download the npm attestation from npm registry.
+
+        Each npm package has two types of attestations:
+
+        * publish with "https://github.com/npm/attestation/tree/main/specs/publish/v0.1" predicateType
+        * SLSA with "https://slsa.dev/provenance/v0.2" predicateType
+
+        We download the SLSA provenance in this method.
```
nathanwn, line 132: I think this method, apart from downloading the attestation, also has some logic to validate the `predicateType` field and pick out the `dsseEnvelope` in the JSON payload. Could you also link to the example payload you showed in the PR description here?
```
https://registry.npmjs.org/-/npm/v1/attestations/@sigstore/mock@0.1.0
```
I think it will greatly help with understanding what this method is trying to do. Thanks, Behnaz.

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+class NPMAttestationAsset(NamedTuple):
+    """An attestation asset hosted on the npm registry.
+
+    The API Documentation can be found here:
+    """
+
+    #: The namespace of the artifact on npm.
+    namespace: str | None
```
nathanwn, line 196: I think it may be helpful to have some explanation of what PURL `namespace` is supposed to be in the case of npm and why it can be `None`. I found the following sources:
* https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#npm. This says
> The namespace is used for the scope of a scoped NPM package.
* This page explains what a scoped NPM package is: https://docs.npmjs.com/cli/v10/using-npm/scope.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
tromai: Pyright reports the same error for me.

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -0,0 +1,265 @@
...
+        attestation_endpoint: str | None = None,
+        request_timeout: int | None = None,
+        enabled: bool = True,
+    ) -> None:
+        """
+        Initialize the npm Registry instance.
+
+        Parameters
+        ----------
+        hostname : str
```
nathanwn, line 43:
```suggestion
        hostname : str | None
```

src/macaron/slsa_analyzer/analyze_context.py
```diff
@@ -91,25 +94,27 @@ def __init__(
...
 
     @property
-    def provenances(self) -> dict:
+    def provenances(self) -> dict[str, list[InTotoV01Statement | InTotoV1Statement]]:
         """Return the provenances data as a dictionary.
 
         Returns
         -------
-        dict
+        dict[str : list[InTotoV01Statement]]
```
nathanwn, line 102:
```suggestion
        dict[str, list[InTotoV01Statement | InTotoV1Statement]]
```

tests/slsa_analyzer/test_analyzer.py
```diff
@@ -103,6 +106,41 @@ def test_resolve_analysis_target(
...
+            "purl": f"pkg:{purl_type}/{namespace}/{artifact_id}@{version}",
+            "path": url,
+            "branch": branch,
+            "digest": digest,
+        }
+    )
+    try:
+        Analyzer.to_analysis_target(config, available_domains)
+    except InvalidPURLError:
+        pass
```
nathanwn, line 141: I wonder what you are trying to test for here. Is it just making sure `Analyzer.to_analysis_target` does not crash unexpectedly with an unknown error type (apart from `InvalidPURLError`)?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -540,6 +541,14 @@ def to_analysis_target(config: Configuration, available_domains: list[str]) -> A
             except ValueError as error:
                 raise InvalidPURLError(f"Invalid input PURL: {config.get_value('purl')}") from error
 
+        # Validate the purl object, which is user-controllable.
+        if parsed_purl:
+            purl_str = parsed_purl.to_string()
+            # The urlparse API sanitizes unsafe characters.
+            parsed_url = urllib.parse.urlparse(purl_str)
+            if purl_str != parsed_url.geturl():
+                raise ValueError("The PURL provided as input has unsafe characters.")
```
nathanwn, line 550: Because this seems important and probably can be tested, I wonder if we can extract this into a function and unit-test it.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -540,6 +541,14 @@ def to_analysis_target(config: Configuration, available_domains: list[str]) -> A
             except ValueError as error:
                 raise InvalidPURLError(f"Invalid input PURL: {config.get_value('purl')}") from error
 
+        # Validate the purl object, which is user-controllable.
+        if parsed_purl:
+            purl_str = parsed_purl.to_string()
+            # The urlparse API sanitizes unsafe characters.
```
tromai, line 547: I think it's also good to document what kind of unsafe characters are we expecting it to have.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -540,6 +541,14 @@ def to_analysis_target(config: Configuration, available_domains: list[str]) -> A
             except ValueError as error:
                 raise InvalidPURLError(f"Invalid input PURL: {config.get_value('purl')}") from error
 
+        # Validate the purl object, which is user-controllable.
+        if parsed_purl:
+            purl_str = parsed_purl.to_string()
```
tromai, line 546: If we are only trying to sanitize the original purl_string provided, do you think we should do it before trying to create a PackageURL object from that purl string?

src/macaron/slsa_analyzer/analyzer.py
tromai: I agree.

tests/slsa_analyzer/test_analyzer.py
```diff
@@ -103,6 +106,41 @@ def test_resolve_analysis_target(
...
+    purl_type: str,
+    namespace: str | None,
+    artifact_id: str,
+    version: str,
+    url: str,
+    branch: str,
+    digest: str,
+    available_domains: list[str],
+) -> None:
+    """Test the analysis target resolution with valid inputs."""
```
tromai, line 129:
```suggestion
    """Test the analysis target resolution with valid inputs."""
```

tests/slsa_analyzer/checks/test_provenance_available_check.py
```diff
@@ -50,6 +55,20 @@ def download_asset(self, url: str, download_path: str) -> bool:
         return False
 
 
+class MockNPMRegistry(NPMRegistry):
+    """Mocj NPMRegistry class."""
```
nathanwn, line 59:
```suggestion
    """Mock NPMRegistry class."""
```

tests/slsa_analyzer/package_registry/test_npm_registry.py
```diff
@@ -0,0 +1,131 @@
...
+from macaron.config.defaults import load_defaults
+from macaron.errors import ConfigurationError
+from macaron.slsa_analyzer.build_tool.base_build_tool import BaseBuildTool
+from macaron.slsa_analyzer.package_registry.npm_registry import NPMAttestationAsset, NPMRegistry
+
+
+def _create_npm_registry() -> NPMRegistry:
+    """Create an npm registry instance.
+
+    This private function is used as parameter for Hypothesis, which otherwise does not accept a fixture.
```
nathanwn, line 20: Hmm, I don't see this being used in any hypothesis tests.

tests/slsa_analyzer/package_registry/test_npm_registry.py
```diff
@@ -0,0 +1,131 @@
...
+    npm_registry.load_defaults()
+    assert npm_registry.is_detected(build_tool=build_tools[build_tool_name]) == expected
+
+
+@pytest.mark.parametrize(
+    (
+        "namespace",
+        "artifact_id",
+        "version",
+        "expected",
```
nathanwn, line 100: Do you think `expected` is the best name here? It is not the expected value of any `assert`, but rather the path path in the URL. I'm thinking `package_path` or something similar may be a better name.

src/macaron/slsa_analyzer/analyzer.py
nathanwn, line 550: This case raises `ValueError`, but `ValueError` is neither documented in the method docstring nor caught at the call site of this method.
I also wonder if `ValueError` here should be converted to `InvalidPURLError` to be consistent with the rest of the method, or should it be a separate error type to be handled differently from the rest.

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -0,0 +1,265 @@
...
+    There is no complete and up-to-date API documentation for the npm registry and the endpoints
+    are discovered by manual inspection of links on https://www.npmjs.com.
+    """
+
+    def __init__(
+        self,
+        hostname: str | None = None,
+        attestation_endpoint: str | None = None,
+        request_timeout: int | None = None,
+        enabled: bool = True,
```
tromai, line 36: This parameter is different from `JfrogMavenRegistry` ([here](https://github.com/oracle/macaron/blob/52cecac033b384d69d3a711033e4b7f8fee49120/src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py)) where the signature of enabled is: `enabled: bool | None = None` when I believe they have the same meaning. I wonder if there is any reason for this difference?

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -0,0 +1,265 @@
...
+            The attestation REST API.
+        request_timeout : int | None
+            The timeout (in seconds) for requests made to the package registry.
+        enabled: bool
+            Shows whether making REST API calls to npm registry is enabled.
+        """
+        self.hostname = hostname or ""
+        self.attestation_endpoint = attestation_endpoint or ""
+        self.request_timeout = request_timeout or 10
+        self.enabled = enabled
```
tromai, line 55: When the instance of this class is created (before loading the values from `defaults.ini`) this attribute will be `True` and during the loading of values in `defaults.ini` it could be set to False if necessary. This approach is opposite to `JfrogMavenRegistry`. Is this desirable and should we keep it consistent both of them have the same `enabled` attribute?

src/macaron/slsa_analyzer/package_registry/npm_registry.py
behnazh-w: This PR adds support for `npm` which has its own requirements. Why do you bring up `JfrogMavenRegistry` in this review?

src/macaron/slsa_analyzer/package_registry/npm_registry.py
behnazh-w: Same answer. `npm` has its own requirements. I don't understand why `JfrogMavenRegistry` is mentioned here.

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -0,0 +1,265 @@
...
+        build_tool : BaseBuildTool
+            A detected build tool of the repository under analysis.
+
+        Returns
+        -------
+        bool
+            ``True`` if the repo under analysis can be published to this package registry,
+            based on the given build tool.
+        """
+        compatible_build_tool_classes = [NPM, Yarn]
```
tromai, line 118: I think we should check if the `self.enabled` is False and return False here.

docs/source/pages/supported_technologies/index.rst
```diff
@@ -86,8 +89,9 @@ Provenances
      - Documentation
    * - `SLSA <https://slsa.dev>`_
      -
-       * `SLSA provenance version 0.2 <https://slsa.dev/spec/v0.2/provenance>`_.
-       * The provenance should be published as a GitHub release asset
+       * | `SLSA provenance version 0.2 <https://slsa.dev/spec/v0.2/provenance>`_. The provenance should be
```
tromai, line 92: I think we should explicitly say that the provenance can `either` be  `published as a GitHub release asset.` or `published on the `npm registry <https://registry.npmjs.org>`_.`

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -0,0 +1,265 @@
...
+            logger.debug("npm registry is disabled in section [{section_name}] of the .ini configuration file.")
+            return
+
+        self.hostname = section.get("hostname")
+        if not self.hostname:
+            raise ConfigurationError(
+                f'The "hostname" key is missing in section [{section_name}] of the .ini configuration file.'
+            )
+
+        self.attestation_endpoint = section.get("attestation_endpoint", "-/npm/v1/attestations")
```
tromai, line 83: I don't think we need to assign the default value of `-/npm/v1/attestations` as a fall back when the key `attestation_endpoint` is not available in `defaults.ini`. Because when en the user don't provide `attestation_endpoint` in their custom `defaults.ini`, for example:
```
[package_registry.npm]
# Set `enabled=False` to disable making REST API calls to the npm registry.
enabled = True
# npm registry host name.
hostname = registry.npmjs.org
# The attestation REST API.
request_timeout = 20
```
The final value of `self.attestation_endpoint` will be `-/npm/v1/attestations` as defined in the `src/macaron/config/defaults.ini` file.

src/macaron/slsa_analyzer/package_registry/npm_registry.py
tromai, line 132: By `download the SLSA provenance`, I wonder whether we are mentioning about the whole attestation payload or we are mentioning about the attestation that has the  `https://slsa.dev/provenance/v0.2` predicateType?

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -0,0 +1,265 @@
...
+        Returns
+        -------
+        str
+        """
+        # Build the path parameters.
+        path_params = [self.npm_registry.attestation_endpoint]
+        if self.namespace:
+            path_params.append(self.namespace)
+        path_params.append(self.artifact_id)
+        path = f'{"/".join(path_params)}'
```
tromai, line 231: I think we can remove the f string here as `"/".join(path_params)` will return a string.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -478,7 +535,7 @@ def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
 
             justification.append("Found provenance in release assets:")
             justification.extend(
-                [asset.name for asset in provenance_assets],
+                [asset.url for asset in provenance_assets],
```
tromai, line 538: I wonder  is the reason the asset's URL is displayed here instead of its artifact id, Behnaz?

tests/slsa_analyzer/package_registry/test_npm_registry.py
tromai: I agree. Or if we want to keep `expected`, the value could change to the exact HTTPS URL that we finally compare too.

tromai (commented):

src/macaron/slsa_analyzer/checks/provenance_available_check.py
nathanwn: I assume with NPM attestation support added, displaying the URL to the attestation instead of the attestation name makes sense because NPM attestations are not files (like what we were dealing with previously) but HTTP response payload. `asset.name` here was always the name of the attestation file being downloaded but it is not applicable for the NPM attestation case.

I think URLs may even be the overall better output type here because they point directly to the place where the attestation can be found. Just the file name may not provide enough information, especially now that we have multiple places where an attestation can be discovered.

@behnazh-w Let us know if you are thinking the same, or if there is anything you want to add.

tests/slsa_analyzer/package_registry/test_npm_registry.py
behnazh-w: Yes, forgot to remove it after moving the hypothesis test.

tests/slsa_analyzer/package_registry/test_npm_registry.py
behnazh-w: I prefer to the name `expected` as it is what the test expects to see as part of the `assert` statement.

tests/slsa_analyzer/test_analyzer.py
```diff
@@ -103,6 +106,41 @@ def test_resolve_analysis_target(
     assert Analyzer.to_analysis_target(config, available_domains) == expect
 
 
+@given(
+    purl_type=st.one_of(st.text(), st.sampled_from(["maven", "npm", "pypi", "github.com"])),
+    namespace=st.one_of(st.none(), st.text()),
+    artifact_id=st.text(),
+    version=st.text(),
+    url=st_pr.urls(),
```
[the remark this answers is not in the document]
behnazh-w: It should be fine. If they make changes that break our test (which is unlikely), we will fix it. The  `urls()` strategy is pretty useful and I prefer to use it.

tests/slsa_analyzer/test_analyzer.py
behnazh-w: It makes sure that random and invalid inputs are handled properly and either the `InvalidPURLError` exception is thrown or there is no other exception.

behnazh-w:
I have answered the questions and addressed comments in [79784f0](https://github.com/oracle/macaron/pull/551/commits/79784f05894629acba02a758c737e3c0d87be6c0).

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -96,14 +96,16 @@ def load_defaults(self) -> None:
             ) from error
 
     def is_detected(self, build_tool: BaseBuildTool) -> bool:
-        """Detect if artifacts of the repo under analysis can possibly be published to this package registry.
+        """Detect if artifacts under analysis can be published to this package registry.
```
tromai, line 99: Should this new py doc string for the `is_detected` method be updated in the base class `PackageRegistry` and its other children as well?

behnazh-w: I think that is better to be done separately to keep this PR focused on `npm` support.

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -128,20 +133,28 @@ def download_attestation_payload(self, url: str, download_path: str) -> bool:
 
         * publish with "https://github.com/npm/attestation/tree/main/specs/publish/v0.1" predicateType
         * SLSA with "https://slsa.dev/provenance/v0.2" predicateType
+        * SLSA with "https://slsa.dev/provenance/v1" predicateType
```
tromai, line 136: If the SLSA provenance v1 is added, I think we should change the above line to `Each npm package has three types of attestations`.

tromai (changes requested):
I have reviewed all the changes related to my comments. There is only one new [comment](https://github.com/oracle/macaron/pull/551#discussion_r1402800276).

src/macaron/slsa_analyzer/package_registry/npm_registry.py
behnazh-w: See commit [bae2fe0](https://github.com/oracle/macaron/pull/551/commits/bae2fe0376631a0a9081d8d507459bf5f14b6d5f).

tromai (approved):
All of my comments have been resolved. The changes related to those LGTM.

nathanwn (approved):
All looks fine to me. Thanks for the PR.

Issue #60, opened 2023-02-09
CycloneDX Gradle plugin has non-detereminitic behavior


behnazh-w:
The [CycloneDX Gradle plugin](https://github.com/CycloneDX/cyclonedx-gradle-plugin) produces different BOM files in different runs, which results in failing [Gradle dependency analysis integration tests](https://github.com/oracle-samples/macaron/actions/runs/4130738272).

PR #552, opened 2023-11-09
test: replace some micronaut-core tests with micronaut-test to reduce integration test time

behnazh-w:
We use [micronaut-core](https://github.com/micronaut-projects/micronaut-core) in the integration tests a few times but this project takes a long time to analyze because it produces many artifacts, which we download and verify as part of the provenance verification.

This PR reduces the execution time of the integration tests by making the following changes:

- Using [micronaut-core](https://github.com/micronaut-projects/micronaut-core) only once and checking the analysis result.
- Using  [micronaut-test](https://github.com/micronaut-projects/micronaut-test) instead of [micronaut-core](https://github.com/micronaut-projects/micronaut-core) in the following tests.
- Commenting out the test for Gradle dependencies as it requires fixing https://github.com/oracle/macaron/issues/60.


tromai (approved):
Thanks for making the changes. LGTM

PR #555, opened 2023-11-14
chore: resolve doctest issues

nathanwn:
While reading through some code, I realized the examples in some docstrings are wrong. These should have been caught by `doctest`.

Currently in `pyproject.toml`, `pytest` is configured as follows:

```toml
[tool.pytest.ini_options]
addopts = "--verbose --doctest-modules -ra --cov macaron"  # Consider adding --pdb
testpaths = [
    "tests",
]
```

Although the `--doctest-modules` flag is enabled, meaning `doctest` should run, `testpaths` is limited to only the `tests` directory. This does not allow `doctest` to pick up the docstring examples in `src/macaron`.

In this PR, I enable `doctest` to discover docstring examples in `src/macaron` as follows:

```diff
testpaths = [
+   "src/macaron",  # this is included to run doctest on examples
    "tests",
]
```

I also fix cases where the example fails `doctest`. There are commonly two reasons why these examples fail:
* Case 1: The example is wrong. One or more of these things happen: the test is syntactically wrong, or the input is wrong, or the output is wrong. In these cases, I fix the example.
* Case 2: The example is only intended to demonstrate how the API looks like, and not how it is actually used and what the corresponding result should be. In addition, the setup required for some of these examples is quite complex (e.g. requires loading a config or making an HTTP request). In these cases, I convert the example from using a Python REPL block, which is what `doctest` picks up, to using a Python code block.

### DIFF
--- a/src/macaron/output_reporter/results.py
+++ b/src/macaron/output_reporter/results.py
@@ -98,8 +98,15 @@ class Record(Generic[RecordNode]):
 
         Examples
         --------
-        >>> record.get_summary()
-        {'id': 'apache/maven', 'description': 'Analysis completed', 'report': 'apache.html', 'status': SCMStatus.AVAILABLE}
+        .. code-block:: python3
+
+            summary = record.get_summary()
+            # summary == {
+            #     'id': 'apache/maven',
+            #     'description': 'Analysis completed',
+            #     'report': 'apache.html',
+            #     'status': SCMStatus.AVAILABLE,
+            # }
         """
         return {
             "id": self.record_id,
--- a/src/macaron/output_reporter/results.py
+++ b/src/macaron/output_reporter/results.py
@@ -98,15 +98,23 @@ class Record(Generic[RecordNode]):
 
         Examples
         --------
-        .. code-block:: python3
-
-            summary = record.get_summary()
-            # summary == {
-            #     'id': 'apache/maven',
-            #     'description': 'Analysis completed',
-            #     'report': 'apache.html',
-            #     'status': SCMStatus.AVAILABLE,
-            # }
+        >>> from pprint import pprint
+        >>> from macaron.config.target_config import Configuration
+        >>> from macaron.output_reporter.scm import SCMStatus
+        >>> from macaron.output_reporter.results import Record
+        >>> record = Record(
+        ...     record_id="apache/maven",
+        ...     description="Analysis completed",
+        ...     pre_config=Configuration({}),
+        ...     status=SCMStatus.MISSING_SCM,
+        ...     context=None,
+        ...     dependencies=[],
+        ... )
+        >>> pprint(record.get_summary())
+        {'description': 'Analysis completed',
+         'id': 'apache/maven',
+         'repo_url_status': <SCMStatus.MISSING_SCM: 'MISSING REPO URL'>,
+         'report': ''}
         """
         return {
             "id": self.record_id,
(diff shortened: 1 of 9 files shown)


tromai (approved):
Thanks for having a look @nathanwn . LGTM.

src/macaron/output_reporter/results.py
```diff
@@ -98,8 +98,15 @@ def get_summary(self) -> dict:
...
+        .. code-block:: python3
+
+            summary = record.get_summary()
+            # summary == {
+            #     'id': 'apache/maven',
+            #     'description': 'Analysis completed',
+            #     'report': 'apache.html',
+            #     'status': SCMStatus.AVAILABLE,
+            # }
         """
```
behnazh-w, line 110: This doctest will not be picked by pytest because according to the documentation:
>The [doctest](https://docs.python.org/3/library/doctest.html#module-doctest) module searches for pieces of text that look like interactive Python sessions, and then executes those sessions to verify that they work exactly as shown.

So, you need to make it interactive, and when you do you will notice that the doctest will fail.
Also, you need to use `pycon` instead of `python3`, otherwise `blacken-docs` will not be able to parse the code.
```suggestion
        .. code-block:: pycon

            >>> from macaron.config.target_config import Configuration
            >>> from macaron.output_reporter.scm import SCMStatus
            >>> from macaron.output_reporter.results import Record
            >>> record = Record(
            ...     record_id="apache/maven",
            ...     description="Analysis completed",
            ...     pre_config=Configuration({}),
            ...     status=SCMStatus.MISSING_SCM,
            ...     context=None,
            ...     dependencies=[],
            ... )
            >>> record.get_summary()
            {'id': 'apache/maven', 'description': 'Analysis completed', 'report': '', 'repo_url_status': <SCMStatus.MISSING_SCM: 'MISSING REPO URL'>}

        """  # noqa: B950
```
This should be a valid doctest that can be checked.

See

nathanwn: I am aware of how doctest works here. As I mentioned in the PR description, I categorized the doctest errors into 2 groups:

* ones that are supposed to be run in Python REPL (interactive mode), and
* ones that are only there to demonstrate how the API should be used.
 
I qualified this to be in the 2nd group, and the way I went about fixing this group of errors was to convert them into Python code blocks without caring about the execution since the setup for these examples can be complex. As you have probably realized in this particular example, it requires a few imports and initialization of an object with 6 different fields before the relevant part of the example can be executed, which is the `get_summary` method. The initial example before my change was not doing the necessary setup for the `get_summary` method, and I was not familiar enough with the `Record` class to make a judgment call about whether I should set up a `Record` object here to actually run the example in interactive mode. Therefore, I opted to keep it almost exactly the same as the previous version and just convert it to a Python code block.

Hopefully, this clarifies my intention here.

I will incorporate your suggestion. Thanks, Behnaz.

behnazh-w: I wasn't sure about your intent here because you were commenting out 
```
            # summary == {
            #     'id': 'apache/maven',
            #     'description': 'Analysis completed',
            #     'report': 'apache.html',
            #     'status': SCMStatus.AVAILABLE,
            # }
```
but leaving `summary = record.get_summary()` uncommented.  If this block was not going to be picked up by pytest anyway, then why to comment and uncomment some lines?

nathanwn: The example was not supposed to be picked up by pytest, but just to give an example of how the API should be used and what the expected result should look like.
The expected result is not valid Python syntax, and therefore I put it in a comment block.

nathanwn: I incorporated your suggestion in 7ab9baee451b5a1f0433b2808f47c625ec40829e.

behnazh-w: > The expected result is not valid Python syntax, and therefore I put it in a comment block.

In such cases, it doesn't make sense to use a `python` code block if the example is not a valid Python code.

behnazh-w (approved):

PR #576, opened 2023-12-19
chore: use GitHub release API to filter possible tag matches

### LINKED ISSUE #537: Improve the commit finding approach
Instead of looking up tags directly, we can get the releases through the GitHub API and retrieve the tag/commit:
```
https://api.github.com/repos/OWNER/REPO/releases
```
For example part of the response for `https://api.github.com/repos/google/guava/releases` is:
```json
    "tag_name": "v32.1.3",
    "target_commitish": "master",
    "name": "32.1.3",
    "draft": false,
    "prerelease": false,
    "created_at": "2023-10-10T20:34:53Z",
    "published_at": "2023-10-10T22:20:41Z",
```

This way we can prune the unrelated tags and further filter the `draft` and `prerelease` results. This seems to be a better approach or at least a good way to improve precision.

If repository is not using GitHub releases, the API returns an empty list. In this case we can fall back to checking the tags.

Note: by finding the corresponding release tag, we can populate the `release_tag` attribute of the `Repository` object.

benmss:
Closes #537 

The GitHub API is called when finding commits for artefacts to retrieve release information in cases where an analysis target has a corresponding GitHub URL. These results are limited to the first 1000 instances by the API. The information retrieved includes two key properties: release name, and release tag. 

The release tags retrieved via the API call can be used to filter down the list of all tags, thereby eliminating potentially spurious entries before the regular expression matching takes place. If that fails, the release names can instead be used as the list of tags for the same purpose, with the linked tag (in the API information) being returned in the case of a successful match. If that also fails, the behaviour is as it was before this change was added, i.e. all repository tags are used for the matching.

### DIFF
--- a/src/macaron/repo_finder/commit_finder.py
+++ b/src/macaron/repo_finder/commit_finder.py
@@ -12,9 +12,10 @@ from gitdb.exc import BadName
 from packageurl import PackageURL
 from pydriller import Commit, Git
 
+from macaron.config.global_config import global_config
 from macaron.repo_finder import repo_finder_deps_dev
 from macaron.repo_finder.repo_finder import to_domain_from_known_purl_types
-from macaron.slsa_analyzer.git_service import GIT_SERVICES
+from macaron.slsa_analyzer.git_service import GIT_SERVICES, api_client
 
 logger: logging.Logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


benmss:
This will be reconsidered and possibly included as part of #644 instead.

PR #579, opened 2023-12-22
ci(test): fix the test descriptions and comparison for micronaut-test

### LINKED ISSUE #577: One micronaut-test integration test is running with a PURL, but we check for dependency report files.
In this integration test case [here](https://github.com/oracle/macaron/blob/1f7f6640ae96791bc47d7083c747c73d6488ef3e/scripts/dev_scripts/integration_tests.sh#L269-L286), we are running the analysis on a PURL without dependency resolving `-purl pkg:maven/io.micronaut/micronaut-test-core@4.1.0 --skip-deps`. However, we check for two dependencies report files `caffeine.json` and `slf4j.json`. 
The reason this test case is still passing is because it's reusing `caffeine.json` and `slf4j.json` from test cases that run before it.

behnazh-w:
Closes #577 

This PR fixes the description of test for `micronaut-test` and also removes the test that uses PURL `pkg:maven/io.micronaut/micronaut-test-core@4.1.0` because for some reason this artifact is not published on Maven central. This test should have been failing but we were checking the wrong output file.

### DIFF
--- a/scripts/dev_scripts/integration_tests.sh
+++ b/scripts/dev_scripts/integration_tests.sh
@@ -257,17 +257,8 @@ $RUN_MACARON analyze -c $WORKSPACE/tests/dependency_analyzer/configurations/micr
 
 check_or_update_expected_output $COMPARE_DEPS $DEP_RESULT $DEP_EXPECTED || log_fail
 
-# TODO: uncomment the test below after resolving https://github.com/oracle/macaron/issues/60.
-# echo -e "\n----------------------------------------------------------------------------------"
-# echo "micronaut-projects/micronaut-test: Check the resolved dependency output with config for cyclonedx gradle plugin (default)."
-# echo -e "----------------------------------------------------------------------------------\n"
-# DEP_EXPECTED=$WORKSPACE/tests/dependency_analyzer/expected_results/cyclonedx_micronaut-projects_micronaut-test.json
-# $RUN_MACARON analyze -c $WORKSPACE/tests/dependency_analyzer/configurations/micronaut_test_config.yaml || log_fail
-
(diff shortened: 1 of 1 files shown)


scripts/dev_scripts/integration_tests.sh
```diff
@@ -257,17 +257,8 @@ $RUN_MACARON analyze -c $WORKSPACE/tests/dependency_analyzer/configurations/micr
...
-# echo "micronaut-projects/micronaut-test: Check the resolved dependency output with config for cyclonedx gradle plugin (default)."
-# echo -e "----------------------------------------------------------------------------------\n"
-# DEP_EXPECTED=$WORKSPACE/tests/dependency_analyzer/expected_results/cyclonedx_micronaut-projects_micronaut-test.json
-# $RUN_MACARON analyze -c $WORKSPACE/tests/dependency_analyzer/configurations/micronaut_test_config.yaml || log_fail
-
-# python $COMPARE_DEPS $DEP_RESULT $DEP_EXPECTED || log_fail
-
 echo -e "\n----------------------------------------------------------------------------------"
-echo "micronaut-projects/micronaut-test: Check the e2e output JSON file with config and no dependency analyzing."
+echo "micronaut-projects/micronaut-test: Check the e2e output JSON file with config when automatic dependency resolution is skipped."
```
tromai, line 261: If this section only check the output files without running any Macaron command ( the Macaron command run is above it), perhaps we should merge this comment with the one above it so that it appears to be one test case.

behnazh-w: No, these are two different log messages for separate comparisons.

tromai (approved):
Thanks for the change. LTGM

PR #586, opened 2024-01-03
chore: disregard branch in commit finder

### LINKED ISSUE #585: Commit finder does not succeed when branch is unavailable
In some cases the commit finder retrieves a commit that appears to have no associated branch. During development these cases were designed to fail as a branch was required by Macaron, and providing a default branch instead would not work. Before the commit finder PR was merged, branches were made optional by PR #548, making this development decision outdated and in need of updating.

benmss:
As in some cases the branch for a commit found by the commit finder cannot also be found, and that branches are not needed to pull a commit, this PR removes all attempts at retrieving branch information as part of the commit finding process. This makes the commit finder return only a commit hash/digest.

Closes #585

### DIFF
--- a/src/macaron/repo_finder/commit_finder.py
+++ b/src/macaron/repo_finder/commit_finder.py
@@ -114,7 +114,7 @@ class AbstractPurlType(Enum):
     UNSUPPORTED = (2,)
 
 
-def find_commit(git_obj: Git, purl: PackageURL) -> str:
+def find_commit(git_obj: Git, purl: PackageURL) -> str | None:
     """Try to find the commit matching the passed PURL.
 
     The PURL may be a repository type, e.g. GitHub, in which case the commit might be in its version part.
@@ -130,13 +130,13 @@ def find_commit(git_obj: Git, purl: PackageURL) -> str:
 
     Returns
     -------
-    str
-        The digest, or an empty string if the commit cannot be correctly retrieved.
+    str | None
+        The digest, or None if the commit cannot be correctly retrieved.
     """
     version = purl.version
     if not version:
         logger.debug("Missing version for analysis target: %s", purl.name)
-        return ""
+        return None
 
     repo_type = determine_abstract_purl_type(purl)
     if repo_type == AbstractPurlType.REPOSITORY:
@@ -144,7 +144,7 @@ def find_commit(git_obj: Git, purl: PackageURL) -> str:
     if repo_type == AbstractPurlType.ARTIFACT:
         return find_commit_from_version_and_name(git_obj, purl.name, version)
     logger.debug("Type of PURL is not supported for commit finding: %s", purl.type)
-    return ""
+    return None
 
 
 def determine_abstract_purl_type(purl: PackageURL) -> AbstractPurlType:
@@ -174,7 +174,7 @@ def determine_abstract_purl_type(purl: PackageURL) -> AbstractPurlType:
         return AbstractPurlType.UNSUPPORTED
 
 
-def extract_commit_from_version(git_obj: Git, version: str) -> str:
+def extract_commit_from_version(git_obj: Git, version: str) -> str | None:
     """Try to extract the commit from the PURL's version parameter.
 
     E.g.
@@ -190,8 +190,8 @@ def extract_commit_from_version(git_obj: Git, version: str) -> str:
 
     Returns
     -------
-    str
-        The digest, or an empty string if the commit cannot be correctly retrieved.
+    str | None
+        The digest, or None if the commit cannot be correctly retrieved.
     """
     # A commit hash is 40 characters in length, but commits are often referenced using only some of those.
     commit: Commit | None = None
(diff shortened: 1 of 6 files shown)


behnazh-w:
We should be able to remove the branch and commit hash of the `guava@32.1.2-jre` in the `integration_tests.sh`:
```
$RUN_MACARON analyze -purl pkg:maven/com.google.guava/guava@32.1.2-jre?type=jar --skip-deps || log_fail
```
Can you please change the corresponding test?

tests/repo_finder/test_commit_finder.py
```diff
@@ -151,71 +136,47 @@ def test_commit_finder() -> None:
...
 
     # Create a tag on the commit that has no branch.
-    tag_no_branch = "0.1.2"
-    git_obj.repo.create_tag(tag_no_branch, ref=commit_with_no_branch.hexsha)
-
-    # Tag with no branch.
-    branch, _ = commit_finder.find_commit(git_obj, PackageURL.from_string(f"pkg:maven/apache/maven@{tag_no_branch}"))
-    assert not branch
+    # tag_no_branch = "0.1.2"
+    # git_obj.repo.create_tag(tag_no_branch, ref=commit_with_no_branch.hexsha)
```
nathanwn, line 140: Should we keep these commented lines?

benmss: We should not, good catch.

nathanwn:
The PR title probably needs to be updated, with the issue number at the end and the description a bit more elaborate. Probably something like this, but just my suggestion:
```
chore: do not try to find a branch in commit finder (#585)
```
Otherwise, I think it's LGTM.

nathanwn (approved):

src/macaron/repo_finder/commit_finder.py
```diff
@@ -114,10 +114,10 @@ class AbstractPurlType(Enum):
     UNSUPPORTED = (2,)
 
 
-def find_commit(git_obj: Git, purl: PackageURL) -> tuple[str, str]:
+def find_commit(git_obj: Git, purl: PackageURL) -> str:
```
behnazh-w, line 117: It's better to change the return type to `str | None` and return `None` instead of empty string to catch type errors.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -174,7 +174,7 @@ def determine_abstract_purl_type(purl: PackageURL) -> AbstractPurlType:
         return AbstractPurlType.UNSUPPORTED
 
 
-def extract_commit_from_version(git_obj: Git, version: str) -> tuple[str, str]:
+def extract_commit_from_version(git_obj: Git, version: str) -> str:
```
behnazh-w, line 177: Same comment as above regarding the return type.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -211,17 +211,12 @@ def extract_commit_from_version(git_obj: Git, version: str) -> tuple[str, str]:
...
-        logger.debug("No valid branch found for commit: %s", commit.hash)
-        return "", ""
+        return ""
 
-    return branch_name, commit.hash
+    return str(commit.hash)
 
 
-def find_commit_from_version_and_name(git_obj: Git, name: str, version: str) -> tuple[str, str]:
+def find_commit_from_version_and_name(git_obj: Git, name: str, version: str) -> str:
```
behnazh-w, line 219: Same comment regarding the return type.

src/macaron/repo_finder/commit_finder.py
```diff
@@ -211,17 +211,12 @@ def extract_commit_from_version(git_obj: Git, version: str) -> tuple[str, str]:
...
-        return "", ""
-
-    branch_name = _get_branch_of_commit(commit)
-    if not branch_name:
-        logger.debug("No valid branch found for commit: %s", commit.hash)
-        return "", ""
+        return ""
 
-    return branch_name, commit.hash
+    return str(commit.hash)
```
behnazh-w, line 216: What if `commit.hash` is `None`? Should we check for that?

src/macaron/repo_finder/commit_finder.py
```diff
@@ -276,26 +271,20 @@ def find_commit_from_version_and_name(git_obj: Git, name: str, version: str) ->
...
-        "Found tag %s with commit %s of branch %s for artifact version %s@%s",
+        "Found tag %s with commit %s for artifact version %s@%s",
         tag,
         hexsha,
-        branch_name,
         name,
         version,
     )
-    return branch_name, hexsha
+    return str(hexsha)
```
behnazh-w, line 287: Same comment. Can `hexsha` be `None`?

src/macaron/repo_finder/commit_finder.py
benmss: I don't believe it can be `None`. The `init` method in the class used by `GitPython` for "Blobs, Trees, Commits and Tags" has an `assert` statement that checks the `binsha` is 20 characters in length.

behnazh-w: In that case, I don't think creating a string object from `commit.hash` is necessary.

You can rewrite this line and the if statement above like this:
```suggestion
    return commit.hash if commit else None
```

src/macaron/repo_finder/commit_finder.py
benmss: The above applies here too as the underlying class is the same.

behnazh-w: So, we don't need to create a `str` object, right?

behnazh-w (approved):

PR #589, opened 2024-01-03
chore: explicitly git init with the `master` branch in tests requiring it

nathanwn:
Improves a unit test that fails on my machine.

The test fails due to the assumption that `git init` always initializes a repo with the branch `master`. Meanwhile, the git `defaultBranch` configured on my machine is not `master`. I think this git config is quite common since GitHub now defaults to `main` as the main branch.

I also reckon if the test is based on the precondition of "the repo should have a `master` branch", then it'd better be stated explicitly. This fix tries to achieve that.

### DIFF
--- a/tests/slsa_analyzer/mock_git_utils.py
+++ b/tests/slsa_analyzer/mock_git_utils.py
@@ -22,8 +22,13 @@ def initiate_repo(repo_path: str | os.PathLike, git_init_options: dict | None = 
 
     Parameters
     ----------
-    repo_path : str or os.PathLike
+    repo_path : str | os.PathLike
         The path to the target repo.
+
+    git_init_options : dict
+        Additional keyword arguments passed to the `git.Repo.init` method.
+        Each key is the name of the argument and each value is the corresponding value
+        for the argument.
 
     Returns
     -------
--- a/tests/slsa_analyzer/mock_git_utils.py
+++ b/tests/slsa_analyzer/mock_git_utils.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2024, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """
@@ -15,7 +15,7 @@ from macaron.database.table_definitions import Analysis, Component, Repository
 from macaron.slsa_analyzer.analyze_context import AnalyzeContext
 
 
-def initiate_repo(repo_path: str | os.PathLike) -> Git:
+def initiate_repo(repo_path: str | os.PathLike, git_init_options: dict | None = None) -> Git:
     """Init the repo at `repo_path` and return a Git wrapper of that repository.
 
     This function will create the directory `repo_path` if it does not exist.
@@ -30,6 +30,8 @@ def initiate_repo(repo_path: str | os.PathLike) -> Git:
     Git
         The wrapper of the Git repository.
     """
+    git_init_options = git_init_options or {}
+
     if not os.path.isdir(repo_path):
         os.makedirs(repo_path)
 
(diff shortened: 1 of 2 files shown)


benmss:
This is an interesting failure case. However, reliance on branch information in the commit finder (and tests) will be removed in PR #586. When that is merged (without major changes), I think this PR can be closed.

nathanwn:
Given the changes in #586, I will close this PR for now.

nathanwn:
I still get this error on my side. I don't think the change made in #586 solved the issue.

The result of the test case:

```
FAILED tests/repo_finder/test_commit_finder.py::test_commit_finder - AttributeError: 'IterableList' object has no attribute 'master'
```

Here is the line causing the issue: https://github.com/oracle/macaron/blob/744335d17a3c3485bbeecd94a22048c409a166d0/tests/repo_finder/test_commit_finder.py#L120.

To reproduce:

* Add the following to your `~/.gitconfig` file:

```
[init]
    defaultBranch = main
```

* Remove the mock repository used in the test:

```
rm -rf ./tests/repo_finder/mock_repos/commit_finder/sample_repo
```

* Re-run the test.

As I have mentioned in the initial PR description, this is not an uncommon config, especially since it is in line with [GitHub's default branch naming practice](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-branches#about-the-default-branch). On the other hand, I don't see why we don't want to explicitly initiate the test repo under the exact condition that the test relies on.

tests/slsa_analyzer/mock_git_utils.py
```diff
@@ -30,6 +30,8 @@ def initiate_repo(repo_path: str | os.PathLike) -> Git:
     Git
```
tromai, line 30: The documentation for the parameter `git_init_options` is missing.

tests/slsa_analyzer/mock_git_utils.py
```diff
@@ -15,7 +15,7 @@
 from macaron.slsa_analyzer.analyze_context import AnalyzeContext
 
 
-def initiate_repo(repo_path: str | os.PathLike) -> Git:
+def initiate_repo(repo_path: str | os.PathLike, git_init_options: dict | None = None) -> Git:
```
tromai, line 18: It seems to me that `git_init_options` contains the key value mapping that will be used as additional arguments for the `git init` command (according to the documentation [here](https://gitpython.readthedocs.io/en/3.1.37/reference.html#git.repo.base.Repo.init)). I wonder if it's worth using `dict[str,str]` as the type annotation?

tromai (changes requested):
Thanks for the changes. I have requested some changes on the documentation of the `initiate_repo` method.

tests/slsa_analyzer/mock_git_utils.py
nathanwn: We can't annotate it as `dict[str, str]` because this is a type error. The reason is that I later unpack this dictionary to keyword arguments for the `Repo.init` method, and some arguments take different types, not just `str` (e.g. as per the documentation, `mkdir` takes a `bool` value).

tests/slsa_analyzer/mock_git_utils.py
nathanwn: Thanks. Fixed in c78edd0734249d495a67c60f9d2be7d60d53b5f3.

nathanwn: Thanks. Fixed in c78edd0734249d495a67c60f9d2be7d60d53b5f3.

tests/slsa_analyzer/mock_git_utils.py
tromai: Ah I see. Was making a mistake in previous comment. It make senses to type annotate it as it is.

tromai (approved):
Thanks for the changes. LGTM.

behnazh-w (approved):

PR #591, opened 2024-01-05
test: fix the GitHub Actions workflow detection test

behnazh-w:
This PR fixes a bug in the GitHub Actions workflow detection test. The assertions in the `test_get_workflows` function were checking the return values of `list.sort()`, which is an in-place function and doesn't return anything. Instead we should compare the sorted lists.

This PR also removes `MacaronTestCase`, which is not necessary anymore.

### DIFF
--- a/tests/slsa_analyzer/ci_service/test_github_actions.py
+++ b/tests/slsa_analyzer/ci_service/test_github_actions.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2024, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """This module tests GitHub Actions CI service."""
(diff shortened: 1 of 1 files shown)


tests/slsa_analyzer/ci_service/test_github_actions.py
```diff
@@ -1,87 +1,101 @@
...
+    workflows = github_actions.get_workflows(str(ga_has_build_kws))
+    workflows.sort()
+    assert workflows == expect
+
+    # Test GitHub Actions workflows that do not contain build commands.
+    expect = [str(path) for path in ga_no_build_kws.joinpath(".github", "workflows").glob("*")]
+    expect.sort()
+    workflows = github_actions.get_workflows(str(ga_no_build_kws))
+    workflows.sort()
+    assert workflows == expect
```
nathanwn, line 98: This could also be
```suggestion
    expect = sorted([str(path) for path in ga_no_build_kws.joinpath(".github", "workflows").glob("*")])
    assert sorted(github_actions.get_workflows(str(ga_no_build_kws)))
    assert workflows == expect
```
which is arguably a bit closer to the original, but I also don't see any problem with using a separate call to `.sort()`.

nathanwn:
I think most test functions in the file are quite repetitive, repeating the same test logic for a few cases within the same test function, and could be refactored into table-driven tests with `@pytest.mark.parametrize`. However, if you don't think it is super necessary right now, we can leave the decision for later. I'd approve the PR then. Let me know.
Otherwise, I don't think there is anything else for this PR.

tests/slsa_analyzer/ci_service/test_github_actions.py
behnazh-w: See commit [416179a](https://github.com/oracle/macaron/pull/591/commits/416179a5d4503986e9cee8179b7e3917909a2080)

behnazh-w:
See commit [416179a](https://github.com/oracle/macaron/pull/591/commits/416179a5d4503986e9cee8179b7e3917909a2080)

nathanwn (approved):
LGTM. Thanks for the PR.

PR #592, opened 2024-01-05
feat: generate Verification Summary Attestation

nathanwn:
This PR adds the Verification Summary Attestation generation feature to Macaron.

During the execution of the `macaron verify-policy` command, Macaron may generate a Verification Summary Attestation (VSA) with the metadata of the verification, if applicable.


tests/vsa/compare_vsa.py
```diff
@@ -0,0 +1,265 @@
...
+log_passed = log_with_tag("PASSED")
+
+
+def log_diff(result: object, expected: object) -> None:
+    """Pretty-print the diff of two Python objects."""
+    log(
+        "".join(
+            context_diff(pformat(result), pformat(expected), "result", "expected"),
+        ),
+    )
```
nathanwn, line 44: Here, I define new log functions instead of using the built-in `logging` module for better control.

* I define some additional custom log tags (`PASSED` and `FAILED`) to indicate the compare result. These tags are not available by default in the `logging` module.
* This also allows the use of f-string without giving lint errors.

The choice of `stderr` is mostly due to convention: the output of this script is for debugging purposes mainly.

docs/source/pages/vsa.rst
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+
+-------
+Example
+-------
+
+The following is an example of Macaron VSA generated from verification on the `slsa-verifier <https://github.com/slsa-framework/slsa-verifier>`_ repository.
+
```
behnazh-w, line 59: Please explain the content of the VSA, what each attribute means, what the included policy has checked, and what should people do with it.

scripts/dev_scripts/integration_tests.sh
```diff
@@ -636,6 +638,32 @@ POLICY_EXPECTED=$WORKSPACE/tests/policy_engine/expected_results/policy_report.js
...
+# Running Macaron without config files
+echo -e "\n=================================================================================="
+echo "Run integration tests for VSA generation"
+echo -e "==================================================================================\n"
+TEST_CASE_DIR="$WORKSPACE/tests/vsa/integration/github_slsa-framework_slsa-verifier"
+OUTPUT_DIR="$TEST_CASE_DIR/output"
+
+rm -rf "$OUTPUT_DIR"  # Make sure we regenerate a fresh database every time.
+macaron --output "$OUTPUT_DIR" analyze \
+    --repo-path "https://github.com/slsa-framework/slsa-verifier" \
```
behnazh-w, line 651: Analyzing `slsa-verifier` in general takes a long time because its release artifacts are quite large and that's why we cache the results and run the policy engine tests without running the `analyze` command again. Note that we already analyze `slsa-verifier` [here](https://github.com/oracle/macaron/blob/8334460935909d7fa8b05ee32fad02107185176d/scripts/dev_scripts/integration_tests.sh#L613). How about using the existing results?

Also, to keep the script consistent, we should either use `$RUN_MACARON` here and follow the same naming conventions for the resources and expected results or change the rest of the script to be consistent with the VSA test.

docs/source/pages/output_files.rst
```diff
@@ -129,37 +133,64 @@ to the directory:
...
+
+    output/
+    └── vsa.intoto.jsonl
+
+
+Users can manually inspect the VSA generated by Macaron with the following command:
+
+.. code-block:: bash
+
+    cat output/vsa.intoto.json | jq -r '.payload' | base64 -d
```
behnazh-w, line 157:
```suggestion
    cat output/vsa.intoto.jsonl | jq -r '.payload' | base64 -d | jq
```

scripts/dev_scripts/integration_tests.sh
```diff
@@ -19,6 +21,7 @@ UPDATE=0
 if [ $# -eq 3 ] && [ "$3" == "--update" ] ; then
     echo "Updating the expected results to match those currently produced by Macaron."
     UPDATE=1
+    COMPARE_VSA="$COMPARE_VSA --update"
```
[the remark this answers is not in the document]
nathanwn: The `compare_vsa.py` script has a `--update` flag to update the expected VSA payload files. 

On this line, the `--update` flag of the `compare_vsa.py` script is turned on if the `integration_tests.sh` script is run with the `--update` flag.

scripts/dev_scripts/integration_tests.sh
```diff
@@ -636,6 +638,32 @@ POLICY_EXPECTED=$WORKSPACE/tests/policy_engine/expected_results/policy_report.js
...
+check_or_update_expected_output "$COMPARE_JSON_OUT" \
+    "$OUTPUT_DIR/reports/github_com/slsa-framework/slsa-verifier/slsa-verifier.json" \
+    "$TEST_CASE_DIR/slsa-verifier.json" || log_fail
+macaron --output "$OUTPUT_DIR" verify-policy \
+    --database "$OUTPUT_DIR/macaron.db" \
+    --file "$TEST_CASE_DIR/policy.dl" || log_fail
+check_or_update_expected_output "$COMPARE_POLICIES" \
+    "$OUTPUT_DIR/policy_report.json" \
+    "$TEST_CASE_DIR/policy_report.json" || log_fail
+python3 "$COMPARE_VSA" \
```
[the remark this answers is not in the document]
nathanwn: Expected VSA payload files differ from the expected result files we have had so far since they only store the decoded payload of the VSA, not the whole VSA itself. Therefore, copying the result file to update the expected file would not work. I implemented the logic to update the expected payload file within the `compare_vsa.py` script. By default, the script does the comparison. If it runs with the `--update` flag, it switches to "update" mode and instead updates the expected payload file.

scripts/dev_scripts/integration_tests.sh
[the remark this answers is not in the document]
behnazh-w: I see, thanks. How about moving it to `check_or_update_expected_output` so that we can consistently use `check_or_update_expected_output` like other tests?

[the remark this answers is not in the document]
nathanwn: I'm not sure how this could be achieved in the `check_or_update_expected_output` function, because the function either runs `python <script> <result_file> <expected_file>` or `cp <result_file> <expected_file>`. When used, this function also comes before the script name, so there is no way to pass the `--update` flag into the `compare_vsa.py` script as the flag must come after the script name.

scripts/dev_scripts/integration_tests.sh
nathanwn: If you think it's OK to reuse the cached result, then I can move this test outside of the isolated directory. My initial intention is to have this integration test isolated to make sure there is only one software component in the database. Meanwhile, there are already some unit tests that take care of "different states of the database", so I think using the cached result is fine.
`$RUN_MACARON` not being used was a side-effect of the initial design of the test to be in a separate directory. The `$RUN_MACARON` command fixes the location of the `output` directory, whereas my first intention was to have a separate `output` directory altogether just for this test.

scripts/dev_scripts/integration_tests.sh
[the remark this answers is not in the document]
behnazh-w: How about checking the script name and passing `--update` instead of running `cp` if the script is `COMPARE_VSA`?

scripts/dev_scripts/integration_tests.sh
behnazh-w: I agree that ideally we should avoid using cached results, but this is purely to reduce the CI runtime. I think it's OK to use the cached result and add a comment to clarify that.

src/macaron/__main__.py
```diff
@@ -144,7 +146,20 @@ def verify_policy(verify_policy_args: argparse.Namespace) -> int:
...
 
-        result = run_policy_engine(verify_policy_args.database, verify_policy_args.file)
+        with open(verify_policy_args.file, encoding="utf-8") as file:
+            policy_content = file.read()
+
+        result = run_policy_engine(verify_policy_args.database, policy_content)
+        vsa = generate_vsa(policy_content=policy_content, policy_result=result)
+        if vsa is not None:
+            vsa_filepath = os.path.join(global_config.output_path, "vsa.intoto.jsonl")
+            logger.info("Generating a VSA to %s", vsa_filepath)
```
behnazh-w, line 156:
```suggestion
            logger.info("Generating the Verification Summary Attestation (VSA) to %s.", vsa_filepath)
            logger.info(
                "To decode and see the VSA payload run `cat %s | jq -r '.payload' | base64 -d | jq`.",
                vsa_filepath,
            )
```

scripts/dev_scripts/integration_tests.sh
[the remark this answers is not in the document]
nathanwn: Hmm, I guess I could do that, although I'm a bit unsure of whether it is indeed better. I guess it provides a more consistent interface for creating tests, with the trade-off of a bit more complex implementation of the `check_or_update_expected_output` function, and aiming for easier/more convenient test addition/creation is a good idea.

src/macaron/vsa/vsa.py
```diff
@@ -0,0 +1,248 @@
...
+class Vsa(TypedDict):
+    """The Macaron Verification Summary Attestation.
+
+    For reference, see:
+    * `SLSA <https://slsa.dev/spec/v1.0/verification_summary>`_.
+    * `in-toto <https://github.com/in-toto/attestation/blob/main/spec/predicates/vsa.md>`_.
+    """
+
+    #: The payload type. Following in-toto, this is always ``application/vnd.in-toto+json``.
+    payloadType: str  # noqa: N815
```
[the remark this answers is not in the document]
nathanwn: Should I move this comment
```
# Disable "N815:mixedCase variable in class scope" to conform with in-toto naming conventions.`
```
to the top of the file, since this applies to a lot of fields in the VSA? or do you think explaining for the first attribute is enough? Let me know.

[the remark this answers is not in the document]
behnazh-w: Yes, that's a good idea.

src/macaron/vsa/vsa.py
```diff
@@ -0,0 +1,248 @@
...
+    subject: list[dict]
+    predicateType: str  # noqa: N815
+    predicate: VsaPredicate
+
+
+class VsaPredicate(TypedDict):
+    """The 'predicate' field in the Statement layer of a Macaron VSA.
+
+    For reference, see:
+    * in-toto Predicate layer specification:
```
behnazh-w, line 51:
```suggestion

    For reference, see:
    * in-toto Predicate layer specification:
 ```

File: src/macaron/vsa/vsa.py [hunk 10]
```diff
@@ -0,0 +1,248 @@
...
+    """The Statement layer of a Macaron VSA.
+
+    For reference, see:
+    * in-toto Statement layer specification: https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md.
+    """
+
+    _type: str
+    subject: list[dict]
+    predicateType: str  # noqa: N815
+    predicate: VsaPredicate
```
behnazh-w, line 44: The documentation for these attributes is missing.

src/macaron/vsa/vsa.py
```diff
@@ -0,0 +1,248 @@
...
+    * SLSA VSA predicate schema:
+    https://slsa.dev/spec/v1.0/verification_summary#schema.
+    """
+
+    #: Identity of the verifier, as a tool carrying out the verification.
+    verifier: Verifier
+
+    #: The timestamp when the verification occurred.
+    #: The field has the type
+    # `Timestamp <https://github.com/in-toto/attestation/blob/main/spec/v1/field_types.md#Timestamp>`_.
```
behnazh-w, line 62:
```suggestion
    #: `Timestamp <https://github.com/in-toto/attestation/blob/main/spec/v1/field_types.md#Timestamp>`_.
```

docs/source/pages/output_files.rst
nathanwn: Thanks. See e4d21c6515c6c67ac3cb255e99806ead40994d9d.

docs/source/pages/vsa.rst
nathanwn: I have added more details in 7a69d314e65e31a44397bf47e0478385e6e14c09.

src/macaron/__main__.py
nathanwn: See 9b6338c637d991af44b97b405f165e3bb4170244.

src/macaron/vsa/vsa.py
```diff
@@ -0,0 +1,248 @@
...
+        ],
+        predicateType="https://slsa.dev/verification_summary/v1",
+        predicate=VsaPredicate(
+            verifier=Verifier(
+                id="https://github.com/oracle/macaron",
+                version={
+                    "macaron": importlib_metadata.version("macaron"),
+                },
+            ),
+            timeVerified=datetime.utcnow().isoformat("T") + "Z",
```
behnazh-w, line 145: Instead of using concatenating concatenation like this, let's create an aware datetime object and call `isoformat()`.

Note that [the RFC allows](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6) `time-numoffset`, so we don't need to concatenate `Z`.

```suggestion
            timeVerified=datetime.now(tz=datetime.timezone.utc).isoformat()
```

src/macaron/vsa/vsa.py
```diff
@@ -0,0 +1,251 @@
...
+
+    #: URI that identifies the resource associated with the software component being verified.
+    #: This field has the type
+    #: `ResourceURI <https://github.com/in-toto/attestation/blob/main/spec/v1/field_types.md#ResourceURI>`_.
+    #: Currently, this has the same value as the subject of the VSA, i.e. the PURL of
+    #: the software component being verified against.
+    resourceUri: str  # noqa: N815
+
+    #: The policy that the subject software component was verified against.
+    #: This field has the type
```
behnazh-w, line 73: Same comment regarding `type`.

src/macaron/vsa/vsa.py
nathanwn: Thanks for the suggestion. See a59da72469db757674fb2dfe5af3851e8057c29c.

`pyupgrade` interestingly made the following change:

```diff
-            timeVerified=datetime.datetime.now(tz=datetime.timezone.utc).isoformat(),
+            timeVerified=datetime.datetime.now(tz=datetime.UTC).isoformat(),
```

behnazh-w: Oh nice. I didn't know the change in Python 3.11.

src/macaron/vsa/vsa.py
```diff
@@ -0,0 +1,251 @@
...
+    #: `ResourceDescriptor <https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md>`_.
+    policy: dict[str, Any]
```
behnazh-w, line 78: This attribute needs a more detailed documentation because it is not clear straightaway why we use `policy` as the `ResourceDescriptor`. Also I'm not sure I understand the choice of `dict[str, Any]` type here. Can you please clarify?

src/macaron/vsa/vsa.py
```diff
@@ -0,0 +1,251 @@
...
+    #: Identity of the verifier, as a tool carrying out the verification.
+    verifier: Verifier
+
+    #: The timestamp when the verification occurred.
+    #: The field has the type
+    # `Timestamp <https://github.com/in-toto/attestation/blob/main/spec/v1/field_types.md#Timestamp>`_.
+    timeVerified: str  # noqa: N815
+
+    #: URI that identifies the resource associated with the software component being verified.
+    #: This field has the type
```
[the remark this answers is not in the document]
nathanwn: Hmm, what do you think about wording it like so?

```diff
-    #: The field has the type
+    #: The field is a
    #: `Timestamp <https://github.com/in-toto/attestation/blob/main/spec/v1/field_types.md#Timestamp>`_.
```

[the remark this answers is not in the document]
behnazh-w: Makes sense, thanks.

src/macaron/vsa/vsa.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+class Verifier(TypedDict):
+    """The 'verifier' field within the Macaron VSA predicate field.
+
+    This field provides the identity of the verifier, as well as the versioning details of its components.
+    """
+
+    #: The identity of the verifier as a value of type
+    #:   `TypeURI <https://github.com/in-toto/attestation/blob/main/spec/v1/field_types.md#TypeURI>`_.
```
behnazh-w, line 96: Same comment regarding the type.

src/macaron/vsa/vsa.py
nathanwn: See 4e168f83169cd10abd9a95a52761d1b05580060b.

src/macaron/vsa/vsa.py
```diff
@@ -0,0 +1,248 @@
...
+
+    #: The payload of the VSA, base64 encoded.
+    payload: str
+
+
+class VsaStatement(TypedDict):
+    """The Statement layer of a Macaron VSA.
+
+    For reference, see:
+    * in-toto Statement layer specification: https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md.
```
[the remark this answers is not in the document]
nathanwn: See 4e168f83169cd10abd9a95a52761d1b05580060b.

src/macaron/vsa/vsa.py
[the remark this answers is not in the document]
nathanwn: See 24896445bedb585bf631f884ae2b8986658e8cc3.

docs/source/pages/vsa.rst
```diff
@@ -0,0 +1,170 @@
...
+
+---------
+Use cases
+---------
+
+The use cases of Macaron VSAs includes, but not limited to:
+
+- **Caching verification results**: It could be expensive or inconvenient to run a full Macaron verification in certain circumstances. A VSA helps with caching and reusing verification results.
+- **Enabling delegated verification**: This allows software consumers to make use of verification results from another party.
```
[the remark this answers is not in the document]
nathanwn: See aa5a9553fac86c6dfe98633bd4c3a7ed132d568e.

src/macaron/vsa/vsa.py
```diff
@@ -0,0 +1,251 @@
...
+        the policy applies to, or ``None`` according to the aforementioned conditions.
+    """
+    component_violates_policy_facts = policy_result.get("component_violates_policy", [])
+    component_satisfies_policy_facts = policy_result.get("component_satisfies_policy", [])
+
+    # key: PURL; value: result with the highest component id
+    component_results: dict[str, tuple[int, VerificationResult]] = {}
+
+    for component_id_string, purl, _ in component_violates_policy_facts:
+        component_id = int(component_id_string)
```
behnazh-w, line 197: To catch unexpected `component.id`s, e.g., if it is changed to UUID but this code is not updated, it would be good to check for `ValueError` and log an error.

src/macaron/vsa/vsa.py
```diff
@@ -0,0 +1,251 @@
...
+        if purl not in component_results:
+            component_results[purl] = (component_id, VerificationResult.FAILED)
+        else:
+            current_component_id, _ = component_results[purl]
+            if component_id > current_component_id:
+                component_results[purl] = (component_id, VerificationResult.FAILED)
+    for component_id_string, purl, _ in component_satisfies_policy_facts:
+        component_id = int(component_id_string)
```
behnazh-w, line 205: Same comment regarding catching `ValueError`.

src/macaron/vsa/vsa.py
```diff
@@ -0,0 +1,251 @@
...
+        if purl not in component_results:
+            component_results[purl] = (component_id, VerificationResult.PASSED)
+        else:
+            current_component_id, _ = component_results[purl]
+            if component_id > current_component_id:
+                component_results[purl] = (component_id, VerificationResult.PASSED)
+
+    if len(component_results) != 1:
```
behnazh-w, line 213: It would be good to differentiate `len(component_results) == 0` vs `len(component_results) > 1`and add a useful info log for the latter.

src/macaron/vsa/vsa.py
```diff
@@ -0,0 +1,251 @@
...
+
+    payload = create_vsa_statement(
+        subject_purl=subject_purl,
+        policy_content=policy_content,
+        verification_result=verification_result,
+    )
+
+    return Vsa(
+        payloadType="application/vnd.in-toto+json",
+        payload=base64.b64encode(json.dumps(payload).encode()).decode("ascii"),
```
behnazh-w, line 253: Looks like the errors raised by `base64.b64encode()`, and  `json.dumps` are not caught here?

src/macaron/vsa/vsa.py
nathanwn: See bcfea7f5aea0124c0421336ff1c85ac131deb0b6.

src/macaron/vsa/vsa.py
nathanwn: Hmm, I should probably extract this out to an encode function and do all error handling there.

src/macaron/vsa/vsa.py
nathanwn: This function currently relies on the assumption that the component id column is auto-incremented to pick the latest analysis results. I don't think UUID is applicable here since there is no ordering on UUID, and if we switch to UUID then the implementation will need to change and this function will most likely become irrelevant.

behnazh-w: Yes that's right. I just want to be precautious and log errors if the assumption made in this implementation is violated.

src/macaron/vsa/vsa.py
nathanwn: This field is a `ResourceDescriptor` following [SLSA VSA Predicate schema](https://slsa.dev/spec/v1.0/verification_summary#schema). Currently, it holds only one inner field of `"content"` which holds the Souffle policy code.

I type-annotated it as `dict[str, Any]` since I didn't care to annotate it further, i.e. as long as it is a dictionary with `str` keys, I was fine with it.

I can probably create an additional `Policy(TypedDict)` class with a single `"content"` subfield and annotate the `policy` here with it.

src/macaron/vsa/vsa.py
nathanwn: See 6870a3ab50fc1342a3727cb0f6aa51ecc0933e33.

src/macaron/vsa/vsa.py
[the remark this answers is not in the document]
nathanwn: See 6870a3ab50fc1342a3727cb0f6aa51ecc0933e33.

src/macaron/vsa/vsa.py
nathanwn: See 60ca8cf3970350aeb6df1a6b260a3d4189571969.

src/macaron/vsa/vsa.py
nathanwn: See 60ca8cf3970350aeb6df1a6b260a3d4189571969.

scripts/dev_scripts/integration_tests.sh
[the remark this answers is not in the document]
nathanwn: See e621ec3a2711205766450727cdd4736d2d70dcb3.

src/macaron/vsa/vsa.py
nathanwn: See 6870a3ab50fc1342a3727cb0f6aa51ecc0933e33.

scripts/dev_scripts/integration_tests.sh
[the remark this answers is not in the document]
nathanwn: Adjusted in e621ec3a2711205766450727cdd4736d2d70dcb3.

src/macaron/vsa/vsa.py
nathanwn: Where do you think I should put the encode function? For simplicity, I would just put it in the same file, before we decide what to do with the `intoto` package, currently under the `slsa_analyzer` package.

src/macaron/vsa/vsa.py
nathanwn: See 450608c15d80c7b731eccba3c177a53872983b5.

src/macaron/vsa/vsa.py
behnazh-w: I think at this point we don't need a function and hoisting the code and handling errors should be fine.

tests/vsa/compare_vsa.py
behnazh-w: I prefer to use the standard `logging` library. You can achieve the points you mentioned with `logging` too:

```python
logging.basicConfig(format="[%(filename)s %(tag)s] %(message)s")

logger.info("The payload of the generated VSA matches the expected payload", extra={"tag": "PASSED"})
```

Will log

```
[compare_vsa.py PASSED] The payload of the generated VSA matches the expected payload.
```

I also prefer to add the line info
```
logging.basicConfig(format="[%(filename)s:%(lineno)s %(tag)s] %(message)s")
```
logs
```
[compare_vsa.py:284 PASSED] The payload of the generated VSA matches the expected payload.
```

src/macaron/vsa/vsa.py
nathanwn: See d5c6a7935e75d7d5cc01e34942a2d2f645fd6830.

tests/vsa/compare_vsa.py
nathanwn: Thanks for the suggestion. See my change in 52c4704d5542cebd35fa3a6de9709781d529e938.

src/macaron/vsa/vsa.py
nathanwn: See my change in ea5aa606f38abc58d8d91378a6af9e78a7ec8e8b.

scripts/dev_scripts/integration_tests.sh
nathanwn: See my adjustments in a272ff9c0cc09840ac87a9e9b1a402a1733a3172.

docs/source/pages/vsa.rst
```diff
@@ -0,0 +1,170 @@
...
+        component_id
+      ).
+
+This policy enforces the subject software component to pass 3 Macaron checks:
+
+* ``mcn_build_as_code_1``
+* ``mcn_provenance_level_three_1``
+* ``mcn_provenance_available_1``
+
+For more details on using the Macaron VSA generation feature and inspecting the resulting VSA, please refer to the :ref:`Output File Guide <output_files_guide>`.
```
behnazh-w, line 170: Please add a reference to the "Output files of macaron verify-policy" section directly.

src/macaron/__main__.py
```diff
@@ -144,7 +146,20 @@ def verify_policy(verify_policy_args: argparse.Namespace) -> int:
...
 
-        result = run_policy_engine(verify_policy_args.database, verify_policy_args.file)
+        with open(verify_policy_args.file, encoding="utf-8") as file:
+            policy_content = file.read()
+
+        result = run_policy_engine(verify_policy_args.database, policy_content)
+        vsa = generate_vsa(policy_content=policy_content, policy_result=result)
+        if vsa is not None:
+            vsa_filepath = os.path.join(global_config.output_path, "vsa.intoto.jsonl")
+            logger.info("Generating the Verification Summary Attestation (VSA) to %s.", vsa_filepath)
```
behnazh-w, line 156: Can you please also add a log to help users decode and read the VSA output?
```
            logger.info(
                "To decode and see the payload run `cat %s | jq -r '.payload' | base64 -d | jq`.",
                vsa_filepath,
            )
```

src/macaron/vsa/vsa.py
behnazh-w: This line still contains `as a value of type`.

src/macaron/__main__.py
nathanwn: Ah, I think I missed this in your suggestion earlier. Thanks for the reminder.

src/macaron/__main__.py
nathanwn: See also 2f3e53480d5cf0b279659c6d8a4d173193c6790b.

docs/source/pages/vsa.rst
```diff
@@ -0,0 +1,170 @@
...
+    Details about the policy that the subject software component was verified against.
+
+* ``policy.content``: string
+    The Souffle datalog policy used for verification, in plain text.
+
+* ``verificationResult``: string, either ``"PASSED"`` or ``"FAILED"``
+    The verification result. The result of ``"PASSED"`` means the subject software component conforms to the policy.
+
+* ``verificationResult``: array (`SlsaResult`_), required
+    Indicates the highest level of each SLSA track verified for the software component (and not its dependencies), or “FAILED” if policy verification failed.
```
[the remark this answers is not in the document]
nathanwn: See 977683d96907d33d068d2dfd3791a82a25e20f17.

src/macaron/vsa/vsa.py
nathanwn: See d2d8bbf31e8d676035776e73618c64fbc4ab4a88.

docs/source/pages/vsa.rst
```diff
@@ -0,0 +1,170 @@
...
+
+
+
+* ``_type``: string (`TypeURI`_)
+    Identifier for the schema of the Statement layer. This follows `in-toto v1 Statement layer schema <https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md>`_ and is always ``https://in-toto.io/Statement/v1``.
+
+* ``subject``: array of `ResourceDescriptor`_ objects
+    Subjects of the VSA. Each entry is a software component being verified by Macaron.
+
+    *Note: In the current version of Macaron, this field only contains one single software component, identified by a `PackageURL`_.*
```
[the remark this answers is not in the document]
nathanwn: The reference was defined [here](https://github.com/oracle/macaron/pull/592/files/02c2dd2ad8814a94ae88b76bba2c000016f94d6a#diff-c1f297c80595821b42e7c9ea0b0d00cd8f69c2ebff27293fee58b5f8bf67a3ddR27) earlier in the same file.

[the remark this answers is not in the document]
nathanwn: Oh, I think I know what happened. Apparently, we can't nest a reference in an italic block in RST. Fixed in 5d7cd6f1279cd0f82ec366f1423cee756cb8a925.

docs/source/pages/vsa.rst
nathanwn: Adjusted in 3ac8eafe8cd6260a3e991e2e8afdeb39bef97760.

behnazh-w (approved):

PR #598, opened 2024-01-11
test: add docker integration test for VSA generation feature

nathanwn:
Adds one additional docker integration test case for the VSA generation feature.


behnazh-w:
Thanks for the test. Can you please change the scope of the PR title to `test:`?

behnazh-w (approved):

PR #599, opened 2024-01-15
feat: support tox to publish artifacts to PyPI

behnazh-w:
This PR adds `tox` as a tool to publish artifacts for Python packages.


benmss (approved):
I have confirmed that it works and therefore approve this change.

PR #605, opened 2024-01-17
chore: improve token import

benmss:
Changes the way in which the GitHub token is input into Macaron.

### DIFF
--- a/scripts/release_scripts/run_macaron.sh
+++ b/scripts/release_scripts/run_macaron.sh
@@ -48,6 +48,7 @@ if [[ -z ${MACARON_IMAGE_TAG:-} ]]; then
     MACARON_IMAGE_TAG="latest"
 fi
 
+# This file is used to store token environment variables that will later be read by Macaron.
 TOKEN_FILE=".macaron_env_file"
 
 IMAGE="ghcr.io/oracle/macaron"
@@ -558,3 +559,5 @@ docker run \
     "${IMAGE}:${MACARON_IMAGE_TAG}" \
     "${entrypoint[@]}" \
     "${macaron_args[@]}"
+
+rm -f "$TOKEN_FILE"
--- a/scripts/release_scripts/run_macaron.sh
+++ b/scripts/release_scripts/run_macaron.sh
@@ -48,6 +48,8 @@ if [[ -z ${MACARON_IMAGE_TAG:-} ]]; then
     MACARON_IMAGE_TAG="latest"
 fi
 
+TOKEN_FILE=".macaron_env_file"
+
 IMAGE="ghcr.io/oracle/macaron"
 
 # Workspace directory inside of the container.
@@ -255,14 +257,14 @@ function mount_file() {
     mounts+=("-v" "${file_on_host}:${file_in_container}:${mount_option}")
 }
 
-# Handle environment token
-if [[ "${GITHUB_TOKEN+1}" ]]; then
-    token_file=".macaron_env_file"
-    echo ${GITHUB_TOKEN} > ${token_file}
-    mount_file "macaron_env_file" $token_file ${MACARON_WORKSPACE}/$token_file "rw,Z"
-else
-    log_warn "GitHub token not found in environment variables."
-fi
+# Handle tokens.
+set +u
+echo "" > ${TOKEN_FILE}
+echo "GITHUB_TOKEN=${GITHUB_TOKEN}" >> ${TOKEN_FILE}
+echo "MCN_GITLAB_TOKEN=${MCN_GITLAB_TOKEN}" >> ${TOKEN_FILE}
+echo "MCN_SELF_HOSTED_GITLAB_TOKEN=${MCN_SELF_HOSTED_GITLAB_TOKEN}" >> ${TOKEN_FILE}
+mount_file "macaron_env_file" ${TOKEN_FILE} ${MACARON_WORKSPACE}/${TOKEN_FILE} "rw,Z"
+set -u
 
 # Parse main arguments.
 while [[ $# -gt 0 ]]; do
@@ -550,8 +552,6 @@ docker run \
     --rm -i "${tty[@]}" \
     -e "USER_UID=${USER_UID}" \
     -e "USER_GID=${USER_GID}" \
-    -e MCN_GITLAB_TOKEN \
-    -e MCN_SELF_HOSTED_GITLAB_TOKEN \
     "${proxy_vars[@]}" \
     "${prod_vars[@]}" \
     "${mounts[@]}" \
--- a/scripts/release_scripts/run_macaron.sh
+++ b/scripts/release_scripts/run_macaron.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2023 - 2024, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 # This script runs the Macaron Docker image.
@@ -254,6 +254,15 @@ function mount_file() {
     file_on_host=$(to_absolute_path "$file_on_host")
     mounts+=("-v" "${file_on_host}:${file_in_container}:${mount_option}")
 }
+
+# Handle environment token
+if [[ "${GITHUB_TOKEN+1}" ]]; then
+    token_file=".macaron_env_file"
+    echo ${GITHUB_TOKEN} > ${token_file}
+    mount_file "macaron_env_file" $token_file ${MACARON_WORKSPACE}/$token_file "rw,Z"
+else
+    log_warn "GitHub token not found in environment variables."
+fi
 
 # Parse main arguments.
 while [[ $# -gt 0 ]]; do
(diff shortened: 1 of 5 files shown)


src/macaron/__main__.py
```diff
@@ -221,6 +222,26 @@ def main(argv: list[str] | None = None) -> None:
...
     """
+    # Handle presence of tokens.
+    token_file = "./.macaron_env_file"  # nosec B105
+    if os.path.exists(token_file):
+        with open(token_file, encoding="utf-8") as file:
+            for line in file:
+                print(f"Line: {line}")
+                if not line or "=" not in line:
+                    continue
+                key, value = line.rstrip().split("=")
```
nicallen, line 233: This will throw ValueError on unpack if there is more than one `=`.
Should change it to `split("=", 1)` so that it always splits on the first `=` just in case a value containing additional `=` characters is ever valid.

src/macaron/__main__.py
```diff
@@ -221,6 +222,26 @@ def main(argv: list[str] | None = None) -> None:
         If ``argv`` is ``None``, argparse automatically looks at ``sys.argv``.
         Hence, we set ``argv = None`` by default.
     """
+    # Handle presence of tokens.
+    token_file = "./.macaron_env_file"  # nosec B105
+    if os.path.exists(token_file):
+        with open(token_file, encoding="utf-8") as file:
+            for line in file:
+                print(f"Line: {line}")
```
nicallen, line 230: Remove print

src/macaron/__main__.py
```diff
@@ -221,6 +222,26 @@ def main(argv: list[str] | None = None) -> None:
...
+                if key and value:
+                    if key == "GITHUB_TOKEN":
+                        global_config.gh_token = value
+                    if key == "MCN_GITLAB_TOKEN":
+                        global_config.gl_token = value
+                    if key == "MCN_SELF_HOSTED_GITLAB_TOKEN":
+                        global_config.gl_self_host_token = value
+        # Overwrite file contents as deleting won't work when in a container.
+        with open(token_file, "w", encoding="utf-8") as file:
+            file.write("")
```
nicallen, line 243: Opening the file for writing erases the content already, writing an empty string is redundant.
Can change to:
```
with open(token_file, "w", encoding="utf-8"):
    pass
```

src/macaron/slsa_analyzer/git_service/gitlab.py
```diff
@@ -81,7 +84,9 @@ def construct_clone_url(self, url: str) -> str:
 
         # Construct clone URL from ``urlparse`` result, with or without an access token.
         # https://docs.gitlab.com/ee/gitlab-basics/start-using-git.html#clone-using-a-token
-        access_token = os.environ.get(self.access_token_env_name)
+        access_token: str | None = self.token_function()
+        if not access_token:
+            access_token = os.environ.get(self.access_token_env_name)
```
nicallen, line 89: Similar to the other token, I think it is preferable to move reading the env var to where the token file is read, making this class rely solely on the provided token_function for getting the token.

src/macaron/__main__.py
```diff
@@ -221,6 +214,38 @@ def main(argv: list[str] | None = None) -> None:
...
+                    if key == "GITHUB_TOKEN":
+                        global_config.gh_token = value
+                    if key == "MCN_GITLAB_TOKEN":
+                        global_config.gl_token = value
+                    if key == "MCN_SELF_HOSTED_GITLAB_TOKEN":
+                        global_config.gl_self_host_token = value
+        # Overwrite file contents as deleting won't work when in a container.
+        with open(token_file, "w", encoding="utf-8") as file:
+            pass
+    else:
```
nicallen, line 235: It' would probably make more sense to fall back to checking env vars on a per-variable basis rather than either all from the file or all from env vars. Also, the current logic has a path where the github token isn't set but no error is reported (token file exists but doesn't contain a GITHUB_TOKEN line).

benmss: I agree the logic should be improved to handle the existence of the GITHUB_TOKEN properly. However, for the other two tokens it seems to me that the only situation where that matters is when a user (or developer) is manually creating the `token_file` before each run of Macaron, and running Macaron without the `run_macaron.sh` script. Otherwise, if they are in the environment variables they will be in the file, and if they are not, it doesn't matter.

src/macaron/__main__.py
```diff
@@ -187,12 +187,13 @@ def perform_action(action_args: argparse.Namespace) -> None:
             sys.exit(verify_policy(action_args))
 
         case "analyze":
-            # Check that the GitHub token is enabled.
-            gh_token = os.environ.get("GITHUB_TOKEN")
-            if not gh_token:
-                logger.error("GitHub access token not set.")
-                sys.exit(os.EX_USAGE)
-            global_config.gh_token = gh_token
+            # If the GitHub token has not already been read, try to read it here.
```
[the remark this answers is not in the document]
benmss: I did not read the last line of your comment correctly and removed the error check entirely from this section. It will be added back in in the next commit.

src/macaron/__main__.py
nicallen: True, for the current use case as a side channel for `run_macaron.sh` to pass data to `macaron`, how it is used is within our control so a mixed situation wouldn't come up. I was just thinking about what makes the most sense as the interface of the Macaron python program in isolation, if one ever needed to use it directly.

nicallen (approved):

src/macaron/__main__.py
```diff
@@ -221,6 +217,26 @@ def main(argv: list[str] | None = None) -> None:
         If ``argv`` is ``None``, argparse automatically looks at ``sys.argv``.
         Hence, we set ``argv = None`` by default.
     """
+    # Handle presence of token file.
+    token_file = "./.macaron_env_file"  # nosec B105
```
behnazh-w, line 221: Instead of a hard-coded path, we can use `os.getcwd()`. Using `/` makes the implementation platform-specific. Instead we can let the `os` package to resolve the path.

Also, please add documentation on how this file is created and the difference in setting token when running Macaron as a container vs. Python package.

```suggestion
    token_file = os.path.join(os.getcwd(), ".macaron_env_file")
```

Looks like with this change Bandit doesn't report a FP hard-coded password error, so we can remove the [`# nosec B105`](https://bandit.readthedocs.io/en/latest/plugins/b105_hardcoded_password_string.html) comment.

src/macaron/__main__.py
```diff
@@ -221,6 +217,26 @@ def main(argv: list[str] | None = None) -> None:
...
+    token_dict = {}
+    if os.path.exists(token_file):
+        # Read values into dictionary.
+        with open(token_file, encoding="utf-8") as file:
```
behnazh-w, line 225: Looks like handling `OSError` is missing`?

src/macaron/__main__.py
```diff
@@ -221,6 +217,26 @@ def main(argv: list[str] | None = None) -> None:
...
+            for line in file:
+                if not line or "=" not in line:
+                    continue
+                key, value = line.rstrip().split("=", 1)
+                token_dict[key] = value
+        # Overwrite file contents.
+        with open(token_file, "w", encoding="utf-8"):
```
behnazh-w, line 232: Same comment regarding `OSError`.

scripts/release_scripts/run_macaron.sh
behnazh-w, line 1: I think it would be good to add `rm -f "$TOKEN_FILE"` after the `docker run` at the end of the script.

src/macaron/slsa_analyzer/git_service/gitlab.py
```diff
@@ -36,10 +37,10 @@
 class GitLab(BaseGitService):
     """This class contains the spec of the GitLab service."""
 
-    def __init__(self, access_token_env_name: str) -> None:
+    def __init__(self, token_function: Callable[[], str]) -> None:
         """Initialize instance."""
```
behnazh-w, line 41: Please add the parameter to the docstring.

src/macaron/slsa_analyzer/git_service/gitlab.py
```diff
@@ -248,9 +249,9 @@ def load_defaults(self) -> None:
         if not self.hostname:
             return
 
-        if not os.environ.get(self.access_token_env_name):
+        if not self.token_function():
             raise ConfigurationError(
-                f"Environment variable '{self.access_token_env_name}' is not set "
+                f"Environment variable for '{self.__class__}' is not set "
                 + f"for private GitLab service '{self.hostname}'."
```
behnazh-w, line 255:
```suggestion
                f"Environment variable for self-hosted GitLab service '{self.hostname}' is not set."
```

src/macaron/__main__.py
```diff
@@ -221,6 +217,36 @@ def main(argv: list[str] | None = None) -> None:
...
+        try:
+            with open(token_file, encoding="utf-8") as file:
+                for line in file:
+                    if not line or "=" not in line:
+                        continue
+                    key, value = line.rstrip().split("=", 1)
+                    token_dict[key] = value
+        except OSError as error:
+            logger.error("Could not open token file %s: %s", token_file, error)
+            sys.exit(1)
```
behnazh-w, line 236:
```suggestion
            sys.exit(os.EX_OSFILE)
```

behnazh-w (approved):

PR #609, opened 2024-01-20
test(docs): improve the doctest configuration

behnazh-w:
This PR 

- moves the configuration to find files for doctests to `.pre-commit-config.yaml` to be consistent with the rest of the hooks.
- It also adds an example doctest for contributors.
- Removes an unnecessary comment in a docstring.


benmss (approved):
Looks good to me.

PR #620, opened 2024-01-31
feat!: introduce confidence scores for check facts

behnazh-w:
## Confidence score and justification type 
- This PR allows specifying confidence scores for check results, which is especially useful when a check reports multiple candidate results. All of these confidence scores are added to the check tables in the database. However, the fact that has the highest confidence is shown in the HTML/JSON report only.  

- The justifications are no longer required to be added manually to the `CheckResultData`. Instead, they are curated directly from the results in the table. If a column has specified `JustificationType` in the column mapping, it will be picked up automatically and rendered as plain text or href depending on the specified type. If a check fails or is skipped, we show a default `Not Available.` justification. This allows to create HTML/JSON reports from the database reproducibly.

## Refactoring
For this feature to work, the following refactorings are done as well:

- The checks are changed to add multiple facts to a result table.
- The confidence score is added to `CheckFacts` ORM mapping, which all the check mappings inherit from. That caused some circular dependency issues for the `Expectation` ORM mapping. I have refactored and improved this mapping accordingly.
- I have improved the check table columns that needed to be added as justification. For example, I have added `asset_url` to `ProvenanceAvailableFacts`. That required adding a new  `SLSAProvenanceData` wrapper class for GitHub release provenances. 

##  Documentation
I have added elaborate instructions for adding a new check and explained the current interface.


docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,162 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
 
 For API reference, see the :doc:`API Reference </pages/developers_guide/apidoc/index>` page.
 
+-------------------
+Writing a New Check
+-------------------
+
```
benmss, line 17: I think all mid-paragraph line breaks should be removed from this file.

behnazh-w: Hm, I have added them to help with git diffs.

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,162 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
+section, we will understand how Macaron checks work and what we need to do to develop one.
+
++++++++++++++++++
+High-level Design
++++++++++++++++++
+
+Before jumping into coding, it is useful to understand how Macaron as a framework works. Macaron is an extensible
+framework designed to make writing new supply chain security analyses easy. It provides an interface
+that you can leverage to access existing models and abstractions instead of implementing everything from scratch. For
+instance, many security checks require to traverse through the code in GitHub Actions configurations. Normally,
```
benmss, line 28:
```suggestion
instance, many security checks require traversing through the code in GitHub Actions configurations. Normally,
```

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,162 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
+Each check needs to be implemented as a Python class in a Python module under ``src/macaron/slsa_analyzer/checks``.
+A check class should subclass the ``BaseCheck`` class in :ref:`base_check module <pages/developers_guide/apidoc/macaron\.slsa_analyzer\.checks:macaron.slsa\\_analyzer.checks.base\\_check module>`.
+
+You need to set the name, description, and other details of your new check in the ``__init__`` method. After implementing
+the initializer, you need to implement the ``run_check`` abstract method. This method provides the context object
+:ref:`AnalyzeContext <pages/developers_guide/apidoc/macaron\.slsa_analyzer:macaron.slsa\\_analyzer.analyze\\_context module>`, which contains various
+intermediate representations and models. The ``dynamic_data`` property would be particularly useful as it contains
+data about the CI service, artifact registry, and build tool used for building the software component.
+
+``component`` is another useful attribute in the :ref:`AnalyzeContext <pages/developers_guide/apidoc/macaron\.slsa_analyzer:macaron.slsa\\_analyzer.analyze\\_context module>` object
```
benmss, line 51:
```suggestion
The ``component`` is another useful attribute in the :ref:`AnalyzeContext <pages/developers_guide/apidoc/macaron\.slsa_analyzer:macaron.slsa\\_analyzer.analyze\\_context module>` object
```

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,162 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
+framework designed to make writing new supply chain security analyses easy. It provides an interface
+that you can leverage to access existing models and abstractions instead of implementing everything from scratch. For
+instance, many security checks require to traverse through the code in GitHub Actions configurations. Normally,
+you would need to find the right repository and commit, clone it, find the workflows, and parse them. With Macaron,
+you don't need to do any of that and can simply write your security check by using the parsed shell scripts that are
+triggered in the CI.
+
+Another important aspect of our design is that all the check results are automatically mapped and stored in a local database.
+By performing this mapping, we make it possible to enforce flexible policies on the results of the checks. While storing
+the check results to the database happens automatically by Macaron's backend, the developer needs to add a brief specification
```
benmss, line 35:
```suggestion
the check results to the database happens automatically in Macaron's backend, the developer needs to add a brief specification
```

docs/source/pages/developers_guide/index.rst
benmss, line 51: I understand the use of `attribute` and `property` is intentional. However, given that the two differ slightly, using "another" as per "another useful attribute" does not follow. I would instead suggest:
```suggestion
The :ref:`AnalyzeContext <pages/developers_guide/apidoc/macaron\.slsa_analyzer:macaron.slsa\\_analyzer.analyze\\_context module>` object contains other useful information, such as ``component``. This acts as a representation of a software component, including it's corresponding ``repository`` and ``dependencies``.
```
Another alternative would be to use the same word for `property` and `attribute` depending on which is more general.
Or perhaps I am simply being too pedantic in this case.

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,162 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
++++++++
+Example
++++++++
+
+In this example, we show how to add a check determine if a software component has a source-code repository.
+Feel free to explore other existing checks under ``src/macaron/slsa_analyzer/checks`` for more examples.
+
+1. First create a module called ``repo_check.py`` under ``src/macaron/slsa_analyzer/checks``.
+
+2. Add a class and specify the columns that you want to store for the check outputs to the database.
```
benmss, line 96:
```suggestion
2. Add a class and specify the columns that you want to store for the check outputs in the database.
```

src/macaron/slsa_analyzer/checks/check_result.py
```diff
@@ -29,6 +25,32 @@ class CheckResultType(str, Enum):
...
+    """
+
+    #: A high confidence score.
+    HIGH = 1.0
+
+    #: A medium confidence score.
+    MEDIUM = 0.7
+
+    #: A low confidence score.
+    LOW = 0.5
```
nathanwn, line 41: I wonder how you picked these values.

behnazh-w: The values themselves are not important. The important part is having three levels, which can definitely change if needed.

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,162 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
++++++++++++++++++++
+
+Each check needs to be implemented as a Python class in a Python module under ``src/macaron/slsa_analyzer/checks``.
+A check class should subclass the ``BaseCheck`` class in :ref:`base_check module <pages/developers_guide/apidoc/macaron\.slsa_analyzer\.checks:macaron.slsa\\_analyzer.checks.base\\_check module>`.
+
+You need to set the name, description, and other details of your new check in the ``__init__`` method. After implementing
+the initializer, you need to implement the ``run_check`` abstract method. This method provides the context object
+:ref:`AnalyzeContext <pages/developers_guide/apidoc/macaron\.slsa_analyzer:macaron.slsa\\_analyzer.analyze\\_context module>`, which contains various
+intermediate representations and models. The ``dynamic_data`` property would be particularly useful as it contains
+data about the CI service, artifact registry, and build tool used for building the software component.
```
nathanwn, line 49: Some example code snippets should be helpful for this part.

nathanwn: To clarify, what I mean here is something like

```
.. code-block:: python
   import ...

   class MyCheck(BaseCheck):
       ...
```

You have some other example snippets after this part in the same section, so it makes sense also to do that here.

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,162 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
+Before jumping into coding, it is useful to understand how Macaron as a framework works. Macaron is an extensible
+framework designed to make writing new supply chain security analyses easy. It provides an interface
+that you can leverage to access existing models and abstractions instead of implementing everything from scratch. For
+instance, many security checks require to traverse through the code in GitHub Actions configurations. Normally,
+you would need to find the right repository and commit, clone it, find the workflows, and parse them. With Macaron,
+you don't need to do any of that and can simply write your security check by using the parsed shell scripts that are
+triggered in the CI.
+
+Another important aspect of our design is that all the check results are automatically mapped and stored in a local database.
+By performing this mapping, we make it possible to enforce flexible policies on the results of the checks. While storing
```
nathanwn, line 34: Nit: The meaning of the word `flexible` in `flexible policies` here feels vague to me. I would prefer something along the line of:

```suggestion
By performing this mapping, we make it possible for users to enforce their use case-specific policies on the results of the checks. While storing
```

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,162 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
+the check results to the database happens automatically by Macaron's backend, the developer needs to add a brief specification
+to make that possible as we will see later.
+
++++++++++++++++++++
+The Check Interface
++++++++++++++++++++
+
+Each check needs to be implemented as a Python class in a Python module under ``src/macaron/slsa_analyzer/checks``.
+A check class should subclass the ``BaseCheck`` class in :ref:`base_check module <pages/developers_guide/apidoc/macaron\.slsa_analyzer\.checks:macaron.slsa\\_analyzer.checks.base\\_check module>`.
```
nathanwn, line 43: You can use [these Sphinx Python roles](https://www.sphinx-doc.org/en/master/usage/domains/python.html#python-roles) to cross-reference Python objects, which is nicer than using the `:ref:` role.
```suggestion
A check class should subclass the :class:`BaseCheck <macaron.slsa_analyzer.checks.base_check.BaseCheck>` class.
```

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,162 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
+
+You need to set the name, description, and other details of your new check in the ``__init__`` method. After implementing
+the initializer, you need to implement the ``run_check`` abstract method. This method provides the context object
+:ref:`AnalyzeContext <pages/developers_guide/apidoc/macaron\.slsa_analyzer:macaron.slsa\\_analyzer.analyze\\_context module>`, which contains various
```
nathanwn, line 47:
```suggestion
:class:`AnalyzeContext <macaron.slsa_analyzer.analyze_context.AnalyzeContext>`, which contains various
```

docs/source/pages/developers_guide/index.rst
nathanwn, line 51:
```suggestion
``component`` is another useful attribute in the :class:`AnalyzeContext <macaron.slsa_analyzer.analyze_context.AnalyzeContext>` object
```

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,162 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
+``component`` is another useful attribute in the :ref:`AnalyzeContext <pages/developers_guide/apidoc/macaron\.slsa_analyzer:macaron.slsa\\_analyzer.analyze\\_context module>` object
+that you should know about. This attribute contains the information about a software component, such
+as it's corresponding ``repository`` and ``dependencies``. Note that ``component`` will also be stored into the database and its attributes
+such as ``repository`` are established as database relationships. You can see the existing tables and their
+relationships in our :ref:`data model <pages/developers_guide/apidoc/macaron.database:macaron.database.table\\_definitions module>`.
+
+Once you implement the logic of your check in the ``run_check`` method, you need to add a class to help
+Macaron handle your check's output:
+
+   * Add a class that subclasses ``CheckFacts`` to map your outputs to a table in the database. The class name should follow the ``<MyCheck>Facts`` pattern.
```
nathanwn, line 60:
```suggestion
   * Add a class that subclasses :class:`CheckFacts <macaron.database.table_definitions>` to map your outputs to a table in the database. The class name should follow the ``<MyCheck>Facts`` pattern.
```

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,162 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
+   * Specify the table name in the ``__tablename__ = "_my_check"`` class variable. Note that the table name should start with ``_`` and it should not have been used by other checks.
+   * Add the ``id`` column as the primary key where the foreign key is ``_check_facts.id``.
+   * Add columns for the check outputs that you would like to store into the database. If a column needs to appear as a justification in the HTML/JSON report, pass ``info={"justification": JustificationType.<TEXT or HREF>}`` to the column mapper.
+   * Add ``__mapper_args__`` class variable and set ``"polymorphic_identity"`` key to the table name.
+
+Next, you need to create a ``result_tables`` list and append check facts as part of the ``run_check`` implementation.
+You should also specify a :ref:`Confidence <pages/developers_guide/apidoc/macaron\.slsa_analyzer\.checks:macaron.slsa\\_analyzer.checks.check\\_result module>`
+score choosing one of  the ``Confidence`` enum values, e.g., ``Confidence.HIGH`` and pass it via keyword
```
nathanwn, line 68:
```suggestion
You should also specify a confidence score by choosing one of  the :class:`Confidence <macaron.slsa_analyzer.checks.check_result.Confidence>` enum values, e.g., :attr:`Confidence.HIGH <macaron.slsa_analyzer.checks.check_result.Confidence.HIGH>` and pass it via keyword
```

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,162 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
+of your check analysis.
+
+.. code-block:: python
+
+   result_tables.append(MyCheckFacts(col_foo=foo, col_bar=bar, confidence=Confidence.HIGH))
+
+This list as well as the check result status should be stored in a :ref:`CheckResultData <pages/developers_guide/apidoc/macaron\.slsa_analyzer\.checks:macaron.slsa\\_analyzer.checks.check\\_result module>`
+object and returned by ``run_check``.
+
+Finally, you need to register your check by adding it to the :ref:`registry module <pages/developers_guide/apidoc/macaron\.slsa_analyzer:macaron.slsa\\_analyzer.registry module>`:
```
nathanwn, line 79:
```suggestion
Finally, you need to register your check by adding it to the :mod:`registry<macaron.slsa_analyzer.registry>` module:
```

nathanwn, line 79: For clarity, I think we should mention that this should be done at the end of the same file.

src/macaron/slsa_analyzer/analyze_context.py
```diff
@@ -93,6 +92,24 @@ def __init__(
...
+        This object is used to pass various models and intermediate representations from the backend
+        in Macaron to checks. A check can also store intermediate results in this object to be used
+        by checks that depend on it. However, please avoid adding arbitrary attributes to this object!
+
+        We recommend to take a look at the attributes in this object before writing a new check. Chances
+        are that what you try to implement is already implemented and the results are available in the
+        `dynamic_data` object.
+
+        Return
+        ------
```
nathanwn, line 108:
```suggestion
        Returns
        -------
```

src/macaron/slsa_analyzer/checks/check_result.py
```diff
@@ -47,18 +69,59 @@ class CheckInfo:
...
 
-    #: List of justifications describing the reasons for the check result.
-    #: If an element in the justification is a string,
-    #: it will be displayed as a string, if it is a mapping,
-    #: the value will be rendered as a hyperlink in the html report.
-    justification: Justification
-
     #: List of result tables produced by the check.
-    result_tables: ResultTables
+    result_tables: list[CheckFacts]
```
nathanwn, line 73: This may be out-of-scope of this PR, but I think this field should be `result_facts` instead of `result_tables`, i.e. the word "tables" here is not appropriate. We may create an issue if you think this should be changed in another PR.

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,162 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
+instance, many security checks require to traverse through the code in GitHub Actions configurations. Normally,
+you would need to find the right repository and commit, clone it, find the workflows, and parse them. With Macaron,
+you don't need to do any of that and can simply write your security check by using the parsed shell scripts that are
+triggered in the CI.
+
+Another important aspect of our design is that all the check results are automatically mapped and stored in a local database.
+By performing this mapping, we make it possible to enforce flexible policies on the results of the checks. While storing
+the check results to the database happens automatically by Macaron's backend, the developer needs to add a brief specification
+to make that possible as we will see later.
+
```
nathanwn, line 37: I think before elaborating on how to implement a check, we may want to clarify at a high level what a check does. Some simple high-level questions that may come to mind for a new contributor could be:
* What are the inputs of a check?
* What are the outputs of a check?
* Is it safe to assume that each check should leave no side-effects, i.e. does it only read the inputs and return the outputs, or does it modify a shared state with other checks?

Some other questions that are not as high-level but still carry a high-level understanding of the checks include:
* Do the checks run in a particular order? What is the meaning of `depends_on` and how does the check dependency relationship work?
* When are these checks executed in Macaron? What does Macaron do before and after the running of the checks? Here, it could even be helpful to emphasize that these checks run as part of `macaron analyze`.

At the same time, it could also be helpful to not just clarify what should be implemented as a check, but also what **should not** be implemented as a check, and one should leverage the policy engine instead. I feel like the `RepoCheck` example that you gave is not particularly representative of what should be implemented as a check since Macaron has already produced this information in the database and one should just write a policy instead of implementing a new check. If you think this is the case, we can either provide a different example, or simply say that "This is just an example for how a check should be implemented, and not an example of what a check should be for. Please open an issue to discuss with us if you want to implement a new check and we can confirm whether your use case is appropriate."

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,162 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
++++++++++++++++++++
+The Check Interface
++++++++++++++++++++
+
+Each check needs to be implemented as a Python class in a Python module under ``src/macaron/slsa_analyzer/checks``.
+A check class should subclass the ``BaseCheck`` class in :ref:`base_check module <pages/developers_guide/apidoc/macaron\.slsa_analyzer\.checks:macaron.slsa\\_analyzer.checks.base\\_check module>`.
+
+You need to set the name, description, and other details of your new check in the ``__init__`` method. After implementing
```
nathanwn, line 45: Here, by `name` I assume you are referring to the check id. Do you think we should mention the naming rule for checks?
The docstring of the `Registry._validate_check_id_format()` method says:
```
        A valid check id must have the following properties:
            - The general format: ``mcn_<name>_<digits>``
            - In ``name``, only lowercase alphabetical letters are allowed. If ``name`` contains multiple \
            words, they must be separated by underscores.
```
I ran into this issue once myself, and a new contributor is likely to run into the same problem of the registry not accepting their check if we do not mention it. On the other hand, if the naming has a convention, I think we should mention it.

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,162 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
+               logger.info("Unable to find a Git repository for %s", ctx.component.purl)
+               # We do not store any results in the database if a check fails. So, just leave result_tables empty.
+               return CheckResultData(result_tables=[], result_type=CheckResultType.FAILED)
+
+           return CheckResultData(
+               result_tables=[RepoCheckFacts(git_repo=ctx.component.repository.remote_path, confidence=Confidence.HIGH)],
+               result_type=CheckResultType.PASSED,
+           )
+
+4. Register your check.
```
nathanwn, line 159: Repeating one of my previous comments: For clarity, I think we should mention that this should be done at the end of the same file.

src/macaron/slsa_analyzer/checks/check_result.py
```diff
@@ -84,7 +147,8 @@ def get_summary(self) -> dict:
             "check_id": self.check.check_id,
             "check_description": self.check.check_description,
             "slsa_requirements": [str(BUILD_REQ_DESC.get(req)) for req in self.check.eval_reqs],
-            "justification": self.result.justification,
+            # The justification report is stored in a heapq where the first element has the highest confidence score.
+            "justification": self.result.justification_report[0][1],
```
nathanwn, line 151: I don't think I'm convinced that a heap should be used here when we have to do things like overriding the comparator while the only thing we need the heap for currently is to get the justification with the highest score. It may be that you have other goals with this heap in mind for later features. Otherwise, I'm not sure how the heap is useful unless we need to maintain a priority queue while allowing both pushing and popping operations throughout its lifetime.

Wouldn't it be simpler just to either sort the list before returning from the property function, or use the `max` function like so?
```suggestion
            "justification": max(self.result.justification_report, key=lambda e: e[0])[1],
```
I also don't think performance-wise we gain anything with a heap.

src/macaron/slsa_analyzer/analyze_context.py
```diff
@@ -93,6 +92,24 @@ def __init__(
             expectation=None,
         )
 
+    @property
+    def dynamic_data(self) -> ChecksOutputs:
```
nathanwn, line 96: Should be out-of-scope of this PR. Having said that, I think we should consider renaming the `ChecksOutputs` class to something like `ChecksDynamicData`, `ChecksSharedData`, `DynamicAnalysisData`, etc. but not `ChecksOutputs`. I think the "outputs" of checks should just be the `CheckResultData` class and we should try to clearly distinguish between these two different things.

src/macaron/slsa_analyzer/checks/check_result.py
nathanwn: Hmm, then my follow-up question would be why do we need 3 levels? I assume the answer is likely as simple as "We define a few set levels just so that we do not have to think about different arbitrary confidence score values when writing checks."

src/macaron/slsa_analyzer/checks/check_result.py
```diff
@@ -47,18 +69,59 @@ class CheckInfo:
...
+        """
+        Return the list of justifications for the check result generated from the tables in the database.
+
+        Note that the elements in the justification will be rendered different based on their types:
+
+        * a :class:`JustificationType.TEXT` element is displayed in plain text in the HTML report.
+        * a :class:`JustificationType.HREF` element is rendered as a hyperlink in the HTML report.
+
+        Return
+        ------
```
nathanwn, line 89:
```suggestion
        Returns
        -------
```

src/macaron/slsa_analyzer/checks/check_result.py
```diff
@@ -29,6 +25,32 @@ class CheckResultType(str, Enum):
...
 
 
+class Confidence(float, Enum):
+    """This class contains confidence score for a check result.
+
+    The scores must be in [0.0, 1.0].
+    """
+
+    #: A high confidence score.
+    HIGH = 1.0
```
nathanwn, line 35: I know we have discussed that the values in this enum are not important. However, IMHO, if we include the value `1.0` i.e. 100%, then I would prefer the name `CERTAIN` over `HIGH` which better expresses that "this is 100% true". We can then have 4 values: `CERTAIN`, `HIGH`, `MEDIUM`, and `LOW`.

src/macaron/slsa_analyzer/checks/check_result.py
nicallen: I agree that if the heap is only used to obtain the max value once after all insertions are done then there is no benefit to using a heap. Using `max` keyed on the confidence score would be sufficient, and allows removing the `__lt__` override (and I don't like that override since it is overly specific to a specific usage, as evidenced by the lengthy comment that is required to explain it by referencing how it is used, which is undesirable as definitions ideally should stand on their own). Though if you really needed the heap, you can still get rid of the `__lt__` override by using a wrapper (defined at the usage) that compares by inverse confidence score instead.

nicallen:
Under this new schema, the confidence score is recorded for a `check_facts` entry, which currently contains multiple pieces of information that would be forced to share a confidence score. For example, the `build_as_code_check` entries contain both `ci_service_name` and `deploy_command` which are derived by different means and hence may have different levels of confidence, which cannot be expressed in the current schema. This could be fixed this by splitting the `build_as_code_check` and similar tables into smaller tables with more granularity of information. Are you intending to do that in later changes?

src/macaron/slsa_analyzer/specs/ci_spec.py
```diff
@@ -38,5 +38,5 @@ class CIInfo(TypedDict):
     Schema: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release.
     """
 
-    provenances: Sequence[InTotoPayload]
-    """The JSON payloads of SLSA provenances in in-toto format."""
+    provenances: Sequence[SLSAProvenanceData]
```
nathanwn, line 41: Shouldn't we use the `DownloadedProvenanceData` interface instead of the specific class `SLSAProvenanceData` here?
```suggestion
    provenances: Sequence[DownloadedProvenanceData]
```
 Provenances may not all be SLSA provenances. In addition, the `DownloadedProvenanceData` interface is also being used in the `PackageRegistryInfo` class which is a parallel class to `CIInfo`.

docs/source/pages/developers_guide/index.rst
behnazh-w: Thanks for the suggestions. I will incorporate them in the documentation. A the same time, I want to keep the description short but useful.

This example check is not meant to be representative and just shows how to create a new check.  More information on what a check should not be is better to be added in a FAQ section in a different PR. Drawing a line between a check and policy and discussing what should be part of a check vs policy is also a topic for a FAQ section.

docs/source/pages/developers_guide/index.rst
behnazh-w: I was looking for a way to reference directly. That's great thanks.

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

src/macaron/slsa_analyzer/checks/check_result.py
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

src/macaron/slsa_analyzer/analyze_context.py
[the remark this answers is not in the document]
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

src/macaron/slsa_analyzer/checks/check_result.py
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

src/macaron/slsa_analyzer/specs/ci_spec.py
behnazh-w: See commit [d25bf57](https://github.com/oracle/macaron/pull/620/commits/d25bf57f3e6f57d29cf4c311207dbcfe1302727e).

docs/source/pages/developers_guide/index.rst
```diff
@@ -113,10 +136,25 @@ Feel free to explore other existing checks under ``src/macaron/slsa_analyzer/che
...
+
+Add a class for your check that subclasses :class:`BaseCheck <macaron.slsa_analyzer.checks.base_check.BaseCheck>`,
+provide the check details in the initializer method, and implement the logic of the check in
+:func:`run_check <macaron.slsa_analyzer.checks.base_check.BaseCheck.run_check>`.
+
+A ``check_id`` should meet the following requirements:
+
+    - The general format: ``mcn_<name>_<digits>``
+    - In ``name``, only lowercase alphabetical letters are allowed. If ``name`` contains multiple \
+        words, they must be separated by underscores.
```
nathanwn, line 154: This currently renders like this on my side:
image

I suspect this is due to the additional indentation on the second line relative to the first, so Sphinx interpreted this as a [definition list item](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#definition-lists). This is instead a [bullet list item](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#bullet-lists) item, so the second line should be indented at the same level as the first line. The backslash at the end of the first line is also not necessary.

```suggestion
    - In ``name``, only lowercase alphabetical letters are allowed. If ``name`` contains multiple
      words, they must be separated by underscores.
```

behnazh-w: > This currently renders like this on my side

See comit [4661c00](https://github.com/oracle/macaron/pull/620/commits/4661c001a72ce289b550fe11591e031f331f943c).

docs/source/pages/developers_guide/index.rst
```diff
@@ -149,9 +149,8 @@ provide the check details in the initializer method, and implement the logic of
 
 A ``check_id`` should meet the following requirements:
 
-    - The general format: ``mcn_<name>_<digits>``
-    - In ``name``, only lowercase alphabetical letters are allowed. If ``name`` contains multiple \
-        words, they must be separated by underscores.
+    - The general format: ``mcn_<name>_<digits>``.
+    - Lowercase alphabetical letters are allowed in ``name`` only. If ``name`` contains multiple words, they must be separated by underscores.
```
nathanwn, line 153: Sorry for nitpicking but this sounds more like "the only part of the check id where lowercase alphabetical letters are allowed is `name`", which is different from "Only lowercase alphabetical letters are allowed in ``name``."

nathanwn (approved):
I think the PR should be good to merge.
Thanks for the PR.

docs/source/pages/developers_guide/index.rst
behnazh-w: When `only` is used at the end of a sentence, it's a modifier for the whole sentence. Here I don't want to focus on `name` or `lowercase`.

nicallen: I agree with Nathan's interpretation of the literal meaning of the sentence. However, "Only lowercase alphabetical letters are allowed in `name`" would also be incorrect because it is immediately contradicted by the following statement that implies that underscore is also allowed.

Presuming that the intention is to convey that name has the form `[a-z]+(_[a-z]+)+`, a precise description would be "`name` must consist solely of words (sequences of lowercase alphabetical letters) separated by underscores" (you could even include the regex if you want to be absolutely precise).

(To be extra nitpicky, "in name only" is a such a common phrase with a specific meaning that I'd avoid invoking it by coincidence, I know that `name` is a specially-quoted identifier that shouldn't be read literally, but it still throws me a little)

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,214 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
 
 For API reference, see the :doc:`API Reference </pages/developers_guide/apidoc/index>` page.
 
+-------------------
+Writing a New Check
+-------------------
+
+As a contributor to Macaron, it is very likely to need to write a new check or modify an existing one at some point. In this
```
benmss, line 18:
```suggestion
Contributors to Macaron are very likely to need to write a new check or modify an existing one at some point. In this
```

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,214 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
+section, we will understand how Macaron checks work and what we need to do to develop one.
```
benmss, line 19:
```suggestion
section, we will explain how Macaron checks work and what needs to be done by contributors working with them.
```

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,214 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
+triggered in the CI.
+
+Another important aspect of our design is that all the check results are automatically mapped and stored in a local database.
+By performing this mapping, we make it possible to enforce use case-specific policies on the results of the checks. While storing
+the check results in the database happens automatically in Macaron's backend, the developer needs to add a brief specification
+to make that possible as we will see later.
+
+Once you get familiar with writing a basic check, you can explore the check dependency feature in Macaron. The checks
+in our framework can be customized to only run if another check has run and returned a specific
+:class:`result type <macaron.slsa_analyzer.checks.check_result.CheckResultType>`. This feature can be used when some checks
```
benmss, line 40:
```suggestion
:class:`result type <macaron.slsa_analyzer.checks.check_result.CheckResultType>`. This feature can be used when checks
```

docs/source/pages/developers_guide/index.rst
```diff
@@ -11,6 +11,214 @@ To follow the project's code style, see the :doc:`Macaron Style Guide </pages/de
...
+can be ordered and have a parent-child relationship, i.e., one check implements a weaker or stronger version of a
```
benmss, line 41:
```suggestion
have an ordering and a parent-child relationship, i.e., one check implements a weaker or stronger version of a
```

src/macaron/database/table_definitions.py
```diff
@@ -415,6 +423,16 @@ class CheckFacts(ORMBase):
     #: The primary key.
     id: Mapped[int] = mapped_column(Integer, primary_key=True, autoincrement=True)  # noqa: A003
 
+    #: The confidence score to estimate the accuracy of the check fact. This value should be in [0.0, 1.0] with
```
benmss, line 426:
```suggestion
    #: The confidence score to estimate the accuracy of the check fact. This value should be in the range [0.0, 1.0] with
```

behnazh-w: See commit [08eb1ce](https://github.com/oracle/macaron/pull/620/commits/08eb1ce77c142753c5fae77b68d5b2b765c50e3d).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [08eb1ce](https://github.com/oracle/macaron/pull/620/commits/08eb1ce77c142753c5fae77b68d5b2b765c50e3d).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [08eb1ce](https://github.com/oracle/macaron/pull/620/commits/08eb1ce77c142753c5fae77b68d5b2b765c50e3d).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [08eb1ce](https://github.com/oracle/macaron/pull/620/commits/08eb1ce77c142753c5fae77b68d5b2b765c50e3d).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [08eb1ce](https://github.com/oracle/macaron/pull/620/commits/08eb1ce77c142753c5fae77b68d5b2b765c50e3d).

docs/source/pages/developers_guide/index.rst
behnazh-w: See commit [08eb1ce](https://github.com/oracle/macaron/pull/620/commits/08eb1ce77c142753c5fae77b68d5b2b765c50e3d).

benmss (approved):
Looks good to me.

nicallen (approved):

PR #622, opened 2024-02-02
feat: use repository url provided as input for finding a commit

### LINKED ISSUE #621: Use repository URL provided as input for finding a commit
In cases that the repository URL for an artifact is not available, we should allow the user to provide the URL as input and let the commit finder find the commit if the input PURL has a version, e.g.,

```bash
macaron analyze -rp https://github.com/assertj/assertj -purl pkg:maven/org.assertj/assertj.core@3.25.1 --skip-deps
```

Right now we require the user to provide the commit in addition to the URL even though the PURL contains a version:

```bash
[ERROR] Please provide the commit digest for the repo at https://github.com/assertj/assertj that matches to the PURL string pkg:maven/org.assertj/assertj.core@3.25.1.
```

benmss:
For cases where the user has provided a PURL with a version but no repository can be found, this PR allows the analysis to succeed as long as the user provides a repository path as well. This differs from the current functionality where if both a PURL and repo are provided, a commit must be too.

An example of this new use case is added in the integration tests:
`macaron analyze -purl pkg:maven/org.tinymediamanager/tinyMediaManager@4.3.13 -rp https://gitlab.com/tinyMediaManager/tinyMediaManager`

Without this new functionality, the above test would fail as the provided version cannot be found by the Repo Finder.

Closes #621

### DIFF
--- a/src/macaron/__main__.py
+++ b/src/macaron/__main__.py
@@ -96,7 +96,11 @@ def analyze_slsa_levels_single(analyzer_single_args: argparse.Namespace) -> None
         if repo_path and purl:
             # To provide the purl together with the repository path, the user must specify the commit digest unless the
             # purl has a version.
-            purl_object = PackageURL.from_string(purl)
+            try:
+                purl_object = PackageURL.from_string(purl)
+            except ValueError as error:
+                logger.debug("Could not parse PURL: %s", error)
+                purl_object = None
             if not purl_object or not (purl_object.version or digest):
                 logger.error(
                     "Please provide the commit digest for the repo at %s that matches to the PURL string %s. Or "
--- a/src/macaron/__main__.py
+++ b/src/macaron/__main__.py
@@ -11,6 +11,7 @@ import sys
 from importlib import metadata as importlib_metadata
 
 from jinja2 import Environment, FileSystemLoader, select_autoescape
+from packageurl import PackageURL
 
 import macaron
 from macaron.config.defaults import create_defaults, load_defaults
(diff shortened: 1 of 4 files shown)


src/macaron/__main__.py
```diff
@@ -92,15 +93,18 @@ def analyze_slsa_levels_single(analyzer_single_args: argparse.Namespace) -> None
...
-                "Please provide the commit digest for the repo at %s that matches to the PURL string %s.",
-                repo_path,
-                purl,
-            )
-            sys.exit(os.EX_USAGE)
+        if repo_path and purl:
+            # To provide the purl together with the repository path, the user must specify the commit digest unless the
+            # purl has a version.
+            purl_object = PackageURL.from_string(purl)
+            if not purl_object or not (purl_object.version or digest):
```
nathanwn, line 100: Having looked at the [implementation of `PackageURL.from_string`](https://github.com/package-url/packageurl-python/blob/f7f41b89a941278e8f76c0aad3a9409c6583eda8/src/packageurl/__init__.py#L481), I don't think the returned object can be falsy or `None`. However, the function may raise `ValueError` which should be handled.

src/macaron/__main__.py
```diff
@@ -92,15 +93,22 @@ def analyze_slsa_levels_single(analyzer_single_args: argparse.Namespace) -> None
...
-            sys.exit(os.EX_USAGE)
+        if repo_path and purl:
+            # To provide the purl together with the repository path, the user must specify the commit digest unless the
+            # purl has a version.
+            try:
+                purl_object = PackageURL.from_string(purl)
+            except ValueError as error:
+                logger.debug("Could not parse PURL: %s", error)
+                purl_object = None
+            if not purl_object or not (purl_object.version or digest):
```
nathanwn, line 104: Instead of setting `purl_object` to `None` and rechecking it in the later `if`, why don't we exit instead?
```suggestion
                sys.exit(os.EX_USAGE)
            if not (purl_object.version or digest):
```
The error message logged inside the following `if` statement does not mention anything about the PURL being invalid, but instead tells the user to provide the commit/version. If the PURL is invalid, this message would be irrelevant since it is pointing to a completely different issue.

behnazh-w: Agreed that exiting immediately if the PURL string is incorrect with a relevant error message would make sense. Then you can keep the following error for the case where neither the version nor the commit hash is provided.

nathanwn (approved):
LGTM. Thanks for the PR.

behnazh-w:
[This note](https://github.com/oracle/macaron/blob/621-accept-repo-url-and-purl-with-version/docs/source/pages/using.rst?plain=1#L134) in the documentation needs to be updated. Also, please add an example in the documentation to show how both a repo path and a PURL containing a version can be used at the same time.

This PR could also be considered as a feature I think.

tromai (approved):
LGTM. Thanks for the changes.

behnazh-w (approved):

PR #627, opened 2024-02-06
chore: improve build command output

nicallen:
Record build command in build service check, and change encoding of build/deploy commands in build as code and build service checks to json to make it easier to parse later.

### DIFF
--- a/src/macaron/slsa_analyzer/checks/build_as_code_check.py
+++ b/src/macaron/slsa_analyzer/checks/build_as_code_check.py
@@ -84,6 +84,21 @@ class BuildAsCodeCheck(BaseCheck):
             result_on_skip=CheckResultType.PASSED,
         )
 
+    def _serialize_command(self, cmd: list[str]) -> str:
+        """Convert a list of command-line arguments to a json-encoded string so that it is easily parsable by later consumers.
+
+        Parameters
+        ----------
+        cmd: list[str]
+            List of command-line arguments.
+
+        Returns
+        -------
+        str
+            The list of command-line arguments as a json-encoded string.
+        """
+        return json.dumps(cmd)
+
     def _has_deploy_command(self, commands: list[list[str]], build_tool: BaseBuildTool) -> str:
         """Check if the bash command is a build and deploy command."""
         # Account for Python projects having separate tools for packaging and publishing.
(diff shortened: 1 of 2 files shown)


src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -117,14 +118,16 @@ def _has_deploy_command(self, commands: list[list[str]], build_tool: BaseBuildTo
                 # If there are no deploy args for this build tool, accept as deploy command.
                 # TODO: Support multi-argument build keywords, issue #493.
                 if not build_tool.deploy_arg:
-                    logger.info("No deploy arguments required. Accept %s as deploy command.", str(com))
-                    return str(com)
+                    com_str = json.dumps(com)
```
behnazh-w, line 121: There is a [warning](https://docs.python.org/3/library/json.html#json.dumps) on `json` documentation page about parsing untrusted input, but unfortunately doesn't give an advice on sanitizing the value.

behnazh-w, line 121: Can you please add a comment to explain why we need to use `json.dumps` here?

behnazh-w: I think we should catch `TypeError` and `ValueError` looking at the implementation of [JSONEncoder](https://github.com/python/cpython/blob/3.12/Lib/json/encoder.py#L74) to be on the safe side.

nicallen: Those exceptions are raised for types and values that cannot be encoded. The type of `com` is `list[str]`, which should always be encodable without errors.

nicallen: I have factored out the `json.dumps` calls into a `_serialize_command` function with a comment explaining the purpose.

nathanwn:
Hi Nick. I just made a minor case change in the PR title just to stay consistent with other PR titles.

src/macaron/slsa_analyzer/checks/build_as_code_check.py
behnazh-w: Based on our offline discussion we should have a consistent solution throughout Macaron, such as putting a limit on the size of inputs. We can work on this issue separately.

behnazh-w (approved):

PR #629, opened 2024-02-06
feat: follow indirect repository URLs

### LINKED ISSUE #626: Apache Indirect SCM URLs
Some Apache projects have a URL in their SCM metadata that is currently ignored by Macaron due to not appearing to be a GitHub or similar address. However, past investigation has revealed that these links are actually indirect references to GitHub addresses, as when entered into a browser a redirect is triggered. Macaron could allow such URLs by performing a GET request on the initial URL and then returning the redirect address for further analysis.

An example indirect URL:
`https://gitbox.apache.org/repos/asf?p=commons-parent.git` -> `https://github.com/apache/commons-parent`

benmss:
This PR allows the Repo Finder to return GitHub, etc. URLs that are initially presented as a redirect URL. URLs that can be considered valid redirects are included within a list in the defaults configuration file. URLs found from redirects are validated as per any other URL. A flag is used to prevent infinite redirect loop shenanigans.

Closes #626

### DIFF
--- a/docs/source/pages/using.rst
+++ b/docs/source/pages/using.rst
@@ -263,10 +263,11 @@ Within the configuration file under the ``repofinder.java`` header, three option
 - ``repo_pom_paths`` (Values: List of POM tags) - Determines where to search for repository information in the POM files. E.g. scm.url.
 - ``find_parents`` (Values: True or False) - When enabled, the Repository Finding feature will also search for repository URLs in parents POM files of the current dependency.
 
-Under the related header ``repofinder``, two more options exist: ``find_repos``, and ``use_open_source_insights``:
+Under the related header ``repofinder``, three more options exist: ``find_repos``, ``use_open_source_insights``, and ``redirect_urls``:
 
 - ``find_repos`` (Values: True or False) - Enables or disables the Repository Finding feature.
 - ``use_open_source_insights`` (Values: True or False) - Enables or disables use of Google's Open Source Insights API.
+- ``redirect_urls`` (Values: List of URLs) - Found URLs that exist in this list will be returned as the URL they redirect to.
 
 .. note:: Finding repositories requires at least one remote call, adding some additional overhead to an analysis run.
 
@@ -279,6 +280,9 @@ An example configuration file for utilising this feature:
     [repofinder]
     find_repos = True
     use_open_source_insights = True
+    redirect_urls =
+        gitbox.apache.org
+        git-wip-us.apache.org
 
     [repofinder.java]
     artifact_repositories = https://repo.maven.apache.org/maven2
--- a/docs/source/pages/using.rst
+++ b/docs/source/pages/using.rst
@@ -267,7 +267,7 @@ Under the related header ``repofinder``, three more options exist: ``find_repos`
 
 - ``find_repos`` (Values: True or False) - Enables or disables the Repository Finding feature.
 - ``use_open_source_insights`` (Values: True or False) - Enables or disables use of Google's Open Source Insights API.
-- ``redirect_urls`` (Values: List of URLs) - Found URLs that exist in this list will be returned as the URL they redirect to.
+- ``redirect_urls`` (Values: List of URLs) - These are URLs that are known to redirect to actual repository URLs.
 
 .. note:: Finding repositories requires at least one remote call, adding some additional overhead to an analysis run.
 
--- a/src/macaron/config/defaults.ini
+++ b/src/macaron/config/defaults.ini
@@ -51,6 +51,10 @@ recursive = False
 [repofinder]
 find_repos = True
 use_open_source_insights = True
+# The list of URLs that will be returned as the location they redirect to.
+redirect_urls =
+    gitbox.apache.org
+    git-wip-us.apache.org
 
 [repofinder.java]
 # The list of maven-like repositories to attempt to retrieve artifact POMs from.
--- a/src/macaron/config/defaults.ini
+++ b/src/macaron/config/defaults.ini
@@ -51,7 +51,7 @@ recursive = False
 [repofinder]
 find_repos = True
 use_open_source_insights = True
-# The list of URLs that will be returned as the location they redirect to.
+# These are URLs that are known to redirect to actual repository URLs.
 redirect_urls =
     gitbox.apache.org
     git-wip-us.apache.org
--- a/src/macaron/repo_finder/repo_validator.py
+++ b/src/macaron/repo_finder/repo_validator.py
@@ -1,10 +1,13 @@
-# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2023 - 2024, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """This module exists to validate URLs in terms of their use as a repository that can be analyzed."""
+import urllib.parse
 from collections.abc import Iterable
 
-from macaron.slsa_analyzer.git_url import get_remote_vcs_url
+from macaron.config.defaults import defaults
+from macaron.slsa_analyzer.git_url import clean_url, get_remote_vcs_url
+from macaron.util import send_get_http_raw
 
 
 def find_valid_repository_url(urls: Iterable[str]) -> str:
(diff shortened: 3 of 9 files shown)


src/macaron/slsa_analyzer/git_url.py
```diff
@@ -616,6 +619,20 @@ def parse_remote_url(
...
+            redirect_list = defaults.get_list("repofinder", "redirect_urls", fallback=[])
+            if redirect_list and parsed_url.netloc in redirect_list:
+                response = send_get_http_raw(parsed_url.geturl(), allow_redirects=False)
+                if not response:
+                    return None
+                redirect_url = response.headers.get("location")
+                if not redirect_url:
+                    return None
+                # Pass the redirect_url through this function and prevent loops by setting the redirect parameter to an
+                # empty string.
```
nathanwn, line 633: Sorry, I'm a bit confused about which empty string you are referring to here. Isn't the redirect parameter of type `bool`?

nathanwn:
I'm unsure if redirection should be handled in the `parse_remote_url` function in the `git_url` module.

It seems like this redirection handling should only happen once at the beginning of the analysis of a target. It may be during/right after repo finding instead (which I guess may be part of your intention, confirmed by the fact that you added some redirection setting fields into the Repo Finder section in the `.ini` config).

Otherwise, I don't see why we should modify the `parse_remote_url` function. As you have mentioned previously, this function is used in different places to split a URL into different parts. Its name does not suggest that it should also handle redirection, and if that is the case then this function is doing too many things at once.

behnazh-w:
@benmss I think this PR can be considered as a new feature.

src/macaron/slsa_analyzer/git_url.py
benmss: I will update this comment to match reality.

src/macaron/repo_finder/repo_validator.py
```diff
@@ -30,3 +43,14 @@ def find_valid_repository_url(urls: Iterable[str]) -> str:
 
     # Report the first valid URL from the end of the list.
     return vcs_list.pop()
+
+
+def _resolve_redirects(parsed_url: urllib.parse.ParseResult) -> str | None:
```
tromai, line 48: Because we are trying to resolve the final redirected URL first before passing it to `get_remote_vcs_url`, I wonder why don't we let `requests` handles the redirection for us (as mentioned [here](https://requests.readthedocs.io/en/latest/user/quickstart/?highlight=allow_redirects#redirection-and-history)) and only care about the final URL that it lands on. In the current implementation, we only stops at the first redirection, I'm curious if there would be any case for multiple redirection in our dataset of repositories :thinking: .

tromai: Right, so we only supports URL with at most 1 redirection. Then I think we can leave the current implementation as it is. 
I agree that following the redirection could potentially add more cost to the operation.

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -566,6 +566,33 @@ def get_remote_vcs_url(url: str, clean_up: bool = True) -> str:
     return url_as_str
 
 
+def clean_url(url: str) -> urllib.parse.ParseResult | None:
```
tromai, line 569: Because we have refactored this logic into its own function. I think it's good if we add some unit tests for it (if it's not out of scope for this PR).

src/macaron/repo_finder/repo_validator.py
benmss: Another consideration is the security aspect, that I'll add here for the sake of completeness. Currently Macaron will only try to resolve URLs within the allowed domain list during that part of the analysis. Allowing `requests` to follow these redirects indiscriminately creates a possibility that some undesirable URL is retrieved. That said, as nothing is done with the retrieved data the threat there may still be minimal.

src/macaron/slsa_analyzer/git_url.py
nathanwn: Hmm, to me at least, this may not be necessary though. This function is extracted from the `parse_remote_url` function, which by itself does not have any unit tests and it is so far OK because this is a utility function called by others, each of which has unit tests already. So if we didn't have unit tests for `parse_remote_url` before and were OK with it, I wouldn't think we need unit tests for this one.

tromai: I think @benmss has already added some unit tests :sweat_smile: . 
I agree with @nathanwn that unit tests are not totally necessary as this logic has been tested through unit tests of other functions that call it. However, I don't think we should revert what @benmss has added as it doesn't cause extra issues (for me). Please let me know what you think @nathanwn.   
With that being said, I will approve the changes.

tromai (approved):

src/macaron/slsa_analyzer/git_url.py
benmss: Indeed, I added some before seeing the follow up message. Probably there is no harm in keeping them now that they exist :)

docs/source/pages/using.rst
```diff
@@ -263,10 +263,11 @@ Within the configuration file under the ``repofinder.java`` header, three option
 - ``repo_pom_paths`` (Values: List of POM tags) - Determines where to search for repository information in the POM files. E.g. scm.url.
 - ``find_parents`` (Values: True or False) - When enabled, the Repository Finding feature will also search for repository URLs in parents POM files of the current dependency.
 
-Under the related header ``repofinder``, two more options exist: ``find_repos``, and ``use_open_source_insights``:
+Under the related header ``repofinder``, three more options exist: ``find_repos``, ``use_open_source_insights``, and ``redirect_urls``:
 
 - ``find_repos`` (Values: True or False) - Enables or disables the Repository Finding feature.
 - ``use_open_source_insights`` (Values: True or False) - Enables or disables use of Google's Open Source Insights API.
+- ``redirect_urls`` (Values: List of URLs) - Found URLs that exist in this list will be returned as the URL they redirect to.
```
behnazh-w, line 270: I found this sentence a bit hard to follow. How about:

```suggestion
- ``redirect_urls`` (Values: List of URLs) - These are the URLs that redirect to the actual repository URL.
```

src/macaron/config/defaults.ini
```diff
@@ -51,6 +51,10 @@ recursive = False
 [repofinder]
 find_repos = True
 use_open_source_insights = True
+# The list of URLs that will be returned as the location they redirect to.
```
behnazh-w, line 54: Same comment here.

src/macaron/slsa_analyzer/git_url.py
behnazh-w: Let's keep the test.

behnazh-w (approved):

tests/slsa_analyzer/test_git_url.py
```diff
@@ -287,3 +287,21 @@ def test_parse_git_branch_output(content: str, expected_output: list[str]) -> No
...
+        ("askjdlkajsdlkajsdlkjasldjlk:scm", ""),
+    ],
+)
+def test_clean_url(url: str, expected: str) -> None:
+    """Test the functionality of the clean_url function."""
+    result = git_url.clean_url(url)
+    if result is None:
+        assert expected == ""
+    else:
+        assert expected == result.geturl()
```
nathanwn, line 307: Hmm, why not have the expected results as `None` instead of empty strings?

nathanwn (approved):
Hi. I just had one minor comment for the test. Otherwise, it looks good.

Thanks for the PR.

tests/slsa_analyzer/test_git_url.py
```diff
@@ -287,3 +287,18 @@ def test_parse_git_branch_output(content: str, expected_output: list[str]) -> No
...
+        ("", None),
+        ("scm:git:git@github.com:FasterXML/jackson-databind.git", "git@github.com:FasterXML/jackson-databind.git"),
+        ("https://github.com/commons-io/commons-io", "https://github.com/commons-io/commons-io"),
+        ("askjdlkajsdlkajsdlkjasldjlk:scm", None),
+    ],
+)
+def test_clean_url(url: str, expected: str | None) -> None:
+    """Test the functionality of the clean_url function."""
+    result = git_url.clean_url(url)
+    assert result.geturl() == expected if result is not None else expected is None
```
behnazh-w, line 304: When `url` is an empty string or invalid, we should check that the `clean_url` function returns `None`. The current usage of `expected` value is strange for these cases and we could replace `""` or `None` with any arbitrary value because it doesn't really check the output of the `clean_url` function.

It would be better to split this test into two tests:

1. testing that return value of `clean_url` is `None` when expected.
2. testing that the `result.geturl()` matches the expectation when `clean_url` is not `None`.

benmss:
```suggestion
    assert result.geturl() == expected if expected is not None else result is None
```
Reordering the variables in the in-line if statement would seem to solve the problem, but perhaps it is better to split it still.

behnazh-w: I was also thinking if we could avoid splitting, but I think have two separate tests makes the properties that are tested more explicit.

behnazh-w (approved):

Issue #426, opened 2023-08-21
Use the Python bindings for in-toto/attestation


behnazh-w:
We could possibly use the [in-toto-attestation](https://github.com/in-toto/attestation/tree/858b9917479327f2d1e88a08394a1534ac1f892d/python) package that contains Python bindings for in-toto/attestation to validate the schema.

benmss:
I looked into the library and feel that it is not suitable for our use case. The library exists to properly validate an artifact's provenance, rather than to verify the validity of a given provenance file. From the website for the most relevant method: 

> Performs complete in-toto supply chain verification for a final product. ([Source](https://in-toto.readthedocs.io/en/latest/api.html#supply-chain-verification))

PR #639, opened 2024-02-13
feat: discover slsa v1 provenances for npm packages

### LINKED ISSUE #549: Discover SLSA v1 provenances for npm packages

benmss:
This PR adds:

- Support for downloading SLSA v1 provenance files for NPM packages
- Support for v1 in-toto attestation files, including content validation

It also removes comments relating to intended use of the in-toto attestation library after the investigation reported in #426 

Closes #549

### DIFF
--- a/src/macaron/slsa_analyzer/provenance/intoto/v01/__init__.py
+++ b/src/macaron/slsa_analyzer/provenance/intoto/v01/__init__.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2023 - 2024, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """This module handles in-toto version 0.1 attestations."""
@@ -39,10 +39,6 @@ def validate_intoto_statement(payload: dict[str, JsonType]) -> TypeGuard[InTotoV
 
     Specification: https://github.com/in-toto/attestation/tree/main/spec/v0.1.0#statement.
 
-    TODO: Consider using the in-toto-attestation package (https://github.com/in-toto/attestation/tree/main/python),
-    which contains Python bindings for in-toto attestation.
-    See issue: https://github.com/oracle/macaron/issues/426.
-
     Parameters
     ----------
     payload : dict[str, JsonType]
@@ -64,9 +60,9 @@ def validate_intoto_statement(payload: dict[str, JsonType]) -> TypeGuard[InTotoV
         raise ValidateInTotoPayloadError(
             "The attribute '_type' of the in-toto statement is missing.",
         )
-    if not isinstance(type_, str):
+    if not isinstance(type_, str) or not type_ == "https://in-toto.io/Statement/v0.1":
         raise ValidateInTotoPayloadError(
-            "The value of attribute '_type' in the in-toto statement is invalid: expecting a string.",
+            "The value of attribute '_type' in the in-toto statement must be: 'https://in-toto.io/Statement/v0.1'",
         )
 
     subjects_payload = payload.get("subject")
@@ -106,10 +102,6 @@ def validate_intoto_subject(subject: JsonType) -> TypeGuard[InTotoV01Subject]:
     """Validate a single subject in the in-toto statement.
 
     See specification: https://github.com/in-toto/attestation/tree/main/spec/v0.1.0#statement.
-
-    TODO: Consider using the in-toto-attestation package (https://github.com/in-toto/attestation/tree/main/python),
-    which contains Python bindings for in-toto attestation.
-    See issue: https://github.com/oracle/macaron/issues/426.
 
     Parameters
     ----------
--- a/src/macaron/slsa_analyzer/provenance/intoto/v1/__init__.py
+++ b/src/macaron/slsa_analyzer/provenance/intoto/v1/__init__.py
@@ -167,7 +167,7 @@ def validate_intoto_subject(subject: JsonType) -> TypeGuard[InTotoV1ResourceDesc
     _validate_property(subject, "name", lambda x: isinstance(x, str))
     _validate_property(subject, "downloadLocation", lambda x: isinstance(x, str))
     _validate_property(subject, "mediaType", lambda x: isinstance(x, str))
-    _validate_property(subject, "annotations", is_valid_annotation_map)
+    _validate_property(subject, "annotations", lambda x: isinstance(x, dict))
 
     return True
 
@@ -194,24 +194,6 @@ def is_valid_digest_set(digest: JsonType) -> bool:
             return False
         if not isinstance(digest[key], str):
             return False
-    return True
-
-
-def is_valid_annotation_map(annotation_map: JsonType) -> bool:
-    """Validate the annotation map which is a dictionary with string keys and JsonType values.
-
-    Parameters
-    ----------
-    annotation_map : JsonType
-        The annotation map dictionary.
-
-    Returns
-    -------
-    bool:
-        ``True`` if the annotation map is valid according to the spec.
-    """
-    if not isinstance(annotation_map, dict):
-        return False
     return True
 
 
--- a/src/macaron/slsa_analyzer/provenance/intoto/v1/__init__.py
+++ b/src/macaron/slsa_analyzer/provenance/intoto/v1/__init__.py
@@ -221,7 +221,7 @@ def _validate_property(
     object_: JsonType,
     key: str,
     required: bool,
-    validator_function: Callable[[JsonType], bool],
+    validator: Callable[[JsonType], bool],
     report_errors: bool = True,
 ) -> _ValidityState:
     """Validate the existence and type of target within the passed Json object."""
@@ -238,7 +238,7 @@ def _validate_property(
             return _ValidityState.MISSING
         return _ValidityState.VALID
 
-    valid = validator_function(value)
+    valid = validator(value)
     if not valid:
         if report_errors:
             raise ValidateInTotoPayloadError(f"The attribute {key} of the in-toto subject is invalid.")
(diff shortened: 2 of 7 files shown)


src/macaron/slsa_analyzer/provenance/intoto/v1/__init__.py
```diff
@@ -1,9 +1,37 @@
...
-from typing import TypedDict
+from __future__ import annotations
+
+from collections.abc import Callable
+from typing import TypedDict, TypeGuard
+
+from macaron.slsa_analyzer.provenance.intoto.errors import ValidateInTotoPayloadError
+from macaron.util import JsonType
+
+VALID_ALGORITHMS = [
```
nathanwn, line 14: Since this is only used in the `is_valid_digest_set` function, and the link to the documentation mentioning these algorithms is also there, do you think it is better to move this list into that function, for closer proximity to the documentation link? Either that or we may need to include the documentation link here because just by looking at this list, I feel it is a bit hard to tell where it came from and what it is for.

src/macaron/slsa_analyzer/provenance/intoto/v1/__init__.py
```diff
@@ -12,3 +40,198 @@ class InTotoV1Statement(TypedDict):
...
+        raise ValidateInTotoPayloadError("The subject name is not valid.")
+
+    # At least one of 'uri', 'digest', and 'content' must be valid and present.
+    uri_valid = _validate_property(subject, "uri", True, lambda x: isinstance(x, str))
+    content_valid = _validate_property(subject, "content", True, lambda x: isinstance(x, str))
+    digest_valid = _validate_property(subject, "digest", True, is_valid_digest_set)
+    if not (uri_valid or content_valid or digest_valid):
+        raise ValidateInTotoPayloadError(
+            "One of 'uri', 'digest', or 'content' must be present and valid within 'subject'."
+        )
```
nathanwn, line 166: Combining different issues into a single error message may not be the best idea since it might be hard to tell what the exact issue is when one comes up.

Since you pass the name of the attribute into the function, it may be better to raise the error within the `_validate_property` function itself with a more specific context and error message, rather than returning a `bool` value, only just to convert it to an exception later. All `if` statements like this one can be removed in that case.

src/macaron/slsa_analyzer/provenance/intoto/v1/__init__.py
```diff
@@ -12,3 +40,198 @@ class InTotoV1Statement(TypedDict):
...
+) -> bool:
+    """Validate the existence and type of target within the passed Json object."""
+    if not isinstance(object_, dict):
+        return False
+
+    target = object_.get(target_name)
+    if not target:
+        return not required
+
+    return validator_function(target)
```
nathanwn, line 237: Nit: `property_name` (or `key`) and `value` would be a more familiar naming scheme for me than `target_name` and `target`. Also in that case, `validator_function` can be called `value_validator`, since it is clearer what is being validated, and the word `function` may not be as necessary since it is already type-annotated as a `Callable`. What do you think?

src/macaron/slsa_analyzer/provenance/intoto/v1/__init__.py
```diff
@@ -12,3 +40,198 @@ class InTotoV1Statement(TypedDict):
...
+    Parameters
+    ----------
+    subject : JsonType
+        The JSON element representing a single subject.
+
+    Returns
+    -------
+    TypeGuard[InTotoSubject]
+        ``True`` if the subject element is valid, in which case its type is narrowed to an
+        ``InTotoSubject``; ``False`` otherwise.
```
nathanwn, line 143:
```suggestion
    Returns
    -------
    TypeGuard[InTotoV1ResourceDescriptor]
        ``True`` if the subject element is valid, in which case its type is narrowed to an
        ``InTotoV1ResourceDescriptor``; ``False`` otherwise.
```

src/macaron/slsa_analyzer/provenance/intoto/v1/__init__.py
```diff
@@ -12,3 +44,211 @@ class InTotoV1Statement(TypedDict):
...
+
+    # At least one of 'uri', 'digest', and 'content' must be valid and present.
+    uri_state = _validate_property(subject, "uri", True, lambda x: isinstance(x, str), False)
+    content_state = _validate_property(subject, "content", True, lambda x: isinstance(x, str), False)
+    digest_state = _validate_property(subject, "digest", True, is_valid_digest_set, False)
+    if _ValidityState.VALID not in (uri_state, content_state, digest_state):
+        raise ValidateInTotoPayloadError(
+            f"One of 'uri', 'digest', or 'content' must be present and valid within 'subject'. URI: {uri_state.name}. "
+            f"Content: {content_state.name}. DIGEST: {digest_state.name}."
+        )
```
nathanwn, line 167: I guess something else you can do instead of using this `report_errors` param, is to return from `_validate_property` the object you are validating (if it is not there, return `None`). Then you can use the values to check for the co-existence of attributes.
```suggestion
    # At least one of 'uri', 'digest', and 'content' must be valid and present.
    uri = _validate_property(subject, "uri", required=False, validator=lambda x: isinstance(x, str))
    content = _validate_property(subject, "content", required=False, validator=lambda x: isinstance(x, str))
    digest = _validate_property(subject, "digest", required=False, validator=is_valid_digest_set)
    if not any([uri, content, digest]):
        raise ValidateInTotoPayloadError("One of 'uri', 'digest', or 'content' must be present and valid within 'subject'")
```

The `_validate_property` function then can be like this:

```python
def _validate_property(
    object_: dict[str, JsonType],
    key: str,
    required: bool,
    validator: Callable[[JsonType], bool],
) -> _ValidityState:
    """Validate the existence and type of target within the passed Json object."""
    value = object_.get(key)

    if not value:
        if required:
            raise ValidateInTotoPayloadError(f"The attribute {key} of the in-toto subject is missing.")
        return None

    if not validator(value):
        raise ValidateInTotoPayloadError(f"The attribute {key} of the in-toto subject is invalid.")

    return value
```

src/macaron/slsa_analyzer/provenance/intoto/v1/__init__.py
```diff
@@ -12,3 +44,211 @@ class InTotoV1Statement(TypedDict):
...
+    validator: Callable[[JsonType], bool],
+    report_errors: bool = True,
+) -> _ValidityState:
+    """Validate the existence and type of target within the passed Json object."""
+    if not isinstance(object_, dict):
+        # The passed object is checked before this function is called, but mypy requires it be explicitly checked here
+        #  too. Also known as: this branch should never be taken.
+        return _ValidityState.INVALID
+
+    value = object_.get(key)
```
nathanwn, line 233: Since you have already checked if `subject` is a `dict`, `object_` here can be type-annotated as `dict`. That way you don't need the `if` statement to narrow the type again to make `mypy` happy.

```suggestion
def _validate_property(
    object_: dict[str, JsonType],
    key: str,
    required: bool,
    validator: Callable[[JsonType], bool],
    report_errors: bool = True,
) -> _ValidityState:
    """Validate the existence and type of target within the passed Json object."""
   value = object_.get(key)
```

src/macaron/slsa_analyzer/provenance/intoto/v1/__init__.py
```diff
@@ -12,3 +43,189 @@ class InTotoV1Statement(TypedDict):
...
+    digest = _validate_property(subject, "digest", is_valid_digest_set)
+    if not any([uri, content, digest]):
+        raise ValidateInTotoPayloadError(
+            "One of 'uri', 'digest', or 'content' must be present and valid within 'subject'."
+        )
+
+    _validate_property(subject, "name", lambda x: isinstance(x, str))
+    _validate_property(subject, "downloadLocation", lambda x: isinstance(x, str))
+    _validate_property(subject, "mediaType", lambda x: isinstance(x, str))
+    _validate_property(subject, "annotations", is_valid_annotation_map)
```
nathanwn, line 170: Wouldn't a lambda be enough in this case, since you only need to check with `isinstance` and this is consistent with the rest of the function?

```suggestion
    _validate_property(subject, "annotations", lambda x: isinstance(x, dict))
```

src/macaron/slsa_analyzer/provenance/intoto/v01/__init__.py
```diff
@@ -39,10 +39,6 @@ def validate_intoto_statement(payload: dict[str, JsonType]) -> TypeGuard[InTotoV
 
     Specification: https://github.com/in-toto/attestation/tree/main/spec/v0.1.0#statement.
 
-    TODO: Consider using the in-toto-attestation package (https://github.com/in-toto/attestation/tree/main/python),
```
nathanwn, line 42: I have a question about this: I vaguely remember that you mentioned looking at this package and seeing some issues with it. I wonder what was the issue you saw and whether we want to reconsider using this package again in the future.

benmss: It does not provide any methods relating to verifying the validity of a given attestation file in the straightforward way that we require.

tests/slsa_analyzer/package_registry/test_npm_registry.py
```diff
@@ -17,14 +17,14 @@
 
 @pytest.fixture(name="npm_registry")
 def create_npm_registry() -> NPMRegistry:
-    """Create an npm registry instance."""
+    """Create an NPM registry instance."""
```
behnazh-w, line 20: `npm` should be lowercase. See https://docs.npmjs.com/cli/v7/using-npm/registry

benmss: But is the comment referring to the class or the external system the class represents? I'll change it, I'm just commenting that it is a bit hard to follow their preferred naming convention when it directly conflicts with our own class naming convention and we ignore it there.

behnazh-w: The comment is referring to the external `npm` registry. But can I see how it gets tricky to figure out the right naming convention.

behnazh-w (approved):

PR #641, opened 2024-02-15
chore: use priority ordering in java repo finder

### LINKED ISSUE #640: Use priority ordering when retrieving SCM fields from POM files
The Java Repo Finder extracts SCM metadata from POM files in order to determine the related repository URL. By default, up to three URLs can be extracted from SCM section, but the list can be extended arbitrarily by Macaron users. In one particular case, it was discovered that one of the three default URLs was different from the other two, and also incorrect. Generally, it seems that the `URL` field is the most reliable choice, and allowing it to be checked first would Repo Finder accuracy in some cases.

benmss:
This PR makes the configuration property `repo_pom_paths` a priority list. When the Repo Finder is resolving URLs to repositories, the first valid URL (based on the ordering of the `repo_pom_paths`) will be returned. By default, the highest priority SCM entry in a POM should be `url`, as this most reliably represents the URL of the source repository.

Closes #640

### DIFF
--- a/src/macaron/repo_finder/repo_validator.py
+++ b/src/macaron/repo_finder/repo_validator.py
@@ -13,6 +13,8 @@ from macaron.util import send_get_http_raw
 def find_valid_repository_url(urls: Iterable[str]) -> str:
     """Find a valid URL from the provided URLs.
 
+    URLs are assumed to be a priority list, with the first match preventing further processing.
+
     Parameters
     ----------
     urls : Iterable[str]
@@ -23,26 +25,17 @@ def find_valid_repository_url(urls: Iterable[str]) -> str:
     str
         A valid URL, or an empty string if none can be found.
     """
-    pruned_list = []
     for url in urls:
         parsed_url = clean_url(url)
         if not parsed_url:
             # URLs that failed to parse can be rejected here.
             continue
         redirect_url = resolve_redirects(parsed_url)
-        # If a redirect URL is found add it, otherwise add the parsed url.
-        pruned_list.append(redirect_url if redirect_url else parsed_url.geturl())
+        checked_url = get_remote_vcs_url(redirect_url if redirect_url else parsed_url.geturl())
+        if checked_url:
+            return checked_url
 
-    vcs_set = {get_remote_vcs_url(value) for value in pruned_list if get_remote_vcs_url(value) != ""}
-
-    # To avoid non-deterministic results we sort the URLs.
-    vcs_list = sorted(vcs_set)
-
-    if len(vcs_list) < 1:
-        return ""
-
-    # Report the first valid URL from the end of the list.
-    return vcs_list.pop()
+    return ""
 
 
 def resolve_redirects(parsed_url: urllib.parse.ParseResult) -> str | None:
--- a/src/macaron/repo_finder/repo_validator.py
+++ b/src/macaron/repo_finder/repo_validator.py
@@ -13,8 +13,6 @@ from macaron.util import send_get_http_raw
 def find_valid_repository_url(urls: Iterable[str]) -> str:
     """Find a valid URL from the provided URLs.
 
-    URLs are assumed to be a priority list, with the first match preventing further processing.
-
     Parameters
     ----------
     urls : Iterable[str]
@@ -23,12 +21,12 @@ def find_valid_repository_url(urls: Iterable[str]) -> str:
     Returns
     -------
     str
-        A valid URL, or an empty string if none can be found.
+        The first valid URL from the iterable, or an empty string if none can be found.
     """
     for url in urls:
         parsed_url = clean_url(url)
         if not parsed_url:
-            # URLs that failed to parse can be rejected here.
+            # URLs that fail to parse can be rejected here.
             continue
         redirect_url = resolve_redirects(parsed_url)
         checked_url = get_remote_vcs_url(redirect_url if redirect_url else parsed_url.geturl())
--- a/tests/repo_finder/test_repo_finder.py
+++ b/tests/repo_finder/test_repo_finder.py
@@ -2,13 +2,14 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """This module tests the repo finder."""
+import os
+from pathlib import Path
 
 import pytest
-from defusedxml.ElementTree import fromstring
 from packageurl import PackageURL
 
+from macaron.config.defaults import load_defaults
 from macaron.config.target_config import Configuration
-from macaron.repo_finder import repo_validator
 from macaron.repo_finder.repo_finder_java import JavaRepoFinder
 from macaron.slsa_analyzer.analyzer import Analyzer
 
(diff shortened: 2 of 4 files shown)


src/macaron/repo_finder/repo_validator.py
```diff
@@ -13,6 +13,8 @@
 def find_valid_repository_url(urls: Iterable[str]) -> str:
```
tromai, line 13: I think it's important to make sure that the order of URLS obtained from the tags described in `repo_pom_paths` is kept when provided to `find_valid_repository_url`. Perhaps we need to add necessary tests to ensure that this behavior is maintained (unless there are existing tests for this behavior that I didn't remember :thinking: ).

src/macaron/repo_finder/repo_validator.py
```diff
@@ -13,6 +13,8 @@
...
     """Find a valid URL from the provided URLs.
 
+    URLs are assumed to be a priority list, with the first match preventing further processing.
```
tromai, line 16: I don't think we need to mention that this function accepted a priority url list because the implementation of this function doesn't care what "priority" would mean for `urls`. It's up to the callers to ensure what they provide has been ordered correctly. 
I think we can just say that this function will return the first valid repository URL from `urls` or something similar.

tests/e2e/repo_finder/repo_finder.py
```diff
@@ -62,6 +62,16 @@ def test_repo_finder() -> int:
     if not find_repo(PackageURL.from_string("pkg:cargo/rand_core")):
         return os.EX_UNAVAILABLE
 
+    # Test priority ordering of Repo Finder using artifact with differing URLs in the POM's SCM section.
```
tromai, line 65: Thanks for adding the test. What we are trying to test here is good. However, I think it relies a lot on the outside states:
- The pom file stored in maven central for `pkg:maven/org.glassfish.jersey/project@3.0.12` which we obtain and inspect.
- The network connection.
Because the behavior we are trying to test doesn't have to rely on those things (we are testing if given a pom file, the order of `repo_pom_paths` is reflected correctly in the order of valid scm urls extracted). I believe it would be even better if we make this a unit test with a pom file (you can reuse from `pkg:maven/org.glassfish.jersey/project@3.0.12` or crafted a custom pom.xml file) so that the test is deterministic (or you can add an additional unit test and keep this e2e test). 
Please let me know what you think about it.

benmss: It does rely on a very specific POM to properly test the ordering. While that POM is unlikely to change, it would be troublesome to find another in the case that it did. Therefore I think you are right that this test can be replaced with a unit test and custom POM.

tests/repo_finder/test_repo_finder.py
```diff
@@ -69,3 +72,36 @@ def test_resolve_analysis_target(
...
+    connection_url = repo_validator.find_valid_repository_url(connection_urls)
+    assert connection_url
+
+    urls = repo_finder._find_scm(pom, ["scm.url", "scm.connection"])  # pylint: disable=W0212
+    assert urls
+    url = repo_validator.find_valid_repository_url(urls)
+    assert url
+
+    # Ensure found URLs differ.
+    assert connection_url != url
```
tromai, line 107: I have a few comments on this test. 
1. I believe what we are testing here is "given the list of tags to search for urls, make sure that the urls found by searching such tags are stored in the same order as their tag in that list"
Therefore, I think we only need to test for `repo_finder._find_scm` in this test function. The method `find_valid_repository_url` is irrelevant and can be removed as it doesn't care about the order of elements in `urls`.
2. Because we have a custom pom file, I think we can assert the return value of `repo_dinfer._find_scm` directly 
```python
    # Retrieve SCM from POM.
    connection_urls = repo_finder._find_scm(pom, ["scm.connection", "scm.url"])  # pylint: disable=W0212
    assert connection_urls == ["scm:git:git@github.com:oracle-samples/macaron.git", "https://github.com/oracle/macaron"]

    urls = repo_finder._find_scm(pom, ["scm.url", "scm.connection"])  # pylint: disable=W0212
    assert urls == ["https://github.com/oracle/macaron", "scm:git:git@github.com:oracle-samples/macaron.git"]
```
A further improvement would be to turn this test function into a parametrized test (optional):
```
@pytest.mark.parametrize(
    ("tag_list", "expected"),
    [
        (
            ["scm.connection", "scm.url"],
            ["scm:git:git@github.com:oracle-samples/macaron.git", "https://github.com/oracle/macaron"],
        ),
        (
            ["scm.url", "scm.connection"],
            ["https://github.com/oracle/macaron", "scm:git:git@github.com:oracle-samples/macaron.git"]
        )
    ],
)
def test_pom_extraction_ordering(tag_list: list[str], expected: list[str]) -> None:
    """Test the ordering of elements extracted from the POM is correct and maintained."""
    pom_text = """
    <project>
        <url>https://example.org</url>
        <scm>
            <connection>scm:git:git@github.com:oracle-samples/macaron.git</connection>
            <developerConnection>scm:git:git@github.com:oracle/macaron.git</developerConnection>
            <url>https://github.com/oracle/macaron</url>
        </scm>
        <properties>
            <url>1.9.15</url>
        </properties>
    </project>
    """
    pom = fromstring(pom_text)
    repo_finder = JavaRepoFinder()

    # Retrieve SCM from POM.
    assert expected == repo_finder._find_scm(pom, tag_list) # pylint: disable=W0212
```
3. Alternatively, we can convert this test case into
```
@pytest.mark.parametrize(
    ("user_config_input", "expected"),
    [
        (
            """
            repo_pom_paths =
                scm.connection
                scm.url
            """,
            ["scm:git:git@github.com:oracle-samples/macaron.git", "https://github.com/oracle/macaron"],
        ),
        (
            """
            repo_pom_paths =
                scm.url
                scm.connection
            """,
            ["https://github.com/oracle/macaron", "scm:git:git@github.com:oracle-samples/macaron.git"]
        )
    ],
)
def test_pom_extraction_ordering(tmp_path: Path,  user_config_input: str, expected: list[str]) -> None:
    """Test the ordering of elements extracted from the POM is correct and maintained."""
    pom_text = """
    <project>
        <url>https://example.org</url>
        <scm>
            <connection>scm:git:git@github.com:oracle-samples/macaron.git</connection>
            <developerConnection>scm:git:git@github.com:oracle/macaron.git</developerConnection>
            <url>https://github.com/oracle/macaron</url>
        </scm>
        <properties>
            <url>1.9.15</url>
        </properties>
    </project>
    """
    user_config_path = os.path.join(tmp_path, "config.ini")
    with open(user_config_path, "w", encoding="utf-8") as user_config_file:
        user_config_file.write(user_config_input)
    load_defaults(user_config_path)
    
    repo_finder = JavaRepoFinder()

    # Retrieve SCM from POM.
    assert expected == repo_finder._read_pom(pom_text) # pylint: disable=W0212
```

tromai:
I have tested your implementation and it seems that because calling `defaults.get_list` method (implementation [here](https://github.com/oracle/macaron/blob/744335d17a3c3485bbeecd94a22048c409a166d0/src/macaron/config/defaults.py#L18)) without `duplicated_ok=True` will first convert the list specified in `defaults.ini` to a set before returning. A set in Python doesn't preserve the inserting order of elements. This will result in the order of elements not following the original order as specified in `defaults.ini` and impact the behavior of `JavaRepoFinder._read_pom`. 
@behnazh-w It sounds to me that we need to improve `defaults.get_list` regarding the order of the elements in the list. `defaults.get_list` altering the order of the list as the default behavior is not what we expected. However, improving that function might be out of scope for this PR. For now I think @benmss can address it by calling 
```python
tags = defaults.get_list("repofinder.java", "repo_pom_paths", duplicated_ok=True)
```
At this [place](https://github.com/oracle/macaron/blob/744335d17a3c3485bbeecd94a22048c409a166d0/src/macaron/repo_finder/repo_finder_java.py#L163).
I don't have a strong opinion on whether we fix it in this PR or in a separate PR.

tests/repo_finder/test_repo_finder.py
tromai: I would prefer the third option as we are testing the behavior of this feature in a similar manner as your original test case.

tests/repo_finder/test_repo_finder.py
```diff
@@ -74,34 +75,49 @@ def test_resolve_analysis_target(
...
+@pytest.mark.parametrize(
+    ("user_config_input", "expected"),
+    [
+        (
+            """
+                [repofinder.java]
+                repo_pom_paths =
+                    scm.connection
+                    scm.url
+                """,
```
tromai, line 87: Hmm is the extra spaces before `"""` here intentional? Otherwise, everything else looks good to me.

tests/repo_finder/test_repo_finder.py
```diff
@@ -69,3 +73,51 @@ def test_resolve_analysis_target(
...
+            ["scm:git:git@github.com:oracle-samples/macaron.git", "https://github.com/oracle/macaron"],
+        ),
+        (
+            """
+                [repofinder.java]
+                repo_pom_paths =
+                    scm.url
+                    scm.connection
+                """,
```
tromai, line 96: Same comment as https://github.com/oracle/macaron/pull/641#discussion_r1501922938

tromai (approved):
Thanks for the changes. LGTM (given the assumption that the issue discussed in https://github.com/oracle/macaron/pull/641#issuecomment-1960742681 is going to be addressed in a separated PR).

behnazh-w:
>@behnazh-w It sounds to me that we need to improve `defaults.get_list` regarding the order of the elements in the list. `defaults.get_list` altering the order of the list as the default behavior is not what we expected. 

Yes, agreed. @tromai can you please work on this issue in a separate PR?

behnazh-w (approved):

PR #643, opened 2024-02-15
chore: update repo finder e2e test

### LINKED ISSUE #642: Update Repo Finder e2e test for new indirect URL feature
Issue #626 added a new feature that resolves indirect repository URLs found in Java metadata. This functionality affects the behaviour of the Repo Finder, and the related e2e test was not updated to reflect this change.

benmss:
This PR updates the Repo Finder e2e test to support the changes made by the new indirect Apache URL support feature.

Closes #642

### DIFF
--- a/tests/e2e/repo_finder/repo_finder.py
+++ b/tests/e2e/repo_finder/repo_finder.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2023 - 2024, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """This script tests the functionality of the repo finder's remote API calls."""
@@ -29,6 +29,7 @@ def test_repo_finder() -> int:
     if not defaults.has_section("repofinder"):
         defaults.add_section("repofinder")
     defaults.set("repofinder", "use_open_source_insights", "True")
+    defaults.set("repofinder", "redirect_urls", "gitbox.apache.org git-wip-us.apache.org")
 
     if not defaults.has_section("git_service.github"):
         defaults.add_section("git_service.github")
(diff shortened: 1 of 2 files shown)


tromai (approved):
Thanks for the change. LGTM

tests/e2e/repo_finder/repo_finder.py
```diff
@@ -62,6 +63,10 @@ def test_repo_finder() -> int:
     if not find_repo(PackageURL.from_string("pkg:cargo/rand_core")):
         return os.EX_UNAVAILABLE
 
+    # Test redirecting URL from Apache commons-io package.
```
tromai, line 66: I believe this new test case was added to test the redirection URL feature that was added in #626. 
However, I am a bit unsure about the behavior we are testing here. 
Are we testing to see
- Whether the repo finder can follow the redirect URL from `pkg:maven/commons-io/commons-io@2.15.1` (no matter the final URL is) 
- Or are we trying to see if the final URL obtained from this PURL `pkg:maven/commons-io/commons-io@2.15.1` is the correct redirected PURL we are looking for (which I believe is `https://github.com/apache/commons-parent`)

benmss: Yes, to your first bullet point. Partially yes, for your second point. The URL returned is validated as part of the Repo Finder process, so we are checking that some valid URL was returned. An indirect URL is not valid, and won't be returned. I could change the test to require an exact URL, but I'm not sure it would make much difference.
The correct URL would be `https://github.com/apache/commons-io`.

If we did decide we wanted it to match an exact URL, then most likely all other tests in the same file would need to be adjusted to match exactly too.

tromai: >  The correct URL would be https://github.com/apache/commons-io.
Oops, you are right, I made a mistake there.
Thanks for the explanation on the reason behind this test case. 
(This might not be in scope of this PR) My further question would be, what if for some reason the URL found for `pkg:maven/commons-io/commons-io@2.15.1` does not lead to a redirection anymore, will this test report that case?
For example, `https://git-wip-us.apache.org/repos/asf?p=commons-io.git` doesn't redirect to `https://github.com/apache/commons-io` anymore, repo finder would take `https://git-wip-us.apache.org/repos/asf?p=commons-io.git` and report that it's not a valid URL right ? :thinking:

tromai (approved):

behnazh-w (approved):

PR #650, opened 2024-02-23
chore: modify check execution from multi-threaded to single-threaded

nathanwn:
This pull request modifies the check execution in Macaron from a multi-threaded implementation to a single-threaded one.

This change is for a couple of reasons:

* The multithreaded solution was initially for performance purposes. However, parallelizing check execution may not be the best approach for speeding up Macaron. In addition, Macaron almost never runs with multithreading (2 or more runners) enabled.
* The new single-threaded implementation makes it easier for debugging and profiling since it is better compatible with native Python toolings (e.g. cProfile and [pdb](https://github.com/python/cpython/issues/85743)).
* The new single-threaded implementation also simplifies the check execution implementation and makes it easier to maintain.


nicallen (approved):

tests/slsa_analyzer/checks/test_registry.py
```diff
@@ -213,7 +193,6 @@ def test_prepare_successfully(self) -> None:
         # at least one runner initialized and no check circular
```
tromai, line 193: I think we can remove `at least one runner initialized` here.

tests/slsa_analyzer/runner/test_runner.py
tromai, line 1: As we have removed the concept of Runner from Macaron, I think we can move this test case into the `tests/slsa_analyzer/checks/test_registry.py` module and make sure to remove any mention to `runner` in the comment. I don't have a strong opinion, as long as the remained mentions to `runner` are removed. For example:
- https://github.com/oracle/macaron/blob/d9de38d4af71eed59f25d3653d2544370649d400/tests/slsa_analyzer/runner/test_runner.py#L44-L45
- https://github.com/oracle/macaron/blob/d9de38d4af71eed59f25d3653d2544370649d400/tests/slsa_analyzer/runner/test_runner.py#L4

tromai (changes requested):
Thanks for the PR. I have left some change requests,

tromai:
https://github.com/oracle/macaron/pull/400 can be closed after this Pull Request is merged.

src/macaron/slsa_analyzer/registry.py
```diff
@@ -507,149 +463,45 @@ def scan(self, target: AnalyzeContext) -> dict[str, CheckResult]:
...
-    def _init_runners(self) -> None:
-        """Initiate runners from values in defaults.ini."""
-        self.runner_num = defaults.getint("runner", "runner_num", fallback=1)
-        self.runner_timeout = defaults.getint("runner", "timeout", fallback=5)
-        if not self.runners:
-            self.runners.extend([Runner(self, i) for i in range(self.runner_num)])
+            try:
+                results[check_id] = check.run(target, skipped_info)
+            except Exception as exc:  # pylint: disable=broad-exception-caught
+                logger.error("Exception in check %s: %s.", check_id, exc)
```
behnazh-w, line 500:
```suggestion
                logger.error("Exception in check %s: %s. Run in verbose mode to get more information.", check_id, exc)
                logger.debug(traceback.format_exc())
```

tests/slsa_analyzer/runner/test_runner.py
nathanwn: See 3349380e5d1981593abb789938affc1dcf20894d. I have renamed the test file to `test_registry_e2e.py` to distinguish it from the other `test_registry.py` file, which seems to me more or less contains unit tests for the internals of the `Registry` class.

src/macaron/slsa_analyzer/registry.py
nathanwn: See 6b0dbab11ecf3cf16be794321294813a48dbbe4b.

tests/slsa_analyzer/checks/test_registry.py
nathanwn: Fixed in 477c99b8929876b4ba6d5766619d266c78d095fb.

tromai (approved):
Thanks for the PR. Everything LGTM.

behnazh-w (approved):

PR #653, opened 2024-02-29
feat: use provenance to find commits for supported PURL types.

benmss:
This PR uses provenance files to find commits and repositories for analysis targets.

Provenance can be discovered as followed:

- In a user defined JFrog repository, for Gradle and Maven targets.
- From npm, for npm libraries built with provenance.
- From the user, if provided as a CLI parameter.

The supported provenance types are SLSA v0.1, v0.2, and v1, Witness v0.1. Supported payloads are in-toto v0.1 and v1.

### DIFF
--- a/src/macaron/repo_finder/provenance_extractor.py
+++ b/src/macaron/repo_finder/provenance_extractor.py
@@ -1,0 +1,276 @@
+# Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""This module contains methods for extracting repository and commit metadata from provenance files."""
+import logging
+from typing import overload
+
+from macaron.slsa_analyzer.provenance import intoto
+from macaron.slsa_analyzer.provenance.intoto import InTotoPayload, InTotoV1Payload, InTotoV01Payload
+from macaron.util import JsonType
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+def extract_repo_and_commit_from_provenance(payload: InTotoPayload) -> tuple[str, str]:
+    """Extract the repository and commit metadata from the passed provenance payload.
+
+    Parameters
+    ----------
+    payload: InTotoPayload
+        The payload to extract from.
+
+    Returns
+    -------
+    tuple[str, str]
+        The repository URL and commit hash if found, a pair of empty strings otherwise.
+    """
+    predicate_type = payload.statement.get("predicateType")
+    if isinstance(payload, InTotoV1Payload):
+        if isinstance(payload, InTotoV1Payload):
+            if predicate_type == "https://slsa.dev/provenance/v1":
+                return _extract_from_slsa_v1(payload)
+    elif isinstance(payload, InTotoV01Payload):
+        if predicate_type == "https://slsa.dev/provenance/v0.2":
+            return _extract_from_slsa_v02(payload)
+        if predicate_type == "https://slsa.dev/provenance/v0.1":
+            return _extract_from_slsa_v01(payload)
+        if predicate_type == "https://witness.testifysec.com/attestation-collection/v0.1":
+            return _extract_from_witness_provenance(payload)
+
+    logger.debug(
+        "Extraction from provenance not supported for versions: predicate_type %s, in-toto %s.",
+        predicate_type,
+        payload.__class__,
+    )
+    return "", ""
+
+
+def _extract_from_slsa_v01(payload: InTotoV01Payload) -> tuple[str, str]:
+    """Extract the repository and commit metadata from the slsa v01 provenance payload."""
+    predicate: dict[str, JsonType] | None = payload.statement.get("predicate")
+    if not predicate:
+        return "", ""
+
+    # The repository URL and commit are stored inside an entry in the list of predicate -> materials.
+    # In predicate -> recipe -> definedInMaterial we find the list index that points to the correct entry.
+    list_index = _json_extract(predicate, ["recipe", "definedInMaterial"], int)
+    if not list_index:
+        return "", ""
+
+    material_list = _json_extract(predicate, ["materials"], list)
+    if not material_list:
+        return "", ""
+
+    if list_index >= len(material_list):
+        return "", ""
+    material = material_list[list_index]
+    if not material or not isinstance(material, dict):
+        return "", ""
+
+    uri = material.get("uri")
+    if not uri:
+        logger.debug("Could not extract repository URL.")
+    repo = _clean_spdx(uri)
+
+    digest_set = material.get("digest")
+    if not digest_set or not isinstance(digest_set, dict):
+        return "", ""
+    commit = _extract_commit_from_digest(digest_set)
+    if not commit:
+        logger.debug("Could not extract commit.")
+        return "", ""
+
+    return repo, commit
+
+
+def _extract_from_slsa_v02(payload: InTotoV01Payload) -> tuple[str, str]:
+    """Extract the repository and commit metadata from the slsa v02 provenance payload."""
+    predicate: dict[str, JsonType] | None = payload.statement.get("predicate")
+    if not predicate:
+        return "", ""
+
+    # The repository URL and commit are stored within the predicate -> invocation -> configSource object.
+    # See https://slsa.dev/spec/v0.2/provenance
+    uri = _json_extract(predicate, ["invocation", "configSource", "uri"], str)
+    if not uri:
+        logger.debug("Could not extract repo URL.")
+        return "", ""
+    repo = _clean_spdx(uri)
+
+    digest_set = _json_extract(predicate, ["invocation", "configSource", "digest"], dict)
+    if not digest_set:
+        return "", ""
+    commit = _extract_commit_from_digest(digest_set)
+    if not commit:
+        logger.debug("Could not extract commit.")
+        return "", ""
+
+    return repo, commit
+
+
+def _extract_from_slsa_v1(payload: InTotoV1Payload) -> tuple[str, str]:
+    """Extract the repository and commit metadata from the slsa v1 provenance payload."""
+    predicate: dict[str, JsonType] | None = payload.statement.get("predicate")
+    if not predicate:
+        return "", ""
+
+    build_def = _json_extract(predicate, ["buildDefinition"], dict)
+    if not build_def:
+        return "", ""
+    build_type = _json_extract(build_def, ["buildType"], str)
+    if not build_type:
+        return "", ""
+
+    # Extract the repository URL.
+    repo = None
+    if build_type == "https://slsa-framework.github.io/gcb-buildtypes/triggered-build/v1":
+        repo = _json_extract(build_def, ["externalParameters", "sourceToBuild", "repository"], str)
+        if not repo:
+            repo = _json_extract(build_def, ["externalParameters", "configSource", "repository"], str)
+    if build_type == "https://slsa-framework.github.io/github-actions-buildtypes/workflow/v1":
+        repo = _json_extract(build_def, ["externalParameters", "workflow", "repository"], str)
+
+    if not repo:
+        logger.debug("Failed to extract repository URL from provenance.")
+        return "", ""
+
+    # Extract the commit hash.
+    commit = None
+    deps = _json_extract(build_def, ["resolvedDependencies"], list)
+    if not deps:
+        return "", ""
+    for dep in deps:
+        if not isinstance(dep, dict):
+            continue
+        uri = dep["uri"]
+        url = _clean_spdx(uri)
+        if url != repo:
+            continue
+        if build_type == "https://slsa-framework.github.io/gcb-buildtypes/triggered-build/v1":
+            commit_dict = _json_extract(dep, ["digest"], dict)
+            if not commit_dict:
+                continue
+            commit = _extract_commit_from_digest(commit_dict)
+        if build_type == "https://slsa-framework.github.io/github-actions-buildtypes/workflow/v1":
+            commit = _json_extract(dep, ["digest", "gitCommit"], str)
+
+    if not commit:
+        logger.debug("Failed to extract commit hash from provenance.")
+        return "", ""
+
+    return repo, commit
+
+
+def _extract_commit_from_digest(digest: dict[str, JsonType]) -> str | None:
+    """Extract the commit from the passed DigestSet.
+
+    The DigestSet is an in-toto object that maps algorithm types to commit hashes (digests).
+    """
+    # TODO decide on a preference for which algorithm to accept.
+    if len(digest.keys()) > 1:
+        logger.debug("DigestSet contains multiple algorithms: %s", digest.keys())
+
+    for key in digest:
+        if key in intoto.v1.VALID_ALGORITHMS:
+            value = digest.get(key)
+            if isinstance(value, str):
+                return value
+    return None
+
+
+def _clean_spdx(uri: str) -> str:
+    """Clean the passed SPDX URI and return the normalised URL it represents.
+
+    A SPDX URI has the form: git+https://example.com@refs/heads/main
+    """
+    url, _, _ = uri.lstrip("git+").rpartition("@")
+    return url
+
+
+def _extract_from_witness_provenance(payload: InTotoV01Payload) -> tuple[str, str]:
+    """Extract the repository and commit metadata from the witness provenance file found at the passed path.
+
+    To successfully return the commit and repository URL, the payload must respectively contain a Git attestation, and
+    either a GitHub or GitLab attestation.
+
+    Parameters
+    ----------
+    payload: InTotoPayload
+        The payload to extract from.
+
+    Returns
+    -------
+    tuple[str, str]
+        The repository URL and commit hash if found, a pair of empty strings otherwise.
+    """
+    predicate: dict[str, JsonType] | None = payload.statement.get("predicate")
+    if not predicate:
+        return "", ""
+    attestations = _json_extract(predicate, ["attestations"], list)
+    if not attestations:
+        return "", ""
+    commit: str | None = None
+    repo: str | None = None
+    for entry in attestations:
+        if not isinstance(entry, dict):
+            continue
+        entry_type = entry.get("type")
+        if not entry_type:
+            continue
+        if entry_type.startswith("https://witness.dev/attestations/git/"):
+            commit = _json_extract(entry, ["attestation", "commithash"], str)
+        elif entry_type.startswith("https://witness.dev/attestations/gitlab/") or entry_type.startswith(
+            "https://witness.dev/attestations/github/"
+        ):
+            repo = _json_extract(entry, ["attestation", "projecturl"], str)
+
+    if not commit or not repo:
+        logger.debug("Could not extract repo and commit from provenance.")
+        return "", ""
+
+    return repo, commit
+
+
+@overload
+def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[int]) -> int | None:
+    ...
+
+
+@overload
+def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[list]) -> list | None:
+    ...
+
+
+@overload
+def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[dict]) -> dict | None:
+    ...
+
+
+@overload
+def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[str]) -> str | None:
+    ...
+
+
+def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[JsonType]) -> JsonType:
+    """Return the value found by following the list of depth-sequential keys inside the passed dictionary.
+
+    The value's type is validated against the passed type.
+    """
+    target = entry
+    for index, key in enumerate(keys):
+        if key not in target:
+            logger.debug("Key not found in JSON: %s", key)
+            return None
+        next_target = target[key]
+        if index == len(keys) - 1:
+            if isinstance(next_target, type_):
+                return next_target
+        else:
+            if not isinstance(next_target, dict):
+                logger.debug("Expected dict found: %s", next_target.__class__)
+                break
+            target = next_target
+
+    logger.debug("Failed to find %s in JSON dictionary", " > ".join(keys))
+    return None
--- a/src/macaron/repo_finder/provenance_extractor.py
+++ b/src/macaron/repo_finder/provenance_extractor.py
@@ -212,7 +212,7 @@ def _extract_commit_from_digest_set(digest_set: dict[str, JsonType], valid_algor
             value = digest_set.get(key)
             if isinstance(value, str):
                 return value
-    raise ProvenanceExtractionException("No valid digest in digest set.")
+    raise ProvenanceExtractionException(f"No valid digest in digest set: {digest_set.keys()} not in {valid_algorithms}")
 
 
 def _clean_spdx(uri: str) -> str:
--- a/src/macaron/repo_finder/provenance_extractor.py
+++ b/src/macaron/repo_finder/provenance_extractor.py
@@ -16,9 +16,9 @@ class ProvenanceExtractionException(MacaronError):
     """When there is an error while extracting from provenance."""
 
 
-SLSA_V01_DIGEST_SET_ALGORITHMS = ["sha1"]
-SLSA_V02_DIGEST_SET_ALGORITHMS = ["sha1"]
-SLSA_V1_DIGEST_SET_ALGORITHMS = ["sha1", "gitCommit"]
+SLSA_V01_DIGEST_SET_GIT_ALGORITHMS = ["sha1"]
+SLSA_V02_DIGEST_SET_GIT_ALGORITHMS = ["sha1"]
+SLSA_V1_DIGEST_SET_GIT_ALGORITHMS = ["sha1", "gitCommit"]
 
 
 def extract_repo_and_commit_from_provenance(payload: InTotoPayload) -> tuple[str, str]:
@@ -90,7 +90,7 @@ def _extract_from_slsa_v01(payload: InTotoV01Payload) -> tuple[str, str]:
     repo = _clean_spdx(uri)
 
     digest_set = json_extract(material, ["digest"], dict)
-    commit = _extract_commit_from_digest_set(digest_set, SLSA_V01_DIGEST_SET_ALGORITHMS)
+    commit = _extract_commit_from_digest_set(digest_set, SLSA_V01_DIGEST_SET_GIT_ALGORITHMS)
 
     if not commit:
         raise ProvenanceExtractionException("Failed to extract commit hash from provenance.")
@@ -112,7 +112,7 @@ def _extract_from_slsa_v02(payload: InTotoV01Payload) -> tuple[str, str]:
     repo = _clean_spdx(uri)
 
     digest_set = json_extract(predicate, ["invocation", "configSource", "digest"], dict)
-    commit = _extract_commit_from_digest_set(digest_set, SLSA_V02_DIGEST_SET_ALGORITHMS)
+    commit = _extract_commit_from_digest_set(digest_set, SLSA_V02_DIGEST_SET_GIT_ALGORITHMS)
 
     if not commit:
         raise ProvenanceExtractionException("Failed to extract commit hash from provenance.")
@@ -153,7 +153,7 @@ def _extract_from_slsa_v1(payload: InTotoV1Payload) -> tuple[str, str]:
         if url != repo:
             continue
         digest_set = json_extract(dep, ["digest"], dict)
-        commit = _extract_commit_from_digest_set(digest_set, SLSA_V1_DIGEST_SET_ALGORITHMS)
+        commit = _extract_commit_from_digest_set(digest_set, SLSA_V1_DIGEST_SET_GIT_ALGORITHMS)
 
     if not commit:
         raise ProvenanceExtractionException("Failed to extract commit hash from provenance.")
--- a/src/macaron/repo_finder/provenance_extractor.py
+++ b/src/macaron/repo_finder/provenance_extractor.py
@@ -6,7 +6,6 @@ import logging
 from typing import TypeVar
 
 from macaron.errors import MacaronError
-from macaron.slsa_analyzer.provenance import intoto
 from macaron.slsa_analyzer.provenance.intoto import InTotoPayload, InTotoV1Payload, InTotoV01Payload
 from macaron.util import JsonType
 
@@ -15,6 +14,11 @@ logger: logging.Logger = logging.getLogger(__name__)
 
 class ProvenanceExtractionException(MacaronError):
     """When there is an error while extracting from provenance."""
+
+
+SLSA_V01_DIGEST_SET_ALGORITHMS = ["sha1"]
+SLSA_V02_DIGEST_SET_ALGORITHMS = ["sha1"]
+SLSA_V1_DIGEST_SET_ALGORITHMS = ["sha1", "gitCommit"]
 
 
 def extract_repo_and_commit_from_provenance(payload: InTotoPayload) -> tuple[str, str]:
@@ -86,7 +90,7 @@ def _extract_from_slsa_v01(payload: InTotoV01Payload) -> tuple[str, str]:
     repo = _clean_spdx(uri)
 
     digest_set = json_extract(material, ["digest"], dict)
-    commit = _extract_commit_from_digest_set(digest_set, intoto.v01.VALID_ALGORITHMS)
+    commit = _extract_commit_from_digest_set(digest_set, SLSA_V01_DIGEST_SET_ALGORITHMS)
 
     if not commit:
         raise ProvenanceExtractionException("Failed to extract commit hash from provenance.")
@@ -108,7 +112,7 @@ def _extract_from_slsa_v02(payload: InTotoV01Payload) -> tuple[str, str]:
     repo = _clean_spdx(uri)
 
     digest_set = json_extract(predicate, ["invocation", "configSource", "digest"], dict)
-    commit = _extract_commit_from_digest_set(digest_set, intoto.v01.VALID_ALGORITHMS)
+    commit = _extract_commit_from_digest_set(digest_set, SLSA_V02_DIGEST_SET_ALGORITHMS)
 
     if not commit:
         raise ProvenanceExtractionException("Failed to extract commit hash from provenance.")
@@ -149,7 +153,7 @@ def _extract_from_slsa_v1(payload: InTotoV1Payload) -> tuple[str, str]:
         if url != repo:
             continue
         digest_set = json_extract(dep, ["digest"], dict)
-        commit = _extract_commit_from_digest_set(digest_set, intoto.v1.VALID_ALGORITHMS)
+        commit = _extract_commit_from_digest_set(digest_set, SLSA_V1_DIGEST_SET_ALGORITHMS)
 
     if not commit:
         raise ProvenanceExtractionException("Failed to extract commit hash from provenance.")
--- a/src/macaron/repo_finder/provenance_extractor.py
+++ b/src/macaron/repo_finder/provenance_extractor.py
@@ -3,7 +3,7 @@
 
 """This module contains methods for extracting repository and commit metadata from provenance files."""
 import logging
-from typing import overload
+from typing import TypeVar
 
 from macaron.slsa_analyzer.provenance import intoto
 from macaron.slsa_analyzer.provenance.intoto import InTotoPayload, InTotoV1Payload, InTotoV01Payload
@@ -238,27 +238,10 @@ def _extract_from_witness_provenance(payload: InTotoV01Payload) -> tuple[str, st
     return repo, commit
 
 
-@overload
-def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[int]) -> int | None:
-    ...
+T = TypeVar("T", bound=JsonType)
 
 
-@overload
-def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[list]) -> list | None:
-    ...
-
-
-@overload
-def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[dict]) -> dict | None:
-    ...
-
-
-@overload
-def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[str]) -> str | None:
-    ...
-
-
-def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[JsonType]) -> JsonType:
+def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[T]) -> T | None:
     """Return the value found by following the list of depth-sequential keys inside the passed dictionary.
 
     The value's type is validated against the passed type.
--- a/src/macaron/repo_finder/provenance_extractor.py
+++ b/src/macaron/repo_finder/provenance_extractor.py
@@ -3,17 +3,13 @@
 
 """This module contains methods for extracting repository and commit metadata from provenance files."""
 import logging
-from typing import TypeVar
 
-from macaron.errors import MacaronError
+from macaron.errors import JsonError, ProvenanceError
+from macaron.json_tools import json_extract
 from macaron.slsa_analyzer.provenance.intoto import InTotoPayload, InTotoV1Payload, InTotoV01Payload
 from macaron.util import JsonType
 
 logger: logging.Logger = logging.getLogger(__name__)
-
-
-class ProvenanceExtractionException(MacaronError):
-    """When there is an error while extracting from provenance."""
 
 
 SLSA_V01_DIGEST_SET_GIT_ALGORITHMS = ["sha1"]
@@ -36,7 +32,7 @@ def extract_repo_and_commit_from_provenance(payload: InTotoPayload) -> tuple[str
 
     Raises
     ------
-    ProvenanceExtractionException
+    ProvenanceError
         If the extraction process fails for any reason.
     """
     repo = ""
@@ -53,9 +49,9 @@ def extract_repo_and_commit_from_provenance(payload: InTotoPayload) -> tuple[str
                 repo, commit = _extract_from_slsa_v01(payload)
             if predicate_type == "https://witness.testifysec.com/attestation-collection/v0.1":
                 repo, commit = _extract_from_witness_provenance(payload)
-    except JsonExtractionException as error:
+    except JsonError as error:
         logger.debug(error)
-        raise ProvenanceExtractionException("JSON exception while extracting from provenance.") from error
+        raise ProvenanceError("JSON exception while extracting from provenance.") from error
 
     if not repo or not commit:
         msg = (
@@ -63,7 +59,7 @@ def extract_repo_and_commit_from_provenance(payload: InTotoPayload) -> tuple[str
             f"predicate_type {predicate_type}, in-toto {str(type(payload))}."
         )
         logger.debug(msg)
-        raise ProvenanceExtractionException(msg)
+        raise ProvenanceError(msg)
 
     logger.debug("Extracted repo and commit from provenance: %s, %s", repo, commit)
     return repo, commit
@@ -73,17 +69,17 @@ def _extract_from_slsa_v01(payload: InTotoV01Payload) -> tuple[str, str]:
     """Extract the repository and commit metadata from the slsa v01 provenance payload."""
     predicate: dict[str, JsonType] | None = payload.statement.get("predicate")
     if not predicate:
-        raise ProvenanceExtractionException("No predicate in payload statement.")
+        raise ProvenanceError("No predicate in payload statement.")
 
     # The repository URL and commit are stored inside an entry in the list of predicate -> materials.
     # In predicate -> recipe -> definedInMaterial we find the list index that points to the correct entry.
     list_index = json_extract(predicate, ["recipe", "definedInMaterial"], int)
     material_list = json_extract(predicate, ["materials"], list)
     if list_index >= len(material_list):
-        raise ProvenanceExtractionException("Material list index outside of material list bounds.")
+        raise ProvenanceError("Material list index outside of material list bounds.")
     material = material_list[list_index]
     if not material or not isinstance(material, dict):
-        raise ProvenanceExtractionException("Indexed material list entry is invalid.")
+        raise ProvenanceError("Indexed material list entry is invalid.")
 
     uri = json_extract(material, ["uri"], str)
 
@@ -93,7 +89,7 @@ def _extract_from_slsa_v01(payload: InTotoV01Payload) -> tuple[str, str]:
     commit = _extract_commit_from_digest_set(digest_set, SLSA_V01_DIGEST_SET_GIT_ALGORITHMS)
 
     if not commit:
-        raise ProvenanceExtractionException("Failed to extract commit hash from provenance.")
+        raise ProvenanceError("Failed to extract commit hash from provenance.")
 
     return repo, commit
 
@@ -102,20 +98,20 @@ def _extract_from_slsa_v02(payload: InTotoV01Payload) -> tuple[str, str]:
     """Extract the repository and commit metadata from the slsa v02 provenance payload."""
     predicate: dict[str, JsonType] | None = payload.statement.get("predicate")
     if not predicate:
-        raise ProvenanceExtractionException("No predicate in payload statement.")
+        raise ProvenanceError("No predicate in payload statement.")
 
     # The repository URL and commit are stored within the predicate -> invocation -> configSource object.
     # See https://slsa.dev/spec/v0.2/provenance
     uri = json_extract(predicate, ["invocation", "configSource", "uri"], str)
     if not uri:
-        raise ProvenanceExtractionException("Failed to extract repository URL from provenance.")
+        raise ProvenanceError("Failed to extract repository URL from provenance.")
     repo = _clean_spdx(uri)
 
     digest_set = json_extract(predicate, ["invocation", "configSource", "digest"], dict)
     commit = _extract_commit_from_digest_set(digest_set, SLSA_V02_DIGEST_SET_GIT_ALGORITHMS)
 
     if not commit:
-        raise ProvenanceExtractionException("Failed to extract commit hash from provenance.")
+        raise ProvenanceError("Failed to extract commit hash from provenance.")
 
     return repo, commit
 
@@ -124,7 +120,7 @@ def _extract_from_slsa_v1(payload: InTotoV1Payload) -> tuple[str, str]:
     """Extract the repository and commit metadata from the slsa v1 provenance payload."""
     predicate: dict[str, JsonType] | None = payload.statement.get("predicate")
     if not predicate:
-        raise ProvenanceExtractionException("No predicate in payload statement.")
+        raise ProvenanceError("No predicate in payload statement.")
 
     build_def = json_extract(predicate, ["buildDefinition"], dict)
     build_type = json_extract(build_def, ["buildType"], str)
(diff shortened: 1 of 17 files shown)


src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -0,0 +1,282 @@
...
+def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[dict]) -> dict | None:
+    ...
+
+
+@overload
+def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[str]) -> str | None:
+    ...
+
+
+def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[JsonType]) -> JsonType:
```
nathanwn, line 261: Nice. I didn't know about using `@overload` like this. Thanks, Ben.

For this case, what I usually gravitate towards is using `TypeVar` instead. In this case, the type property we need to maintain is that the return type should be similar to the value of the parameter `type_`, which can be achieved with:

```suggestion
T = TypeVar("T", bound=JsonType)


def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[T]) -> T | None:
```
Here, the upper-bound of the type variable `T` is `JsonType`, meaning that `type_` can be any type that is a subtype of `T`, including all of the types we encounter here: `int`, `list`, `dict`, and `str` (not sure if I miss any).

I would probably prefer this since it is more concise, but up to your preference.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -292,8 +294,9 @@ def run_single(
         """
         repo_id = config.get_value("id")
         component = None
+        provenance_finder = ProvenanceFinder()
         try:
-            component = self.add_component(config, analysis, existing_records)
+            component = self.add_component(config, analysis, provenance_finder, existing_records, provenance_payload)
```
tromai, line 299: If the instance of `ProvenanceFinder` is created here and pass to `self.add_component` directly, there is no case where `provenance_payload` is None. I wonder why is the `provenance_payload` parameter in `self.add_component` optional here optional (or may be I haven't seen the case where `provenance_finder` is None yet)? 
```python
def add_component(
        self,
        config: Configuration,
        analysis: Analysis,
        provenance_finder: ProvenanceFinder,
        existing_records: dict[str, Record] | None = None,
        provenance_payload: InTotoPayload | None = None,
    ) -> Component:
```

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -0,0 +1,265 @@
...
+    Returns
+    -------
+    tuple[str, str]
+        The repository URL and commit hash if found, a pair of empty strings otherwise.
+    """
+    repo = ""
+    commit = ""
+    predicate_type = payload.statement.get("predicateType")
+    if isinstance(payload, InTotoV1Payload):
+        if isinstance(payload, InTotoV1Payload):
```
tromai, line 32: I think these two if conditions are identical :thinking:

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -0,0 +1,265 @@
...
+        if predicate_type == "https://slsa.dev/provenance/v0.1":
+            repo, commit = _extract_from_slsa_v01(payload)
+        if predicate_type == "https://witness.testifysec.com/attestation-collection/v0.1":
+            repo, commit = _extract_from_witness_provenance(payload)
+
+    if not repo or not commit:
+        logger.debug(
+            "Extraction from provenance not supported for versions: predicate_type %s, in-toto %s.",
+            predicate_type,
+            payload.__class__,
```
tromai, line 47: nit: I would prefer `str(type(payload))` to get the type of a class instance.

src/macaron/repo_finder/provenance_finder.py
```diff
@@ -0,0 +1,232 @@
...
+        except ValueError as error:
+            logger.debug("Failed to parse default value as int: %s", error)
+            return None
+
+        jfrog_registry = JFrogMavenRegistry(
+            defaults.get(jfrog_section, "hostname"),
+            defaults.get(jfrog_section, "repo"),
+            request_timeout,
+            download_timeout,
+        )
```
nathanwn, line 161: You shouldn't need to create a new `JFrogMavenRegistry` instance and load configurations here.
Instead, I think the `JFrogMavenRegistry` instance created [here](https://github.com/oracle/macaron/blob/8ba2e388f5c2410252b6562183f83d682b71f205/src/macaron/slsa_analyzer/package_registry/__init__.py#L11) should be used instead. This is because this instance has already had its configurations loaded [at the beginning when we run `macaron analyze`](https://github.com/oracle/macaron/blob/8ba2e388f5c2410252b6562183f83d682b71f205/src/macaron/__main__.py#L220C1-L221C53).

I think what should be done here is to get the `JFrogMavenRegistry` instance from the `PACKAGE_REGISTRIES` list, check if it is enabled through the `.enabled` field, and proceed accordingly.

src/macaron/repo_finder/provenance_finder.py
```diff
@@ -0,0 +1,232 @@
...
+            "provenance_extensions",
+            fallback=["intoto.jsonl"],
+        )
+
+        provenance_assets = jfrog_registry.fetch_assets(
+            group_id=purl.namespace if purl.namespace else "",
+            artifact_id=purl.name,
+            version=purl.version if purl.version else "",
+            extensions=set(provenance_extensions),
+        )
```
nathanwn, line 174: I don't think the `.fetch_assets` method would accept empty strings for `group_id` and `version`. It is probably better to check if they are empty and stop proceeding before calling the method.

I designed the API of the `JFrogMavenRegistry` class to be: if an argument is required, it is assumed to be a valid non-empty string. Otherwise, the argument should be optional, and takes `str | None`.

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -0,0 +1,265 @@
...
+            commit = _json_extract(dep, ["digest", "gitCommit"], str)
+
+    if not commit:
+        logger.debug("Failed to extract commit hash from provenance.")
+        return "", ""
+
+    return repo, commit
+
+
+def _extract_commit_from_digest(digest: dict[str, JsonType]) -> str | None:
```
tromai, line 171:
```suggestion
def _extract_commit_from_digest_set(digest: dict[str, JsonType]) -> str | None:
```

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -0,0 +1,265 @@
...
+    # TODO decide on a preference for which algorithm to accept.
+    if len(digest.keys()) > 1:
+        logger.debug("DigestSet contains multiple algorithms: %s", digest.keys())
+
+    for key in digest:
+        if key in intoto.v1.VALID_ALGORITHMS:
+            value = digest.get(key)
+            if isinstance(value, str):
+                return value
+    return None
```
tromai, line 185: As I can see, we are only getting the list of valid algorithms from `intoto.v1` specifically while this function is used in handling both `InToToV1Payload` and `InToToV01Payload` .
In my opinion, to be explicit about it, we can modify this function as follow:
```
def _extract_commit_from_digest_set(digest_set: dict[str, JsonType], valid_algorithms: list[str]) -> str | None:
    ...
```
In additions, we can define the list of valid algorithms for `intoto.v01`. Together they can be used as follows:
```
# For InToToV1Payload
commit = _extract_commit_from_digest_set(digest_set, intoto.v1.VALID_ALGORITHMS)

# For InToToV01Payload
commit = _extract_commit_from_digest_set(digest_set, intoto.v01.VALID_ALGORITHMS)
```
Even though the value in  `intoto.v01.VALID_ALGORITHMS` can be identical to what we currently have in `intoto.v1.VALID_ALGORITHMS`. Please let me know what you think.

src/macaron/repo_finder/provenance_finder.py
```diff
@@ -0,0 +1,232 @@
...
+from macaron.slsa_analyzer.provenance.witness import is_witness_provenance_payload, load_witness_verifier_config
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class ProvenanceFinder:
+    """This class is used to find and retrieve provenance files from supported registries."""
+
+    def __init__(self) -> None:
+        self.last_provenance_payload: InTotoPayload | None = None
```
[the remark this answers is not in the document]
benmss: Yes, but it's not just that we don't have access to the `AnalyzeContext` there, it's that we cannot. At least without larger refactoring as the returned object of the `add_component` call that `Analyzer_to_analysis_target` is a child of, is required when the `AnalyzeContext` object is instantiated.

src/macaron/slsa_analyzer/analyzer.py
benmss: I think you are right. I'll change it.

src/macaron/repo_finder/provenance_extractor.py
benmss: You are right, and I meant to change it at some point. I like your suggestion so I'll go with that.

src/macaron/repo_finder/provenance_finder.py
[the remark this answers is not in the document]
tromai: Yep I think I have the same understanding of the problem. I will think more about it. Thanks for the clarification Ben.

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -0,0 +1,265 @@
...
+        if predicate_type == "https://witness.testifysec.com/attestation-collection/v0.1":
+            repo, commit = _extract_from_witness_provenance(payload)
+
+    if not repo or not commit:
+        logger.debug(
+            "Extraction from provenance not supported for versions: predicate_type %s, in-toto %s.",
+            predicate_type,
+            payload.__class__,
+        )
+        return "", ""
```
tromai, line 49: I wonder if there is any reason why we want to return an empty string to indicate errors instead of None or raising an exception :thinking: ?

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -0,0 +1,265 @@
...
+
+    if list_index >= len(material_list):
+        return "", ""
+    material = material_list[list_index]
+    if not material or not isinstance(material, dict):
+        return "", ""
+
+    uri = material.get("uri")
+    if not uri:
+        logger.debug("Could not extract repository URL.")
```
tromai, line 79: If `uri` is not available, I think we should return error instead of proceeding with `repo = _clean_spdx(uri)`

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -0,0 +1,265 @@
...
+        if key not in target:
+            logger.debug("Key not found in JSON: %s", key)
+            return None
+        next_target = target[key]
+        if index == len(keys) - 1:
+            if isinstance(next_target, type_):
+                return next_target
+        else:
+            if not isinstance(next_target, dict):
+                logger.debug("Expected dict found: %s", next_target.__class__)
```
tromai, line 260:
```suggestion
                logger.debug("Expected dict, found: %s", next_target.__class__)
```

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -0,0 +1,265 @@
...
+
+    # Extract the commit hash.
+    commit = None
+    deps = _json_extract(build_def, ["resolvedDependencies"], list)
+    if not deps:
+        return "", ""
+    for dep in deps:
+        if not isinstance(dep, dict):
+            continue
+        uri = dep["uri"]
```
tromai, line 152: I think we need to handle the case where `uri` doesn't exist in the `dep` dictionary.

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -0,0 +1,265 @@
...
+        url = _clean_spdx(uri)
+        if url != repo:
+            continue
+        if build_type == "https://slsa-framework.github.io/gcb-buildtypes/triggered-build/v1":
+            commit_dict = _json_extract(dep, ["digest"], dict)
+            if not commit_dict:
+                continue
+            commit = _extract_commit_from_digest(commit_dict)
+        if build_type == "https://slsa-framework.github.io/github-actions-buildtypes/workflow/v1":
+            commit = _json_extract(dep, ["digest", "gitCommit"], str)
```
tromai, line 162: I believe that each entry in `resolvedDependencies` is of type `ResourceDescriptor` (according to the doc [here](https://slsa.dev/spec/v1.0/provenance#builddefinition)). Therefore the varible `dep` is a dictionary of type `ResrouceDescriptor` and the value of the key `digest` in `dep` is of type `DigestSet` (see [ResourceDescriptor](https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md)). 
I think we need to handle them using the same logic for extract commit from digest set.

src/macaron/slsa_analyzer/provenance/intoto/v1/__init__.py
```diff
@@ -11,9 +11,10 @@
...
 from macaron.util import JsonType
 
-# The full list of cryptographic algorithms supported in SLSA v1 provenance. These are used as keys within the digest
-#  set of the resource descriptors within the subject.
-# See: https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md
+# The full list of cryptographic algorithms supported in in-toto v0.1 and v1 provenance.
+# These are used as keys within the digest set of the resource descriptors within the subject.
+# For v1 see: https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md
+# For v0.1 see: https://github.com/in-toto/attestation/blob/main/spec/v0.1.0/field_types.md#DigestSet
 VALID_ALGORITHMS = [
```
tromai, line 18: This might not be in scope of this PR.
I have a look at the definition for DigestSet in the v1 specification ([here](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md)) and I see three types:
1. [The standard crypto graphic hash algorithms](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#sha256-sha224-sha384-sha512-sha512_224-sha512_256-sha3_224-sha3_256-sha3_384-sha3_512-shake128-shake256-blake2b-blake2s-ripemd160-sm3-gost-sha1-md5) which is the ones that you put in this list (in https://github.com/oracle/macaron/issues/639)
2. [dirHash](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#dirhash)
3. [git related hash](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#gitcommit-gittree-gitblob-gittag). This section contains the `gitCommit` value that you used for getting the commit hash of the source for SLSA provenance v1.0
I wonder if we want to support the second and third type as well :thinking:

src/macaron/repo_finder/provenance_extractor.py
benmss: Handling `tuple[str, str] | None` can be a bit unpleasant when expanding returned values. However, I will switch to using exceptions.

src/macaron/slsa_analyzer/provenance/intoto/v1/__init__.py
benmss: I think adding support for type three is in scope.

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -41,12 +50,12 @@ def extract_repo_and_commit_from_provenance(payload: InTotoPayload) -> tuple[str
...
-        logger.debug(
-            "Extraction from provenance not supported for versions: predicate_type %s, in-toto %s.",
-            predicate_type,
-            payload.__class__,
+        msg = (
+            f"Extraction from provenance not supported for versions: "
+            f"predicate_type {predicate_type}, in-toto {str(type(payload))}."
         )
-        return "", ""
+        logger.error(msg)
```
tromai, line 57: I think Macaron don't terminate on error if we failed to perform the extraction here (or in another terms, it's not considered an error to stop the analysis). Therefore, I think we should log this DEBUG instead.

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -232,34 +181,56 @@ def _extract_from_witness_provenance(payload: InTotoV01Payload) -> tuple[str, st
...
+    """
+    url, _, _ = uri.lstrip("git+").rpartition("@")
+    return url
+
+
 T = TypeVar("T", bound=JsonType)
 
 
-def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[T]) -> T | None:
+def _json_extract(entry: dict[str, JsonType], keys: list[str], type_: type[T]) -> T:
```
tromai, line 218: I am sorry for saying this after you have changed this function to raise the ProvenanceExtraction instead :sweat_smile: . 
For me I was find with this private method returning None for error. It's supposed to be a generic function that as long as `entry` is a JSON-liked object. 
I can see some issues with raising `ProvenanceExtractionException` here:
- Because this is supposed to be a generic function, raising a context specific exception `ProvenanceExtractionException` is not a good idea. 
- The exceptions raised here are bubbled up to the caller function (for example at this [place](https://github.com/oracle/macaron/blob/56e680e36bc87170109972ce7f114b71abd69b33/src/macaron/repo_finder/provenance_extractor.py#L72)), which to me is not very explicit.
I do have some suggestions:
- If we want to keep the exception here, I think we can rename it to something more generic (let's called it `JsonExtractionException` for example). But it also means that we have to explicitly catch this exception whenever we use it.
- We can make this function return None on error and check for None on usage without having to catch any errors.
I am fine with both suggestions so please let me know what you think.

src/macaron/repo_finder/repo_finder_java.py
```diff
@@ -24,6 +24,7 @@ class JavaRepoFinder(BaseRepoFinder):
     def __init__(self) -> None:
         """Initialise the Java repository finder instance."""
         self.pom_element: Element | None = None
+        self.repositories: list[str] | None = None
```
tromai, line 27: I wonder where this attribute `repositories` is being used. I can't seem to find it :sweat_smile:

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -199,6 +201,44 @@ def download_attestation_payload(self, url: str, download_path: str) -> bool:
...
+        url = f"{url}/{name}/latest"
+
+        response = send_get_http_raw(url, timeout=self.request_timeout)
+
+        if not response or not response.text:
+            logger.debug("No valid response from NPM server for latest version.")
+            return None
+
+        json_data = json.loads(response.text)
+        version = json_data["version"]
```
tromai, line 234: I think we need to handle the case where the key `version` doesn't exist in the JSON data.

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2024, Oracle and/or its affiliates. All rights reserved.
```
tromai, line 1: I wonder why we need to update the copyright year of this file while its content is not changed :thinking:

tests/repo_finder/test_provenance_extractor.py
```diff
@@ -0,0 +1,231 @@
+# Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved.
```
tromai, line 1: I think it would be great if we can add some unit test cases here for invalid cases:
- There is no repository url.
- There is a repository url but the commit cannot be found.
- Both repository url and commit cannot be found.

Optional, for the `_json_extract` function, we could make use of unit test cases for error cases (where currently exceptions are raised).

tromai (changes requested):
I have finished my first round of review. Thanks for the changes.

src/macaron/repo_finder/provenance_extractor.py
benmss: I agree, although utilising exceptions for the JSON extraction allows for cleaner code in the provenance extractor functions, in my opinion. My compromise is using suggestion 1 above, but the catching of the JSON extraction exception occurs in the public function of the file `extract_repo_and_commit_from_provenance`, which itself still only raises `ProvenanceExtractionException`.

tromai: Yep, that sounds good to me. If I understand your comment correctly, you were mentioning about catching and re-raising the exceptions from this `_json_extract` method within the public function:
```python
try:
    list_index = _json_extract(predicate, ["recipe", "definedInMaterial"], int) 
except JsonExtractException as error:
    raise ProvenanceExtractionException(...) from error
```

tests/repo_finder/test_provenance_extractor.py
```diff
@@ -0,0 +1,457 @@
...
+
+    # Set repository to an empty string.
+    _json_modify(payload, ["predicate", "buildDefinition", "externalParameters", "workflow", "repository"], "")
+    with pytest.raises(ProvenanceExtractionException):
+        _perform_provenance_comparison(payload, "", "")
+
+    # Remove repository key.
+    _json_modify(payload, ["predicate", "buildDefinition", "externalParameters", "workflow", "repository"], None)
+    with pytest.raises(ProvenanceExtractionException):
+        _perform_provenance_comparison(payload, "", "")
```
nathanwn, line 239: Please use `@pytest.mark.parametrize` for tests like these where you want to test multiple different cases of input with the same test logic. It will make the test easier to manage since it separates test data from test logic, as well as allowing for the test logic to be defined once instead of repeatedly. I am also not a big fan of mutating test data like this to test different cases, and I prefer to create the input once each time with `@pytest.mark.parametrize`. For example, each case in this test can take a pair of `(keys, new_value)` as input, modify the payload accordingly only once, and do the assertion. I think that is enough for the logic of a unit test function like this. 

In addition, if the test logic deviates between cases, we should separate them into separate test functions. This test, for example, can also be split into two test functions: one to test for valid provenance cases, and the other for invalid cases.

tests/repo_finder/test_provenance_extractor.py
```diff
@@ -0,0 +1,457 @@
...
+@pytest.fixture(name="target_commit")
+def target_commit_() -> str:
+    """Return the target commit hash."""
+    return "51aa22a42ec1bffa71518041a6a6d42d40bf50f0"
+
+
+def test_slsa_v1_gcb_1(slsa_v1_gcb_1_provenance: str, target_repository: str, target_commit: str) -> None:
+    """Test SLSA v1 provenance with build type gcb and sourceToBuild."""
+    payload = json.loads(slsa_v1_gcb_1_provenance)
+    assert isinstance(payload, dict)
```
nathanwn, line 175: I think this deserialization should be done within the fixture function, not within the test function.

src/macaron/repo_finder/provenance_finder.py
```diff
@@ -0,0 +1,225 @@
...
+                continue
+
+            provenances.append(provenance_payload)
+
+        if not provenances:
+            logger.debug("No payloads found in provenance files.")
+            return None
+
+        # TODO decide what to do when multiple provenance payloads are present.
+        provenance = provenances[0]
```
nathanwn, line 223: Although the API of `JfrogMavenRegistry::fetch_assets` returns multiple assets, it does not necessarily mean there will be multiple provenances. For our internal use case, I think you can safely assume that: there can only be 1 provenance per a triple of GAV coordinates, and the provenance discovery you do here has a triple of GAV coordinates as input. IMO you can state the assumption here in the comment and remove the TODO.

src/macaron/slsa_analyzer/provenance/loader.py
tromai: I think this has not been addressed :sweat_smile:

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -174,8 +174,8 @@ def find_provenance_assets_on_package_registries(
                     # adversary provides a super large malicious file.
 
                     # TODO: refactor the size checking in this check and the `provenance_l3_check`
-                    # so that we have consistent behavior when checking provenance size.
-                    # The schema of the ini config also needs changing.
+                    #  so that we have consistent behavior when checking provenance size.
+                    #  The schema of the ini config also needs changing.
```
nathanwn, line 178: Nit: Please revert this change. I know it is not that important, but since it messes with `git blame` I figure we should not keep it.

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -199,6 +201,47 @@ def download_attestation_payload(self, url: str, download_path: str) -> bool:
 
         return False
 
+    def get_latest_version(self, namespace: str, name: str) -> str | None:
+        """Try to retrieve the latest version of a package from the registry.
+
+        Parameters
+        ----------
+        namespace: str
+            The optional namespace of the package.
```
tromai, line 210: If this `namespace` value is optional, should we make it an optional parameter? Or is there a reason why the type of `namespace` must be `str` instead of `str | None` :thinking:

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -295,11 +295,44 @@ def run_single(
...
+        if not provenance_payload and parsed_purl and not config.get_value("path"):
+            # Try to find the provenance file for the parsed PURL.
+            provenance_payload = ProvenanceFinder().find_provenance(parsed_purl)
+
+        # Create the analysis target.
+        msg = ""
+        available_domains = [git_service.hostname for git_service in GIT_SERVICES if git_service.hostname]
+        try:
+            analysis_target = Analyzer.to_analysis_target(config, available_domains, parsed_purl, provenance_payload)
+        except InvalidPURLError as error:
```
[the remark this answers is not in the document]
nathanwn: I think after the change in 4b39ae8148c50b9798fa1640fa86b7298c98a1ab, the `to_analysis_target` method no longer raises the `InvalidPURLError` because it no longer does PURL parsing. Therefore, this `try-except` is probably not necessary.
The docstring of the `to_analysis_target` method also needs updating.

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -199,6 +201,47 @@ def download_attestation_payload(self, url: str, download_path: str) -> bool:
...
+
+        if not response or not response.text:
+            logger.debug("No valid response from NPM server for latest version.")
+            return None
+
+        json_data = json.loads(response.text)
+        try:
+            version = json_data["version"]
+        except KeyError:
+            version = ""
```
nathanwn, line 237: Using `.get` is much nicer than trying to access with the `[]` operator and catching `KeyError`.
```suggestion
        version = json_data.get("version")
```

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -199,6 +201,47 @@ def download_attestation_payload(self, url: str, download_path: str) -> bool:
...
+        if not version:
+            logger.debug("No version found in response from NPM server.")
+            return None
+
+        logger.debug("Found version for NPM artifact: %s", version)
+        return version if isinstance(version, str) else str(version)
```
nathanwn, line 243: When is `version` not a `str`, and is it even a valid result in that case?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -282,18 +287,52 @@ def run_single(
...
+            logger.debug("Invalid input PURL: %s", error)
+            msg = "Invalid input PURL."
+            analysis_target = None
```
nathanwn, line 323: Why don't we just return a `Record` in the `except` branch, instead of reassigning `analysis_target` to `None` just to check for it later? This way we don't need the additional `msg` variable, and the sequential `if` statement is simpler.

src/macaron/slsa_analyzer/provenance/loader.py
benmss: I was waiting to do a force push update to fix this issue. Otherwise I think our checks will continue to "fix" the year.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -282,18 +287,52 @@ def run_single(
...
+
+        if not analysis_target or (not analysis_target.parsed_purl and not analysis_target.repo_path):
+            return Record(
+                record_id=repo_id,
+                description=msg or "Cannot determine the analysis as PURL and/or repository path is not provided.",
+                pre_config=config,
+                status=SCMStatus.ANALYSIS_FAILED,
+            )
```
nathanwn, line 331: I would suggest moving this `not analysis_target.parsed_purl and not analysis_target.repo_path` check into the `Analyzer.to_analysis_target` method and raise an error. My argument is this looks like another piece of case handling and most of the input case handling is already being done in the `Analyzer.to_analysis_target` method.

This also looks like a state that the `analysis_target` object should not be in (both PURL and repo URL are empty), so there is no good reason for it to exist in the first place if this case happens. It reminds me of this item in the C++ Core Guidelines: [If a constructor cannot construct a valid object, throw an exception](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c42-if-a-constructor-cannot-construct-a-valid-object-throw-an-exception).

src/macaron/slsa_analyzer/provenance/loader.py
tromai: Oh I see. Thanks for letting me know. 
Hm that's interesting, I have never encountered this issue before :sweat_smile:

tests/repo_finder/test_provenance_extractor.py
```diff
@@ -455,3 +465,10 @@ def _json_modify(entry: dict[str, JsonType], keys: list[str], new_value: JsonTyp
             if not isinstance(next_target, dict):
                 raise JsonExtractionException(f"Cannot extract value from non-dict type: {str(type(next_target))}")
             target = next_target
+
+
+def _load_and_validate_josn(payload: str) -> dict[str, JsonType]:
```
tromai, line 470:
```suggestion
def _load_and_validate_json(payload: str) -> dict[str, JsonType]:
```

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -587,10 +651,27 @@ def to_analysis_target(config: Configuration, available_domains: list[str]) -> A
...
+
+                    if repo and digest:
+                        return Analyzer.AnalysisTarget(
+                            parsed_purl=parsed_purl,
+                            repo_path=repo,
+                            branch="",
+                            digest=digest,
+                        )
+
+                    # The commit was not found from provenance. Proceed with Repo and Commit Finder.
```
tromai, line 674: From what I see the statements below this comment only use the repo finder to find the repository url while the commit finder is not used. 
I believe the reason why we cannot use the commit finder at this point is because we need to obtain the `git_obj` from the cloned repository first.

tromai: Do you think it would be better if we change this comment to 
`The commit was not found from provenance. Proceed with Repo Finder.` instead?

benmss: The comment is more referring to the larger process that will take place. Yes, the commit finder cannot be used at this point, but it will be called later on given the current parameters.

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -0,0 +1,263 @@
...
+        next_target = target[key]
+        if index == len(keys) - 1:
+            if next_target and isinstance(next_target, type_):
+                return next_target
+        else:
+            if not isinstance(next_target, dict):
+                raise JsonExtractionException(f"Cannot extract value from non-dict type: {str(type(next_target))}")
+            target = next_target
+
+    raise JsonExtractionException(f"Failed to find '{' > '.join(keys)}' as type '{type_}' in JSON dictionary.")
```
nathanwn, line 263: I have a couple of questions as well as suggestions for this function.

### About the final `raise`

I believe the final `raise` only happens in two cases:
* When the extracted value is either falsy or not of the given `type_`, i.e. when the branch `if next_target and isinstance(next_target, type_)` is not taken. This is a type issue.
* When the `keys` list is empty. This seems to me more of a usage issue. Should this function assume that it always takes a non-empty list? If this is the case, then we should clarify it in the docstring at least.

I think the message "Failed to find ... as type ... in JSON dictionary" can be a bit confusing since it does not exactly point out what is the actual problem. The error message could be a bit more specific.

### About the extracted value must be truthy

I wonder why this function requires this. I can see legitimate cases where the value is falsy. An empty list or 0 can be legitimate values.

### About the error message

The error messages in the first two `raise`s give some info about what went wrong, but the info is quite limited. With a hierarchical structure like this, I think it is nice to also include the path to the problematic value in the error messages.

### About the logical flow of this function

I wonder if we could make the logical flow of this function a bit simpler. My intuition initially tells me that the use of the `if index == len(keys) - 1` branching and the `next_target` variable looks a bit clunky.

If I understand it right, there are two phases for this function:

* Traversing down the nested dictionary following the given list of keys.
* Validating the type of the value.

I think we can definitely separate the logic into these two phases which simplifies the implementation.

With all of these points, I would personally implement the function like so:

```suggestion
    target = entry
    target_path = lambda _keys: f".{'.'.join(_keys)}"

    for i, key in enumerate(keys):
        if not isinstance(target, dict):
            raise JsonExtractionException(f"Expect the value {target_path(keys[:i])} to be a dict.")
        if key not in target:
            raise JsonExtractionException(f"JSON key '{key}' not found in {target_path(keys[:i])}.")
        target = target[key]

    if isinstance(target, type_):
        return target
    else:
        raise JsonExtractionException(f"Expect the value {target_path(keys)} to be of type '{type_}'.")
```

I think this is nicer because we got rid of both the `if i == len - 1` and the `next_target` variable. We also introduce a nice invariant for `target` across all iterations that it must initially be a `dict`. I think this is nicer to reason about.

src/macaron/slsa_analyzer/package_registry/npm_registry.py
benmss: This was because of `mypy` but has been replaced by a type hint on the `version` instantiation line.

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -99,11 +103,16 @@ def _extract_from_slsa_v02(payload: InTotoV01Payload) -> tuple[str, str]:
     # The repository URL and commit are stored within the predicate -> invocation -> configSource object.
     # See https://slsa.dev/spec/v0.2/provenance
     uri = json_extract(predicate, ["invocation", "configSource", "uri"], str)
+    if not uri:
```
nathanwn, line 106: Why do we need to check for an empty string here? Have you encountered cases where there can be empty strings?

nathanwn: Oh. I think I understand why. Nevermind.

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -0,0 +1,271 @@
...
+
+    return repo, commit
+
+
+def _extract_commit_from_digest_set(digest_set: dict[str, JsonType], valid_algorithms: list[str]) -> str:
+    """Extract the commit from the passed DigestSet.
+
+    The DigestSet is an in-toto object that maps algorithm types to commit hashes (digests).
+    """
+    # TODO decide on a preference for which algorithm to accept.
```
nathanwn, line 207: Git hashes are always SHA1 as far as I know, but to be vigilant I did a bit of digging through the `slsa-github-generator` code. If you look [here](https://github.com/slsa-framework/slsa-github-generator/blob/b595e069f78187eaead9bac15429f87237be550a/slsa/buildtype.go#L222), it clearly shows that the `.predicate.invocation.configSource.digest` field always contains a `"sha1"` key, which should be where we get the commit hash.

nathanwn: Update: I found a case where the git hash is not provided through the "sha1" key: npm SLSA v1 provenances.
These provenances have the git hash in the "gitDigest" key ([npm provenance generator implementation](https://github.com/npm/provenance/blob/8a3ab1cbc231e72587cdbb5cd8113a592bfd401a/packages/generate-provenance/src/index.ts#L55C1-L55C43)).
I think it is probably best to assume it's either "sha1" or "gitDigest" unless we encounter any other exceptions (Have you encountered any?). I think this seems at least more precise than accepting a whole collection of hash algorithms.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -554,26 +586,58 @@ def to_analysis_target(config: Configuration, available_domains: list[str]) -> A
...
-            parsed_purl = config.get_value("purl")
-        else:
-            try:
-                # Note that PackageURL.from_string sanitizes the unsafe characters in the purl string,
-                # which is user-controllable, by calling urllib's `urlsplit` function.
-                parsed_purl = PackageURL.from_string(config.get_value("purl"))
-            except ValueError as error:
-                raise InvalidPURLError(f"Invalid input PURL: {config.get_value('purl')}") from error
+            return None
+        purl = config.get_value("purl")
```
nathanwn, line 591: Nit:
```diff
-        if config.get_value("purl") is None or config.get_value("purl") == "":
-            return None
-        purl = config.get_value("purl")
+        purl = config.get_value("purl")
+        if purl is None or purl == "":
+            return None
```

tests/repo_finder/test_provenance_extractor.py
```diff
@@ -0,0 +1,474 @@
...
+        ("https://in-toto.io/Statement/v1", "https://slsa.dev/provenance/v0.2"),
+        ("https://in-toto.io/Statement/v1", "https://slsa.dev/provenance/v0.1"),
+        ("https://in-toto.io/Statement/v1", "https://witness.testifysec.com/attestation-collection/v0.1"),
+    ],
+)
+def test_invalid_type_payloads(type_: str, predicate_type: str) -> None:
+    """Test payloads with invalid type combinations."""
+    payload_text = '{ "_type": ' + f'"{type_}",' + ' "predicateType": ' + f'"{predicate_type}",'
+    payload_text = f"{payload_text}" + '"subject": [], "predicate": {} }'
+    payload = json.loads(payload_text)
```
nathanwn, line 428: This string construction looks ugly. We can create a `dict` here.
```suggestion
    payload: dict[str, JsonType] = {
        "_type": type_,
        "predicateType": predicate_type,
        "subject": [],
        "predicate": {},
    }
```

tests/repo_finder/test_provenance_extractor.py
```diff
@@ -0,0 +1,474 @@
...
+            if new_value is None:
+                # Remove value.
+                del target[key]
+            else:
+                # Replace value
+                target[key] = new_value
+        else:
+            if not isinstance(next_target, dict):
+                raise JsonExtractionException(f"Cannot extract value from non-dict type: {str(type(next_target))}")
+            target = next_target
```
nathanwn, line 467: Nit: This may not really matter because this is test code anyway. However, I would have the same comment as the one I had with the `json_extract` function. I feel like the `for` loop is trying to do too many things which results in unnecessary branching within it. It can definitely be separated into a two-step process: first traversing the dict, then modifying the value.

Also, I personally think for test code, we do not need to be super strict about throwing and catching exceptions. If anything, I would personally go against trying to error-handling test code like it's production code. Errors in test code mean the test is wrongly set up, and not trying to swallow them or converting them to some custom error type is a good thing as it helps us catch our own mistakes.

tests/repo_finder/test_provenance_extractor.py
```diff
@@ -0,0 +1,474 @@
...
+def test_invalid_type_payloads(type_: str, predicate_type: str) -> None:
+    """Test payloads with invalid type combinations."""
+    payload_text = '{ "_type": ' + f'"{type_}",' + ' "predicateType": ' + f'"{predicate_type}",'
+    payload_text = f"{payload_text}" + '"subject": [], "predicate": {} }'
+    payload = json.loads(payload_text)
+    with pytest.raises(ProvenanceExtractionException):
+        _perform_provenance_comparison(payload, "", "")
+
+
+def _perform_provenance_comparison(payload: dict[str, JsonType], expected_repo: str, expected_commit: str) -> None:
```
nathanwn, line 433: `expected_repo` and `expected_commit` can be made optional parameters so that we don't have to pass in empty strings when no comparison is needed.

Also, this function is literally testing the `extract_repo_and_commit_from_provenance` function. So I think why not call it `_test_extract_repo_and_commit_from_provenance`. It is much more clear than `_perform_provenance_comparison`.

```suggestion
def _test_extract_repo_and_commit_from_provenance(
    payload: dict[str, JsonType],
    expected_repo: str | None = None,
    expected_commit: str | None = None,
) -> None:
```

nathanwn (changes requested):
I think I have gone through everything and will most likely not have any further comments. Thanks, Ben.

src/macaron/slsa_analyzer/provenance/witness/__init__.py
```diff
@@ -77,7 +77,6 @@ def is_witness_provenance_payload(
     bool
         ``True`` if the payload is a witness provenance payload, ``False`` otherwise.
     """
-    # TODO: add support for in-toto v1 provenances.
```
nathanwn, line 80: Why is this TODO removed?

src/macaron/repo_finder/provenance_extractor.py
benmss: I will restrict it to `sha1` and `gitCommit` (for `in-toto` v1) and throw a detailed error for other values if they ever show up.

src/macaron/slsa_analyzer/provenance/witness/__init__.py
benmss: Witness provenances only use `in-toto` v0.1 at the moment. Adding support for `v1` is therefore meaningless. I don't believe the spec for Witness `v1` has been released yet, but even if it has and it uses `in-toto` v1 (or will), this TODO is not required as the changes required for supporting that will clearly require changes here regardless.

nathanwn: The context of why this TODO is here in the first place was that `predicateType` for a Witness provenance is "configurable". The `predicate_types` set is from the user ini config, which was not my decision. Otherwise, I could just have hard-code the corresponding `predicateType` here and not have the TODO which was how I was doing it initially, because this function is only meant for v0.1 (and I would agree with your points here in that case), but it is not, and hence the TODO.

In addition, this file also includes similar TODO comments in a couple of places. If this comment is removed, then we can make the case to remove other TODOs as well, which I am unsure of.

I think the best thing to do is to leave the TODOs as is and concretely decide on what to do with these Witness provenances later. In general, these TODOs are less about whether v1 is currently supported or we should support v1, but more about the fact that `predicateType` is configurable.

src/macaron/slsa_analyzer/provenance/intoto/v01/__init__.py
```diff
@@ -10,11 +10,29 @@
 from macaron.slsa_analyzer.provenance.intoto.errors import ValidateInTotoPayloadError
 from macaron.util import JsonType
 
-# The list of cryptographic algorithms supported in in-toto v0.1 provenance.
+# The full list of cryptographic algorithms supported in in-toto v0.1 provenance.
 # These are used as keys within the digest set of the resource descriptors within the subject.
-# For the full v0.1 list see: https://github.com/in-toto/attestation/blob/main/spec/v0.1.0/field_types.md#DigestSet
+# For v0.1 see: https://github.com/in-toto/attestation/blob/main/spec/v0.1.0/field_types.md#DigestSet
 VALID_ALGORITHMS = [
+    "sha256",
```
tromai, line 17: Hmm if we have decided to only accept:
- `sha1` for intoto v0.1 DigestSet
- `sha1` and `gitDigest` for intoto v1.0 DigestSet

Why do we still need to include other algorithms as `VALID_ALGORITHM` for both v0.1 and v1.0?
From what I can see:
- The `VALID_ALGORITHMS` variable of `provenance/intoto/v01/__init__.py` is not being used anywhere in the codebase.
- The `VALID_ALGORITHMS` variable of `provenance/intoto/v1/__init__.py` is being used in this method [is_valid_digest_set](https://github.com/oracle/macaron/blob/e36578ee974471f156919bebdfbce5134479982a/src/macaron/slsa_analyzer/provenance/intoto/v1/__init__.py#L176).
- There are some differences between the [intoto/v1/\_\_init\_\_.py::is_valid_digest_set](https://github.com/oracle/macaron/blob/e36578ee974471f156919bebdfbce5134479982a/src/macaron/slsa_analyzer/provenance/intoto/v1/\_\_init\_\_.py#L176) and [intoto/v01/\_\_init\_\_.py::is_valid_digest_set](https://github.com/oracle/macaron/blob/e36578ee974471f156919bebdfbce5134479982a/src/macaron/slsa_analyzer/provenance/intoto/v01/__init__.py#L173).
  - The return type is different.
  - The one for v01 doesn't check the keys in DigestSet against the list of `VALID_ALGORITHMS`.

From my understanding, the usages of `sha1` and/or `gitDigest` are only for obtaining the git commit hash from the DigestSet of provenance. Therefore, they are different from the `VALID_ALGORITHMS` we used to validate the content of DigestSet for the provenance payload. Am I understanding it correctly? If that is True, then I still think we need to agree on whether we should resolve this inconsistencies between the two `is_valid_digest_set` :thinking: 
@nathanwn can you let me know the original reason why the `is_valid_digest_set` doesn't incorporate check for valid algorithms. Was it because we want to accept un-recognized fields as suggested by the intoto payload definition?

tests/repo_finder/test_provenance_extractor.py
```diff
@@ -0,0 +1,473 @@
...
+
+    if last_target is None:
+        return False
+
+    if new_value is None:
+        del last_target[keys[len(keys) - 1]]
+    else:
+        last_target[keys[len(keys) - 1]] = new_value
+
+    return True
```
nathanwn, line 466: I would personally do it like this.

```suggestion
def _json_modify(entry: JsonType, keys: list[str], new_value: JsonType) -> None:
    """Modify the value found by following the list of depth-sequential keys inside the passed JSON dictionary.

    The found value will be overwritten by the `new_value` parameter.
    If `new_value` is `None`, the value will be removed.
    If the final key does not exist, it will be created as `new_value`.
    """
    target = entry

    for key in keys[:-1]:
        target = target[key]

    if new_value is None:
        del target[keys[-1]]
    else:
        target[keys[-1]] = new_value
```

No reason to do any validation and return `bool`. Just let the test raise errors and give us the stack trace.

src/macaron/slsa_analyzer/provenance/intoto/v01/__init__.py
nathanwn: My understanding:
* The `VALID_ALGORITHMS` sets and the `is_valid_digest_set` functions are used to validate values of the `DigestSet` type (in-toto spec: [v0.1.0](https://github.com/in-toto/attestation/blob/main/spec/v0.1.0/field_types.md), [v1](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md)). Previously, the `is_valid_digest_set` function for `v01` did not validate the hash algorithms. The new `is_valid_digest_set` function for `v1` now does, though. We probably want to unify these two functions somehow. Not sure if it should be part of this PR. What do you think @benmss?
* The `"sha1"` and `"gitCommit"` keys are only used in the context of extracting commit hashes from provenances, and are not used for validating `DigestSet` values in general.

benmss: I agree with unifying the two functions. I'll add it to this PR.

src/macaron/slsa_analyzer/provenance/witness/__init__.py
benmss: Given that the way in which the `in-toto` spec is used depends upon the provenance it is contained within, I believe that this (and the other TODO in the file) provides no value. We cannot add support until we know what we are adding it for. This would be better tracked in an issue.

nathanwn: I agree with you here. In that case, we should remove all similar TODOs in the file.

In connection with this, unless there is any objection from anyone else, I also want to suggest hard-coding the `predicateType` check for Witness provenances instead of making it a configuration option, which was the initial reason for all of these TODOs. If we know for sure we are only supporting one specific version of Witness provenance, there is no reason to make it a configurable option. This is not a suggestion for this PR. If it is agreed, a fix can be created for it later.

tests/repo_finder/test_provenance_extractor.py
benmss: I see where you are coming from, but this would require disabling multiple `mypy` checks to be usable.

src/macaron/slsa_analyzer/provenance/witness/__init__.py
benmss: I agree with that also.

tests/repo_finder/test_provenance_extractor.py
benmss: I've made use of the `json_extract` function here instead.

nathanwn: I assume it should be OK if we don't type-annotate `entry` as `JsonType`, which is too wide of a type in this context because test input must always be nested dictionaries anyway? Either annotating it as `dict` or even `Any` is fine for me here.

tromai (approved):
Thanks for the change. I have no further comment.

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -0,0 +1,274 @@
...
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class ProvenanceExtractionException(MacaronError):
+    """When there is an error while extracting from provenance."""
+
+
+SLSA_V01_DIGEST_SET_ALGORITHMS = ["sha1"]
+SLSA_V02_DIGEST_SET_ALGORITHMS = ["sha1"]
+SLSA_V1_DIGEST_SET_ALGORITHMS = ["sha1", "gitCommit"]
```
nathanwn, line 21: The name should mention "commit hash" because these hash algorithms are specifically for git commit hash and not just any general digest set value.

nathanwn (approved):
I don't have any other comments.

Thanks for your hard work for the PR.

src/macaron/slsa_analyzer/provenance/intoto/v01/__init__.py
```diff
@@ -161,7 +186,9 @@ def is_valid_digest_set(digest: dict[str, JsonType]) -> TypeGuard[dict[str, str]
         ``True`` if the digest set is valid according to the spec, in which case its type
         is narrowed to a ``dict[str, str]``; ``False`` otherwise.
     """
-    for value in digest.values():
-        if not isinstance(value, str):
+    for key in digest:
+        if key not in VALID_ALGORITHMS:
+            return False
```
nathanwn, line 191: Sorry for missing this earlier.

According to the [Guidelines section on the Digest Set page](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#guidelines):

> Consumers MUST only accept algorithms that they consider secure and MUST ignore unrecognized or unaccepted algorithms.

Therefore, I don't think we should restrict the digest set to only a set of known algorithms.

For Witness provenances, they also have `"gitoid:sha1"` and `"gitoid:sha256"` as keys in each subject's digest set, although these are not mentioned on the Digest Set page in the Supported Algorithms section. This valid algorithm restriction will cause Macaron to error while validating a Witness provenance payload.

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -0,0 +1,274 @@
...
+
+
+class JsonExtractionException(MacaronError):
+    """When there is an error while extracting from JSON."""
+
+
+T = TypeVar("T", bound=JsonType)
+
+
+def json_extract(entry: JsonType, keys: list[str], type_: type[T]) -> T:
```
behnazh-w, line 238: This is a useful function that can be used outside this module.

Can you please move it to `src/macaron/json_tools.py` module?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -920,3 +1000,7 @@ def __init__(self, *args: Any, context: AnalyzeContext | None = None, **kwargs:
         """
         super().__init__(*args, **kwargs)
         self.context: AnalyzeContext | None = context
+
+
+class InvalidAnalysisTargetError(MacaronError):
```
behnazh-w, line 1005: Can you please move this error as well?

behnazh-w:
@benmss Can you please add an integration test, disabling repo finder to test if the repo and commit is found through the provenance? This test worked for me:

```bash
macaron -dp <path-to--defaults.ini> analyze -purl pkg:npm/semver@7.6.0
```

Please make sure to add it the block that checks if npm tests are enabled.

behnazh-w (approved):

PR #660, opened 2024-03-05
fix: preserve the order of elements of the list extract from defaults.ini

tromai:
This PR is created to addressed to issue mentioned here https://github.com/oracle/macaron/pull/641#issuecomment-1960742681.

This issue is addressed by removing the parameter `duplicate_ok` in the `get_list` method https://github.com/oracle/macaron/blob/023e0952a2c885b8e31a813b32c1274d210d5629/src/macaron/config/defaults.py#L18-L26. This essentially will make the `get_list` method not modifying the order of elements in the return list of strings. 

As a result, the unit tests in [tests/config/test_defaults.py](https://github.com/oracle/macaron/blob/fb302eba14bf5eab637d788c28dfa1cd492ef890/tests/config/test_defaults.py) are also re-implemented to reflect this new behavior, hence a major modification within that file. 

In additions, I have also rename the parameter `item` to `option` according to the documentation of ConfigParser - https://docs.python.org/3/library/configparser.html#module-configparser

### DIFF
--- a/src/macaron/config/defaults.py
+++ b/src/macaron/config/defaults.py
@@ -22,7 +22,7 @@ class ConfigParser(configparser.ConfigParser):
         delimiter: str | None = "\n",
         fallback: list[str] | None = None,
         strip: bool = True,
-        duplicated_ok: bool = False,
+        remove_duplicates: bool = True,
     ) -> list[str]:
         r"""Parse and return a list of strings from an ``option`` for ``section`` in ``defaults.ini``.
 
@@ -37,7 +37,7 @@ class ConfigParser(configparser.ConfigParser):
         If ``strip`` is True  (default: True), strings are whitespace-stripped and empty strings
         are removed from the final result.
 
-        If `duplicated_ok` is False, duplicated elements which come after the their first instances will
+        If `remove_duplicates` is True, duplicated elements which come after the their first instances will
         be removed from the list. This operation happens after ``strip`` is handled.
 
         The order of non-empty elements in the list is preserved.
@@ -55,8 +55,8 @@ class ConfigParser(configparser.ConfigParser):
             The fallback value in case of errors.
         strip : bool
             If True, strings are whitespace-stripped and any empty strings are removed.
-        duplicated_ok : bool
-            If True, duplicated elements are removed from the list.
+        remove_duplicates : bool
+            If True, duplicated elements will be removed from the list.
 
         Returns
         -------
@@ -88,7 +88,7 @@ class ConfigParser(configparser.ConfigParser):
                 if strip:
                     content = [x.strip() for x in content if x.strip()]
 
-                if duplicated_ok:
+                if not remove_duplicates:
                     return content
 
                 values = []
--- a/src/macaron/config/defaults.py
+++ b/src/macaron/config/defaults.py
@@ -48,7 +48,7 @@ class ConfigParser(configparser.ConfigParser):
         section : str
             The section in ``defaults.ini``.
         option : str
-            The option whose value will be split into the return list of strings.
+            The option whose values will be split into the a list of strings.
         delimiter : str | None
             The delimiter used to split the strings.
         fallback : list | None
--- a/src/macaron/config/defaults.py
+++ b/src/macaron/config/defaults.py
@@ -18,13 +18,12 @@ class ConfigParser(configparser.ConfigParser):
     def get_list(
         self,
         section: str,
-        item: str,
+        option: str,
         delimiter: str | None = "\n",
         fallback: list[str] | None = None,
-        duplicated_ok: bool = False,
         strip: bool = True,
     ) -> list[str]:
-        r"""Parse and return a list of strings from an item in ``defaults.ini``.
+        r"""Parse and return a list of strings from an ``option`` for ``section`` in ``defaults.ini``.
 
         This method uses str.split() to split the value into list of strings.
         References: https://docs.python.org/3/library/stdtypes.html#str.split.
@@ -37,22 +36,19 @@ class ConfigParser(configparser.ConfigParser):
         If ``strip`` is True  (default: True), strings are whitespace-stripped and empty strings
         are removed from the final result.
 
-        If ``duplicated_ok`` is True (default: False), duplicated values are not removed from the final list.
-
+        The order of non-empty elements in the list is preserved.
         The content of each string in the list is not validated and should be handled separately.
 
         Parameters
         ----------
         section : str
             The section in ``defaults.ini``.
-        item : str
-            The item to parse the list.
+        option : str
+            The option whose value will be split into the return list of strings.
         delimiter : str | None
             The delimiter used to split the strings.
         fallback : list | None
             The fallback value in case of errors.
-        duplicated_ok : bool
-            If True allow duplicate values.
         strip: bool
             If True, strings are whitespace-stripped and any empty strings are removed.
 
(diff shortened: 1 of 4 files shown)


nathanwn:
I have a question: why did we have this de-duplicating logic in the first place, and if we remove it does it break any of the current behaviors (i.e. those that may not be verified in any unit tests)?

tests/config/test_defaults.py
```diff
@@ -41,93 +42,194 @@ def test_create_defaults_without_permission() -> None:
...
+        (
+            """
+            [test.list]
+            list = github.com comma_ended, space string space string
+            """,
             None,
             True,
-            True,
-            ["comma_ended,", "github.com", "space", "space", "string", "string"],
+            ["github.com", "comma_ended,", "space", "string", "space", "string"],
```
benmss, line 100: I feel like this third set of inputs could be removed as it doesn't really test anything new. That multiple list items can be read per line is already handled by the above two inputs.

behnazh-w:
> I have a question: why did we have this de-duplicating logic in the first place, and if we remove it does it break any of the current behaviors (i.e. those that may not be verified in any unit tests)?

That's useful when the user makes a mistake in the configuration file while we expect to have unique values. We shouldn't remove this feature.

tromai:
> That's useful when the user makes a mistake in the configuration file while we expect to have unique values. We shouldn't remove this feature.

Hmm interesting, I wonder what would be the examples in Macaron where the caller of `defaults.get_list` is expecting a unique values @behnazh-w? I was not aware of places that removing the de-duplication would break the existing logic. Therefore, resolving those places too would need to be done in this PR.
In my implementation was following the principle that `defaults.get_list` will only return the same value as defined in `defaults.ini`, hence the de-duplication was removed from `defaults.get_list`. 
If we were to preserve the order of elements in the returned list and de-duplicate the list inside `defaults.get_list`, it would be quite complicated. This is because `defaults.get_list` doesn't know what is the context of the values in the list. Therefore, it cannot de-duplicate the values without breaking the original order (as defined in `defaults.ini`). For example:
```
[section]
option = 
    foo
    bar
    foo
```
`defaults.get_list` won't know whether it should return `["bar", "foo"]` or `["foo", "bar"]` because essentially both are the same and without any further context, we wouldn't know which of them should be the order the user is expecting :thinking:. 
Therefore, my suggestion is that we remove the de-duplication from `defaults.get_list` and instead we handle the de-duplication separated at places where we call `defaults.get_list` and expect a list with unique elements. And to do that, I think I need to go through the places where `defaults.get_list` is called again to see at which location, a list with unique elements from `defaults.get_list` is expected.

behnazh-w:
@tromai 
> Therefore, my suggestion is that we remove the de-duplication from `defaults.get_list` and instead we handle the de-duplication separated at places where we call `defaults.get_list` and expect a list with unique elements. 

I'm not sure I agree. Checking for duplicates as a way to catch mistakes in the configuration can be the default and be done in one place. There is no point to repeat that outside the function call. I would argue that if a config needs to allow duplicates, that's a specific case and should be turned on explicitly.

In general, I prefer to be as explicit as possible. So, I suggest to add a new argument to preserve the order, like `preserve_order`, which can be turned off by default. If both `duplicated_ok` and `preserve_order` are turned on, we can remove the duplicate item that comes last and explain that in the docstring.

nathanwn:
My suggestion: I think that we should **always keep the order** (which is a sensible default behavior while there is no good reason to not keep the order), while **optionally allowing for duplication**.

The logic would be different from the original one. I am thinking something similar to this (which is probably similar to what Behnaz is suggesting):

```python3
# content is the list of values read from the config and split apart with the delimiter
# Therefore, content retains the initial order.

if duplication_allowed:
    return content
else:
    values = []
    for item in content:
        if item in values:
            continue
        values.append(item)
    return values
```
A dedicated set, for example `already = set()`, can be used to check for duplication instead of checking on `values`. However, we need to use the `values` list to store and return the values anyway, and asymptotic time complexity is not an issue here because configurations cannot be very large.

tromai:
I think I agree with most of what Nathan suggested. 
The issue of ordering arises because I used a set to de-duplicate the list of strings. At the beginning, I don't have any use case for not preserving the original order of elements, therefore after we fix the order, I don't know if `preserve_order` would still need to be there.
I agree with @behnazh-w that we could now put the de-duplication inside `defaults.get_list` and always remove the duplicate item that comes last. I was thinking of a similar solution as proposed by @nathanwn .

src/macaron/config/defaults.py
```diff
@@ -37,22 +36,19 @@ def get_list(
...
+        The order of non-empty elements in the list is preserved.
         The content of each string in the list is not validated and should be handled separately.
 
         Parameters
         ----------
         section : str
             The section in ``defaults.ini``.
-        item : str
-            The item to parse the list.
+        option : str
```
nathanwn, line 46: Nit: I would call this `key` instead of `option`, as suggested by the [configparser documentation](https://docs.python.org/3/library/configparser.html#supported-ini-file-structure) and [wikipedia](https://en.wikipedia.org/wiki/INI_file).

tromai: I think the reason I decided to use `option` is because I was following a similar interface as these `get` method in the `configparser` documentation - https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.get
But if it's more correct to use the same terminology for the INI file here then I am happy to change too.

nathanwn: I see. I think if you have chosen it for a reason then I think it's fine.

nathanwn (approved):
LGTM. Thanks for the fix.

tests/config/test_defaults.py
tromai: The reason for this test case was for when `None` is used as the delimiter. `None` as the delimiter used in `str.split` will split the string on any white spaces, new lines, etc ([docs](https://docs.python.org/3/library/stdtypes.html#str.split)). This test case is for validating if the string is splits correctly on spaces. 
I agree that it doesn't test anything new in the context of `defaults.get_list` and therefore it can be removed.

tromai: Removed in https://github.com/oracle/macaron/pull/660/commits/8a3f5ce40e5a0e8288b2205ff445d075aaa8830a

benmss (approved):

src/macaron/config/defaults.py
```diff
@@ -37,24 +37,26 @@ def get_list(
...
+            The option whose value will be split into the return list of strings.
```
behnazh-w, line 51:
```suggestion
            The option whose values will be split into a list of strings.
```

src/macaron/config/defaults.py
```diff
@@ -37,24 +37,26 @@ def get_list(
...
             The delimiter used to split the strings.
         fallback : list | None
             The fallback value in case of errors.
-        duplicated_ok : bool
-            If True allow duplicate values.
-        strip: bool
+        strip : bool
             If True, strings are whitespace-stripped and any empty strings are removed.
+        duplicated_ok : bool
+            If True, duplicated elements are removed from the list.
```
behnazh-w, line 59: The description should be 
```
         If True, the duplicated elements will not be removed from the list.
```
?
I think this argument name is confusing, so maybe change it to

```suggestion
        remove_duplicates : bool
            If True, duplicated elements will be removed from the list.
```

tests/config/test_defaults.py
```diff
@@ -41,93 +42,234 @@ def test_create_defaults_without_permission() -> None:
...
 )
-def test_get_str_list_with_default_delimiter(
-    section: str, item: str, strip: bool, duplicated_ok: bool, fallback: list[str], expect: list[str]
+def test_get_str_list_default(
+    user_config_input: str,
+    expect: list[str],
+    tmp_path: Path,
 ) -> None:
-    """Test getting a list of strings from defaults.ini using the default delimiter."""
+    """Test default behavior of getting a list of strings from an option in defaults.ini."""
```
behnazh-w, line 149: Can you please be more specific what the default behavior is?

tromai: Addressed in https://github.com/oracle/macaron/pull/660/commits/87cd0f9b38681b1b741bf6a2c43a1f271d036247

src/macaron/config/defaults.py
tromai: Modified in https://github.com/oracle/macaron/pull/660/commits/3b7cd97fd7b245ad8f05994aad7f4e3ff38f8427

src/macaron/config/defaults.py
tromai: Fixed in https://github.com/oracle/macaron/pull/660/commits/ed2a337ad980489722f351fbac97d879329cd0fa

behnazh-w (approved):

PR #668, opened 2024-03-14
chore: check tags using GitHub API before cloning.

### LINKED ISSUE #644: Check GitHub repository tags via API calls before cloning
Currently tags for a GitHub repository are retrieved as part of the repository cloning operation. By using GitHub's API, tags alone can be retrieved, allowing the Commit Finder to perform its function before the repository is cloned. In cases where the Commit Finder would be unable to match a version to a tag and find a commit (meaning analysis will not continue), this allows cloning the repository to be skipped entirely.

benmss:
For analysis targets that would require the cloning of a GitHub repository, and have a provided version but no provided digest, this PR adds an additional pre-step that performs the Commit Finder tag matching process on the first page of tag results returned via the GitHub API for that repository. 
One of the following outcomes result from this process:

- The version matched a tag within the results -> No change
- The version did not match a tag within the results, but there are more tags than the first page of results contains -> No change
- The version did not match a tag within the results, and the tags within the results are an exhaustive list -> Do not clone the repository

Closes #644 
Does not include #576

### DIFF
--- a/src/macaron/repo_finder/commit_finder.py
+++ b/src/macaron/repo_finder/commit_finder.py
@@ -12,9 +12,10 @@ from gitdb.exc import BadName
 from packageurl import PackageURL
 from pydriller import Commit, Git
 
+from macaron.config.global_config import global_config
 from macaron.repo_finder import repo_finder_deps_dev
 from macaron.repo_finder.repo_finder import to_domain_from_known_purl_types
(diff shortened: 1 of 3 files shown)


tromai:
As we have discussed internally to stop exploring this approach, should we close this PR?

PR #675, opened 2024-03-21
chore: use dunder function __str__ when requiring a string serialization of a PackageURL object

jenstroeger:
The `PackageURL` class implements the [`__str__()`](https://docs.python.org/3/reference/datamodel.html#object.__str__) function [here](https://github.com/package-url/packageurl-python/blob/ef3747a2e6c136f3fbe1742c81d6da295cee4546/src/packageurl/__init__.py#L416-L417):
```python
    def __str__(self, *args: Any, **kwargs: Any) -> str:
        return self.to_string()
```
That `__str__()` function is invoked automatically when using `%s` in a format string (as is the case when logging, [docs](https://docs.python.org/3/library/string.html#format-specification-mini-language)) or when calling the [str()](https://docs.python.org/3/library/stdtypes.html#str) built-in.

This change removes the `to_string()` use such that the plain `PackageURL` objects are passed to logging and `str()` and thereby their `__str__()` function is called instead. That guarantees that the `PackageURL` class is responsible for returning the “‘informal’ or nicely printable string representation” of the object. And, lucky enough, that’s the value produced by `to_string()` anyway.

### DIFF
--- a/src/macaron/repo_finder/repo_finder.py
+++ b/src/macaron/repo_finder/repo_finder.py
@@ -70,11 +70,11 @@ def find_repo(purl: PackageURL) -> str:
     ]:
         repo_finder = DepsDevRepoFinder()
     else:
-        logger.debug("No Repo Finder found for package type: %s of %s", purl.type, purl.to_string())
+        logger.debug("No Repo Finder found for package type: %s of %s", purl.type, purl)
         return ""
 
     # Call Repo Finder and return first valid URL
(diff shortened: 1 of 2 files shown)


behnazh-w:
Thanks @jenstroeger for the changes. They make sense :+1:

behnazh-w (approved):

behnazh-w (approved):

behnazh-w (approved):

PR #673, opened 2024-03-20
feat: extend static analysis and compute confidence scores for deploy commands

behnazh-w:
This PR extends and makes changes to the static analysis of CI configurations in Macaron with the high-level goal of finding build and deploy commands more accurately. To achieve that, some of the abstractions had to be replaced to allow writing customized analyses, such as detecting build language setup in a GitHub Actions workflow or detecting reachable secrets. The list of extensions and changes are as follows:

- Replaces `BashCommands` which consisted of build tool commands collected after analyzing GitHub Actions and passed to checks with `BuildToolCommand`. With this PR `BuildToolCommand` are created by each build tool instead, and can be detected by calling a build tool API as part of the check. This delegation of analysis to each build tool allows to customize them per build tool. Moreover, now each build tool has direct access to the parsed AST of GitHub Actions and bash scripts, which opens opportunities to add new analyses more easily.
- The build related checks are refactored and simplified to use `BuildToolCommand`.
- The callgraph analysis, which needs to be implemented for each CI service, is extended with new node types for GitHub Actions. The callgraph plays the role of Intermediate Representation and is available to all checks.
- The `mcn_build_script_1` check does not depend on any checks and always runs by default based on a customer request.
- The `mcn_build_as_code_1` check now reports deploy commands with confidence scores. The confidence scores are computed based on additional facts collected from GitHub Actions, such as the CI event, reachable secrets, and name of the workflows.
- New analysis is added to resolve the value of expression variables, which is used for other analysis, such as reachable secrets and build language detection.
- New abstractions are added to model third-party GitHub Actions. This feature is used to collect data about build Language setup.
- Fixes this parsing issue: https://github.com/oracle/macaron/issues/107
- Changes the check result log level to debug mode.

PR #676, opened 2024-03-21
chore: exclude provenance check in micronaut-core temporarily

behnazh-w:
This PR temporarily excludes `mcn_provenance_available_1` check for `micronaut-core` integration test because provenances have failed to publish due to an [issue](https://github.com/slsa-framework/slsa-github-generator/issues/3350) in `generator_generic_slsa3.yml@v1.9.0`.

It also excludes `mcn_infer_artifact_pipeline_1`, which is due to a non-deterministic behavior in deploy command detection, which will be fixed in https://github.com/oracle/macaron/pull/673.


tests/e2e/defaults/micronaut-core.ini
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2022 - 2024, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+[analysis.checks]
+# By default, we don't exclude any checks.
```
benmss, line 5: Maybe a note about why these checks are being excluded would be better here instead of this.

behnazh-w: See commit [6275820](https://github.com/oracle/macaron/pull/676/commits/627582061efd0d22991e34e3a3f38e61f524c213)

benmss (approved):

PR #679, opened 2024-03-22
chore: update deps.dev repo finder to use latest API.

### LINKED ISSUE #666: Update the Repo Finder to work with deps.dev official v3 release
The initial version of the public deps.dev API was `v3alpha`. A recent update has created a stable version `v3` that comes with certain guarantees, including: no breaking changes, and a six-month deprecation warning. Updating the Repo Finder to use this version would provide long term stability.

Alongside this update, the deps.dev API has added new features to the `v3alpha` version. The most important feature among these is support for using PURLs directly, instead of needing to convert into the specific API format which differs per language type. The Repo Finder code could be simplified if using this new style.

benmss:
This PR simplifies code in the `DepsDevRepoFinder` class thanks to the changes made to the API as part of a new `deps.dev` release. Previously, each language (Python, Rust, C#, and JavaScript) had to follow specific rules to allow for the API to handle each of them. Now, the PURL itself can be sent instead.

Closes #666

### DIFF
--- a/src/macaron/repo_finder/repo_finder_deps_dev.py
+++ b/src/macaron/repo_finder/repo_finder_deps_dev.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2023 - 2024, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """This module contains the PythonRepoFinderDD class to be used for finding repositories using deps.dev."""
@@ -9,6 +9,8 @@ from urllib.parse import quote as encode
 
 from packageurl import PackageURL
 
+from macaron.errors import JsonError
+from macaron.repo_finder.provenance_extractor import json_extract
 from macaron.repo_finder.repo_finder_base import BaseRepoFinder
 from macaron.repo_finder.repo_validator import find_valid_repository_url
 from macaron.util import send_get_http_raw
@@ -47,7 +49,7 @@ class DepsDevRepoFinder(BaseRepoFinder):
         str :
             The URL of the found repository.
         """
-        request_urls = self._create_urls(purl.namespace or "", purl.name, purl.version or "", purl.type)
+        request_urls = self._create_urls(purl)
         if not request_urls:
             logger.debug("No urls found for: %s", purl)
             return ""
@@ -70,7 +72,7 @@ class DepsDevRepoFinder(BaseRepoFinder):
 
         return ""
 
-    def _create_urls(self, namespace: str, name: str, version: str, type_: str) -> list[str]:
+    def _create_urls(self, purl: PackageURL) -> list[str]:
         """
         Create the urls to search for the metadata relating to the passed artifact.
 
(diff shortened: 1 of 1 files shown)


src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -70,51 +72,54 @@ def find_repo(self, purl: PackageURL) -> str:
...
+            The PURL of an artifact.
 
         Returns
         -------
         list[str]
             The list of created URLs.
         """
-        base_url = self._create_type_specific_url(namespace, name, type_)
+        # See https://docs.deps.dev/api/v3alpha/
+        base_url = f"https://api.deps.dev/v3alpha/purl/{encode(purl.to_string()).replace('/', '%2F')}"
```
behnazh-w, line 92: You don't need to call `purl.to_string()` directly. You can either use the PURL in the format string or call `str()`. See https://github.com/oracle/macaron/pull/675

src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -70,51 +72,53 @@ def find_repo(self, purl: PackageURL) -> str:
...
+            metadata: dict = json.loads(response.text)
+        except ValueError as error:
+            logger.debug("Failed to parse response from deps.dev: %s", error)
+            return []
 
-        if latest_version:
-            logger.debug("Found latest version: %s", latest_version)
-            return [f"{base_url}/versions/{latest_version}"]
+        try:
+            versions_keys = ["package", "versions"] if "package" in metadata else ["version"]
```
tromai, line 113: nit: I think this line can be move outside of the try catch block because it doesn't raise `JsonError`.

src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -151,54 +155,21 @@ def _read_json(self, json_data: str) -> list[str]:
...
+            logger.debug("Failed to parse response from deps.dev: %s", error)
+            return []
 
-        if not parsed["links"]:
-            logger.debug("Metadata had no URLs: %s", parsed["versionKey"])
+        try:
+            links_keys = ["version", "links"] if "version" in parsed else ["links"]
+            links = json_extract(parsed, links_keys, list)
+        except JsonError as error:
+            logger.debug("Could not extract 'links' from deps.dev response: %s", error)
```
tromai, line 168: nit: The debug command only mentions `links` while in some scenarios it could be `version` too.

src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -151,54 +155,21 @@ def _read_json(self, json_data: str) -> list[str]:
...
             return []
 
         result = []
-        for item in parsed["links"]:
+        for item in links:
             result.append(item.get("url"))
```
tromai, line 173: Because the return type of this function is `list[str]`. If an `item` doesn't have `url` field, a `None` will be appended to the return list, which violates the type `list[str]`. I think we need to check if `url` exists and have a value of type string before appending to the returned list (unless we made an assumption that each item will always have `url` field that contains a string value).

tromai (approved):
LGTM. Thanks!

behnazh-w (approved):

PR #685, opened 2024-03-26
feat: improve VSA generation with digest for each subject

nathanwn:
This pull request improves upon the VSA generation feature in Macaron, by populating each subject with a corresponding SHA256 digest if certain conditions are met. This change helps VSA generation in Macaron become more in line with in-toto attestation specifications. More importantly, this also improves the VSA's quality in uniquely identifying artifacts verified by Macaron.

Specifically, built on top of Macaron's recent feature of taking a provenance file as input, this pull request identifies if the PackageURL provided through the command-line argument `--package-url/-purl` of `macaron analyze` identifies an artifact being one of the subjects in the provenance. If such a subject is found, the corresponding SHA256 digest is stored in the Macaron database, which can then be retrieved and populated in Macaron-generated VSAs.

For example, given the following analyze command

```
macaron analyze \
    --provenance-file ./multiple.intoto.jsonl \
    --package-url pkg:maven/io.micronaut/micronaut-router@4.1.1?type=jar
```

Macaron will check if there is a jar file for the `micronaut-router` package among the subjects in the provenance file `./multiple.intoto.jsonl`.

In this PR, only PackageURL of type `maven` is supported. Other PackageURL types can be added in future pull requests.

Some other related changes include
* Disabling provenance discovery in the `mcn_provenance_available_1` check in case a user-provided provenance is available.
* Making the `mcn_provenance_expectation_1` check also validate the user-provided provenance if it is available.

Below is a comprehensive list detailing the changes made in this PR, spanning across a few of commits:

* b3eb53a `fix: disable provenance discovery in case a provenance is already available`

* 8c49c69 `fix: enable provenance expectation validation for user-provided provenances`

* 1b82d5e `chore: add a table to the database schema to store sha256 digests of provenance subjects`

A new table is created to store the SHA256 hash. One option being considered was to add another column to the `_component` table. However, since this will introduce breaking changes to the current policy library, I decided against it at this time. Another option considered was the `ReleaseArtifact` and `HashDigest` table. However, they are now currently storing a different kind of data to what this pull request wants to achieve. Future revision of the database schema may clarify whether we want to unify these tables and how we want an eventual schema to look like.

* f2ed7f7 `chore: introduce maven artifact types and utilities`

* 12066e3 `refactor: improve extraction of subjects being build artifacts from witness provenances`

There was an existing utility function for extracting subjects from witness provenance. It was originally created for the Witness provenance check. It has then been modified, taking into account the [product attestor](https://github.com/in-toto/witness/blob/main/docs/attestors/product.md) to only recognize subjects that are actual artifacts. This utility function is also now used for matching a subject to a given PURL, introduced in the next commit.

* 7fa98d6 `chore: match a witness provenance subject against a maven artifact purl`

* 1fbe7ec `chore: modify vsa generation to populate each subject with a sha256 digest if it exists`


src/macaron/artifact/maven.py
```diff
@@ -0,0 +1,145 @@
...
+        purl_qualifiers={"type": "pom"},
+    )
+    JAVADOC = _MavenArtifactType(
+        filename_pattern="{artifact_id}-{version}-javadoc.jar",
+        purl_qualifiers={"type": "javadoc"},
+    )
+    JAVA_SOURCE = _MavenArtifactType(
+        filename_pattern="{artifact_id}-{version}-sources.jar",
+        purl_qualifiers={"type": "sources"},
+    )
```
tromai, line 53: I wonder if the `type` key within `purl_qualifiers` maps directly to the `type` column in the Default Artifact Handlers Reference table [here](https://maven.apache.org/ref/3.9.6/maven-core/artifact-handlers.html)?

tests/artifact/test_maven.py
```diff
@@ -0,0 +1,205 @@
+# Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""Tests for types and utilities for Maven artifacts."""
+
+import pytest
+from packageurl import PackageURL
+
+from macaron.artifact.maven import MavenArtifact, MavenArtifactType
+# , MavenSubjectPURLMatcher
```
tromai: Should this line be removed?

src/macaron/artifact/maven.py
nathanwn: Yes, that's correct. As stated in the [Maven section of the PackageURL doc](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#maven) which I have left a link to:
> Known qualifiers keys are: classifier and type as defined in the POM documentation.

Hope this clarifies.

tromai: Hmmm right, I have some questions regarding this:
- Why is the `type` field for `JAVA_SOURCE` is `sources` while in the `type` column [here](https://maven.apache.org/ref/3.9.6/maven-core/artifact-handlers.html), it's `java-source`.
> Known qualifiers keys are: classifier and type as defined in the POM documentation.
- If it's mentioned that there are two type of qualifiers `type` and `classifier` for a Maven PURL. I wonder why for `JAVADOC` and `JAVA_SOURCE` we only have one quantifier. Shouldn't it be something like this :thinking: ?
```
    JAVADOC = _MavenArtifactType(
        filename_pattern="{artifact_id}-{version}-javadoc.jar",
        purl_qualifiers={"type": "javadoc", "classifier": "javadoc"},
    )
    JAVA_SOURCE = _MavenArtifactType(
        filename_pattern="{artifact_id}-{version}-sources.jar",
        purl_qualifiers={"type": "java-source", "classifier": "sources},
    )
```
In additions, we perform a dictionary comparison here 
https://github.com/oracle/macaron/blob/7fa98d6e8118dca219f927cfaf123a368c54fd54/src/macaron/artifact/maven.py#L100-L102
It would mean that we expect the qualifier from the input package url must only contains one key-value which is `{ "type": "<value>" }`.
From the examples given [here](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#maven). It seems to me that a Maven package URL can have both `type` and/or `classifier` quantifiers. 
Are we only supporting the package URL with only `type` quantifier in this PR? :thinking:

src/macaron/artifact/maven.py
```diff
@@ -0,0 +1,203 @@
...
+        return None
+
+
+class MavenSubjectPURLMatcher:
+    """A matcher matching a PURL identifying a Maven artifact to a provenance subject."""
+
+    @staticmethod
+    def get_subject_in_provenance_matching_purl(
+        provenance_payload: InTotoPayload, purl: PackageURL
+    ) -> InTotoV01Subject | InTotoV1ResourceDescriptor | None:
```
tromai, line 163: I don't think this method returns `InTotoV1ResourceDescriptor` instances. 
```
artifact_subjects = extract_build_artifacts_from_witness_subjects(provenance_payload)
```
The returned subject is an element in `artifact_subjects` whose type is `list[InTotoV01Subject]` 
https://github.com/oracle/macaron/blob/12066e31b59e64ecad29c59c703515083baf7468/src/macaron/slsa_analyzer/provenance/witness/__init__.py#L124

tromai: Hmm I think it's because we are following the protocol here https://github.com/oracle/macaron/blob/7fa98d6e8118dca219f927cfaf123a368c54fd54/src/macaron/slsa_analyzer/provenance/intoto/__init__.py#L128

If that's the case then I don't think there is any issue.

src/macaron/artifact/maven.py
nathanwn: Thanks for pointing out the issue with the `type` qualifier of `java-source`. You are correct that it should be `java-source` and not `sources`. Fixed in [ebf8667](https://github.com/oracle/macaron/pull/685/commits/ebf866729da04f25871a8125519eb665f2dcee32).

I have decided to only support the `type` qualifier for now. I added a comment to clarify that in the same commit. The `type` qualifier is enough for uniquely identifying the artifact type of each artifact while the `classifier` qualifier does not have this property. In addition, supporting `classifier` will make the implementation quite a bit more complex, while there is no concrete usage for it for now, so I figure supporting only the `type` qualifier at the moment is probably sufficient.

tromai: Sounds good to me. Thanks.

src/macaron/artifact/maven.py
nathanwn: Yeah, the function signature of `get_subject_in_provenance_matching_purl` returns a union of multiple different subject types since it is meant for extracting the subject of any provenance. This will probably require some rethinking later on and I am not 100% that this interface is ideal. As you have seen the only provenance type that we are supporting is Witness, which is still in-toto v0.1.

src/macaron/vsa/vsa.py
```diff
@@ -135,11 +144,42 @@ class VerificationResult(StrEnum):
...
 
+def get_common_purl_from_artifact_purls(purl_strs: Iterable[str]) -> str | None:
+    """Get a single common PackageURL given some artifact PackageURLs.
+
+    Assumption: A package may have more than one artifact. If each artifact is identified
+    by a PackageURL, these PackageURLs still share the type, namespace, name, and
+    version values. The common PackageURL contains these values.
+    """
+    try:
+        purls = [PackageURL.from_string(_) for _ in purl_strs]
```
tromai, line 155: Interesting, I didn't know that using `_` like this could work! 
nit: I think it  would be clearer to use a variable name instead.

src/macaron/vsa/vsa.py
```diff
@@ -135,11 +144,42 @@ class VerificationResult(StrEnum):
...
+    except ValueError:
+        return None
+
+    purl_type = purls[0].type
```
tromai, line 159: Are we assuming that `purls` will never be empty?

src/macaron/artifact/maven.py
```diff
@@ -0,0 +1,210 @@
...
+            group_id=package_url.namespace,
+            artifact_id=package_url.name,
+            version=package_url.version,
+            artifact_type=maven_artifact_type,
+        )
+
+    @classmethod
+    def from_artifact_name(
+        cls,
+        artifact_name: str,
```
nicallen, line 122: `artifact_name` seems a bit ambiguous. Is it the filename?

src/macaron/vsa/vsa.py
```diff
@@ -210,53 +285,39 @@ def get_subject_verification_result(policy_result: dict) -> tuple[str, Verificat
...
+        component id of that component.
     """
     component_violates_policy_facts = policy_result.get("component_violates_policy", [])
     component_satisfies_policy_facts = policy_result.get("component_satisfies_policy", [])
 
+    if len(component_violates_policy_facts) > 0:
+        logger.info("Encountered software component failing the policy. No VSA is generated.")
+        return None
+
     # key: PURL; value: result with the highest component id
```
tromai, line 301: Do the values of this dictionary store the *highest component id number* that has a passing result, or the *result* of the highest component id? 
From my understanding of the implementation, the values are the `highest component id number`, is that correct?

src/macaron/artifact/maven.py
nathanwn: It is indeed the filename. Thanks for pointing out.
I have renamed `artifact_name` to `artifact_filename` in 263414bbd7cdd94a9613adcc97a66453d988a22f.

src/macaron/vsa/vsa.py
nathanwn: I realized I should have updated this comment now that the implementation has changed. Thanks for pointing out.
See my updated comment in 96da6a92e2703fea016413bee29c209ee8655c3c. Let me know if it clarifies things.

tromai: Yeah, it's clearer too me know. Thanks.

src/macaron/vsa/vsa.py
nathanwn: This function currently does not take an empty collection of PURL strings, because it always gets the result from the `get_components_passing_policy` function, which returns `None` if the result is empty.

That said, I also don't think adding a check here will do any harm. It may be even better to do it anyway. See my commit 077adc05a1a77f2299d418f267a85a97ae2b6e37. I also added some unit tests while I was at it.

src/macaron/vsa/vsa.py
nathanwn: See 6f948e6a56d897cdd91999ce3c847c0eb021a539.

Besides the point, but I initially coded this as `list(map(PackageURL.from_string, purl_strs))`, then changed to this since list comprehension, since is usually regarded as being more idiomatic in Python. The `_` was something I learned from reading some Scala code a while back, and then trying it in Python.

tromai (approved):
LGTM. Thanks for the changes.

src/macaron/slsa_analyzer/provenance/witness/__init__.py
```diff
@@ -133,28 +134,25 @@ def extract_witness_provenance_subjects(witness_payload: InTotoPayload) -> set[W
 
     Returns
     -------
-    dict[str, str]
+    list[InTotoV01Subject]
         A dictionary in which each key is a subject name and each value is the corresponding SHA256 digest.
```
nicallen, line 138: Need to update comment.

nicallen: Also need to remove non-existent `extensions` parameter from comment above.

nathanwn: Thanks. Fixed in 9e6f98f27a3e7255e6b5333907ad3b626e2220a8.

nicallen (approved):

src/macaron/artifact/maven.py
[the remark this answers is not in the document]
behnazh-w: Yes, that's true. I didn't recommend to use it because it has fixed endpoints unfortunately. But I wanted to mention this implementation as an alternative. We shouldn't think about `maven` artifacts only and also think how easy it is to add support for more artifact types. This module supports several artifacts types if you take a look and each of them is implemented in a simple function.

src/macaron/artifact/maven.py
```diff
@@ -0,0 +1,210 @@
...
+        purl_qualifiers={"type": "pom"},
+    )
+    JAVADOC = _MavenArtifactType(
+        filename_pattern="{artifact_id}-{version}-javadoc.jar",
+        purl_qualifiers={"type": "javadoc"},
+    )
+    JAVA_SOURCE = _MavenArtifactType(
+        filename_pattern="{artifact_id}-{version}-sources.jar",
+        purl_qualifiers={"type": "java-source"},
+    )
```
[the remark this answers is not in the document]
nathanwn: `-javadoc.jar` and `-sources.jar` are not file extensions, but I get the question.

This pattern is currently being used in the `MavenArtifact.from_artifact_filename` class method to construct a regex to extract the artifact id from an artifact filename.

An alternative I see is to change `filename_pattern` to something like `filename_suffix`, which will involve also changing the implementation of the `MavenArtifact.from_artifact_filename` class method. I am not sure if doing this is exactly simplifying by any means while I chose to do it this was initially to explicitly specify the naming convention of Maven artifacts, but I can change the current implementation to this if you think it makes sense. If you have any other suggestions please let me know.

src/macaron/artifact/maven.py
```diff
@@ -0,0 +1,210 @@
...
+            return cls(
+                group_id=group_id,
+                artifact_id=artifact_id,
+                version=version,
+                artifact_type=maven_artifact_type,
+            )
+        return None
+
+
+class MavenSubjectPURLMatcher:
```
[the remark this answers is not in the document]
nathanwn: The generic function you are mentioning is defined [here](https://github.com/oracle/macaron/blob/nathanwn/improve-vsa-generation/src/macaron/database/table_definitions.py#L569C1-L607C20) and being used [here in `analyze.py`](https://github.com/oracle/macaron/blob/9e6f98f27a3e7255e6b5333907ad3b626e2220a8/src/macaron/slsa_analyzer/analyzer.py#L573C1-L577C14) to create a `ProvenanceSubject` for a `Component`.

This class is a concrete subject-PURL matcher for provenances of Maven build, which has Maven-specific logic and follows the generic interface [here](https://github.com/oracle/macaron/blob/nathanwn/improve-vsa-generation/src/macaron/slsa_analyzer/provenance/intoto/__init__.py#L128C1-L153C12). This was originated from the conversation we had previously about thinking of, for example, what if we want to support not just Maven artifacts but also Python artifacts, so I created a generic interface for subject-PURL matching, then implements a concrete matcher for Maven.

src/macaron/slsa_analyzer/provenance/witness/__init__.py
```diff
@@ -120,41 +121,38 @@ def extract_repo_url(witness_payload: InTotoPayload) -> str | None:
...
+    if not isinstance(witness_payload, InTotoV01Payload):
+        return []
+
+    subjects = witness_payload.statement["subject"]
+    artifact_subjects = []
+    for subject in subjects:
+        # Filter all subjects attested by the product attestor, which records all changed and
+        # created files in the build process.
+        # Documentation: https://github.com/in-toto/witness/blob/main/docs/attestors/product.md
+        if not subject["name"].startswith("https://witness.dev/attestations/product/v0.1/file:"):
```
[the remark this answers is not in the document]
nathanwn: This can match files that are not eventually published artifacts because these are files recorded by the Witness product attestors, which include all files created by the build.
However, this is sufficient for our internal use case that we are trying to support.

src/macaron/artifact/maven.py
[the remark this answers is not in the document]
nathanwn: In ff22105d7d2a62b5637f337fbbf7612a0554353a, I have removed the `MavenArtifact` type and enum.

behnazh-w (approved):

PR #687, opened 2024-03-27
chore: improve commit finder accuracy

### LINKED ISSUE #684: Improve Commit Finder Matching Accuracy
Further investigation of the Commit Finder's matching process has revealed two cases where a better match than currently found is available:

- Matching version '5.0.0.preview0009' finds tag 'v5.0.0-dataattributes001' instead of 'v5.0.0-preview009' because of a different number of zeros in the trailing number.
- Matching version '2.3.0-rc4' finds tag 'v2.3.0' instead of 'v2.3.0-RC4' because of a difference in case in the textual suffix.

benmss:
This PR improves the Commit Finder so it can match in two additional tag usage cases:

- Where part of a suffix has a different case to the version it represents. E.g. rc4 vs RC4
- Where part of a suffix contains a number string with a different amount of preceding zeros to the version it represents.

Tests for these cases have been added to the regression data set.

Closes #684

### DIFF
--- a/src/macaron/repo_finder/commit_finder.py
+++ b/src/macaron/repo_finder/commit_finder.py
@@ -100,7 +100,7 @@ validation_pattern = re.compile("^[0-9a-z]+$", flags=re.IGNORECASE)
 alphabetic_only_pattern = re.compile("^[a-z]+$", flags=re.IGNORECASE)
 hex_only_pattern = re.compile("^[0-9a-f]+$", flags=re.IGNORECASE)
 numeric_only_pattern = re.compile("^[0-9]+$")
-versioned_string = re.compile("^[a-z]+[0-9]+$", flags=re.IGNORECASE)  # e.g. RC1, M5, etc.
+versioned_string = re.compile("^([a-z]+)(0*)([1-9]+[0-9]*)$", flags=re.IGNORECASE)  # e.g. RC1, M5, etc.
 
 
 class AbstractPurlType(Enum):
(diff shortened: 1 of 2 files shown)


src/macaron/repo_finder/commit_finder.py
```diff
@@ -473,42 +473,73 @@ def _compute_tag_version_similarity(tag_version: str, tag_suffix: str, version_p
...
             count = count - 1
 
     # Try to reduce the count further based on the tag suffix.
     if tag_suffix:
-        last_part = version_parts[-1]
+        last_part = version_parts[-1].lower()
         # The tag suffix might consist of multiple version parts, e.g. RC1.RELEASE
         suffix_split = split_pattern.split(tag_suffix)
+        # Try to match suffix parts to version.
+        versioned_string_match = False
```
tromai, line 489: I am a bit unclear on the meaning of this `versioned_string_match` variable. Could you please give me some explanation on its role for matching the suffix parts?
Additionally, in this block:
```
                if versioned_string_match:
                    count = count + 1
                    continue
```
Why does a match would increase the count number if the lower the count, the better :thinking:

src/macaron/repo_finder/commit_finder.py
```diff
@@ -473,42 +473,73 @@ def _compute_tag_version_similarity(tag_version: str, tag_suffix: str, version_p
...
+
+        variable_suffix_pattern = _create_suffix_tag_comparison_pattern(tag_suffix)
+        if variable_suffix_pattern:
+            if re.match(variable_suffix_pattern, last_part):
+                count = count - 1
+            else:
+                count = count + 1
         else:
-            count = count - 1
+            count = count + 1
```
tromai, line 522: I not sure what the "last_part" here means in the case of "5.0.0.preview0009":
- Does it mean the whole "preview0009"
- Or just the non-zero numbers after the list of 0(s) "9"
:thinking:

src/macaron/repo_finder/commit_finder.py
benmss: `versioned_string_match` relates to a part of the suffix that is alphanumeric and is assumed to be some kind of versioning scheme, e.g. RC1, M4, etc. A single one of these is allowed to alter the overall score per comparison (if it matches part of the version), hence why the match is stored as a boolean.

The count increases if a match of that kind has already been found. Essentially the same as a non-matching part.

tromai: Ah I see. So essentially `v0.5.0-RC1` should be treated as good while `v0.5.0-RC1-RC2-RC3` would be treat as bad, and how bad it is depends on the number of alphanumeric parts after the first matched one (i.e RC1 is the first match one, then RC2, RC3 would be treated as bad).

src/macaron/repo_finder/commit_finder.py
```diff
@@ -473,42 +473,73 @@ def _compute_tag_version_similarity(tag_version: str, tag_suffix: str, version_p
...
-                        count = count - 1
-                        versioned_string_match = True
-                    else:
-                        count = count + 1
-        if tag_suffix != last_part:
-            count = count + 1
+
+                variable_suffix_pattern = _create_suffix_tag_comparison_pattern(suffix_part)
+                if not variable_suffix_pattern:
+                    continue
```
tromai, line 501: Hmm why do we ignore the scenario where the suffix_part is in the form that we don't support (i.e. `_create_suffix_tag_comparison_pattern` returns None) Should we treat this tag as bad if it's malformed?

benmss: If it returns None, that means the suffix is not an alphanumeric versioned string, and it's not an alphabetic string that matches the version part. In that case we do assume it is essentially junk. For a check that evaluates tags that could be an important thing to flag, but for these comparisons I don't think it is important.

src/macaron/repo_finder/commit_finder.py
benmss: Yes, that's it. And with the above two examples you pose, assuming the version we are after is `RC1` then this part of the code helps ensure that the tag with more suffixes is counted as being a worse match.

tromai (approved):

nathanwn (approved):
Thanks for the PR.

As we have discussed, we may want to consider unit-testing smaller components within the tag-matching algorithm if it makes sense. That may be addressed later though.

Issue #416, opened 2023-08-15
Rename the attribute `domain` to `hostname` for in config of Git services


nathanwn:
In the Git service-related sections in the `.ini` config, we are using the attribute `domain` to obtain the hostnames for Git services. This may not be the most accurate terminology in accordance with [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986.html) and [Python documentation for URL parsing](https://docs.python.org/3/library/urllib.parse.html#url-parsing).

Current ini config:
```
[git_service.github]
domain = github.com

[git_service.gitlab.publicly_hosted]
domain = gitlab.com
```

It is better to have:
```
[git_service.github]
hostname = github.com

[git_service.gitlab.publicly_hosted]
hostname = gitlab.com
```

Note: When using the `urlparse`/`urlsplit` function from the `urllib.parse` package, we get the following:

```
>>> from urllib.parse import urlsplit
>>> parse_result = urlsplit("https://github.com/owner/repo.git")
>>> parse_result
SplitResult(scheme='https', netloc='github.com', path='/owner/repo.git', query='', fragment='')
>>> parse_result.hostname
'github.com'
```

Therefore, it is probably best to use the name `hostname` as it is consistent with the naming of the library.

nathanwn:
Closed in #453.

PR #694, opened 2024-04-08
feat: allow defining a git service from defaults.ini  

### LINKED ISSUE #332: Add support for Bitbucket
Due to the changes in this [PR](https://github.com/oracle/macaron/pull/316)  we have to add a section for Bitbucket in `defaults.ini` to add the "allowed" Git hosts if cloned manually and provided as a local path.

We can also consider to support cloning the repositories.

### LINKED ISSUE #693: Support defining a generic Git Service through defaults.ini for build tool detection
# Problem
In Macaron, at this place https://github.com/oracle/macaron/blob/a3842b4bbd121c8f0b7e9cd1597830d1b41e419f/src/macaron/slsa_analyzer/analyzer.py#L912-L921

The build tool detection logic is only run if the git service of the target repository is supported by Macaron. This mean that `git_service` must be either `GitLab` or `GitHub`.

The git service detection requires the remote url to be of a known service (e.g. [github.com](http://github.com/) or [gitlab.com](http://gitlab.com/)). See implementation here:
https://github.com/oracle/macaron/blob/71accbf81e849947ca238ebfa3a9a6ef88fd01f6/src/macaron/slsa_analyzer/git_service/base_git_service.py#L71-L95

For a user to analyze an repository that is hosted on an unknown/not supported git service (e.g. `bitbucket.org`), they must first clone the repository into their filesystem and provide its local path to Macaron (i.e [analyzing-a-locally-cloned-repository](https://oracle.github.io/macaron/pages/using.html#analyzing-a-locally-cloned-repository)). Macaron will detect the remote URL of this local repository as it was from an unknown git service and will not run the build tool detection. _This is an expected behavior_.

However, there are some problems to this behavior:
- Theoretically, the build tool detection logic, which only looks statically into the files of the target repository, doesn't rely on the type of git or CI service for a repository.
- Given the scenario above, there aren't any checks that could possibly pass (other than the `version_control_system`, which always passes).

# Solution
Due to the following reasons:
- A build tool can be detected regardless of the git service or CI service. However our current checks (e.g. build and provenance related checks) all rely on reachability via CI (which currently, we have great support for GitHub but not as much for other Git and CI service). 
- Therefore, unless we come up with a check that doesn't rely on the git or CI service of a repo. The "if" statement here still makes sense https://github.com/oracle/macaron/blob/a3842b4bbd121c8f0b7e9cd1597830d1b41e419f/src/macaron/slsa_analyzer/analyzer.py#L912

With those reasons, we have decided to:
- **Support a GenericGitService with the host name that the user define in `defaults.ini`. This will allow the build tool detection to run for the scenario of analyzing a local repository with unknown git service.** (will be handled in this issue).
- Add a check called `has_build_configs` which relies only on the build tool detection and doesn't rely on git or CI service (TBD in a separated issue)
- Rename and modify the description of checks in Macaron to further reflects the changes (TBD in a separated issue)

tromai:
Closes #693 
Closes #332

- [x] Add generic git service
- [x] Rename the class and module to `LocalRepoGitService`.
- [x] Update the analyzing local repo tutorial to include the step to define the git service in `defaults.ini` . We should also document that if a domain is already supported in other services, it doesn't need to be allowed in this service.
- [x] Remember to note that this git service should only be used for local repos. 
- [x] Add integration test case.

### DIFF
--- a/docs/source/pages/using.rst
+++ b/docs/source/pages/using.rst
@@ -238,7 +238,7 @@ With the example above, the generated output reports can be seen here:
 Analyzing dependencies in the SBOM without the main software component
 ''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
 
-In the case where the repository URL of the main software component is not available (e.g. the repository is in a private domain where Macaron cannot access),
+In the case where the repository URL of the main software component is not available (e.g. the repository is in a self-hosted git service instance where Macaron cannot access),
 Macaron can still run the analysis on the dependencies listed in the SBOM.
 To do that, you must first create a PURL to present the main software component. This is so that this software component could be referenced later in the :ref:`verify-policy <verify-policy-command-cli>` command.
 For example: ``pkg:private_domain.com/org/name``.
@@ -332,7 +332,7 @@ In which ``hostname`` contains the hostname of the git service URL. In this exam
 
 .. note::
 
-  This ``defaults.ini`` section must only be used for analyzing a repository on the local file system. If the domain name has already been supported in other services, it doesn't need to be defined again here.
+  This ``defaults.ini`` section must only be used for analyzing a repository on the local file system. If the hostname has already been supported in other services, it doesn't need to be defined again here.
 
 Assume that the dir tree at the current working directory has the following structure:
 
--- a/docs/source/pages/using.rst
+++ b/docs/source/pages/using.rst
@@ -51,7 +51,7 @@ Take the same example as above, to disable analyzing `micronaut-core <https://gi
 
   ./run_macaron.sh analyze -rp https://github.com/micronaut-projects/micronaut-core -b 4.0.x -d 82d115b4901d10226552ac67b0a10978cd5bc603 --skip-deps
 
-.. note:: By default, Macaron would generate report files into the ``output`` directory in the current workspace. To understand the structure of this directory please see :ref:`Output Files Guide <output_files_guide>`.
+.. note:: By default, Macaron would generate report files into the ``output`` directory in the current working directory. To understand the structure of this directory please see :ref:`Output Files Guide <output_files_guide>`.
 
 With the example above, the generated output reports can be seen here:
 
@@ -321,7 +321,7 @@ Analyzing a repository whose git service is not supported by Macaron
 If the repository remote URL is from an unknown git service (see :ref:`Git Services <supported_git_services>` for a list of supported git services in Macaron), Macaron won't recognize it when analyzing the repository.
 
 You would need to tell Macaron about that git service through the ``defaults.ini`` config.
-For example, let's say you want to analyze a repository hosted at ``https://git.example.com/foo/target``. First, you need to create a ``defaults.ini`` file in the current workspace with the following content:
+For example, let's say you want to analyze a repository hosted at ``https://git.example.com/foo/target``. First, you need to create a ``defaults.ini`` file in the current working directory with the following content:
 
 .. code-block:: ini
 
@@ -334,7 +334,7 @@ In which ``hostname`` contains the domain of the git service URL. In this exampl
 
   This ``defaults.ini`` section must only be used for analyzing a locally cloned repository. If the domain name has already been supported in other services, it doesn't need to be defined again here.
 
-Assume that the dir tree at the current workspace has the following structure:
+Assume that the dir tree at the current working directory has the following structure:
 
 .. code-block:: shell
 
@@ -360,7 +360,7 @@ Analyzing a local repository with supported git service
 
 If the local repository you want to analyze has a remote origin hosted on a supported git service, you can run the analysis directly without having to prepare ``defaults.ini`` as above.
 
-Assume that the dir tree at the current workspace has the following structure:
+Assume that the dir tree at the current working directory has the following structure:
 
 .. code-block:: shell
 
--- a/docs/source/pages/using.rst
+++ b/docs/source/pages/using.rst
@@ -300,9 +300,9 @@ An example configuration file for utilising this feature:
 
 
 
--------------------------------------
-Analyzing a locally cloned repository
--------------------------------------
+-----------------------------------------------
+Analyzing a repository on the local file system
+-----------------------------------------------
 
 .. warning::
   During the analysis, Macaron can check out different commits, which can reset the index and working tree of the repository.
@@ -312,7 +312,7 @@ Analyzing a locally cloned repository
 .. note::
   We assume that the ``origin`` remote exists in the cloned repository and checkout the relevant commits from ``origin`` only.
 
-If you have a local repository that you want to analyze, Macaron also supports running the analysis against a local repository.
+Macaron supports analyzing a repository on the local file system.
 
 ''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
 Analyzing a repository whose git service is not supported by Macaron
@@ -328,11 +328,11 @@ For example, let's say you want to analyze a repository hosted at ``https://git.
   [git_service.local_repo]
   hostname = git.example.com
 
-In which ``hostname`` contains the domain of the git service URL. In this example it's ``git.example.com``.
+In which ``hostname`` contains the hostname of the git service URL. In this example it is ``git.example.com``.
 
 .. note::
 
-  This ``defaults.ini`` section must only be used for analyzing a locally cloned repository. If the domain name has already been supported in other services, it doesn't need to be defined again here.
+  This ``defaults.ini`` section must only be used for analyzing a repository on the local file system. If the domain name has already been supported in other services, it doesn't need to be defined again here.
 
 Assume that the dir tree at the current working directory has the following structure:
 
--- a/docs/source/pages/using.rst
+++ b/docs/source/pages/using.rst
@@ -314,7 +314,53 @@ Analyzing a locally cloned repository
 
 If you have a local repository that you want to analyze, Macaron also supports running the analysis against a local repository.
 
-Assume that the dir tree at the local repository has the following components:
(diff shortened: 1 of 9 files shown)


docs/source/pages/using.rst
```diff
@@ -314,7 +314,53 @@ Analyzing a locally cloned repository
...
 
-Assume that the dir tree at the local repository has the following components:
+''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+Analyzing a repository whose git service is not supported by Macaron
+''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+If the repository remote URL is from an unknown git service (see :ref:`Git Services <supported_git_services>` for a list of supported git services in Macaron), Macaron won't recognize it when analyzing the repository.
+
+You would need to tell Macaron about that git service through the ``defaults.ini`` config.
+For example, let's say you want to analyze the Bitbucket repository at ``https://bitbucket.org/snakeyaml/snakeyaml``. First, you need to create a ``defaults.ini`` file in the current workspace with the following content:
```
nicallen, line 324: I'm not sure using  `bitbucket.org` as the example is a good idea, since this is a well-known public git service that in principle we should support at some point, rendering this example meaningless (and the same goes for any other real public git service one might choose). Using a nonexistent hostname as the example (e.g. `git.example.com`) would probably be better, at the cost of making the example not concretely executable.

docs/source/pages/using.rst
```diff
@@ -326,13 +372,13 @@ We can run Macaron against the local repository at ``target`` by using this comm
...
 .. code-block:: shell
 
-  ./run_macaron.sh -lr path/to/boo/foo analyze -rp target <rest_of_args>
+  ./run_macaron.sh --local-repos-path ./boo/foo analyze -rp target <rest_of_args>
 
-With ``rest_of_args`` being the arguments to the ``analyze`` command (e.g. ``-b``, ``-d`` or ``--skip-deps`` similar to two previous examples)
+With ``rest_of_args`` being the arguments to the ``analyze`` command (e.g. ``-b``, ``-d`` or ``--skip-deps`` similar to two previous examples).
 
-The ``-lr`` flag configure Macaron to looks into ``path/to/boo/foo`` for local repositories. For more information, please see :ref:`Command Line Usage <cli-usage>`.
+The ``-lr`` flag tells Macaron to look into ``path/to/boo/foo`` for local repositories. For more information, please see :ref:`Command Line Usage <cli-usage>`.
```
nicallen, line 379: `path/to/boo/foo` -> `./boo/foo` to be consistent with the change above.

docs/source/pages/using.rst
tromai: I agree. Using `bitbucket.org` is not the best way to demonstrate an unknown generic git service.

docs/source/pages/using.rst
tromai: Fixed in https://github.com/oracle/macaron/pull/694/commits/a05ef37bd7f4509e6d36378eae137153eb423cbe

docs/source/pages/using.rst
```diff
@@ -314,7 +314,53 @@ Analyzing a locally cloned repository
...
+
+The ``-lr`` flag tells Macaron to look into ``./boo/foo`` for local repositories. For more information, please see :ref:`Command Line Usage <cli-usage>`.
+
+.. note:: If ``-lr`` is not provided, Macaron will looks inside ``<current_working_directory>/output/git_repos/local_repos/`` whenever you provide a local path to ``-rp``.
+
+'''''''''''''''''''''''''''''''''''''''''''''''''''''''
+Analyzing a local repository with supported git service
+'''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+If the local repository you want to analyze has remote origin hosted on a supported git service, you can run the analysis directly without having to prepare ``defaults.ini`` as above.
```
benmss, line 361:
```suggestion
If the local repository you want to analyze has a remote origin hosted on a supported git service, you can run the analysis directly without having to prepare ``defaults.ini`` as above.
```

tromai: Fixed in https://github.com/oracle/macaron/pull/694/commits/559041db4f8db569f9098ab6df056c5b3446d999

docs/source/pages/using.rst
```diff
@@ -314,7 +314,53 @@ Analyzing a locally cloned repository
...
+
+We can run Macaron against the local repository at ``target`` by using this command:
+
+.. code-block:: shell
+
+  ./run_macaron.sh --local-repos-path ./boo/foo --defaults-path ./defaults.ini analyze -rp target <rest_of_args>
+
+With ``rest_of_args`` being the arguments to the ``analyze`` command (e.g. ``-b``, ``-d`` or ``--skip-deps`` similar to two previous examples).
+
+The ``-lr`` flag tells Macaron to look into ``./boo/foo`` for local repositories. For more information, please see :ref:`Command Line Usage <cli-usage>`.
```
nathanwn, line 353: The example command uses `--local-repos-path` but `-lr` is mentioned instead here.
It is probably best to mention both the long and short versions.
```suggestion
The ``--local-repos-path/-lr`` flag tells Macaron to look into ``./boo/foo`` for local repositories. For more information, please see :ref:`Command Line Usage <cli-usage>`.
```

docs/source/pages/using.rst
```diff
@@ -314,7 +314,53 @@ Analyzing a locally cloned repository
...
 
-Assume that the dir tree at the local repository has the following components:
+''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+Analyzing a repository whose git service is not supported by Macaron
+''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+If the repository remote URL is from an unknown git service (see :ref:`Git Services <supported_git_services>` for a list of supported git services in Macaron), Macaron won't recognize it when analyzing the repository.
+
+You would need to tell Macaron about that git service through the ``defaults.ini`` config.
+For example, let's say you want to analyze a repository hosted at ``https://git.example.com/foo/target``. First, you need to create a ``defaults.ini`` file in the current workspace with the following content:
```
nathanwn, line 324: I think the word `current workspace` here may be confusing. It is probably better to say `current working directory`.

docs/source/pages/using.rst
```diff
@@ -314,7 +314,53 @@ Analyzing a locally cloned repository
...
+
+.. code-block:: ini
+
+  [git_service.local_repo]
+  hostname = git.example.com
+
+In which ``hostname`` contains the domain of the git service URL. In this example it's ``git.example.com``.
```
nathanwn, line 331:
```suggestion
In which ``hostname`` contains the hostname of the git service URL. In this example, it is ``git.example.com``.
```

The term "domain" used here may not be the most accurate. We had something similar to this before. See #416.

docs/source/pages/using.rst
```diff
@@ -314,7 +314,53 @@ Analyzing a locally cloned repository
...
+
+.. note::
+
+  This ``defaults.ini`` section must only be used for analyzing a locally cloned repository. If the domain name has already been supported in other services, it doesn't need to be defined again here.
```
nathanwn, line 335: A bit nitpicking, but I think this is more general:
```suggestion
  This ``defaults.ini`` section must only be used for analyzing a repository on the local file system. If the domain name has already been supported in other services, it doesn't need to be defined again here.
```
A local repository may just have been created on the local machine and has not been pushed to any remote.

src/macaron/slsa_analyzer/git_service/local_repo_git_service.py
```diff
@@ -0,0 +1,42 @@
...
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class LocalRepoGitService(BaseGitService):
+    """This class contains the spec of the local repo git service."""
+
+    def __init__(self) -> None:
+        """Initialize instance."""
+        super().__init__("generic")
```
nathanwn, line 22: I assume the name `"generic"` is used here because this class was initially named `GenericGitService`? Should it be changed to something more similar to the class name, e.g. `"local_repo"`?

tromai: That's true. It should be `local_repo` with the latest update of this PR.

docs/source/pages/using.rst
tromai: Fixed in https://github.com/oracle/macaron/pull/694/commits/7c360b00111c1de6c824b8f90598c9c2551732aa

docs/source/pages/using.rst
tromai: Fixed in https://github.com/oracle/macaron/pull/694/commits/52e50f7d6799676ff8095cbc3c5082a158c2da8a

src/macaron/slsa_analyzer/git_service/local_repo_git_service.py
tromai: Fixed in https://github.com/oracle/macaron/pull/694/commits/d80e19b7d7242c82aea78b23fa846e1b0ad08548

docs/source/pages/using.rst
tromai: Fixed in https://github.com/oracle/macaron/pull/694/commits/ace778c0b7ee12da54461140ddee63ff2ebb1f13

docs/source/pages/using.rst
tromai: Fixed in https://github.com/oracle/macaron/pull/694/commits/ace778c0b7ee12da54461140ddee63ff2ebb1f13 and https://github.com/oracle/macaron/pull/694/commits/a96ab2bbd8d6ff29f32e30b77619e8326cc31fa9

benmss (approved):
Looks good to me.

nathanwn (approved):

nicallen (approved):

PR #696, opened 2024-04-09
chore: disable redirect resolution in tests

### LINKED ISSUE #686: Repo finder configuration does not disable the redirect feature in tests
The [`test_convert_components_to_artifacts` test](https://github.com/oracle/macaron/blob/8ba2e388f5c2410252b6562183f83d682b71f205/tests/dependency_analyzer/cyclonedx/test_cyclonedx.py#L64) still tries to make network calls when URLs are redirected, and therefore running the test in offline mode is not possible.

benmss:
Updates the configuration file for the cyclonedx unit test that was previously performing remote calls due to containing URLs that required following redirection to resolve. The related test resource file is reverted to how it looked before the redirection resolution feature was added.

Closes #686


tromai (approved):
LGTM. Thanks for the changes.

PR #698, opened 2024-04-10
chore: add provenance file handling into run_macaron.sh

### LINKED ISSUE #697: Update run_macaron.sh to mount the provenance file into the Docker container file system.
#654 added `--provenance-file` as an input option. 
However, because we didn't handle mounting this file into the Macaron Docker container when the user run `run_macaron.sh`, the file content will not be available for Macaron to read.
```
./run_macaron.sh analyze --provenance-file prov.json <rest_of_args>
```
This error is shown
```
...
2024-04-09 05:19:37,692 [ERROR] Error while loading the input provenance file: Cannot deserialize the file content as JSON.
```

This issue tracks the fix for mounting the provenance file into the Docker container.

tromai:
Closes #697


scripts/release_scripts/run_macaron.sh
```diff
@@ -439,6 +443,17 @@ if [[ -n "${arg_prov_exp:-}" ]]; then
     fi
 fi
 
+# Determine the provenance expectation path to be mounted into ${MACARON_WORKSPACE}/prov_files/${pf_name} where pf_name is a file name.
+if [[ -n "${arg_prov_file:-}" ]]; then
+    prov_file_path="${arg_prov_file}"
+    assert_path_exists "${prov_file_path}" "-pf/--provenance-expectation"
```
nathanwn, line 449: This `assert_path_exists` check is not necessary. It exists only in the `--provenance_expectation` case because that argument expects either a file or a directory.
In this case of `--provenance-file`, it is always a file, which means we should use the `mount_file` function. This function already encapsulates path checking, so there is no need to do the `assert_path_exists` here.
```suggestion
```

tromai: Removed in https://github.com/oracle/macaron/pull/698/commits/74f7bcf00de4f0f7eba0ae18de3a4be2790102a0

nathanwn (approved):
Thanks. LGTM.

behnazh-w:
@tromai Now that https://github.com/oracle/macaron/pull/685 is merged, can you please add a test to ingest the witness provenance and check the results in both `integration_tests_docker.sh` and `integration_tests.sh`?

scripts/release_scripts/run_macaron.sh
```diff
@@ -439,6 +443,16 @@ if [[ -n "${arg_prov_exp:-}" ]]; then
     fi
 fi
 
+# Determine the provenance expectation path to be mounted into ${MACARON_WORKSPACE}/prov_files/${pf_name} where pf_name is a file name.
```
behnazh-w, line 446:
```suggestion
# Mount the provenance file into ${MACARON_WORKSPACE}/prov_files/${pf_name} where pf_name is a file name.
```

scripts/release_scripts/run_macaron.sh
```diff
@@ -446,7 +446,6 @@ fi
 # Determine the provenance expectation path to be mounted into ${MACARON_WORKSPACE}/prov_files/${pf_name} where pf_name is a file name.
 if [[ -n "${arg_prov_file:-}" ]]; then
     prov_file_path="${arg_prov_file}"
-    assert_path_exists "${prov_file_path}" "-pf/--provenance-expectation"
```
behnazh-w: Why is this check removed?

scripts/release_scripts/run_macaron.sh
tromai: Fixed in https://github.com/oracle/macaron/pull/698/commits/bf2f4078e4d4c6f85e29ffb1dd7cb962b5f0ec6c

nathanwn:
Not a comment for this PR in particular, but a general comment: We should probably rethink where the expected result files should be stored. To me, it makes more sense to store all expected files for a particular test case in the same directory instead of scattering it through multiple different directories, making it hard to understand which expected file belongs to which test.

tromai:
I will remove the `--digest` CLI input from the added integration test after https://github.com/oracle/macaron/pull/708 is merged.

tromai:
@behnazh-w @nathanwn I have updated the integration test case:
1. Remove the provided digest as input through `--digest`. With this move, we are hoping for the digest to be obtained from the provided witness provenance, combined with the repository path provided as input to form the main analysis target. (see https://github.com/oracle/macaron/pull/698/commits/60d0cab8966d6b7e02105ba79a6594b0d8d0b65f).
2. Update the policy file to enforce more check results in the test case (see https://github.com/oracle/macaron/pull/698/commits/b1bc20c41fea8aaa4e92d2a7fac49ac0f02a9f96)

For the `1.` change, however, just removing the `--digest` won't make Macaron behave as we expected. This is because for this test case scenario, these values are provided as input:
- The witness provenance that only contains the commit hash
- The path to the repository on local file system.
- The PURL of the main target analysis. 
This combination will be handled inside this block https://github.com/oracle/macaron/blob/e214326697aee6e31f0c9e406b11344c8619581e/src/macaron/slsa_analyzer/analyzer.py#L702-L708

However, its logic doesn't extract the commit hash from provenance, which leave `Analyzer.AnalysisTarget.digest` as empty. Therefore, when preparing the repository for the analysis, Macaron will try to find the commit from the given PURL here 
https://github.com/oracle/macaron/blob/e214326697aee6e31f0c9e406b11344c8619581e/src/macaron/slsa_analyzer/analyzer.py#L819-L825

which will fail because the target repository doesn't have any release on GitHub (see https://github.com/behnazh-w/example-maven-app/releases). 
I have proposed a fix (see https://github.com/oracle/macaron/pull/698/commits/82e241b7272731206cc16c1fb6234f14f56bdc33) which make the test case work again. However, I am not sure that this fix is the best solution, and whether it's out of scope of this PR. There are other solutions:
- Create releases on the `example-maven-project` repository so that the commit finder can work. 
- Still using `--digest` to provide digest for the integration test case and we don't need to fix anything.

nicallen:
> @behnazh-w @nathanwn I have updated the integration test case:
> ... I have proposed a fix (see [82e241b](https://github.com/oracle/macaron/commit/82e241b7272731206cc16c1fb6234f14f56bdc33)) which make the test case work again. However, I am not sure that this fix is the best solution, and whether it's out of scope of this PR. ...

I think the proposed fix is the correct behaviour, but I think that commit should be separated into its own PR and merged prior to finishing this one (since it isn't really related to this change, more a continuation of #708 that happens to be a prerequisite for this one).

nathanwn:
> I have proposed a fix (see [82e241b](https://github.com/oracle/macaron/commit/82e241b7272731206cc16c1fb6234f14f56bdc33)) which make the test case work again.

In your fix, if I understand it right, the user-provided digest takes precedence over the commit in the provenance.
If this is as expected, then the fix looks OK to me.
One thing I realized from the fix: this final `case` branch here is not necessary and should be removed (duplicating the first `case` branch)

https://github.com/oracle/macaron/blob/697-update-run_macaronsh-to-mount-the-provenance-file-into-the-docker-container-file-system/src/macaron/slsa_analyzer/analyzer.py#L725-L729.

tromai:
Thanks everyone! I will bring the fix commit to a difference Pull request shortly.
> In your fix, if I understand it right, the user-provided digest takes precedence over the commit in the provenance.
> If this is as expected, then the fix looks OK to me.

This is True, the digest provided from the user takes precedence.

> One thing I realized from the fix: this final case branch here is not necessary and should be removed (duplicating the first case branch)
Yes this is True. I meant to put it there originally as a "default" case because mypy doesn't recognize that we have covered all possible cases.

nathanwn:
I now remember why the last `case` branch is necessary. It is related to [the issue of `mypy` cannot type-narrow tuples correctly](https://github.com/python/mypy/pull/16905) (seems to be recently fixed but not yet released).
We can keep it around I guess.

tromai:
@nicallen @nathanwn Thanks for your feedback. I have created a Pull Request for fixing the issue here - https://github.com/oracle/macaron/pull/711. After that PR is merged, I will rebase and fix any conflicts here.

nicallen (approved):

PR #699, opened 2024-04-11
fix: make build script check fail when no repo is found

nicallen:
Change build script check to fail on skip instead of pass, which became incorrect when the check dependency relationship was changed so that the build script check is skipped when the version control system check fails (and thus should likewise fail as no source or build script is available), rather than being skipped when the stronger build checks pass (which previously meant that build script would then pass by default).

Added an integration check to confirm that all checks fail for a purl referring to a nonexistent artifact.


behnazh-w (approved):

tromai (approved):
LGTM. Thanks!

PR #701, opened 2024-04-12
fix: fix a compatibility issue in run_macaron.sh for macOS

behnazh-w:
macOS does not support the `--parents` option in `mkdir`. This PR fixes the `run_macaron.sh` script by using the `-p` option instead to be compatible with macOS.


tromai (approved):
LGTM.

PR #704, opened 2024-04-15
feat: add checks to determine if repo and commit came from provenance

### LINKED ISSUE #677: Differentiate when a repo commit is found from the provenance vs tags
It is very useful to differentiate when a repository and commit is found from the provenance vs registry metadata and repository tags because the former is more trustworthy. We can report that as a new check result.

benmss:
This PR adds a new check that succeeds if the repository URL and commit of the analysis target match those that can be extracted from the provenance. If the repository or provenance do not exist, or do not contain the needed information, or are not identical, this check will fail.

Closes #677

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -82,10 +82,10 @@ the requirements that are currently supported by Macaron.
    * - 3
      - **Provenance expectation** - Check if the provenance meets an expectation.
      - The user can provide an expectation for the provenance as a CUE policy, which will be compared against the SLSA provenance.
-   * - 4
+   * - 3
      - **Provenance derived repo** - Check if the analysis target's repository matches the repository in the provenance.
      - If there is no provenance, this check will fail.
-   * - 4
+   * - 3
      - **Provenance derived commit** - Check if the analysis target's commit matches the commit in the provenance.
      - If there is no commit, this check will fail.
 
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -82,6 +82,12 @@ the requirements that are currently supported by Macaron.
    * - 3
      - **Provenance expectation** - Check if the provenance meets an expectation.
      - The user can provide an expectation for the provenance as a CUE policy, which will be compared against the SLSA provenance.
+   * - 4
+     - **Provenance derived repo** - Check if the analysis target's repository matches the repository in the provenance.
+     - If there is no provenance, this check will fail.
+   * - 4
+     - **Provenance derived commit** - Check if the analysis target's commit matches the commit in the provenance.
+     - If there is no commit, this check will fail.
 
 ----------------------
 How does Macaron work?
--- a/src/macaron/slsa_analyzer/analyze_context.py
+++ b/src/macaron/slsa_analyzer/analyze_context.py
@@ -46,6 +46,10 @@ class ChecksOutputs(TypedDict):
     """The package registries for the target software component."""
     provenance: InTotoPayload | None
     """The provenance payload for the target software component."""
+    provenance_repo: str | None
+    """The repository URL extracted from provenance, if applicable."""
+    provenance_commit: str | None
+    """The commit digest extracted from provenance, if applicable."""
 
 
 class AnalyzeContext:
@@ -97,6 +101,8 @@ class AnalyzeContext:
             is_inferred_prov=True,
             expectation=None,
             provenance=None,
+            provenance_repo=None,
+            provenance_commit=None,
         )
 
     @property
--- a/src/macaron/slsa_analyzer/analyze_context.py
+++ b/src/macaron/slsa_analyzer/analyze_context.py
@@ -46,9 +46,9 @@ class ChecksOutputs(TypedDict):
     """The package registries for the target software component."""
     provenance: InTotoPayload | None
     """The provenance payload for the target software component."""
-    provenance_repo: str | None
+    provenance_repo_url: str | None
     """The repository URL extracted from provenance, if applicable."""
-    provenance_commit: str | None
+    provenance_commit_digest: str | None
     """The commit digest extracted from provenance, if applicable."""
 
 
@@ -101,8 +101,8 @@ class AnalyzeContext:
             is_inferred_prov=True,
             expectation=None,
             provenance=None,
-            provenance_repo=None,
-            provenance_commit=None,
+            provenance_repo_url=None,
+            provenance_commit_digest=None,
         )
 
     @property
--- a/src/macaron/slsa_analyzer/checks/provenance_commit_check.py
+++ b/src/macaron/slsa_analyzer/checks/provenance_commit_check.py
@@ -1,0 +1,79 @@
+# Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""This module adds a check that determines whether the repository URL came from provenance."""
+import logging
+
+from sqlalchemy import ForeignKey, String
+from sqlalchemy.orm import Mapped, mapped_column
+
+from macaron.database.table_definitions import CheckFacts
+from macaron.slsa_analyzer.analyze_context import AnalyzeContext
+from macaron.slsa_analyzer.checks.base_check import BaseCheck
+from macaron.slsa_analyzer.checks.check_result import CheckResultData, CheckResultType, Confidence, JustificationType
+from macaron.slsa_analyzer.registry import registry
+from macaron.slsa_analyzer.slsa_req import ReqName
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class ProvenanceDerivedCommitFacts(CheckFacts):
+    """The ORM mapping for justifications in the commit from provenance check."""
+
+    __tablename__ = "_provenance_derived_commit_check"
+
+    #: The primary key.
+    id: Mapped[int] = mapped_column(ForeignKey("_check_facts.id"), primary_key=True)  # noqa: A003
+
+    #: The state of the commit.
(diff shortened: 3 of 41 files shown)


behnazh-w:
@benmss The scope of this PR should be `feat`.

src/macaron/slsa_analyzer/checks/provenance_commit_check.py
```diff
@@ -0,0 +1,135 @@
...
+        Returns
+        -------
+        CheckResultData
+            The result of the check.
+        """
+        provenance = ctx.dynamic_data["provenance"]
+        if not provenance:
+            return CheckResultData(
+                result_tables=[
+                    ProvenanceDerivedCommitFacts(commit_digest="No provenance found.", confidence=Confidence.HIGH)
```
nicallen, line 72: Why is this field called `commit_digest`? It appears to be a explanatory message instead.

src/macaron/slsa_analyzer/checks/provenance_repo_check.py
```diff
@@ -0,0 +1,136 @@
...
+        Returns
+        -------
+        CheckResultData
+            The result of the check.
+        """
+        provenance = ctx.dynamic_data["provenance"]
+        if not provenance:
+            return CheckResultData(
+                result_tables=[
+                    ProvenanceDerivedRepoFacts(repository_url="No provenance found.", confidence=Confidence.HIGH)
```
nicallen, line 72: Similar to `commit_digest`, why is this called `repository_url` when it is a message instead?

src/macaron/slsa_analyzer/checks/provenance_commit_check.py
```diff
@@ -0,0 +1,135 @@
...
+        ----------
+        ctx : AnalyzeContext
+            The object containing processed data for the target repo.
+
+        Returns
+        -------
+        CheckResultData
+            The result of the check.
+        """
+        provenance = ctx.dynamic_data["provenance"]
```
behnazh-w, line 68: You should also check if the provenance is inferred or not using `ctx.dynamic_data["is_inferred_prov"]`. If a provenance is not found the build checks populate the provenance and `ctx.dynamic_data["is_inferred_prov"]` and will be set to `True`.

src/macaron/slsa_analyzer/checks/provenance_repo_check.py
```diff
@@ -0,0 +1,136 @@
...
+        ----------
+        ctx : AnalyzeContext
+            The object containing processed data for the target repo.
+
+        Returns
+        -------
+        CheckResultData
+            The result of the check.
+        """
+        provenance = ctx.dynamic_data["provenance"]
```
behnazh-w, line 68: Same comment for checking `ctx.dynamic_data["is_inferred_prov"]`.

src/macaron/slsa_analyzer/checks/provenance_commit_check.py
```diff
@@ -0,0 +1,145 @@
...
+class ProvenanceDerivedCommitFacts(CheckFacts):
+    """The ORM mapping for justifications in the commit from provenance check."""
+
+    __tablename__ = "_provenance_derived_commit_check"
+
+    #: The primary key.
+    id: Mapped[int] = mapped_column(ForeignKey("_check_facts.id"), primary_key=True)  # noqa: A003
+
+    #: The reason this test failed to pass, when applicable.
+    error: Mapped[str] = mapped_column(String, nullable=True, info={"justification": JustificationType.TEXT})
```
tromai, line 31: I believe you have switched to this `error` attribute to store the explanatory message from the check (as discussed in this conversation https://github.com/oracle/macaron/pull/704#discussion_r1593378468). Should we remove the `commit_digest` attribute as it's not being used anymore :thinking: ? This same comment applies for : 
https://github.com/oracle/macaron/blob/50b6e2b708b91518be965d60416c97acc0f996bd/src/macaron/slsa_analyzer/checks/provenance_repo_check.py#L22

benmss: The `commit_digest` attribute is still being used in the `PASSED` case.

src/macaron/slsa_analyzer/checks/provenance_commit_check.py
```diff
@@ -0,0 +1,145 @@
...
+            )
+
+        provenance = ctx.dynamic_data["provenance"]
+        if not provenance:
+            return CheckResultData(
+                result_tables=[ProvenanceDerivedCommitFacts(error="No provenance found.", confidence=Confidence.HIGH)],
+                result_type=CheckResultType.FAILED,
+            )
+
+        if not ctx.component or not ctx.component.repository:
```
tromai, line 87: It seems to me that the type of `component`, `repository` and `commit_sha` are not nullable. I wonder if this is True as I have seen it being used as if it was non nullable here - https://github.com/oracle/macaron/blob/a57d282e3e0dec541464a45f0f8d1a9307512d91/src/macaron/slsa_analyzer/checks/build_service_check.py#L132
Therefore, I wonder if we need to check for None value here and down below for `commit_sha`.

src/macaron/slsa_analyzer/checks/provenance_repo_check.py
```diff
@@ -0,0 +1,146 @@
...
+                result_tables=[
+                    ProvenanceDerivedRepoFacts(
+                        error="The Analysis Context did not contain a Component or Repository",
+                        confidence=Confidence.HIGH,
+                    )
+                ],
+                result_type=CheckResultType.FAILED,
+            )
+
+        current_repository = ctx.component.repository.remote_path
```
tromai, line 98: As indicated here - https://github.com/oracle/macaron/blob/5e7ccdc68fb9f37345a32a2373fa0a1a67a0ce0b/src/macaron/database/table_definitions.py#L290 the field of `remote_path` is not nullable. Should we move the check for None value here too? :thinking:

tromai (approved):
Thanks for the change. LGTM.

tests/e2e/expected_results/purl/npm/semver/semver.json
```diff
@@ -89,62 +93,14 @@
                         }
                     }
                 ],
-                "npm Registry": [
```
behnazh-w, line 92: Any idea why the provenance is removed from the JSON report? It might not be related to this PR but I wonder what caused it.

src/macaron/slsa_analyzer/analyze_context.py
```diff
@@ -46,6 +46,10 @@ class ChecksOutputs(TypedDict):
     """The package registries for the target software component."""
     provenance: InTotoPayload | None
     """The provenance payload for the target software component."""
+    provenance_repo: str | None
```
behnazh-w, line 49:
```suggestion
   provenance_repo_url: str | None
```

tests/e2e/expected_results/purl/npm/semver/semver.json
benmss: Looks like it is caused by the `provenance_available` check modification that was done to allow it to return early if provenance was already found. If that happens the release assets are not collected, so this will report as being empty.

behnazh-w: Thanks for looking into the issue. I guess once we move all the provenance finding code from the check to the new `ProvenanceFinder.find_provenance`  method, this issue will be resolved.

behnazh-w:
@benmss Please update the list of checks in `docs/source/index.rst` at `Current checks in Macaron` and run `make docs-api`.

docs/source/index.rst
```diff
@@ -82,6 +82,12 @@ the requirements that are currently supported by Macaron.
    * - 3
      - **Provenance expectation** - Check if the provenance meets an expectation.
      - The user can provide an expectation for the provenance as a CUE policy, which will be compared against the SLSA provenance.
+   * - 4
+     - **Provenance derived repo** - Check if the analysis target's repository matches the repository in the provenance.
+     - If there is no provenance, this check will fail.
+   * - 4
+     - **Provenance derived commit** - Check if the analysis target's commit matches the commit in the provenance.
+     - If there is no commit, this check will fail.
```
behnazh-w, line 90: It's better to keep these levels to SLSA 3 similar to Provenance expectations to be consistent.

src/macaron/slsa_analyzer/checks/provenance_commit_check.py
```diff
@@ -0,0 +1,88 @@
...
+
+class ProvenanceDerivedCommitCheck(BaseCheck):
+    """This check tries to extract the repo from the provenance and compare it to what is in the context."""
+
+    def __init__(self) -> None:
+        """Initialize instance."""
+        check_id = "mcn_provenance_derived_commit_1"
+        description = "Check whether the commit came from provenance."
+        depends_on: list[tuple[str, CheckResultType]] = []
+        eval_reqs = [ReqName.SECURITY]
```
behnazh-w, line 44: I think `ReqName.EXPECTATION` would be more suitable?

src/macaron/slsa_analyzer/checks/provenance_repo_check.py
```diff
@@ -0,0 +1,88 @@
...
+
+class ProvenanceDerivedRepoCheck(BaseCheck):
+    """This check tries to extract the repo from the provenance and compare it to what is in the context."""
+
+    def __init__(self) -> None:
+        """Initialize instance."""
+        check_id = "mcn_provenance_derived_repo_1"
+        description = "Check whether the repo came from provenance."
+        depends_on: list[tuple[str, CheckResultType]] = []
+        eval_reqs = [ReqName.SECURITY]
```
behnazh-w, line 44: I think `ReqName.EXPECTATION` would be more suitable?

behnazh-w (approved):

Issue #707, opened 2024-04-15
Allow repo and commit extraction from provenance to be independent


benmss:
In the current implementation, the provenance extractor will only succeed if it can retrieve both the repo URL and the commit. In some cases it may make sense to allow one of these to be absent, meaning extraction could conclude with a partial success: A repo URL, or a commit.

PR #708, opened 2024-04-16
chore: allow independent extraction of repo and commit from provenance

benmss:
This PR allows the provenance extractor to return partial results when either the repo URL or commit cannot be found. Tests have been updated to ensure this change works as intended while retaining the previous non-overlapping functionality, i.e. returning both, or neither on critical errors.

This change does not apply to SLSA v1 as these use the repo URL as a key to access the correct commit. To support this in SLSA v1 would require the repo URL be provided through some other means before provenance extraction begins.

Updates #707

### DIFF
--- a/src/macaron/repo_finder/provenance_extractor.py
+++ b/src/macaron/repo_finder/provenance_extractor.py
@@ -4,7 +4,7 @@
 """This module contains methods for extracting repository and commit metadata from provenance files."""
 import logging
 
-from macaron.errors import JsonError, ProvenanceError
+from macaron.errors import ProvenanceError
 from macaron.json_tools import json_extract
 from macaron.slsa_analyzer.provenance.intoto import InTotoPayload, InTotoV1Payload, InTotoV01Payload
 from macaron.util import JsonType
@@ -17,7 +17,7 @@ SLSA_V02_DIGEST_SET_GIT_ALGORITHMS = ["sha1"]
 SLSA_V1_DIGEST_SET_GIT_ALGORITHMS = ["sha1", "gitCommit"]
 
 
-def extract_repo_and_commit_from_provenance(payload: InTotoPayload) -> tuple[str, str]:
+def extract_repo_and_commit_from_provenance(payload: InTotoPayload) -> tuple[str | None, str | None]:
     """Extract the repository and commit metadata from the passed provenance payload.
 
     Parameters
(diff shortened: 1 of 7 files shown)


src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -53,7 +53,7 @@ def extract_repo_and_commit_from_provenance(payload: InTotoPayload) -> tuple[str
         logger.debug(error)
```
tromai, line 53: Because we have changed the implementation of `_extract_from_*` functions, there are 2 duplicated logic within the public `extract_repo_and_comit_from_provenance`, which are already handled in `_extract_from_*` functions:
1. catching `JsonError` https://github.com/oracle/macaron/blob/6cae4fc5c32f3ceb214698312abc6b5f0927ee19/src/macaron/repo_finder/provenance_extractor.py#L52-L54
2. raising `ProvenanceError` when both repo and commit are not available:
https://github.com/oracle/macaron/blob/6cae4fc5c32f3ceb214698312abc6b5f0927ee19/src/macaron/repo_finder/provenance_extractor.py#L56-L62

I think it's okay to remove catching `JsonError` from `extract_repo_and_comit_from_provenance` function as we are not treating missing repo url and/or commit as a critical error anymore (for example, a witness provenance can have commit hash but no repo).

At the moment, I can see some differences in handling missing repo url and/or commit hash between `_extract_from_*` functions:
- Provenance has repo and no commit hash: 
  -  [`_extract_from_slsa_v1`](https://github.com/oracle/macaron/blob/6cae4fc5c32f3ceb214698312abc6b5f0927ee19/src/macaron/repo_finder/provenance_extractor.py#L139) raises `ProvenanceError`.
  - other `_extract_from_*` functions will return `(<repo_url>, <empty_string_for_commit_hash>)`
- Provenance has repo and commit hash: all of them return `(<repo_url>, <commit_hash>)`
- Provenance has no repo and commit hash: 
  - [`_extract_from_slsa_v1`](https://github.com/oracle/macaron/blob/6cae4fc5c32f3ceb214698312abc6b5f0927ee19/src/macaron/repo_finder/provenance_extractor.py#L139) raises `ProvenanceError`. This is because for slsa provenance v1 we need the repo url for obtaining the commit hash. 
  - other `_extract_from_*` functions will return `(<empty_string_for_repo_url>, <commit_hash>)`
- Provenance has no repo and no commit hash: all of them will raise `ProvenanceError`.

For me, even though most of the differences are implementation details, they could make it difficult for us to maintain because we have to take care of the differences in behaviors within `extract_repo_and_comit_from_provenance` because:
- An empty string is used to represent repo url or commit hash is missing
- A raised `ProvenanceError` is also used to represent repo_url or commit hash is missing 

I think to improve it we must decide on the what to use for indicating:
- A critical error that prevent us from reading provenances for repo url and commit (e.g. the type of provenance is not supported) -> I think we can raise `ProvenanceError`
- We support and can read the content of the provenance but repo url and/or commit hash are not available, as long as it's different from the usage of `ProvenanceError` above.
I don't have any concrete ideas on the implementation details but as long as the issues I mentioned above is resolve, it's fine by me. Please let me know what you think about it.

nicallen (approved):

tromai (dismissed):

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -668,34 +668,33 @@ def to_analysis_target(
...
-                            digest=digest,
-                        )
+                            logger.debug("Failed to extract repo or commit from provenance: %s", error)
+
+                            return Analyzer.AnalysisTarget(
+                                parsed_purl=parsed_purl,
+                                repo_path=repo or "",
+                                branch="",
+                                digest=digest or "",
+                            )
```
tromai, line 687: This return would only run when a `ProvenanceError` is raised. Is this intended? :thinking:

benmss: It was not intended.

tromai (approved):

behnazh-w (approved):

PR #709, opened 2024-04-16
chore: support versions with one alphanumeric part and a prefix in the commit finder

benmss:
This PR includes a fix for the commit finder so that is supports the case where a version has only one part, but also consists of alphanumeric characters. Previously this would cause the only version part to become optional, and therefore allow invalid matches such as matching the entire tag as a prefix.

E.g. 
Version: `2012c`, Tag: `release_2012c`
`2012c` becomes optional allowing `release_2012c` to match as the prefix thereby preventing a proper match because the tag version part is empty.

### DIFF
--- a/src/macaron/repo_finder/commit_finder.py
+++ b/src/macaron/repo_finder/commit_finder.py
@@ -341,8 +341,10 @@ def _build_version_pattern(name: str, version: str) -> tuple[Pattern | None, lis
             # - 3.1.test.2 -> 'test' and '2' become optional.
             has_non_numeric_suffix = True
 
-        if count == len(split) - 1 and has_trailing_zero or has_non_numeric_suffix:
-            # This part will be made optional in the regex, hence the grouping bracket.
+        # This part will be made optional in the regex if it matches the correct requirements.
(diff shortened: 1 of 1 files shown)


src/macaron/repo_finder/commit_finder.py
```diff
@@ -341,8 +341,10 @@ def _build_version_pattern(name: str, version: str) -> tuple[Pattern | None, lis
             # - 3.1.test.2 -> 'test' and '2' become optional.
             has_non_numeric_suffix = True
 
-        if count == len(split) - 1 and has_trailing_zero or has_non_numeric_suffix:
-            # This part will be made optional in the regex, hence the grouping bracket.
+        # This part will be made optional in the regex if it matches the correct requirements.
```
tromai, line 344: nit: I think `correct requirements.` in this comment is a bit ambiguous. I think it's also good to put an example to demonstrate the requirements as the case we are handling here is quite specific.

tromai (approved):
Thanks for the change. LGTM.

nathanwn (approved):

Issue #398, opened 2023-08-04
Retry HTTP requests if they time out


behnazh-w:
The [util](https://github.com/oracle/macaron/blob/fa2622b588546888041cbf3a0c07a67841a98080/src/macaron/util.py) module contains helper functions for sending HTTP requests. We need to allow a retry option to send the requests n times (default could be 3) when there is a timeout. This will allow handling failures in Macaron `analyze` action more deterministically.

jenstroeger:
It looks like this check

https://github.com/oracle/macaron/blob/fa2622b588546888041cbf3a0c07a67841a98080/src/macaron/util.py#L48-L49

is required because the loop here

https://github.com/oracle/macaron/blob/fa2622b588546888041cbf3a0c07a67841a98080/src/macaron/util.py#L42

hammers on the target host which can be interpreted as a [DOS attack](https://en.wikipedia.org/wiki/Denial-of-service_attack) from the user’s machine.[^1]

[^1]: And worst case that gets the user’s host (temporarily) banned from accessing the target host altogether.

It would probably make sense to add a [time.sleep()](https://docs.python.org/3/library/time.html#time.sleep) (maybe with exponential backoff) into the loop.

PR #710, opened 2024-04-18
feat: allow provenance files to be files containing a URL pointing to…

nicallen:
… the actual provenance file which will be transparently downloaded

URL link file format is the Windows .url file format i.e. an ini file containing a "URL" field inside an "InternetShortcut" section. When processing the URL link provenance file, the content will be transparently downloaded if the hostname of the URL is configured as an allowed hostname in the defaults.ini file.

### DIFF
--- a/src/macaron/slsa_analyzer/provenance/loader.py
+++ b/src/macaron/slsa_analyzer/provenance/loader.py
@@ -1,48 +1,75 @@
-# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2024, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """This module contains the loaders for SLSA provenances."""
 
 import base64
+import configparser
 import gzip
+import io
 import json
 import zlib
+from urllib.parse import urlparse
 
+import requests
+
+from macaron.config.defaults import defaults
 from macaron.slsa_analyzer.provenance.intoto import InTotoPayload, validate_intoto_payload
 from macaron.slsa_analyzer.provenance.intoto.errors import LoadIntotoAttestationError, ValidateInTotoPayloadError
 from macaron.util import JsonType
 
 
-def load_provenance_file(filepath: str) -> dict[str, JsonType]:
-    """Load a provenance file and obtain the payload.
+def _try_read_url_link_file(file_content: bytes) -> str | None:
+    parser = configparser.ConfigParser()
+    try:
+        parser.read_file(io.TextIOWrapper(io.BytesIO(file_content), encoding="utf-8"))
+        return parser.get("InternetShortcut", "url", fallback=None)
+    except (configparser.Error, UnicodeDecodeError):
+        return None
 
-    Inside a provenance file is a DSSE envelope containing a base64-encoded
-    provenance JSON payload. See: https://github.com/secure-systems-lab/dsse.
-    If the file is gzipped, it will be transparently decompressed.
 
-    Parameters
-    ----------
-    filepath : str
-        Path to the provenance file.
+def _download_url_file_content(url: str, url_link_hostname_allowlist: list[str]) -> bytes:
+    hostname = urlparse(url).hostname
+    if hostname is None or hostname == "":
+        raise LoadIntotoAttestationError("Cannot resolve URL link file: invalid URL")
+    if hostname not in url_link_hostname_allowlist:
+        raise LoadIntotoAttestationError(
+            "Cannot resolve URL link file: target hostname '" + hostname + "' is not in allowed hostnames."
+        )
 
-    Returns
-    -------
-    dict[str, JsonType]
-        The provenance JSON payload.
+    try:
+        # TODO download size limit?
+        timeout = defaults.getint("downloads", "timeout", fallback=120)
+        response = requests.get(url=url, timeout=timeout)
+        if response.status_code != 200:
+            raise LoadIntotoAttestationError(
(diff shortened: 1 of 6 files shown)


src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -1,48 +1,75 @@
...
+        )
 
-    Returns
-    -------
-    dict[str, JsonType]
-        The provenance JSON payload.
+    try:
+        # TODO download size limit?
+        timeout = defaults.getint("downloads", "timeout", fallback=120)
+        response = requests.get(url=url, timeout=timeout)
```
behnazh-w, line 43: We should use the [send_get_http_raw](https://github.com/oracle/macaron/blob/ba3fcb0c84d6727de343c247a3181908fcd78410/src/macaron/util.py#L59) helper function so that we can improve the implementation in one place as explained in this [issue](https://github.com/oracle/macaron/issues/398).

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -1,48 +1,75 @@
...
 from macaron.slsa_analyzer.provenance.intoto.errors import LoadIntotoAttestationError, ValidateInTotoPayloadError
 from macaron.util import JsonType
 
 
-def load_provenance_file(filepath: str) -> dict[str, JsonType]:
-    """Load a provenance file and obtain the payload.
+def _try_read_url_link_file(file_content: bytes) -> str | None:
+    parser = configparser.ConfigParser()
+    try:
+        parser.read_file(io.TextIOWrapper(io.BytesIO(file_content), encoding="utf-8"))
```
nathanwn, line 25: Hmm, I assume we can use [`parser.read_string()`](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read_string) instead since we have `file_content` as input instead of a file object?
```suggestion
        parser.read_string(file_content.decode())
```
The [`bytes.decode()`](https://docs.python.org/3/library/stdtypes.html#bytes.decode) method converts `bytes` to `str` with `utf-8` as the default encoding.

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -1,48 +1,75 @@
...
     try:
         try:
-            with gzip.open(filepath, mode="rt", encoding="utf-8") as file:
-                provenance = json.load(file)
-        except (gzip.BadGzipFile, EOFError, zlib.error):
-            with open(filepath, encoding="utf-8") as file:
-                provenance = json.load(file)
-    except (OSError, json.JSONDecodeError, TypeError) as error:
+            decompressed_file_content = gzip.decompress(file_content)
+            provenance = json.load(io.TextIOWrapper(io.BytesIO(decompressed_file_content), encoding="utf-8"))
```
nathanwn, line 69: Similarly to [my previous comment](https://github.com/oracle/macaron/pull/710/files#r1570067495), I think we can do this:
```suggestion
            provenance = json.loads(decompressed_file_content.decode())
```

src/macaron/slsa_analyzer/provenance/loader.py
nicallen: Done.

src/macaron/slsa_analyzer/provenance/loader.py
nicallen: I didn't know about `bytes.decode()`, that is much simpler. Done.

src/macaron/slsa_analyzer/provenance/loader.py
nicallen: Done.

nathanwn (approved):
Thanks for the PR.

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -1,48 +1,71 @@
...
         try:
-            with gzip.open(filepath, mode="rt", encoding="utf-8") as file:
-                provenance = json.load(file)
-        except (gzip.BadGzipFile, EOFError, zlib.error):
-            with open(filepath, encoding="utf-8") as file:
-                provenance = json.load(file)
-    except (OSError, json.JSONDecodeError, TypeError) as error:
+            decompressed_file_content = gzip.decompress(file_content)
+            provenance = json.loads(decompressed_file_content.decode())
+        except (gzip.BadGzipFile, EOFError, zlib.error, configparser.NoOptionError):
```
tromai, line 66: I wonder from which statement are we catching `configparser.NoOptionError` from :thinking:  ?

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -1,48 +1,71 @@
...
+            provenance = json.loads(file_content.decode())
+    except (json.JSONDecodeError, TypeError) as error:
```
tromai, line 68: Should we catch `UnicodeDecodeError` here similar to https://github.com/oracle/macaron/blob/3e5e27d363ec76afafc6dd206c8d09aff71551bc/src/macaron/slsa_analyzer/provenance/loader.py#L24 because we are decoding `decompressed_file_content` which is of type `bytes`?

tromai (changes requested):
Thanks for the changes. I have left some comments. Other than those, everything else looks good to me.

src/macaron/slsa_analyzer/provenance/loader.py
nicallen: One that isn't there anymore. I have removed it.

src/macaron/slsa_analyzer/provenance/loader.py
nicallen: Yes we should. Done.

tromai (approved):
Thanks! LGTM

PR #711, opened 2024-04-19
fix: extract digest from provenance when repo path is provided but digest is not provided from the user

tromai:
This Pull Request fixes the issue mentioned here - https://github.com/oracle/macaron/pull/698#issuecomment-2065609375
This can be seen as a continuation of https://github.com/oracle/macaron/pull/708

# Description
When the user provide only a repository path from `--repo-path` to Macaron without providing `--digest`. The expected behavior is that Macaron will use the repository path from the user provided `--repo-path` and try to extract the digest extracted from provenance if it's available (provided from user via `--provenance-file` or obtained from the PURL using provenance finder). 

This combination will be handled inside this block https://github.com/oracle/macaron/blob/e214326697aee6e31f0c9e406b11344c8619581e/src/macaron/slsa_analyzer/analyzer.py#L702-L708

However, its logic doesn't extract the commit hash from provenance, which leave `Analyzer.AnalysisTarget.digest` as empty. 

# Solution
- Use the provenance extractor to obtain the digest if the provenance payload is available.
- Note: the digest provided from the user still takes precedence. Therefore, if the user has already provided a value through `--digest`, the digest will not be extracted from the provenance.

### DIFF
--- a/src/macaron/slsa_analyzer/analyzer.py
+++ b/src/macaron/slsa_analyzer/analyzer.py
@@ -655,15 +655,6 @@ class Analyzer:
                     "Cannot determine the analysis target: PURL and repository path are missing."
                 )
 
-            case (None, _):
-                # If only the repository path is provided, we will use the user-provided repository path to create the
-                # ``Repository`` instance. Note that if this case happen, the software component will be initialized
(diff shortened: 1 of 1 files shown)


nathanwn (approved):
LGTM.

nicallen (approved):

PR #719, opened 2024-04-26
chore: update provenance status justification note in html report

tromai:
Update the provenance status justification in the HTML report from `repository` to `the target software component`.


vinkris01:
Instead of _the target component_, can we have the actual name of the artefact/repo url that'd the target?

behnazh-w:
> Instead of _the target component_, can we have the actual name of the artefact/repo url that'd the target?

@vinkris01 Each report has a target software component that is shown at the top. So, mentioning the name of the target here again might not be necessary here.

behnazh-w (approved):

PR #734, opened 2024-05-11
fix: do not fetch from origin/HEAD for local repo targets

behnazh-w:
For analyzing local repositories in offline mode, we should not fetch from `origin/HEAD` because user might want to run Macaron on the local changes. This PR changes the `check_out_repo_target` function to fix this issue: if `branch_name` and a commit are not provided, this function will not do anything and the latest local commit will be analyzed. If there are uncommitted local changes, the latest commit will appear in the report but the repo with local changes will be analyzed. We leave it up to the user to decide whether to commit the changes or not.

### DIFF
--- a/src/macaron/slsa_analyzer/git_url.py
+++ b/src/macaron/slsa_analyzer/git_url.py
@@ -124,7 +124,7 @@ def check_out_repo_target(
     with the repository specified by origin remote.
 
     If ``offline_mode`` is True and neither ``branch_name`` nor commit are provided, this function will not do anything
-    and the latest local commit will be analyzed. If there are uncommitted local changes, the latest commit will
+    and the HEAD commit will be analyzed. If there are uncommitted local changes, the HEAD commit will
     appear in the report but the repo with local changes will be analyzed. We leave it up to the user to decide
     whether to commit the changes or not.
 
@@ -233,7 +233,7 @@ def check_out_repo_target(
         logger.critical("The current HEAD at %s. Expect %s.", final_head_commit.hexsha, digest)
         return False
 
-    logger.info("The latest commit is %s.", final_head_commit.hexsha)
+    logger.info("The HEAD commit is %s.", final_head_commit.hexsha)
     return True
 
 
--- a/src/macaron/slsa_analyzer/git_url.py
+++ b/src/macaron/slsa_analyzer/git_url.py
@@ -123,9 +123,10 @@ def check_out_repo_target(
     will prune and update all references (e.g. tags, branches) to make sure that the local repository is up-to-date
     with the repository specified by origin remote.
 
-    If ``branch_name`` and a commit are not provided, this function will not do anything and the latest local commit
-    will be analyzed. If there are uncommitted local changes, the latest commit will appear in the report but the repo
-    with local changes will be analyzed. We leave it up to the user to decide whether to commit the changes or not.
+    If ``offline_mode`` is True and neither ``branch_name`` nor commit are provided, this function will not do anything
+    and the latest local commit will be analyzed. If there are uncommitted local changes, the latest commit will
+    appear in the report but the repo with local changes will be analyzed. We leave it up to the user to decide
+    whether to commit the changes or not.
 
     If ``branch_name`` is provided and a commit is not provided, this function will checkout that branch from origin
     remote (i.e. origin/<branch_name).
@@ -182,6 +183,15 @@ def check_out_repo_target(
             logger.error("Unable to fetch from the origin remote of the repository.")
             return False
 
+        # By default check out the commit at origin/HEAD only when offline_mode is False.
+        if not branch_name and not digest:
+            try:
+                git_obj.repo.git.checkout("--force", "origin/HEAD")
+            except GitCommandError:
+                logger.debug("Cannot checkout the default branch at origin/HEAD")
+                return False
+
+    # The following checkout operations will be done whether offline_mode is False or not.
     if branch_name and not digest:
         try:
             git_obj.repo.git.checkout("--force", f"origin/{branch_name}")
@@ -223,7 +233,7 @@ def check_out_repo_target(
         logger.critical("The current HEAD at %s. Expect %s.", final_head_commit.hexsha, digest)
         return False
 
-    logger.info("Successfully checked out commit %s.", final_head_commit.hexsha)
+    logger.info("The latest commit is %s.", final_head_commit.hexsha)
     return True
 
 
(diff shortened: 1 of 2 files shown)


src/macaron/slsa_analyzer/git_url.py
```diff
@@ -229,7 +233,7 @@ def check_out_repo_target(
         logger.critical("The current HEAD at %s. Expect %s.", final_head_commit.hexsha, digest)
         return False
 
-    logger.info("Successfully checked out commit %s.", final_head_commit.hexsha)
+    logger.info("The latest commit is %s.", final_head_commit.hexsha)
```
tromai, line 236: Because this INFO log is displayed to the user, I don't think using `latest commit` here is a good idea. This is because `final_head_commit` is  the commit which HEAD of the repository is pointing to after perform checking out (if any). Therefore, it may be the latest commit, or it might be the commit whose hash is provided by the user. I am good with changing it to anything as long as it reflects the details that this is the HEAD commit of the repository, which Macaron will analyze.

behnazh-w: >Therefore, it may be the latest commit, or it might be the commit whose hash is provided by the user. 

What is your definition of the latest commit. My definition and the reason for this message is to indicate the top or last commit that is currently present and Macaron has identified.

tromai: For me, "latest" here indicates time. "Latest commit" could mean the last commit that you have created on the git history of a branch.
```
# Run within the Macaron repository
> git log --format="%H%ad" origin/staging~5..origin/staging
d5920fcc16dd6ee4632f00fc030c4aa54696d390 Wed May 8 12:31:26 2024 +1000
e8ce555c128cf60ff74b2e384e1415c6556ec669 Mon Apr 29 02:33:20 2024 +0000
fece3827dd7e69d471750f8fd6d48a53e7df5bbe Mon Apr 29 11:19:53 2024 +1000
5b7f6b2227a927e172c5093dd46c67d055e56358 Mon Apr 29 11:19:36 2024 +1000
0e14e06ceaac14a9a0340c4bfb1b75a11fabcbf5 Fri Apr 26 12:27:40 2024 +1000
```
Here the latest commit is `d5920fcc16dd6ee4632f00fc030c4aa54696d390` which is the commit `origin/staging` is pointing to. If Macaron are analyzing this particular commit, we can safely say that it's analyzing the latest commit of branch `origin/staging`.

However, in the case where the user provides a specific commit hash for Macaron to checkout, or when Macaron checks out the commit for a specific tag. For example let say we checkout `fece3827dd7e69d471750f8fd6d48a53e7df5bbe`:
- That commit comes before `d5920fcc16dd6ee4632f00fc030c4aa54696d390` in the commit history.
- The latest commit here is still `d5920fcc16dd6ee4632f00fc030c4aa54696d390`. HEAD of the repository only temporarily points to a different commit. 

Therefore, I think using `latest commit` could be ambiguous because people might have different understanding of what `latest` means, therefore I think it's not the best idea. With the purpose of indicating the commit which Macaron has checked out and now performs the analysis, I would suggest something like:
```
Macaron checked out <commit-hash> and now performs the analysis on this commit.
``` 
Or if this is too analysis-related and not relevant to the standalone function for checking out a commit hash, we can also move this log into the `analyze.py` module too.

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -123,8 +123,10 @@ def check_out_repo_target(
     will prune and update all references (e.g. tags, branches) to make sure that the local repository is up-to-date
     with the repository specified by origin remote.
 
-    If ``branch_name`` and a commit are not provided, this function will checkout the latest commit of the
-    default branch (i.e. origin/HEAD).
+    If ``offline_mode`` is True and neither ``branch_name`` nor commit are provided, this function will not do anything
+    and the latest local commit will be analyzed. If there are uncommitted local changes, the latest commit will
+    appear in the report but the repo with local changes will be analyzed. We leave it up to the user to decide
+    whether to commit the changes or not.
```
tromai, line 129:
```suggestion
    If ``offline_mode`` is True and neither ``branch_name`` nor commit are provided, this function will not do anything
    and leave the commit which HEAD of the local repository is pointing to and any local changes AS IS. If there are uncommitted local changes, the commit which HEAD is pointing to will
    appear in the report but the local changes are involved in the analysis. We leave it up to the user to decide
    whether to commit the changes or not.
```
Because we are not modifying the HEAD of the local repository, I don't think we can say that it's going to always analyze the latest local commit. The user can manually check out a specific commit `B` different from the latest commit `A` (HEAD currently points to `B`), and then run Macaron without providing a branch or commit hash on it. Macaron would analyze the commit `B`, which is not the latest local commit. In my opinion, I think we should avoid using `latest commit` as `latest` can mean different things (e.g. a latest local commit vs a latest commit from origin - `origin/HEAD`). 

nit: `If there are uncommitted local changes, the commit which HEAD is pointing to will
    appear in the report but the local changes are involved in the analysis. We leave it up to the user to decide
    whether to commit the changes or not.` - I think these sentences are not relevant to what this function is doing, hence it's can be removed from the doc string of the function. Perhaps we can move it to the `analyzer.py` module or make it a normal comment
```python
# If there are uncommitted local changes, the commit which HEAD is pointing to will
# appear in the report but the local changes are involved in the analysis. We leave it up to the user to decide
# whether to commit the changes or not.
```

behnazh-w: > I think these sentences are not relevant to what this function is doing, hence it's can be removed from the doc string of the function. Perhaps we can move it to the analyzer.py module or make it a normal comment

I prefer to keep this documented in the docstring of this function because the commit that appears in the report is related to the checkout operation and communicating the behavior to the caller of the function is helpful.

src/macaron/slsa_analyzer/git_url.py
behnazh-w: I agree that it's better to avoid using `latest` and `last` here.

src/macaron/slsa_analyzer/git_url.py
tromai: Sounds good to me. Thanks!

behnazh-w: See commit [a19a2b0](https://github.com/oracle/macaron/pull/734/commits/a19a2b052c13e758c7a5700a1edd8720acb4517e)

src/macaron/slsa_analyzer/git_url.py
behnazh-w: See commit [a19a2b0](https://github.com/oracle/macaron/pull/734/commits/a19a2b052c13e758c7a5700a1edd8720acb4517e)

tromai (approved):
Thanks for the changes. LGTM

PR #739, opened 2024-05-14
chore: validate input repo, commit, provenance to ensure they match

benmss:
Adds a validation step to analysis that ensures the passed repo and commit match those found in the passed provenance, if all of those exist.

### DIFF
--- a/src/macaron/json_tools.py
+++ b/src/macaron/json_tools.py
@@ -5,14 +5,13 @@
 import logging
 from typing import TypeVar
 
-from macaron.util import JsonType
-
+JsonType = int | float | str | None | bool | list["JsonType"] | dict[str, "JsonType"]
 T = TypeVar("T", bound=JsonType)
 
 logger: logging.Logger = logging.getLogger(__name__)
 
 
-def json_extract(entry: JsonType, keys: list[str], type_: type[T]) -> T | None:
+def json_extract(entry: JsonType, keys: list, type_: type[T]) -> T | None:
     """Return the value found by following the list of depth-sequential keys inside the passed JSON dictionary.
 
     The value must be of the passed type.
@@ -21,8 +20,8 @@ def json_extract(entry: JsonType, keys: list[str], type_: type[T]) -> T | None:
     ----------
     entry: JsonType
         An entry point into a JSON structure.
-    keys: list[str]
-        The list of depth-sequential keys within the JSON.
+    keys: list
+        The list of depth-sequential keys within the JSON. Can be dict keys or list indices.
     type: type[T]
         The type to check the value against and return it as.
 
@@ -34,12 +33,22 @@ def json_extract(entry: JsonType, keys: list[str], type_: type[T]) -> T | None:
     target = entry
 
     for index, key in enumerate(keys):
-        if not isinstance(target, dict):
-            logger.debug("Expect the value .%s to be a dict.", ".".join(keys[:index]))
+        if isinstance(target, dict):
+            if key not in target:
+                logger.debug("JSON key '%s' not found in .%s", key, ".".join(keys[:index]))
+                return None
+        elif isinstance(target, list):
+            if isinstance(key, str):
+                logger.debug("JSON key '%s' of type 'str' cannot extract from list.", key)
+                return None
+
+            if key < 0 or key >= len(target):
+                logger.debug("JSON key '%s' is outside of list bounds %s.", key, len(target))
+                return None
+        else:
+            logger.debug("Expect the value .%s to be a dict or list.", ".".join(keys[:index]))
             return None
-        if key not in target:
-            logger.debug("JSON key '%s' not found in .%s", key, ".".join(keys[:index]))
-            return None
+
         target = target[key]
 
     if isinstance(target, type_):
--- a/src/macaron/json_tools.py
+++ b/src/macaron/json_tools.py
@@ -3,6 +3,7 @@
 
 """This module provides utility functions for JSON data."""
 import logging
+from collections.abc import Sequence
 from typing import TypeVar
 
 JsonType = int | float | str | None | bool | list["JsonType"] | dict[str, "JsonType"]
@@ -11,17 +12,17 @@ T = TypeVar("T", bound=JsonType)
 logger: logging.Logger = logging.getLogger(__name__)
 
 
-def json_extract(entry: JsonType, keys: list, type_: type[T]) -> T | None:
+def json_extract(entry: dict | list, keys: Sequence[str | int], type_: type[T]) -> T | None:
     """Return the value found by following the list of depth-sequential keys inside the passed JSON dictionary.
 
     The value must be of the passed type.
 
     Parameters
     ----------
-    entry: JsonType
+    entry: dict | list
         An entry point into a JSON structure.
-    keys: list
-        The list of depth-sequential keys within the JSON. Can be dict keys or list indices.
+    keys: Sequence[str | int]
+        The sequence of depth-sequential keys within the JSON. Can be dict keys or list indices.
     type: type[T]
         The type to check the value against and return it as.
 
@@ -30,29 +31,28 @@ def json_extract(entry: JsonType, keys: list, type_: type[T]) -> T | None:
     T | None:
         The found value as the type of the type parameter.
     """
-    target = entry
-
-    for index, key in enumerate(keys):
-        if isinstance(target, dict):
+    target: JsonType = entry
+    for key in keys:
+        if isinstance(target, dict) and isinstance(key, str):
             if key not in target:
-                logger.debug("JSON key '%s' not found in .%s", key, ".".join(keys[:index]))
+                logger.debug("JSON key '%s' not found in dict target.", key)
                 return None
-        elif isinstance(target, list):
-            if isinstance(key, str):
-                logger.debug("JSON key '%s' of type 'str' cannot extract from list.", key)
-                return None
-
+        elif isinstance(target, list) and isinstance(key, int):
             if key < 0 or key >= len(target):
-                logger.debug("JSON key '%s' is outside of list bounds %s.", key, len(target))
+                logger.debug("JSON list index '%s' is outside of list bounds %s.", key, len(target))
                 return None
         else:
-            logger.debug("Expect the value .%s to be a dict or list.", ".".join(keys[:index]))
+            logger.debug("Cannot index '%s' (type: %s) in target (type: %s).", key, type(key), type(target))
             return None
 
-        target = target[key]
+        # If statement required for mypy to not complain. The else case can never happen because of the above if block.
+        if isinstance(target, dict) and isinstance(key, str):
+            target = target[key]
+        elif isinstance(target, list) and isinstance(key, int):
+            target = target[key]
 
     if isinstance(target, type_):
         return target
 
-    logger.debug("Expect the value .%s to be of type %s", ".".join(keys), type_)
+    logger.debug("Found value of incorrect type: %s instead of %s.", type(target), type(type_))
     return None
--- a/src/macaron/repo_finder/provenance_extractor.py
+++ b/src/macaron/repo_finder/provenance_extractor.py
@@ -5,9 +5,8 @@
 import logging
 
 from macaron.errors import ProvenanceError
-from macaron.json_tools import json_extract
+from macaron.json_tools import JsonType, json_extract
 from macaron.slsa_analyzer.provenance.intoto import InTotoPayload, InTotoV1Payload, InTotoV01Payload
-from macaron.util import JsonType
 
 logger: logging.Logger = logging.getLogger(__name__)
 
@@ -67,16 +66,8 @@ def _extract_from_slsa_v01(payload: InTotoV01Payload) -> tuple[str | None, str |
     if not list_index:
         return None, None
 
-    material_list = json_extract(predicate, ["materials"], list)
-    if not material_list:
-        return None, None
-
-    if list_index >= len(material_list):
-        logger.debug("Material list index outside of material list bounds.")
-        return None, None
-
-    material = material_list[list_index]
-    if not material or not isinstance(material, dict):
+    material = json_extract(predicate, ["materials", list_index], dict)
+    if not material:
         logger.debug("Indexed material list entry is invalid.")
         return None, None
 
--- a/src/macaron/repo_finder/provenance_extractor.py
+++ b/src/macaron/repo_finder/provenance_extractor.py
@@ -235,7 +235,7 @@ def _clean_spdx(uri: str) -> str:
     return url
 
 
-def _check_if_input_repo_commit_provenance_conflict(
+def check_if_input_repo_commit_provenance_conflict(
     repo_path_input: str | None,
     digest_input: str | None,
     provenance_repo_url: str | None,
@@ -282,7 +282,7 @@ def _check_if_input_repo_commit_provenance_conflict(
     return False
 
 
-def _check_if_input_purl_provenance_conflict(
+def check_if_input_purl_provenance_conflict(
     git_obj: Git,
     repo_path_input: bool,
     digest_input: bool,
--- a/src/macaron/repo_finder/provenance_extractor.py
+++ b/src/macaron/repo_finder/provenance_extractor.py
@@ -3,9 +3,19 @@
 
 """This module contains methods for extracting repository and commit metadata from provenance files."""
 import logging
+import urllib.parse
+
+from packageurl import PackageURL
+from pydriller import Git
 
 from macaron.errors import ProvenanceError
 from macaron.json_tools import JsonType, json_extract
+from macaron.repo_finder.commit_finder import (
+    AbstractPurlType,
+    determine_abstract_purl_type,
+    extract_commit_from_version,
+)
+from macaron.repo_finder.repo_finder import to_domain_from_known_purl_types
 from macaron.slsa_analyzer.provenance.intoto import InTotoPayload, InTotoV1Payload, InTotoV01Payload
 
 logger: logging.Logger = logging.getLogger(__name__)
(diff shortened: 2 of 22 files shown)


src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -324,6 +324,19 @@ def run_single(
                 provenance_repo_url, provenance_commit_digest = extract_repo_and_commit_from_provenance(
                     provenance_payload
                 )
+                # Check the provenance repo commit match the input repo commit
```
tromai, line 327: Just some general questions. I can see that this PR only handles the case where repo url, commit hash are available as CLI options, and obtainable within the input provenance:
- Are we planning to perform the validation when the repo url or commit hash are obtainable from the input PURL string too?
- Are we planning to handle the case where repo url is provide but not commit hash ? 
(they could be out of scope, and could be handled in a different Pull request)
:thinking:

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -324,6 +324,19 @@ def run_single(
...
+                repo_path_input: str = config.get_value("path")
+                digest_input: str = config.get_value("digest")
+                if (repo_path_input and provenance_repo_url and repo_path_input != provenance_repo_url) or (
+                    digest_input and provenance_commit_digest and digest_input != provenance_commit_digest
+                ):
+                    return Record(
+                        record_id=repo_id,
+                        description="The repository URL or commit digest provided as input do not match what exists "
```
tromai, line 335: Do you think it would be good for the user if we print out the different values of repo url or commit hash if they don't match with each other?

src/macaron/json_tools.py
```diff
@@ -34,12 +33,22 @@ def json_extract(entry: JsonType, keys: list[str], type_: type[T]) -> T | None:
...
-        if key not in target:
-            logger.debug("JSON key '%s' not found in .%s", key, ".".join(keys[:index]))
+        if isinstance(target, dict):
+            if key not in target:
+                logger.debug("JSON key '%s' not found in .%s", key, ".".join(keys[:index]))
+                return None
+        elif isinstance(target, list):
+            if isinstance(key, str):
+                logger.debug("JSON key '%s' of type 'str' cannot extract from list.", key)
+                return None
```
tromai, line 43: I think this check doesn't cover the case where `key` is not a `str` and not `int`. Perhaps it's better to return None if the `key` is not `int`, for example:
```
if not isinstance(key, int):
...
```

src/macaron/json_tools.py
```diff
@@ -21,8 +20,8 @@ def json_extract(entry: JsonType, keys: list[str], type_: type[T]) -> T | None:
     ----------
     entry: JsonType
         An entry point into a JSON structure.
-    keys: list[str]
-        The list of depth-sequential keys within the JSON.
+    keys: list
```
tromai, line 23: Are we expecting elements of `keys` has any other types besides `str` and `int`. If not, I think we can use the type annotation `list[str | int]` here.

src/macaron/json_tools.py
```diff
@@ -34,12 +33,22 @@ def json_extract(entry: JsonType, keys: list[str], type_: type[T]) -> T | None:
...
+            if key not in target:
+                logger.debug("JSON key '%s' not found in .%s", key, ".".join(keys[:index]))
+                return None
+        elif isinstance(target, list):
+            if isinstance(key, str):
+                logger.debug("JSON key '%s' of type 'str' cannot extract from list.", key)
+                return None
+
+            if key < 0 or key >= len(target):
+                logger.debug("JSON key '%s' is outside of list bounds %s.", key, len(target))
```
tromai, line 46: nit: I think it should be called `JSON list index`  in the debug message :thinking:

src/macaron/json_tools.py
nathanwn: `list[str | int]` may not be the best choice here. tldr: In this particular case, I think `Sequence[str | int]` makes the most sense.

Although `str | int` is a supertype of both `str` and `int`, intuitively we may think `list[str | int]` is also a supertype of both `list[str]` and `list[int]`. However, this is not true, because `list` in Python is invariant. For the above property to hold, `list` needs to be covariant, which it is not (See [this section](https://mypy.readthedocs.io/en/stable/generics.html#variance-of-generic-types) in the mypy doc as they explain it much better than I can). You can try the following example with mypy:

```
from typing import reveal_type

def do_it(l: list[int | str]) -> None:
    return

a = [1, 2, 3]
b = ["foo", "bar", "baz"]
c = [1, "foo"]
reveal_type(a)
reveal_type(b)
reveal_type(c)
do_it(a)  # type error
do_it(b)  # type error
do_it(c)  # no type error
```

A covariant alternative usually used in a scenario like this would be `Sequence`. If you change the type annotation of the param `l` of `do_it` to `Sequence`, there will be no type error when you call `do_it(a)` or `do_it(b)`. One caveat is that `c` needs to be type-annotated as `list[str | int]`, as `mypy` infers `c` to be a `list[object]` and raises a type error otherwise. Here is the complete example with no type errors.

```python3
from typing import Sequence, reveal_type


def do_it(l: Sequence[int | str]) -> None:
    return


a = [1, 2, 3]
b = ["foo", "bar", "baz"]
c: list[int | str] = [1, "foo"]
reveal_type(a)
reveal_type(b)
reveal_type(c)
do_it(a)
do_it(b)
do_it(c)
```

src/macaron/util.py
```diff
@@ -262,17 +262,8 @@ def copy_file_bulk(file_list: list, src_path: str, target_path: str) -> bool:
...
-        elif isinstance(this, int) and isinstance(doc, list) and 0 <= this < len(doc):
-            doc = doc[this]
-        else:
-            return None
-    return doc
+def compare_urls_for_domain_path_equality(url_a: str, url_b: str) -> bool:
+    """Compare two URLs for equality based on the netloc/domain/hostname and path."""
+    parsed_a = urllib.parse.urlparse(url_a)
+    parsed_b = urllib.parse.urlparse(url_b)
+    return parsed_a.netloc == parsed_b.netloc and parsed_a.path == parsed_b.path
```
nathanwn, line 269: Is this function necessary? I don't see it being used anywhere.

src/macaron/json_tools.py
nathanwn: I agree that in the case of lists, as we always expect `int` keys, it's probably best to check if the key is `int` instead of checking for a different type like `str`.

src/macaron/json_tools.py
tromai: Ah I see. Thanks for the explanation. @nathanwn suggestion sounds good to me.

src/macaron/util.py
benmss: Good catch. I almost added another unused `util` function right after removing one.

src/macaron/json_tools.py
benmss: This is no longer applicable after most recent changes.

tests/e2e/expected_results/slsa-verifier/slsa-verifier_explicitly_provided_cue_PASS.json
```diff
@@ -1,50 +1,50 @@
...
-            "branch": "main",
-            "commit_hash": "fc50b662fcfeeeb0e97243554b47d9b20b14efac",
-            "commit_date": "2022-10-04T01:00:02+00:00"
+            "branch": null,
+            "commit_hash": "6fb4f7e2dd9c2f5d4f55fa88f6796278a7bba6d6",
+            "commit_date": "2022-08-25T11:37:20-05:00"
         },
         "provenances": {
-            "is_inferred": true,
+            "is_inferred": false,
```
tromai, line 47: This is not related to this Pull Request :sweat_smile:. 
It looks like even though we provide a provenance as input, it's not displayed here in the report, but rather the default inferred provenance content instead. I believe this is because we haven't tidied up different places where provenances are stored.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -536,31 +615,23 @@ def add_component(
...
-                # TODO: use both the repo URL and the commit hash to check.
-                if (
-                    existing_records
-                    and (existing_record := existing_records.get(git_url.get_remote_origin_of_local_repo(git_obj)))
-                    is not None
-                ):
-                    raise DuplicateCmpError(
-                        f"{analysis_target.repo_path} is already analyzed.", context=existing_record.context
-                    )
+        if git_obj:
```
tromai, line 618: If the type annotation of `git_obj` is not nullable, I wonder why we need to check for null-ability of `git_obj` here.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -341,11 +362,66 @@ def run_single(
...
+                        logger.debug(msg)
+                        return Record(
+                            record_id=repo_id,
+                            description=msg,
+                            pre_config=config,
+                            status=SCMStatus.ANALYSIS_FAILED,
+                        )
+                if not digest_input and provenance_commit_digest:
+                    # Check the commit from purl.
+                    digest_input = extract_commit_from_version(git_obj, parsed_purl.version)
```
tromai, line 405: From the signature of this `extract_commit_from_version` function, `git_obj` is not nullable - https://github.com/oracle/macaron/blob/8b0903b9b99304ba27874abd93c93f1e33899bbf/src/macaron/repo_finder/commit_finder.py#L177.
However, the `git_obj` variable here is still nullable, due to we not check for None after we prepare the repo here 
https://github.com/oracle/macaron/blob/8b0903b9b99304ba27874abd93c93f1e33899bbf/src/macaron/slsa_analyzer/analyzer.py#L366-L374

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -341,11 +362,66 @@ def run_single(
...
+                            pre_config=config,
+                            status=SCMStatus.ANALYSIS_FAILED,
+                        )
+
         # Create the component.
         try:
             component = self.add_component(
                 analysis,
                 analysis_target,
+                git_obj,
```
tromai, line 424: Similar comment as https://github.com/oracle/macaron/pull/739/files#r1610845420
The `add_component` method requires `git_obj` parameter to not be None.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -341,11 +362,66 @@ def run_single(
...
+                analysis_target.parsed_purl,
+            )
+
+        # Check if only one of the repo or digest came from direct input, and a versioned purl exists for the other.
+        if (provenance_repo_url or provenance_commit_digest) and parsed_purl and parsed_purl.version:
+            if determine_abstract_purl_type(parsed_purl) == AbstractPurlType.REPOSITORY:
+                if not repo_path_input and provenance_repo_url:
+                    # Check the repo from purl.
+                    possibly_expanded_type = parsed_purl.type
+                    if defaults.has_section("package_url"):
```
tromai, line 382: I wonder if resolving the mapping between the PURL type for git services (e.g. `github` or `bitbucket`) to its domain name is in the same manner as handled by this function https://github.com/oracle/macaron/blob/f21bb6fb38a10fee47b0512f07d7ecfb3fb05da2/src/macaron/repo_finder/repo_finder.py#L81 Or are these mappings in `defaults.ini` also serve another purpose :thinking:

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -341,11 +362,66 @@ def run_single(
...
+                        possibly_expanded_type = defaults.get(
+                            "package_url", parsed_purl.type, fallback=parsed_purl.type
+                        )
+                    repo_url_input = urllib.parse.urlparse(repo_path_input)
```
tromai, line 386: Because this line is within the if block
```
if not repo_path_input and provenance_repo_url:
```
I think `repo_url_input` will always be
```
>>> import urllib.parse
>>> urllib.parse.urlparse("")
ParseResult(scheme='', netloc='', path='', params='', query='', fragment='')
```

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -341,11 +362,66 @@ def run_single(
...
+                    purl_path = parsed_purl.name
+                    if parsed_purl.namespace:
+                        purl_path = f"{parsed_purl.namespace}/{purl_path}"
+                    if repo_url_input.netloc != possibly_expanded_type or repo_url_input.path != purl_path:
```
tromai, line 390: If I understand correctly, within this if block, we are comparing the repo url obtained from the PURL against the repo url obtained from the provenance. Should this comparison between them instead of the repo url from user input ?

src/macaron/slsa_analyzer/analyzer.py
benmss: There is sort of an argument for having it be user configurable, but that can be a discussion for another time. I will use this instead.

tromai: Oh okay, so there is a reason behind this decision. I will leave it to you for the final decision. Both ways are totally good for me. Thanks for the clarification.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -341,11 +386,75 @@ def run_single(
...
+                    # Check the repo from purl.
+                    possibly_expanded_type = to_domain_from_known_purl_types(parsed_purl.type)
+                    parsed_provenance_repo_url = urllib.parse.urlparse(provenance_repo_url)
+                    purl_path = parsed_purl.name
+                    if parsed_purl.namespace:
+                        purl_path = f"{parsed_purl.namespace}/{purl_path}"
+                    # Compare provenance repo url with purl repo url. Note that the urllib method includes the "/"
+                    # before path while the PURL method does not.
+                    provenance_url = f"{parsed_provenance_repo_url.netloc}{parsed_provenance_repo_url.path}"
+                    input_url = f"{possibly_expanded_type or parsed_purl.type}/{purl_path}"
```
nathanwn, line 413: If I understand it right, this piece of code constructs a repo URL from a repo-type PURL. Can we extract this out to a function and unit test it?

tromai: I wonder if it's doing anything differently from https://github.com/oracle/macaron/blob/f21bb6fb38a10fee47b0512f07d7ecfb3fb05da2/src/macaron/repo_finder/repo_finder.py#L103

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -318,12 +325,50 @@ def run_single(
...
+                        status=SCMStatus.ANALYSIS_FAILED,
+                    )
+                logger.debug(
+                    "Validated input repository and/or commit against input provenance and/or commit. "
+                    "Repo: %s - %s. Commit: %s - %s",
+                    repo_path_input,
+                    provenance_repo_url,
+                    digest_input,
+                    provenance_commit_digest,
+                )
```
nathanwn, line 371: I don't think this whole code block should be under the `try`-`except`. The `try`-`except` should only catch the error raised by the `extract_repo_and_commit_from_provenance` function call.
In addition, this block of code can also be moved out of the `if provenance_payload` statement altogether, since the logic inside it does not depend on whether a user-provided provenance exists or not, as we already have the initialization `provenance_repo_url = provenance_commit_digest = None` before this.

src/macaron/slsa_analyzer/analyzer.py
behnazh-w: @benmss I was also about to point the `to_repo_path` function that @tromai has mentioned. Looks like the logic here is duplicated?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -318,12 +325,50 @@ def run_single(
...
+                    if digest_input == provenance_commit_digest:
+                        digest_match = True
+                    elif (
+                        len(digest_input) < len(provenance_commit_digest)
+                        and 7 >= len(digest_input) <= 40
+                        and provenance_commit_digest.startswith(digest_input)
+                    ):
+                        digest_match = True
+
+                if repo_match or digest_match:
```
nathanwn, line 351: I don't understand this `if` statement. Shouldn't it be `if (not repo_match) or (not digest_match)`?

In addition, why don't we have two separate `if` statements and log the specific error for each case separately? i.e. one `if` statement to check and log if repo URLs not matching, and another `if` statement for commit hashes not matching? I don't see the advantage of combining multiple different statuses in the same log message. Meanwhile, having separate ones makes the log message much less noisy.

src/macaron/slsa_analyzer/analyzer.py
nathanwn: I did look at the `to_repo_path` function before making the comment because I did have the same feeling that there is some duplication going on. However, I don't think `to_repo_path` is relevant because it triggers the repo finder, which is not what we are doing here in this particular block of code.

nathanwn: Not to mention, the `to_repo_path` is currently being used for a different purpose, has no unit tests, and contains a `TODO`, meaning that its behavior may not be super well-defined for use in another context. If we want to make it useable for this case, we probably also need to revise where this function was used previously. I'm not sure if it's what we want to do in this PR.

benmss: I'm leaning towards improving and repurposing `to_domain_from_known_purl_types`. Its current form is a bit odd, and could easily be changed to work with this new requirement without issue.

benmss: Repurposing `to_domain_from_known_purl_types` and finishing the `TODO` in it is a little more involved than I expected, requiring a refactoring split of `_prepare_repo`. As such, I think I will create an issue to address that in a separate PR. For now I will keep the new similar code but still move it to a standalone function with unit tests. It can be merged/removed with existing functionality later.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -391,6 +442,136 @@ def run_single(
...
+        """
+        # Check the provenance repo against the input repo.
+        if repo_path_input and provenance_repo_url and repo_path_input != provenance_repo_url:
+            logger.debug(
+                "The repository URL from input does not match what exists in the provenance. "
+                "Input Repo: %s, Provenance Repo: %s.",
+                repo_path_input,
+                provenance_repo_url,
+            )
+            return True
```
tromai, line 478: An early return here mean that we accept the case where the repo url from provenance and user input are similar, but the digest from both sources can be different, which I don't think is the intended behavior.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -391,6 +442,136 @@ def run_single(
...
+    @staticmethod
+    def _check_if_input_purl_provenance_conflict(
+        git_obj: Git,
+        repo_path_input: bool,
+        digest_input: bool,
+        provenance_repo_url: str | None,
+        provenance_commit_digest: str | None,
+        purl: PackageURL,
+    ) -> bool:
+        """Test if the input repo and commit match the contents of the provenance.
```
tromai, line 501: nit: it would be good if we can put a comment here saying that we only support `purl` with Repository type.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -391,6 +442,136 @@ def run_single(
...
+        git_obj: Git
+            The Git object.
+        repo_path_input: bool
+            True if there is a repo as input.
+        digest_input: str
+            True if there is a commit as input.
+        provenance_repo_url: str | None
+            The repo url from provenance.
+        provenance_commit_digest: str | None
+            The commit digest from provenance.
```
tromai, line 514: The document for parameter `purl` for this method is missing.

src/macaron/slsa_analyzer/analyzer.py
benmss: The return here happens when there is a conflict between input repo and provenance repo, meaning analysis will fail.

tromai: Oh I see, my bad. Yes you are right.

tromai (approved):
Thanks for addressing the comments. LGTM

tests/slsa_analyzer/test_analyzer.py
```diff
@@ -155,3 +155,18 @@ def test_resolve_analysis_target_no_purl_or_repository() -> None:
     """Test creation of an Analysis Target when no PURL or repository path is provided."""
     with pytest.raises(InvalidAnalysisTargetError):
         Analyzer.to_analysis_target(Configuration(), [], None)
+
+
+@pytest.mark.parametrize(
+    ("url", "purl_string", "expected"),
+    [
+        ("https://github.com:9000/oracle/macaron", "pkg:github/oracle/macaron", True),
+        ("http://user:pass@github.com/oracle/macaron", "pkg:github.com/oracle/macaron", True),
```
nathanwn, line 164: Are you sure this is a valid URL? I tried passing this through the `parse_remote_url` function in `src/macaron/slsa_analyzer/git_url.py` and it did not parse successfully.

benmss: It is valid, however, it seems that many browsers have removed support for the feature, or marked it as deprecated. See : [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#access_using_credentials_in_the_url)

The reason it fails with `parse_remote_url` is because that function compares `netloc` against the list of domains, e.g. `github.com`, but `netloc` contains the auth part, the domain, and the port. So most likely that function should be updated to use `hostname` instead.

nathanwn (approved):
Thanks for the PR.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -341,11 +363,40 @@ def run_single(
...
+            git_obj = self._prepare_repo(
+                os.path.join(self.output_path, self.GIT_REPOS_DIR),
+                analysis_target.repo_path,
+                analysis_target.branch,
+                analysis_target.digest,
+                analysis_target.parsed_purl,
+            )
+
+        # Check if only one of the repo or digest came from direct input, and a versioned purl exists for the other.
+        if git_obj and (provenance_repo_url or provenance_commit_digest) and parsed_purl and parsed_purl.version:
```
behnazh-w, line 378: What if `parsed_purl.version` is `None`? Shouldn't we check for mismatch of the repo related parts the purl object?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -391,6 +442,138 @@ def run_single(
             context=analyze_ctx,
         )
 
+    @staticmethod
+    def _check_if_input_repo_commit_provenance_conflict(
```
behnazh-w, line 446: Isn't it better to move this function to `provenance_extractor.py` module?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -391,6 +442,138 @@ def run_single(
...
+                "Input Commit: %s, Provenance Commit: %s.",
+                digest_input,
+                provenance_commit_digest,
+            )
+            return True
+
+        return False
+
+    @staticmethod
+    def _check_if_input_purl_provenance_conflict(
```
behnazh-w, line 493: Same comment regarding moving this function to `provenance_extractor.py` module.

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -232,3 +233,133 @@ def _clean_spdx(uri: str) -> str:
...
+            "Input Commit: %s, Provenance Commit: %s.",
+            digest_input,
+            provenance_commit_digest,
+        )
+        return True
+
+    return False
+
+
+def _check_if_input_purl_provenance_conflict(
```
behnazh-w, line 285: Same comment regarding the private function.

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -232,3 +233,133 @@ def _clean_spdx(uri: str) -> str:
     """
     url, _, _ = uri.lstrip("git+").rpartition("@")
     return url
+
+
+def _check_if_input_repo_commit_provenance_conflict(
```
behnazh-w, line 238: Why is this function marked to be private? Since it is called from the `analyzer` module, I think it shouldn't be private.

behnazh-w (approved):

Issue #706, opened 2024-04-15
Potential problem for commit finder when packages do not use semver.


jenstroeger:
Package versioning in the Python ecosystem doesn’t always follow [semantic versioning](https://semver.org/); in fact, alternative [versioning schemes](https://en.wikipedia.org/wiki/Software_versioning#Schemes) are documented [here](https://packaging.python.org/en/latest/specifications/version-specifiers/#compatibility-with-other-version-schemes).[^1] For example for final release versions,

> The release segment consists of one or more non-negative integer values, separated by dots:
> ```
> N(.N)*
> ```
> Final releases within a project MUST be numbered in a consistently increasing fashion, otherwise automated tools will not be able to upgrade them correctly.

and then

> Date based release segments are also permitted.

For example, [black](https://github.com/psf/black) or [flake8-bugbear](https://github.com/PyCQA/flake8-bugbear) are popular packages following a date based version scheme. It might be interesting to pull metadata of existing packages from the [PyPI API](https://warehouse.pypa.io/api-reference/) and correlate a package’s release date with its version number, and perhaps other non-semver versioning schemes… [^2] 🤔

With that in mind, I suspect the commit finder might need to be reviewed?

[^1]: See also [romantic versioning](https://github.com/romversioning/romver) and [sentimental versioning](https://github.com/dominictarr/sentimental-versioning).

[^2]: Unfortunately, the [current pyproject.toml spec](https://packaging.python.org/en/latest/specifications/pyproject-toml/#declaring-project-metadata-the-project-table) doesn’t seem to contain a “versioning scheme” entry, so Macaron might be stuck with a few heuristics-based wild-wild guesses…

behnazh-w:
Macaron's commit finder can correctly find the commit for [black](https://github.com/psf/black) and [flake8-bugbear](https://github.com/PyCQA/flake8-bugbear) even though they follow a date-based version pattern. But these versions fail to communicate breaking changes via major version. It would be nice to have a check in Macaron that reports if a project does not follow semantic versioning?

jenstroeger:
@behnazh-w have you tried [`pytz`](https://pythonhosted.org/pytz/) which [supposedly](https://packaging.python.org/en/latest/specifications/version-specifiers/#olson-database-versioning) follows IANA’s timezone versioning ([link](https://github.com/eggert/tz/tags), [docs](https://data.iana.org/time-zones/tz-link.html)) — a year followed by a lower-case letter.

behnazh-w:
[...]

PR #741, opened 2024-05-17
chore: improve commit finder and input handling

benmss:
This PR improves the commit finder so that it can handle cases where numeric version parts have preceding zeroes. E.g. Version `2024.2.2` -> Tag `2024.02.02`. The trailing zero evaluation (for the version as a whole) has also been updated to accept "long" zeros, e.g. `1.00`, `1.000`.

In addition, PURLs with an invalid type as defined in the spec are rejected before analysis begins.

Partially relates to discussion in #706

### DIFF
--- a/src/macaron/repo_finder/commit_finder.py
+++ b/src/macaron/repo_finder/commit_finder.py
@@ -101,6 +101,7 @@ alphabetic_only_pattern = re.compile("^[a-z]+$", flags=re.IGNORECASE)
 hex_only_pattern = re.compile("^[0-9a-f]+$", flags=re.IGNORECASE)
 numeric_only_pattern = re.compile("^[0-9]+$")
 versioned_string = re.compile("^([a-z]+)(0*)([1-9]+[0-9]*)$", flags=re.IGNORECASE)  # e.g. RC1, M5, etc.
+multiple_zero_pattern = re.compile("^0+$")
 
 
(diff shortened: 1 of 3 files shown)


src/macaron/repo_finder/commit_finder.py
```diff
@@ -356,6 +358,10 @@ def _build_version_pattern(name: str, version: str) -> tuple[Pattern | None, lis
         elif count > 1:
             this_version_pattern = this_version_pattern + INFIX_3
 
+        if numeric_only:
+            # Allow for any number of preceding zeros when the part is numeric only. E.g. 000 + 1, 0 + 20
+            this_version_pattern = this_version_pattern + "0*"
```
tromai, line 363: If we are trying to match the preceding zeros, I wonder why we append the pattern for "0s" at the end, but not at the front.

nicallen (approved):

src/macaron/repo_finder/commit_finder.py
benmss: The version part gets added after this, so this addition is a preceding zero for that.

tromai: Oh I see. Thanks for the clarification.

tromai (approved):
Thanks for the change. LGTM

behnazh-w (approved):

PR #742, opened 2024-05-20
chore: replace use of actionlint parser in golang for GitHub Workflow…

nicallen:
… files, with a combination of yaml parsing and jsonschema validation in python.


src/macaron/parsers/github_workflow_model.py
```diff
@@ -0,0 +1,426 @@
...
+def as_reusable_workflow_call_job(job: Job) -> ReusableWorkflowCallJob | None:
+    if "uses" in job:
+        return cast(ReusableWorkflowCallJob, job)
+    return None
+
+
+def as_run_step(step: Step) -> RunStep | None:
+    if "run" in step:
+        return cast(RunStep, step)
+    return None
```
nathanwn, line 417: This is one way to do this kind of type-casting and I think it's fine.

I'm more used to using [type guards](https://peps.python.org/pep-0647/) in situations like these. Here for example, the type guard function can be:

```python3
def is_run_step(step: Step) -> TypeGuard[RunStep]:
    return "run" in step
```

in which case the following function

```python3
def get_run_step(step: Step) -> str | None:
    run_step = as_run_step(step)
    if run_step is not None:
        return run_step["run"]
    return None
```

can be rewritten as:

```python3
def get_run_step(step: Step) -> str | None:
    if is_run_step(step):
        return step["run"]  # here, `step` is no longer just `Step` and has been type-narrowed to `RunStep`
    return None
```

One drawback of this is that it requires Python 3.10. I also do not know if there is any specific intention you have in mind to have this kind of API, so this is just a suggestion and up to you to decide.

benmss: Python 3.13 introduces `TypeIs`, which appears to be recommended over `TypeGuard` in most cases. See https://peps.python.org/pep-0742/

nathanwn: Nice. I did not know about `TypeIs`. Thanks, Nick.

nathanwn (approved):
Thanks for the PR.

src/macaron/parsers/github_workflow_model.py
nicallen: I didn't know about either of them. I have changed it to use `TypeGuard`, which does seem simpler. `TypeIs` does look better than `TypeGuard` (and in using `TypeGuard`, I did encounter the issue that `TypeIs` addresses where it doesn't narrow in the else block), but we currently have 3.11 as our min python version, so until we raise that we can't use `TypeIs`.

benmss (approved):

pyproject.toml
```diff
@@ -141,7 +144,7 @@ omit = [
 # https://flit.pypa.io/en/latest/pyproject_toml.html#sdist-section
 # See also: https://github.com/pypa/flit/issues/565
 [tool.flit.sdist]
-include = []
+include = ["src/macaron/resources/schemastore/*"]
```
behnazh-w, line 147: `flit` should include all the files under `src/macaron` by default. I wonder if you faced any issues and needed to add this path explicitly?

nicallen: `flit` only includes files that are in the git repo. These files are downloaded as part of `make setup`, so need to be included explicitly.

behnazh-w: Hm, we also generate for example the `bashparser` binary during `make setup` and store it under `src/macaron/resources`, which gets included in the package. Is there a difference here?

src/macaron/resources/.gitignore
```diff
@@ -7,3 +7,6 @@ gradle-*
 
 # Ignore Gradle build output directory
 build
+
+# Ignore downloaded schemastore files
+schemastore
```
behnazh-w, line 12: I think you needed to include the path to schemastore because it's ignored here?

pyproject.toml
nicallen: Looks like the "only files in the git repo" behaviour only applies to the sdist, for the wheel it does include all files under the directory. So built binaries such as `bashparser` are packaged into the wheel but not the sdist. I suppose that makes sense, as from my limited understanding, the wheel is supposed to be the fully prebuilt ready-to-use package and the sdist is supposed to be the source distribution, though supposedly you are also supposed to be able to install the package from the sdist, which I don't think you can do with ours because it doesn't contain any build script to produce the binaries, so I'm not sure what the purpose of our sdist is if you have to use the wheel for installing. 

The original issue that I encountered trying to build the package with `flit` was that it wouldn't build if there were untracked files in the repo, I fixed that by adding the schemastore files to gitignore but I also just assumed that they were supposed to be included in the sdist, but now I'm not so sure, we could just leave them out as well since they are not our source code (regardless, they will be included in the wheel).

src/macaron/resources/.gitignore
nicallen: Without the gitignore, flit won't let you build the package at all, it aborts if there are untracked files.

pyproject.toml
behnazh-w: Yes, that makes sense. If someone wants to build the Macaron Python package from source, just using the `sdist` will not be enough and they need to use the `Makefile`. But we haven't made an attempt to allow automatically build the package from source using `pip install`. If/when we release Macaron on PyPI, we should look into this.

behnazh-w (approved):

Issue #678, opened 2024-03-21
Add integration test alongside provenance extraction validation


benmss:
Extracting the repository URL and commit for an analysis target requires validating the results against those found within the repository. As part of adding this feature, tests will be needed to verify it with real world data. For example, retrieving an npm provenance, extracting from it, and comparing the results with the repository URL and commits found within it for the specific version.

PR #747, opened 2024-05-24
feat: verify npm SLSA provenance against signed npm provenance

benmss:
This PR verifies the non-verified SLSA provenance retrieved from npm by comparing the subject hash against the hash found in the verified provenance that is also provided by npm.

The way in which provenance is found and verified has been changed so that the specifics of the finding and verifying are handled by the `Callable` functions passed to the Provenance Finder. For finding provenance, multiple functions can be passed resulting in the first found provenance being returned. By default, finding and verifying make use of the pre-existing methods introduced before, and as part of, this PR. In the future these can be replaced with other platform/provenance specific finding and/or verifying methods.

Relates to #678 
Closes https://github.com/oracle/macaron/issues/550


scripts/dev_scripts/integration_tests.sh
```diff
@@ -112,6 +112,8 @@ if [[ -z "$NO_NPM_TEST" ]]; then
     $RUN_POLICY -d $DB -f $OUTPUT_POLICY || log_fail
 fi
 
+exit
```
behnazh-w, line 115: I'm not sure why this line is added?

src/macaron/repo_finder/provenance_finder.py
```diff
@@ -110,19 +117,89 @@ def find_npm_provenance(purl: PackageURL, npm_registry: NPMRegistry) -> InTotoPa
...
-                except LoadIntotoAttestationError as loadintotoerror:
-                    logger.error("Error while loading provenance %s", loadintotoerror)
-                    return None
 
-                return provenance_payload
+                    signed_download_path = f"{download_path}.signed"
+                    if not os.path.exists(signed_download_path):
+                        return provenance_payload, False
+
+                    # Compare to signed provenance.
```
behnazh-w, line 130: The provenances can be published in various places, e.g., `npm` registry, GitHub Attestations, GitHub releases, each requiring a different mechanism to verify, e.g., the comparison implemented in `_compare_unsigned_signed_provenance`, using SLSA verifier, etc. I'm not sure if the current implementation and design is flexible enough to allow extending for different verification methods. 

One design that I can think of is to separate discovery from verification. 

- The discovery function accepts a list of callables that discover a provenance, and it calls them one by one and stops as soon as it can discover a provenance. 
- The verification function accepts a provenance and a callable that implements the suitable verification. This function can call the verification method that is passed at the callsite. The verfication can also be turned off through the `.ini` configuration.

Using this approach, we can control at the callsite which discovery and verification functions are suitable for a specific `purl` type.

Please let me know what you think

benmss: As per the updated PR description, finding and verifying accept callable functions to perform their respective tasks. A set of default functions are provided to retain the pre-existing functionality for npm and Gradle. The provenance verification can be disabled in the config.

src/macaron/repo_finder/provenance_finder.py
```diff
@@ -35,191 +37,350 @@ def __init__(self) -> None:
                 elif isinstance(registry, JFrogMavenRegistry):
                     self.jfrog_registry = registry
 
-    def find_provenance(self, purl: PackageURL) -> InTotoPayload | None:
-        """Find the provenance files of the passed PURL.
+    def find_provenance(
```
tromai, line 40: To me, this method looks like it only returns either:
- An empty list
- An instance of IntotoPayload (from my understanding, we go through the list of `discovery_functions` and return the very first IntotoPayload that we find).
Therefore, I wonder why we need the return value of this method to be `list[InTotoPayload]`?

src/macaron/repo_finder/provenance_finder.py
```diff
@@ -35,191 +37,350 @@ def __init__(self) -> None:
...
-
-        return None
-
-    @staticmethod
-    def find_npm_provenance(purl: PackageURL, npm_registry: NPMRegistry) -> InTotoPayload | None:
-        """Find and download the NPM based provenance for the passed PURL.
+        if not discovery_functions:
+            if determine_abstract_purl_type(purl) == AbstractPurlType.REPOSITORY:
+                # Do not perform default discovery for repository type targets.
+                return []
```
tromai, line 65: When we provide a list of `discovery_functions`, do we also need to handle repository type PURL, or we let the discovery functions decide whether the they want to support it or not?

src/macaron/repo_finder/provenance_finder.py
```diff
@@ -35,191 +37,350 @@ def __init__(self) -> None:
                 elif isinstance(registry, JFrogMavenRegistry):
                     self.jfrog_registry = registry
 
-    def find_provenance(self, purl: PackageURL) -> InTotoPayload | None:
-        """Find the provenance files of the passed PURL.
+    def find_provenance(
+        self,
+        purl: PackageURL,
+        discovery_functions: list[Callable[..., list[InTotoPayload]]] | None = None,
+        parameters: list[Any] | None = None,
```
tromai, line 44: From the implementation of this method, I can see that:
- `parameters` is provided as positional arguments to `discovery_function`. 
- `parameters` is provided the same way for each function in `discovery_function`.
This mean that the interface of each function must be the same. If their interfaces are different (e.g the order of arguments are different, the number of arguments are different, etc.) we can have a lot issues.
Is is okay to assume that all discovery functions are going to have the same interface and return type signature? :thinking:

src/macaron/repo_finder/provenance_finder.py
tromai: Ah nevermind, it looks like the list of InTotoPayload is used here because we decide that each discovery function must returns a list of InTotoPayload instances.

src/macaron/repo_finder/provenance_finder.py
```diff
@@ -35,191 +37,350 @@ def __init__(self) -> None:
...
+            return False
+
+        if not parameters:
+            logger.debug("No parameter arguments for provenance verification function.")
+
+        if parameters:
+            provenance_verified = verification_function(*parameters)
+        else:
+            # This will never be reached, and exists only for mypy.
+            return False
```
tromai, line 146: I think the `else` branch can still be reached because we didn't exit in the `if not parameters` statement.
I wonder if `parameters` is None, do we still run `verification_function()` or we exit on False :thinking: ?

src/macaron/repo_finder/provenance_finder.py
```diff
@@ -35,191 +37,350 @@ def __init__(self) -> None:
...
+        return provenance_verified
+
+
+def find_npm_provenance(purl: PackageURL, registry: NPMRegistry) -> list[InTotoPayload]:
+    """Find and download the NPM based provenance for the passed PURL.
+
+    Two kinds of attestation can be retrieved from npm: "Provenance" and "Publish". The "Provenance" attestation
+     contains the important information Macaron seeks, while the "Publish" attestation is pre-verified. Comparison
+     of the signed vs unsigned at the subject level, allows the unsigned to be verified too.
+     See: https://docs.npmjs.com/generating-provenance-statements
```
tromai, line 158: nit: 
```suggestion
    contains the important information Macaron seeks, while the "Publish" attestation is pre-verified. Comparison
    of the signed vs unsigned at the subject level, allows the unsigned to be verified too.
    See: https://docs.npmjs.com/generating-provenance-statements
```

src/macaron/repo_finder/provenance_finder.py
```diff
@@ -35,191 +37,350 @@ def __init__(self) -> None:
...
+            logger.debug("Provenance could not be verified.")
+        return provenance_verified
+
+
+def find_npm_provenance(purl: PackageURL, registry: NPMRegistry) -> list[InTotoPayload]:
+    """Find and download the NPM based provenance for the passed PURL.
+
+    Two kinds of attestation can be retrieved from npm: "Provenance" and "Publish". The "Provenance" attestation
+     contains the important information Macaron seeks, while the "Publish" attestation is pre-verified. Comparison
+     of the signed vs unsigned at the subject level, allows the unsigned to be verified too.
```
tromai, line 157: nit: I think it would be clearer if we mention which of `Provenance` and `Publish` is signed, and which of them is unsigned.

src/macaron/repo_finder/provenance_finder.py
benmss: Yes, and they return a list because the pre-existing npm function has to return two provenances. In most other cases I expect it will be a single item list.

src/macaron/repo_finder/provenance_finder.py
benmss: You raise a good point. I think it would make more sense to have a separate set of parameters for each function call.

src/macaron/repo_finder/provenance_finder.py
benmss: When provided, the intention is to let them decide what to do with different PURL types. However, from other work I've been doing, I'm not sure there is a possibility of utilising repository PURLs at this point in the analysis.

src/macaron/repo_finder/provenance_finder.py
```diff
@@ -35,191 +37,350 @@ def __init__(self) -> None:
...
+        The provenances to verify.
+
+    Returns
+    -------
+    bool
+        True if the provenance was verified, or False otherwise.
+    """
+    if len(provenance) != 2:
+        logger.debug("Expected unsigned and signed provenance.")
+        return False
```
tromai, line 242: Am I understanding this correctly: this function can only work on `provenance: list[InTotoPayload` generated from `find_npm_provenance` (due to specific semantic requirements of that list).

src/macaron/repo_finder/provenance_finder.py
benmss: I should have trusted mypy. This time.

src/macaron/repo_finder/provenance_finder.py
benmss: Yes. It is an example of a paired discovery and verification function.

tests/integration/cases/semver/policy.dl
```diff
@@ -11,6 +11,7 @@ Policy("test_policy", component_id, "") :-
     check_passed(component_id, "mcn_provenance_derived_commit_1"),
     check_passed(component_id, "mcn_provenance_derived_repo_1"),
     check_passed(component_id, "mcn_version_control_system_1"),
+    check_passed(component_id, "mcn_provenance_verified_1"),
```
tromai, line 14: I wonder if we need to update `policy.dl` for this test case too - https://github.com/oracle/macaron/blob/08183263a690dfc870ed750b53d93cf6a31a8096/tests/integration/cases/sigstore_mock/policy.dl.

benmss: Based on the current implementation, we do not. However, this is because the presence of the `-rp` command line option in this test prevents provenance finding from occurring. Perhaps that should be changed, and then this can be updated.

tromai: Ah I see. I don't think it's a problem, it's just be wondering why we don't need it for this test case. Thanks for the clarification.

src/macaron/repo_finder/provenance_finder.py
```diff
@@ -83,13 +84,19 @@ def find_provenance(
...
+        for index, discovery_function in enumerate(discovery_functions):
+            parameter_list = parameter_lists[index] if parameter_lists and index < len(parameter_lists) else [purl]
+            function_signature = signature(discovery_function)
+            if len(function_signature.parameters) != len(parameter_list):
+                logger.debug(
+                    "Mismatch between function and parameters: %s vs. %s",
+                    len(function_signature.parameters),
+                    len(parameter_list),
+                )
+                continue
```
tromai, line 96: Hmmm I can see that:
- Because we are passing the function pointers and the corresponding parameters separately, we need to do a lot of checking. 
- We also need to handle the default parameter when the parameter list is empty for each individual discovery function, which is obscured from the caller of `find_provenance`. 
```python
parameter_list = parameter_lists[index] if parameter_lists and index < len(parameter_lists) else [purl]
```

I think to address such problems, we can instead provide the function pointer and its corresponding parameters together :thinking: 

This is what I had in mind (logs are removed for clarity):
```python3
    def find_provenance(
        self,
        purl: PackageURL,
        discovery_functions: list[partial[list[InTotoPayload]]] | None = None,
    ) -> list[InTotoPayload]:
        if not discovery_functions:
            if determine_abstract_purl_type(purl) == AbstractPurlType.REPOSITORY:
                # Do not perform default discovery for repository type targets.
                return []
            if purl.type == "npm":
                if not self.npm_registry:
                    logger.debug("Missing npm registry to find provenance in.")
                    return []
                discovery_functions = [
                    partial(
                        find_npm_provenance,
                        *(purl, self.npm_registry),
                    ),
                ]
            elif purl.type in ["gradle", "maven"]:
                if not self.jfrog_registry:
                    logger.debug("Missing npm registry to find provenance in.")
                    return []

                discovery_functions = [
                    partial(
                        find_gav_provenance,
                        *(purl, self.jfrog_registry),
                    )
                ]
        if not discovery_functions:
            return []

        for discovery_function in discovery_functions:
            provenance = discovery_function()
            if provenance:
                return provenance

        logger.debug("No provenance found.")
        return []
```
In this example, I leveraged `partial` (https://docs.python.org/3/library/functools.html#functools.partial) which support the use case of defining the function pointer and its parameter lists for being called at a later stage. Doing it this way, we are essentially:
- Leave the responsibility of defining the function call to the caller of `find_provenance` . For example, this is how someone would use this function if they want to use specific discovery functions instead of the default one:
```python
provenances = ProvenanceFinder.find_provenance(
    purl,
    [
        partial(discovery1, a, b, c),
        partial(discovery2, a, e, f),
    ],
)
```
- `partial` supports both positional and keyword arguments, which I think can allow us flexibly define the discovery function interface.  
- We can still have the default usage as `provenances = ProvenanceFinder.find_provenance(purl)`
- In the example, I also removed setting the default parameter list to `[purl]` if the parameter list provided is empty. The caller of `find_provenance` know the signature of the discovery functions, so they should take responsibility to provide the necessary parameters (from what I see, `find_provenance` don't create or obtain any extra parameters required for the discover functions, so I guess it is fine). Please let me know what you think about this idea.

Furthermore, I think we can separate `find_provenance` as I described above into 2 methods:
- `find_provenance(purl: PackageURL)` - the default find provenance logic
- `find_provenance_custom(purl: PackageURL, discovery_functions: list[partial[list[InTotoPayload]]]` - the more advanced method to iterate through `discovery_functions` and return the first list of provenances found.
However, I don't have a strong opinion on this change.

tromai: @behnazh-w Please let me know what you think about it too, and whether it's inline with your initial design.

benmss: I have addressed this mostly as suggested. The key difference is the `find_provenance_custom` equivalent being an internal function, as having it being public made little sense when there was no mechanism to supply the custom discovery functions. Instead, the `find_provenance` function will again be responsible for calling the appropriate function(s), and can be extended for other types as required moving forwards.

tromai (dismissed):
LGTM

src/macaron/slsa_analyzer/checks/provenance_verified_check.py
```diff
@@ -0,0 +1,78 @@
...
+class ProvenanceVerifiedFacts(CheckFacts):
+    """The ORM mapping for justifications in the provenance verified check."""
+
+    __tablename__ = "_provenance_verified_check"
+
+    #: The primary key.
+    id: Mapped[int] = mapped_column(ForeignKey("_check_facts.id"), primary_key=True)  # noqa: A003
+
+    #: The verified status of the provenance.
+    verified: Mapped[str] = mapped_column(String, nullable=True, info={"justification": JustificationType.TEXT})
```
behnazh-w, line 29: This field doesn't add any value as it is. I think it would make more sense to report the SLSA level here which can be any of 0, 1, 2, 3, 4. If a provenance doesn't exist, according to SLSA [it is level 0](https://slsa.dev/spec/v1.0/levels#build-l0), if it exists but cannot be verified, according to SLSA [it is level 1](https://slsa.dev/spec/v1.0/levels#build-l1). If a provenance is generated by [SLSA GitHub generator](https://github.com/slsa-framework/slsa-github-generator) and verified, it is level 3 due to isolation enhancements, and otherwise it is level 2. We should be able to assign a level based on the `buildType`.

Please also document this reasoning for the user. I think it's best to create a separate page on the website for this check and explain the rules and logic of it, and what the results mean.

docs/source/pages/checks/slsa_builds.rst
```diff
@@ -0,0 +1,20 @@
...
+
+SLSA Build Levels report on various security aspects of a project, to provide a score that represents its overall trustworthiness and completeness.
+See `SLSA Levels <https://slsa.dev/spec/v1.0/levels>`_.
+
+Macaron's ``Provenance verified`` check uses the criteria of SLSA Build Levels to output a result that matches the correct level for a given artifact.
+
+- Build Level 0: There is no provenance for the artifact.
+- Build Level 1: There is provenance for the artifact but it cannot be verified.
+- Build Level 2: There is provenance for the artifact, and it has been verified.
+- Build Level 3: There is provenance for the artifact, it has been verified, and it was created by the SLSA GitHub Generator.
```
behnazh-w, line 18:
```suggestion
- Build Level 3: There is provenance for the artifact, it has been verified, and the build service isolates provenance generation in the control plane from the untrusted build process.
```

behnazh-w:
@benmss Can you please add this check to the policy spec of other relevant integration tests that generate provenances?

src/macaron/slsa_analyzer/checks/provenance_verified_check.py
```diff
@@ -0,0 +1,118 @@
...
+class ProvenanceVerifiedFacts(CheckFacts):
+    """The ORM mapping for justifications in the provenance verified check."""
+
+    __tablename__ = "_provenance_verified_check"
+
+    #: The primary key.
+    id: Mapped[int] = mapped_column(ForeignKey("_check_facts.id"), primary_key=True)  # noqa: A003
+
+    #: The SLSA build level of the provenance.
+    build_level: Mapped[str] = mapped_column(String, nullable=True, info={"justification": JustificationType.TEXT})
```
behnazh-w, line 30: I think this should be an integer to allow checking the value of the level in policies. Also, any particular reason to set `nullable` to `True`?

src/macaron/slsa_analyzer/package_registry/npm_registry.py
```diff
@@ -187,10 +191,20 @@ def download_attestation_payload(self, url: str, download_path: str) -> bool:
...
 
+            download_path_extended = download_path
+            if att.get("predicateType") == "https://github.com/npm/attestation/tree/main/specs/publish/v0.1":
+                download_path_extended = download_path_extended + ".signed"
+            else:
+                if downloaded:
+                    # In case there are multiple unsigned provenance, log it and continue.
+                    logger.debug("Found multiple unsigned provenance.")
+                    continue
+                downloaded = True
```
tromai, line 202: Hmm the `downloaded` variable is only set to True if we encounter the first unsigned provenance. Can we rename it to something that's less generic to prevent confusion with the signed provenance? I am thinking of something like `downloaded_unsigned`.

tromai: In additions, do we assume that there is always at least an unsigned provenance if a signed provenance is available?
For example, if `res_obj.get("attestations")` only contains the signed provenance predicate, then the return value of this function will be `False` even though we managed to download a signed provenance :thinking:

benmss: I expect that both kinds of provenance should exist. If only the signed is found, there is nothing we can do with it and reporting `False` makes the most sense. I will add a message to the log though.

tromai (approved):
LGTM. Thanks for the change!

behnazh-w (approved):

Issue #429, opened 2023-08-23
Add dependency analyzer for Python


benmss:
Issue #428 enables use of deps.dev for finding repositories of non-Java analysis targets. For this to work, some means of finding dependencies for these other languages is needed. As Macaron already supports analysis of Python projects, implementing a dependency analyzer for it is the next logical step.

PR #748, opened 2024-05-25
feat: add dependency resolution for Python

behnazh-w:
This PR adds a new feature to automatically analyze the dependencies of a Python package by providing the path to environment where the package is installed:

```shell
run_macaron.sh analyze -purl pkg:pypi/django@5.0.6 --python-venv <path-to-the-virtual-environment>
```
- [x] To implement this feature, it was necessary to detect the target software component in an SBOM to find its dependencies. So, the `metadata ` section of an SBOM is only required for multi-module Java projects and for other projects we should be able to resolve the dependencies even if `metadata` is not present in the SBOM.
- [x] To enhance the processing of the SBOMs, two new dependencies are added:
* cyclonedx-bom
* cyclonedx-python-lib

`cyclonedx-bom` is used to automatically generate the SBOM from a virtual environment and `cyclonedx-python-lib` is used to improve deserialization of SBOMs and processing the components.
- [x] Because the the `DependencyAnalyzer` class has CycloneDX-specific implementations, I have moved it to the `cyclonedx.py` module to avoid confusions.
- [x] The dependency analyzer of each build tool needs to implement a new function `get_purl_from_cdx_component` to customize how a PURL should be constructed.
- [x] The private Apache Maven integration test has been fixed to correctly match with the target software component.
- [x] I have added the dependency analysis tutorials on the website in the integration tests.

Closes https://github.com/oracle/macaron/issues/429


src/macaron/dependency_analyzer/cyclonedx_python.py
```diff
@@ -0,0 +1,111 @@
...
+                root_bom_path=Path(global_config.output_path, self.file_name),
+                recursive=defaults.getboolean(
+                    "dependency.resolver",
+                    "recursive",
+                    fallback=False,
+                ),
+            )
+        )
+
+    def remove_sboms(self, dir_path: str) -> bool:
```
tromai, line 80: It might not be relevant to this PR, but it looks to me that the return value of this method is not used anywhere. 
The only place where this method is being used (as far as I know) - https://github.com/oracle/macaron/blob/4e2c3ab8bceef002c534691910218773cf665a29/src/macaron/dependency_analyzer/cyclonedx.py#L426.
I guess we could consider remove the return value of this method if we are not using it.

For this method, is the reason we always return True is because `cyclonedx-bom` only produces one file as we specified from the command line options, instead of producing `bom.json` in multiple directories like for mvn and gradle?

docker/user.sh
```diff
@@ -52,6 +52,28 @@ then
     mkdir --parents "$HOME"/output
 fi
 
+# Prepare the python virtual environment. We copy the mounted directory if it exists to `.python_venv` to fix the symbolic
```
tromai, line 55: nit: I think it's not clear to me what virtual environment we are preparing here, is that the virtual environment where Macaron is installed and run, or is it the virtual environment provided from the user.

docs/source/pages/using.rst
```diff
@@ -251,6 +267,32 @@ Then the analysis can be run with:
...
+Analyzing dependencies using Python virtual environment
+-------------------------------------------------------
+
+Macaron can automatically identify and analyze the dependencies of a Python package if you provide the path to the virtual environment where the package is installed.
+
+Let's say you want to analyze ``django@5.0.6`` and its dependencies. First create a virtual environment and install ``django@5.0.6``:
+
+.. code-block:: shell
+
+  python3.11 -m venv /tmp/.django_venv
```
tromai, line 280: Does using `python3.11` here indicate that for this particular use case, Python3.11 must be used?

docker/user.sh
```diff
@@ -52,6 +52,28 @@ then
...
+# links to the Python interpreter in the container without affecting the files on host.
+if [[ -d "$HOME/python_venv" ]];
+then
+    cp -r "$HOME/python_venv" "$HOME/.python_venv"
```
tromai, line 59: I wonder if it would be a good idea to name these two `.python_venv` and `python_venv` differently?
It's not quite clear to me at the moment which of it is read-only (i.e. provided by the user) and read-write-able (the one we copied to `.python_venv`) from reading this file alone.
For example:
- `python_venv` -> `input_python_venv`: The virtual env mounted from the host file system
- `.python_venv` -> `deps_python_venv`: The write-able virtual env that only exists in the container.

docs/source/pages/using.rst
```diff
@@ -251,6 +267,32 @@ Then the analysis can be run with:
...
+.. code-block:: shell
+
+  ./run_macaron.sh analyze -purl pkg:pypi/django@5.0.6 --python-venv "/tmp/.django_venv"
+
+Where ``--python-venv`` is the path to virtual environment.
+
+Alternatively, you can create an SBOM for the python package and provide it to Macaron as input as explained :ref:`here <with-sbom>`.
+
+.. note:: We only support Python 3.11 for this feature of Macaron. Please make sure to install the package using this version of Python.
+
```
tromai, line 295: As Python3.11 is a prerequisite for running dependency resolution for PYPI packages, could we add this in https://oracle.github.io/macaron/pages/installation.html#prerequisites as an optional prerequisite and link back to this section?

src/macaron/config/global_config.py
```diff
@@ -79,6 +101,20 @@ def load_expectation_files(self, exp_path: str) -> None:
...
+        """
+        Load Python virtual environment.
+
+        Parameters
+        ----------
+        venv_path : str
+            The path to the Python virtual environment of the target software component.
+        """
+        if os.path.isdir(venv_path):
+            logger.info("Successfully loaded %s", venv_path)
```
benmss, line 114: I can't help but feel that checking that the path exits does not really count as a "load" operation.

docker/user.sh
```diff
@@ -52,6 +52,28 @@ then
...
+then
+    python_binaries=(
+        "python"
+        "python3"
+        "python3.11"
+    )
+    for p in "${python_binaries[@]}"; do
+        if [[ -f "$HOME/.venv/bin/${p}" ]];
+        then
+            ln -sf "$HOME/.venv/bin/${p}" "$HOME/.python_venv/bin/${p}"
```
tromai, line 72: I believe it would be useful for adding a comment about why we need to fix the python interpreters within `.python_venv/bin/` (i.e. `cyclone-dx` requires valid python interpreters when it's analyzing a virtual environment).

tromai: I also wonder why we symlink to the Python interpreters in `$HOME/.venv/bin/${p}` (what we are doing here works fine at the moment, however it doesn't sit right with me).
With virtual environment created by [venv](https://docs.python.org/3/library/venv.html). The Python interpreters are all symlinks to the base Python interpreters that were used to created that environment. 
This will mean that if we creates 2 virtual environments from the Python installation at `/usr/local/bin/python3.11`:
- The Python interpreters of those 2 virtual environments will symlink to `/usr/local/bin/python3.11`
- Both virtual environments should not symlink to each other, because that defeat the idea of isolated virtual environment.
Therefore, I think we should symlink from "$HOME/.python_venv/bin/${p}" to `realpath "$HOME/.venv/bin/python"` (which resolves to `/usr/local/bin/python3.11`, the base Python installation that comes with the base Docker image).

scripts/dev_scripts/integration_tests.sh
```diff
@@ -457,6 +458,27 @@ $RUN_MACARON -lr $WORKSPACE/output/git_repos/github_com analyze -rp apache/maven
...
+POLICY_FILE=$WORKSPACE/tests/policy_engine/resources/policies/django/test_dependencies.dl
+POLICY_RESULT=$WORKSPACE/output/policy_report.json
+POLICY_EXPECTED=$WORKSPACE/tests/policy_engine/expected_results/django/test_dependencies.json
+
+$RUN_POLICY -f "$POLICY_FILE" -d "$WORKSPACE/output/macaron.db" || log_fail
+check_or_update_expected_output $COMPARE_POLICIES "$POLICY_RESULT" "$POLICY_EXPECTED" || log_fail
+
+# Clean up and remove the virtual environment.
+rm -rf "$VIRTUAL_ENV_PATH"
+
```
tromai, line 481: I wonder if it's worth adding tests for invalid input virtual environments? I can see that there are 2 scenarios:
- The virtual environment doesn't contain valid Python interpreters
- The virtual environment directory doesn't exist.

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -1,31 +1,42 @@
-# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2024, Oracle and/or its affiliates. All rights reserved.
```
tromai, line 1: I wonder why the start date of this file is changed :thinking:

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -35,155 +46,608 @@ def deserialize_bom_json(file_path: Path) -> dict:
...
     if not os.path.exists(file_path):
         raise CycloneDXParserError(f"Unable to locate any BOM files at: {str(file_path.parent)}.")
 
+    # We use the `cyclonedx-python-library` library for deserialization following the example here:
+    # https://cyclonedx-python-library.readthedocs.io/en/v7.3.4/examples.html
     with open(file_path, encoding="utf8") as file:
+        json_data = file.read()
+        if defaults.getboolean("dependency.resolver", "validate", fallback=True):
+            try:
+                schema_version = defaults.get("dependency.resolver", "schema", fallback="1.6")
```
tromai, line 66: I wonder if we are intending to catch `ValueError` from `defaults.get(...)` too?. If not, I think this line can be moved outside of the try catch block.

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -35,155 +46,608 @@ def deserialize_bom_json(file_path: Path) -> dict:
...
+                my_json_validator = JsonStrictValidator(SchemaVersion.from_version(schema_version))
+            except ValueError as error:
+                raise CycloneDXParserError(f"Unable to find schema validator for {schema_version}: {error}") from error
+            try:
+                validation_errors = my_json_validator.validate_str(json_data)
+            except MissingOptionalDependencyException as error:
+                # Allow MissingOptionalDependencyException.
+                logger.debug("JSON-validation was skipped due to %s", error)
+            except json.JSONDecodeError as error:
+                raise CycloneDXParserError(f"BOM file is invalid: {error}") from error
```
tromai, line 76: Looking at the documentation for `validate_str` of the JsonStrictValidator https://cyclonedx-python-library.readthedocs.io/en/latest/autoapi/cyclonedx/validation/json/index.html#cyclonedx.validation.json.JsonStrictValidator
I have the following understanding:
- If `json_data` is not a valid JSON content -> `json.JSONDecodeError` is raised from `validate_str`
- If `json_data` is a valid JSON content, but violates the schema for cyclonedx, those errors will be returned as `validation_errors`
Is my assumption correct?

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -35,155 +46,608 @@ def deserialize_bom_json(file_path: Path) -> dict:
...
+
+            if validation_errors:
+                logger.debug("BOM file is invalid: %s", repr(validation_errors))
+                raise CycloneDXParserError(f"BOM file is invalid: {repr(validation_errors)}")
```
tromai, line 80: From the assumption I made in https://github.com/oracle/macaron/pull/748/files#r1628602723
I currently don't see any test that triggers this block. I believe to trigger this block, we must have a test with a BOM file that contains additional or unknown JSON properties. (as mentioned [here](https://cyclonedx-python-library.readthedocs.io/en/latest/autoapi/cyclonedx/validation/json/index.html#cyclonedx.validation.json.JsonStrictValidator))
Please let me know whether I missed something.

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -35,155 +46,608 @@ def deserialize_bom_json(file_path: Path) -> dict:
...
+            try:
+                schema_version = defaults.get("dependency.resolver", "schema", fallback="1.6")
+                my_json_validator = JsonStrictValidator(SchemaVersion.from_version(schema_version))
+            except ValueError as error:
+                raise CycloneDXParserError(f"Unable to find schema validator for {schema_version}: {error}") from error
+            try:
+                validation_errors = my_json_validator.validate_str(json_data)
+            except MissingOptionalDependencyException as error:
+                # Allow MissingOptionalDependencyException.
+                logger.debug("JSON-validation was skipped due to %s", error)
```
tromai, line 74: I wonder what `MissingOptionalDependency` means?
From reading the documentation:
- https://cyclonedx-python-library.readthedocs.io/en/latest/autoapi/cyclonedx/exception/index.html#cyclonedx.exception.MissingOptionalDependencyException
- https://cyclonedx-python-library.readthedocs.io/en/latest/install.html#extras
It seems to me that this exception is only raised if we try to perform the json/xml validation without installing the extras alongside with `cyclonedx-python-lib`. 
However, looking at our `pyproject.toml`, I am not sure if we are installing the `validation` extras required for this feature.

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -35,155 +46,608 @@ def deserialize_bom_json(file_path: Path) -> dict:
...
-    # TODO: use the official CycloneDX library to validate and deserialize BOM files
-    # once merged: https://github.com/CycloneDX/cyclonedx-python-lib/pull/290
     if not os.path.exists(file_path):
         raise CycloneDXParserError(f"Unable to locate any BOM files at: {str(file_path.parent)}.")
 
+    # We use the `cyclonedx-python-library` library for deserialization following the example here:
+    # https://cyclonedx-python-library.readthedocs.io/en/v7.3.4/examples.html
     with open(file_path, encoding="utf8") as file:
+        json_data = file.read()
+        if defaults.getboolean("dependency.resolver", "validate", fallback=True):
```
tromai, line 64: I wonder if it's better to provide a `validate: bool` flag to this `deserialize_bom_json` function instead of using the value of `defaults` directly inside this function's body?
This way, it allows us to easily unit test this feature, and communicate clearly in the interface about what  `deserialize_bom_json` will do. To me, `deserialize_bom_json` function seems to be a good generic function that doesn't need to depend on Macaron configuration in `defaults.ini` and should work on all json input.

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -35,155 +46,608 @@ def deserialize_bom_json(file_path: Path) -> dict:
...
+            # Based on the example provided by cyclonedx-python-lib this type issue needs to be suppressed.
+            # This method is injected into the Bom class that is annotated by ``serializable`` but mypy is not
+            # able to detect that.
+            bom_from_json = Bom.from_json(json.loads(json_data))  # type: ignore[attr-defined]
+        except (ValueError, AttributeError, json.JSONDecodeError) as error:
             raise CycloneDXParserError(f"Could not process the dependencies at {file_path}: {error}") from None
 
+        if isinstance(bom_from_json, Bom):
+            logger.debug("Successfully deserialized the BOM file: %s", repr(bom_from_json))
+            return bom_from_json
```
tromai, line 95: I wonder in which scenario that `bom_from_json` is not of type `Bom`. Is it correct to say `bom_from_json` is not of type `Bom` when no validation happens, and `json_data` is not a valid cyclonedx format :thinking:

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -35,155 +46,608 @@ def deserialize_bom_json(file_path: Path) -> dict:
...
+class DependencyInfo(TypedDict):
+    """The information of a resolved dependency."""
+
+    purl: PackageURL
+    url: str
+    note: str
+    available: SCMStatus
+
+
+class DependencyAnalyzer(ABC):
```
tromai, line 117: Is it okay for me to assume these details:
- `DependencyAnalyzer` is relocated from `dependency_resolver.py`, with **no change** to its behavior
- Old functions like `get_root_component`, `get_deps_components`, `get_deps_from_sbom` and `convert_components_to_artifacts` are moved inside this class with **no change** to their behaviors.

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -35,155 +46,608 @@ def deserialize_bom_json(file_path: Path) -> dict:
...
+
+            item["url"] = find_repo(item["purl"])
+            if item["url"] == "":
+                logger.debug("Failed to find url for purl: %s", item["purl"])
+            else:
+                # TODO decide how to handle possible duplicates here
+                item["available"] = SCMStatus.AVAILABLE
+                item["note"] = ""
+
+    def get_root_component(self, root_bom_path: Path) -> CDXComponent | None:
```
tromai, line 481: `self` is not used within this method, therefore we can consider making it a `@staticmethod`.

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -35,155 +46,608 @@ def deserialize_bom_json(file_path: Path) -> dict:
...
+        """
+        for cmp in root_bom.components:
+            if not isinstance(cmp, CDXComponent):
+                continue
+            cmp_purl = self.get_purl_from_cdx_component(cmp)
+            if str(cmp_purl) == target_component.purl:
+                logger.debug("Found the target CycloneDX component: %s", cmp.bom_ref.value)
+                return cmp
+        if root_bom.metadata:
+            return root_bom.metadata.component
```
tromai, line 529: I wonder why are we returning `root_bom.metadata.component`, which I believe is the [target component](https://cyclonedx.org/docs/1.6/json/#metadata_component) that this BOM document describes, without comparing to see if it matches `target_component` :thinking:

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -35,155 +46,608 @@ def deserialize_bom_json(file_path: Path) -> dict:
...
+            for node in root_bom.dependencies:
+                if not isinstance(node, CDXDependency):
+                    continue
+                if recursive or (
+                    target_cdx_component and target_cdx_component.bom_ref and node.ref == target_cdx_component.bom_ref
+                ):
+                    if dep_on := node.dependencies:
+                        dependencies.extend(dep_on)
+        except AttributeError as error:
+            logger.debug(error)
```
tromai, line 581: I wonder which part of the try block would raise this `AttributeError`?

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -35,155 +46,608 @@ def deserialize_bom_json(file_path: Path) -> dict:
...
+
+        # Find dependencies in child BOMs if they exist. Multi-module Java projects need this resolution.
```
tromai, line 583: Hmmm if only Multi-module Java projects need this resolution, should we move this logic into `cyclonedx_gradle` and `cyclonedx_mvn`?

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -35,155 +46,608 @@ def deserialize_bom_json(file_path: Path) -> dict:
...
+            try:
+                target_bom_ref = bom.metadata.component.bom_ref
+                if target_bom_ref and target_bom_ref.value:
+                    modules.add(target_bom_ref.value)
+                for node in bom.dependencies:
+                    if not isinstance(node, CDXDependency):
+                        continue
+                    if node.ref == target_bom_ref or recursive:
+                        dependencies.extend(node.dependencies)
+                        break
```
tromai, line 606: I wonder why for this part, `recursive` is put behind the OR while in this part, it's the opposite :thinking: 
https://github.com/oracle/macaron/blob/4e2c3ab8bceef002c534691910218773cf665a29/src/macaron/dependency_analyzer/cyclonedx.py#L575-L579

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -204,53 +668,124 @@ def convert_components_to_artifacts(
...
-
-    try:
-        with open(os.path.join(global_config.output_path, "sbom_debug.json"), "w", encoding="utf8") as debug_file:
-            debug_file.write(json.dumps(all_versions, indent=4))
-    except OSError as error:
-        logger.error(error)
-
-    return latest_deps
+            # except (KeyError, AttributeError) as error:
+            #     logger.debug(error)
```
tromai, line 688: Should we remove these comments?

tests/dependency_analyzer/cyclonedx/resources/bom_no_group.json
```diff
@@ -63,16 +63,6 @@
       ],
       "type" : "library",
       "bom-ref" : "pkg:maven/joda-time/joda-time@2.6?type=jar"
-    },
```
tromai, line 66: I wonder what is the issue of this component to be removed from this BOM file. This same question applies to changes in `tests/dependency_analyzer/cyclonedx/resources/bom_no_version.json`

tests/dependency_analyzer/cyclonedx/test_cyclonedx.py
```diff
@@ -6,26 +6,23 @@
...
 
 
-def test_deserialize_bom_json(snapshot: dict) -> None:
+def test_deserialize_bom_json(snapshot: list[str]) -> None:
     """Test deserializing a bom.json file."""
     # Deserialize a valid bom.json.
     path = Path(RESOURCES_DIR, "valid_bom.json")
-    assert snapshot == deserialize_bom_json(path)
+    bom = deserialize_bom_json(path)
+    assert snapshot == [str(cmp.bom_ref) for cmp in bom.components if isinstance(cmp, CDXComponent)]
```
tromai, line 25: I wonder why we moved away from testing the whole `bom` object, when we are de-serializing a static file that doesn't change. Is is intended to let this test pass, even in cases where the `deserialize_bom_json` method introduces other objects that is not type `CDXComponent` into the returned `bom` object?

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -35,155 +46,608 @@ def deserialize_bom_json(file_path: Path) -> dict:
...
+        tool_version : str
+            The version of the dependency analyzer.
+        repo_path: str
+            The path to the target repo.
+        """
+        self.resources_path: str = resources_path
+        self.file_name: str = file_name
+        self.tool_name: str = tool_name
+        self.tool_version: str = tool_version
+        self.repo_path: str = repo_path
```
tromai, line 140: I wonder what is this `repo_path` attributes being used for and where it's being used. I might miss something but I can't seem to find the place where it's being used yet.

src/macaron/dependency_analyzer/cyclonedx.py
tromai: Hmm but if we are not installing the optional dependencies as part of Macaron dev environment, how can we still perform the json and xml validation (as seen from the unit tests) :thinking:

tests/dependency_analyzer/cyclonedx/test_cyclonedx.py
tromai: I think it would be good to have a small comment here about this issue that you encountered.

src/macaron/dependency_analyzer/cyclonedx.py
behnazh-w: I suspect parts of the validation requires the optional dependency. Anyway, I will add the `extra` but prefer to keep the exception handling following the example to be on the safe side.

behnazh-w: See commit 5a38400417f92ad803993c36d0b25508fcd72f83

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -35,155 +46,633 @@ def deserialize_bom_json(file_path: Path) -> dict:
...
+                    DependencyAnalyzer._resolve_more_dependencies(deps_resolved)
 
-    return None
+                return deps_resolved
 
+            if not main_ctx.component.repository:
+                logger.info(
+                    "Unable to find a repository and no SBOM is provided as input. Analyzing the dependencies will be skipped."
+                )
+                return {}
```
behnazh-w, line 412: @tromai FYI: I fixed a bug here when neither a repository nor SBOM is available.

tromai: Hmmm it might not be relevant to this PR. This block operates somewhat irrelevant to the build tool detected for the software component and can be moved outside of the for loop too
https://github.com/oracle/macaron/blob/569b511ae5a1a22d74b0c2a185643c5ba79ced50/src/macaron/dependency_analyzer/cyclonedx.py#L397-L412

behnazh-w: The `dep_analyzer` is obtained from the build tool though: https://github.com/oracle/macaron/blob/569b511ae5a1a22d74b0c2a185643c5ba79ced50/src/macaron/dependency_analyzer/cyclonedx.py#L385

tests/dependency_analyzer/cyclonedx/test_cyclonedx.py
behnazh-w: See commit de28bafbf82894f10d157bbfaccaf65d1a6ecd83

behnazh-w:
@tromai I have added the tutorial examples on the website that were related to dependency analysis to the integration tests: 569b511ae5a1a22d74b0c2a185643c5ba79ced50

tromai (approved):
LGTM. Thanks for the changes!

docs/source/pages/using.rst
```diff
@@ -208,48 +208,82 @@ where ``micronaut-core.cue`` file can contain:
...
 
-..
-  TODO: Remove the -rp path after https://github.com/oracle/macaron/issues/108 is merged.
+.. code-block:: shell
+
+  ./run_macaron.sh analyze -purl pkg:maven/org.apache.maven/maven@3.9.7?type=pom -sbom <path_to_sbom>
+
+Where ``path_to_sbom`` is the path to the SBOM you want to use.
+
+For Python projects, you can use `cyclonedx-py <https://github.com/CycloneDX/cyclonedx-python>`_ to generate the SBOM. First install the package in a virtual environment, and then use ``cyclonedx-py`` to generate an SBOM for it. Here is an example:
```
benmss, line 231: A subheading here would be good to have.

docs/source/pages/using.rst
```diff
@@ -208,48 +208,82 @@ where ``micronaut-core.cue`` file can contain:
...
 
 ''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
 Analyzing dependencies in the SBOM without the main software component
 ''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
 
 In the case where the repository URL of the main software component is not available (e.g. the repository is in a self-hosted git service instance where Macaron cannot access),
 Macaron can still run the analysis on the dependencies listed in the SBOM.
-To do that, you must first create a PURL to present the main software component. This is so that this software component could be referenced later in the :ref:`verify-policy <verify-policy-command-cli>` command.
-For example: ``pkg:private_domain.com/org/name``.
+To do that, you must first create a PURL to present the main software component, e.g.,
```
benmss, line 251:
```suggestion
To do that, you must first create a PURL to represent the main software component, e.g.,
```

scripts/dev_scripts/integration_tests_docker.sh
```diff
@@ -105,6 +116,26 @@ run_macaron_clean analyze -purl pkg:maven/apache/maven -rp https://github.com/ap
...
+$MAKE_VENV "$VIRTUAL_ENV_PATH"
+"$VIRTUAL_ENV_PATH"/bin/pip install django==5.0.6
+run_macaron_clean analyze -purl pkg:pypi/django@5.0.6 --python-venv "$VIRTUAL_ENV_PATH" || log_fail
+
+# Check the dependencies using the policy engine.
+POLICY_FILE=$WORKSPACE/tests/policy_engine/resources/policies/django/test_dependencies.dl
+POLICY_RESULT=$WORKSPACE/output/policy_report.json
+POLICY_EXPECTED=$WORKSPACE/tests/policy_engine/expected_results/django/test_dependencies.json
+
+$RUN_MACARON_SCRIPT verify-policy -f "$POLICY_FILE" -d "$WORKSPACE/output/macaron.db" || log_fail
```
benmss, line 133:
```suggestion
$RUN_POLICY -f "$POLICY_FILE" -d $DB || log_fail
```

src/macaron/dependency_analyzer/cyclonedx_python.py
benmss: I think having a comment here stating that there are no files to remove would be useful.

scripts/dev_scripts/integration_tests_docker.sh
behnazh-w: See commit [a300d7d](https://github.com/oracle/macaron/pull/748/commits/a300d7d6c9573bde46dc0f6225a651d174e1e4ae)

docs/source/pages/using.rst
behnazh-w: See commit [a300d7d](https://github.com/oracle/macaron/pull/748/commits/a300d7d6c9573bde46dc0f6225a651d174e1e4ae)

docs/source/pages/using.rst
behnazh-w: See commit [a300d7d](https://github.com/oracle/macaron/pull/748/commits/a300d7d6c9573bde46dc0f6225a651d174e1e4ae)

src/macaron/dependency_analyzer/cyclonedx_python.py
behnazh-w: See commit [a300d7d](https://github.com/oracle/macaron/pull/748/commits/a300d7d6c9573bde46dc0f6225a651d174e1e4ae)

benmss (approved):

PR #750, opened 2024-05-26
feat: add a check to analyze malicious Python packages

Yao-Wen-Chang:
**PR Message:**

Implemented PyPI's heuristics check to detect suspicious packages using seven key heuristics.

**TODO List:**

- [x] Implement PyPI's heuristics check to detect the suspicious packages with seven heuristics
- [x] Implement confidence policy
- [x] Implement PyPI package registry 
- [x] Design the DB table to store the proper result
- [x] Create `src/macaron/malware_analyzer/` directory and move this check there.
- [x] Update documentation to reflect the new heuristics and their usage
- [x] Write unit tests to ensure the reliability and accuracy of the heuristics implementation


src/macaron/slsa_analyzer/pypi_heuristics/metadata/high_release_frequency.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def _get_releases(self) -> dict | None:
+        """Get all releases of the package.
+
+        Returns
+        -------
+            dict | None: Version to metadata.
+        """
+        releases: dict | None = self.api_client.get_releases()
+        return releases
```
tromai, line 37: If this method is a private method, and it doesn't do anything with the results from `self.api_client.get_releases`, I think we can remove `_get_releases` entirely and call `self.api_client.get_releases` directly.

src/macaron/slsa_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
```diff
@@ -0,0 +1,127 @@
...
+            package_name (str): PyPI package's name
+            version (str): The version of the package
+            destination_dir (str): The location of the tmp
+
+        Returns
+        -------
+            str | None: Source code.
+        """
+        # Create a temporary directory to store the downloaded source
+        temp_dir = tempfile.mkdtemp()
```
tromai, line 47: I usually prefer creating a temporary directory in a context manager. This way, you don't need to care about cleaning up the directory:
```python
# create a temporary directory using the context manager
with tempfile.TemporaryDirectory() as tmpdirname:
    print('created temporary directory', tmpdirname)
# directory and contents have been removed
```
Source - https://docs.python.org/3/library/tempfile.html#tempfile-examples

src/macaron/slsa_analyzer/pypi_heuristics/metadata/high_release_frequency.py
Yao-Wen-Chang: Good idea !

src/macaron/slsa_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
Yao-Wen-Chang: I will modify this, thanks.

pyproject.toml
```diff
@@ -31,7 +31,9 @@ dependencies = [
     "defusedxml >=0.7.1,<1.0.0",
     "packageurl-python >= 0.11.1,<1.0.0",
     "ruamel.yaml >= 0.18.6,<1.0.0",
-    "jsonschema >= 4.22.0,<5.0.0"
+    "jsonschema >= 4.22.0,<5.0.0",
+    "beautifulsoup4==4.12.3",
+    "types-beautifulsoup4==4.12.0.20240511"
```
tromai, line 36: We prefer setting a range for the dependencies (as the dependencies above) instead of pinning the exact version. This is because we want to let pip handle the dependency resolution and find the optimal version for each dependency. Forcing it to use a specific version might cause conflicts between the packages when you run `pip install`. 
In this case, you can specify it something like     
```
"beautifulsoup4 >= 4.12.0,<5.0.0",
"types-beautifulsoup4 >= 4.12.0,<5.0.0"
```
In additions, please move the entry for `types-beautifulsoup4` to the `[project.optional-dependencies].dev` section, where you also find other "types" packages defined there.

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -0,0 +1,181 @@
...
+        ctx : AnalyzeContext
+            The object containing processed data for the target repo.
+
+        Returns
+        -------
+        CheckResultData
+            The result of the check.
+        """
+        # https://www.imperva.com/learn/application-security/cve-cvss-vulnerability/
+        package = "requests"
```
tromai, line 152: Is this hard coding only for testing at the moment?

src/macaron/slsa_analyzer/pypi_heuristics/base_analyzer.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def __init__(
+        self,
+        name: str = "",
+        heuristic: HEURISTIC | None = None,
+        depends_on: list[tuple] | None = None,
+    ) -> None:
+        self.name: str = name
+        self.heuristic: HEURISTIC | None = heuristic
+        self.depends_on: list[tuple] | None = depends_on
```
tromai, line 21: Hmmm why do we need to specify the dependency between different heuristics? If I remember correctly, each heuristic should be isolated and independent :thinking:

src/macaron/slsa_analyzer/pypi_heuristics/analysis_result.py
```diff
@@ -0,0 +1,83 @@
+# Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved.
```
tromai, line 1: This is more of a discussion comment, not a request for change (yet :smile: ).
As I understand, this `Analysis` class is used similarly as a dictionary to store the result of each heuristic.
An instance of `Analysis` is used here https://github.com/oracle/macaron/blob/b523cacdd9965ccd09fb8fd772208f6e1120ec23/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py#L119-L136
If we only need a data structure to store the heuristic result during runtime, I think using a simple dictionary here  https://github.com/oracle/macaron/blob/b523cacdd9965ccd09fb8fd772208f6e1120ec23/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py#L120
is okay, and we can simplify by removing this `Analysis` class. Please let me know what you think, and whether my understanding of this `Analysis` class is correct. Thanks!

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -0,0 +1,181 @@
...
+    def _analyze(self, api_client: PyPIApiClient) -> dict:
+        analysis = Analysis()
+        results: dict = {}
+        detail_infos = {}
+        for _analyzer in ANALYZERS:
+            analyzer = _analyzer(api_client)
+            depends_on = analyzer.depends_on
+
+            if depends_on and any(analysis.get_result(heuristic[0]) is not heuristic[1] for heuristic in depends_on):
+                continue
```
tromai, line 128: As a general rule, it better to stay away from list/dict comprehension if the logic is getting a bit complicated. This is because:
- It's quite difficult to read as list/dict comprehension tend to get very long.
- It's difficult to test for code coverage because all the "logic branches" of the list/dict comprehension are in one expression, therefore you don't know which "logic branch" is covered, which is not.

For this line, one better way is to refactor `any(analysis.get_result(heuristic[0]) is not heuristic[1] for heuristic in depends_on)` to a method in `DetectMaliciousMetadataCheck` and use for loop instead. For example:
```
def should_skip(self, analysis, depends_on) -> bool:
    ...
```

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
Yao-Wen-Chang: Yes. Does the package input exist? If so, could you please specify it?

src/macaron/slsa_analyzer/pypi_heuristics/base_analyzer.py
Yao-Wen-Chang: Some heuristic depends on another one.
For example, if the one_release heuristic FAIL, it means there is only one release. Then, the high_release_frequency heuristic will not be operated. 
In this case, the high_release_frequency depends on the one_release.

tromai: Hmmm after having a closer look at the implementation, looks like the dependency between heuristics are used within this method https://github.com/oracle/macaron/blob/b523cacdd9965ccd09fb8fd772208f6e1120ec23/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py#L119
Could you please let me know more about the relationship between heuristics? (e.g. which heuristic depends on which, how are they depended on each other, etc.)

src/macaron/slsa_analyzer/pypi_heuristics/analysis_result.py
Yao-Wen-Chang: I don't notice it is duplicated to `results`.
I will modify this.

src/macaron/slsa_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
```diff
@@ -0,0 +1,127 @@
...
+                except zipfile.LargeZipFile as large_zip_exception:
+                    # Handle the LargeZipFile exception
+                    logger.debug("Zip file too large to read: %s", large_zip_exception)
+
+            for root, _, files in os.walk(temp_dir):
+                for file in files:
+                    if file == "setup.py":
+                        file_path = os.path.join(root, file)
+                        with open(file_path, encoding="utf-8") as py_file:
+                            return py_file.read()
```
tromai, line 85: This implementation will proceed with the very first `setup.py` it discovers. Are there any cases where the source somehow contains multiple `setup.py` in different directories ? :thinking:

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
Yao-Wen-Chang: All right, I will make the code clean.

src/macaron/slsa_analyzer/pypi_heuristics/metadata/unreachable_project_links.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def _get_project_links(self) -> dict | None:
+        """Implement the method to get the project links.
+
+        Returns
+        -------
+            dict | None: Link name to url.
+        """
+        project_links: dict | None = self.api_client.get_project_links()
+        return project_links
```
tromai, line 37: Same comment as https://github.com/oracle/macaron/pull/750/files#r1619944559

src/macaron/slsa_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
```diff
@@ -0,0 +1,127 @@
...
+
+        try:
+            sourcecode_url = self.api_client.get_sourcecode_url()
+            if sourcecode_url is None:
+                return None
+            response = requests.get(sourcecode_url, stream=True, timeout=40)
+            response.raise_for_status()
+            filename = sourcecode_url.split("/")[-1]
+            with open(os.path.join(temp_dir, filename), "wb") as f:
+                f.write(response.content)
```
tromai, line 57: From your evaluation, what is your impression on the size of the source code that you need to download here? Was any big impact on the running time observed for any of the PYPI packages that you encountered?

src/macaron/slsa_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
Yao-Wen-Chang: Could be.
To my knowledge, the setup.py should be located in the root directory.
https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/
As it mentions, `The most important file is setup.py which exists at the root of your project directory.`.

Yao-Wen-Chang: Also, if analyzing the imported module in other files with the pattern is required, it can be extended.

src/macaron/slsa_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
Yao-Wen-Chang: The packages related to LLM is bigger, but I haven't tested it yet. 
I add the exception to handle this.

pyproject.toml
Yao-Wen-Chang: Modified

src/macaron/slsa_analyzer/pypi_heuristics/metadata/unreachable_project_links.py
Yao-Wen-Chang: Fixed

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
tromai: Yes, to be able to obtain the name of the package, you can look into `ctx.component.purl`, which is declared here https://github.com/oracle/macaron/blob/5e7ccdc68fb9f37345a32a2373fa0a1a67a0ce0b/src/macaron/database/table_definitions.py#L140
That will give you the PackageURL string of the software component you are analyzing. You then would need to extract the necessary information from that string. Let's go through an example.
First, let's run Macaron against the PYPI requests package, using [a PYPI type PURL string](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#pypi) for it:
```
macaron analyze -purl pkg:pypi/requests@2.32.3 --skip-deps
```
Then within your check implementation, use this block to extract the package name from the purl string:
```
# Remember to import this library.
from packageurl import PackageURL

...

purl_string = ctx.component.purl
parsed_purl = PackageURL.from_string(purl_string)
pypi_package_name = parsed_purl.name
logger.info("TESTING: %s", pypi_package_name)
```
Please give it a try and let me know if you have any issue with it.

tromai:
@Yao-Wen-Chang When you address an issue mentioned in a comment, you can put the link to the commit next to the comment indicating that it has been resolve. Something like this would be extremely helpful for my review. Thanks!
```
Fixed this in https://github.com/oracle/macaron/pull/750/commits/d6684088dbe7792731240709d73a17de1eba73ad
```

pyproject.toml
tromai: Thanks. Even though the range issue is fixed, `types-beautifulsoup4` entry hasn't been moved to `[project.optional-dependencies].dev` in `pyproject.toml`, just to let you know :smile:

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
Yao-Wen-Chang: Fixed this in https://github.com/oracle/macaron/pull/750/commits/7255812b138562a950c90fa7d5f7b80e29d24842

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -0,0 +1,192 @@
...
+    def __init__(self) -> None:
+        """Initialize a check instance."""
+        check_id = "mcn_detect_malicious_metadata_1"
+        description = "Check whether the features of package adhere to the heurisic."
+        super().__init__(
+            check_id=check_id,
+            description=description,
+        )
+
+    def _should_skip(self, results: dict, dependency_heuristic: tuple[HEURISTIC, RESULT]) -> bool:
```
tromai, line 120: It would be better if we annotate the element types of list, dict or similar datastructures. For example, let's say that the `results` dictionary has keys of type `int` and values of type `str`, it should be annotated like this:
```
_should_skip(self, results: dict[int, str], dependency_heuristic: tuple[HEURISTIC, RESULT]) -> bool:
```

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -0,0 +1,192 @@
...
+        if results.get(dependency_heuristic[0], None) is not dependency_heuristic[1]:
+            return True
+        return False
+
+    def _analyze(self, api_client: PyPIApiClient) -> dict:
+        results: dict = {}
```
tromai, line 126: Same comment as https://github.com/oracle/macaron/pull/750/files#r1632536180

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -0,0 +1,192 @@
...
+        detail_infos = {}
```
tromai, line 127: Same comment as https://github.com/oracle/macaron/pull/750/files#r1632536180

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -0,0 +1,198 @@
...
+            Shows whether making REST API calls to npm registry is enabled.
+        """
+        self.hostname = hostname or ""
+        self.attestation_endpoint = attestation_endpoint or ""
+        self.request_timeout = request_timeout or 10
+        self.enabled = enabled
+        super().__init__("pypi Registry")
+
+
+class PyPIApiClient:
```
tromai, line 55: Hmm I think this class design is not in a same manner as what we are doing for [src/macaron/slsa_analyzer/package_registry/npm_registry.py](https://github.com/oracle/macaron/blob/28131cd745114f8d7d51444eb272ee8d71be2fb1/src/macaron/slsa_analyzer/package_registry/npm_registry.py).
The idea is that only one instance of the "registry" class needs to be created. It will be reused to obtain information for all packages that it supports (i.e. the "package name" should be provided as parameters, instead of being an attribute of the class instance). 
I think the logic for obtaining API data you have here is okay. What we need to do right now is merging `PyPIRegistry` and `PyPIApiClient` together:
Here is an example implementation that you can extend upon
1. `src/macaron/config/defaults.ini`
```ini
...
[package_registry.pypi]
# Set `enabled=False` to disable making REST API calls to the PyPI registry.
enabled = True
# pypi registry host name.
hostname = pypi.org
request_timeout = 2
...
```

2. `src/macaron/slsa_analyzer/package_registry/__init__.py`
```python
PACKAGE_REGISTRIES: list[PackageRegistry] = [JFrogMavenRegistry(), MavenCentralRegistry(), NPMRegistry(), PYPIRegistry(PackageResgistry)]
```

3. `src/macaron/slsa_analyzer/package_registry/pypi_registry.py`
```python
class PyPIRegistry(PackageRegistry):
    ...
    def __init__(
        self,
        hostname: str | None = None,
        request_timeout: int | None = None,
        enabled: bool = True,
        ...
    ) -> None:
        self.hostname = hostname or ""
        self.request_timeout = request_timeout or 10
        self.enabled = enabled
        ...
        super().__init__("pypi Registry")

    def load_defaults(self) -> None:
        section_name = "package_registry.pypi"
        if not defaults.has_section(section_name):
            self.enabled = False
            return
        section = defaults[section_name]

        if not section.getboolean("enabled", fallback=True):
            self.enabled = False
            logger.debug("pypi registry is disabled in section [%s] of the .ini configuration file.", section_name)
            return

        self.hostname = section.get("hostname")
        if not self.hostname:
            raise ConfigurationError(
                f'The "hostname" key is missing in section [{section_name}] of the .ini configuration file.'
            )

        try:
            self.request_timeout = section.getint("request_timeout", fallback=10)
        except ValueError as error:
            raise ConfigurationError(
                f'The "request_timeout" value in section [{section_name}]'
                f"of the .ini configuration file is invalid: {error}",
            ) from error

     # Implementing this method is a must because Macaron will run this method when trying to detect which build tool is supported by PyPI.
     def is_detected(self, build_tool: BaseBuildTool) -> bool:
        if not self.enabled:
            logger.debug("Support for the pypi registry is disabled.")
            return False
        compatible_build_tool_classes = [Pip, Poetry]
        for build_tool_class in compatible_build_tool_classes:
            if isinstance(build_tool, build_tool_class):
                return True
        return False
   
    # Please move all methods of `PyPIApiClient` to become `PyPIRegistry` 's methods. Let's see an example
    def get_base_metadata(self, package_name: str) -> dict:
        """Get the metadata through PyPI official API.

        Returns
        -------
            dict | None: PyPI official metadata.
        """
        endpoint = os.path.join(self.base_url, "pypi", package_name, "json")
        response: dict = send_get_http(endpoint, headers=self.headers)
        return response
```
Some additional notes:
- Because one instance of `PyPIRegistry` is used again and again, all methods that you implements must have "pypi package related information" passed in as parameters (see an example for `get_base_metadata` above). 
- You don't need to create an instance of `PyPIRegistry` within the check implementation. This is being done already in `src/macaron/slsa_analyzer/package_registry/__init__.py`. When you run Macaron, at this stage - https://github.com/oracle/macaron/blob/621fad7184e05995c61bffcc3d8b7501c04f07d5/src/macaron/slsa_analyzer/analyzer.py#L1015-L1026
The `PyPIRegistry` will be assigned to `analyze_ctx.dynamic_data["package_registries"]` automatically. This allow you to obtain and use `PyPIRegistry` in the check like this: 
```python
def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
    ...
    package_registry_info_entries = ctx.dynamic_data["package_registries"]
    for package_registry_info_entry in package_registry_info_entries:
        match package_registry_info_entry:
            case PackageRegistryInfo(
                build_tool=Pip() | Poetry(),
                package_registry=PyPIRegistry() as pypi_registry,
            ) as pypi_info_entry:
                #  Analyzing the heuristics here
                result: dict = self._analyze(pypi_registry)
                ...
    
    #  If the code reaches here, it means that `ctx` is not pointing to a PyPI package, hence the UNKNOWN status.
    return CheckResultData(result_tables=[], result_type=CheckResultType.UNKNOWN)
``` 
Doing it this way, we can safely skip this check when we are analyzing software components that are not PyPI packages. In the future, if this check supports more ecosystems for example, the `match...case` can be extended upon.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -0,0 +1,198 @@
...
+    """PyPI's official and customed API client."""
+
+    def __init__(self, package: str) -> None:
+        self.headers = {"Host": "pypi.org", "Accept": "application/json"}
```
tromai, line 59: I wonder if those headers are strictly necessary for us to use the PyPI API?

tests/malware_analyzer/pypi/test_empty_project_link_analyzer.py
```diff
@@ -0,0 +1,54 @@
+# Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved.
```
tromai, line 1: Thanks for adding the tests!
However, we have started to move away from using `unittest.TestCase` for our tests. Instead, we define test cases in test functions, which will be picked up by `pytest` - https://docs.pytest.org/en/stable/how-to/assert.html#how-to-write-and-report-assertions-in-tests. Please see this file for a simple example - https://github.com/oracle/macaron/blob/a3842b4bbd121c8f0b7e9cd1597830d1b41e419f/tests/slsa_analyzer/checks/test_check_results.py
For the logic within `setUp`, you can consider using `fixtures` - https://docs.pytest.org/en/stable/how-to/fixtures.html#how-to-use-fixtures. 
The required fixtures can be defined in this file [`tests/conftest.py`](https://github.com/oracle/macaron/blob/51a1aad6bc8cf57f8318e2a82069b1b4939126b9/tests/conftest.py) and then used within the test module.
For example, you can define a fixture
```
@pytest.fixture(autouse=True)
def my_fixture() -> dict:
    return {"A": 1, "B": 2} 
```
and it can be used within the test module without the need for importing from `tests/conftest.py`
```
def test_my_fixture(my_fixture) -> None:
    assert my_fixture["A"] == 1
```

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
tromai: In additions, it would be great to add doc string to this method as this is an important logic for our check.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def get_releases(self) -> dict | None:
+        """Get all releases.
+
+        Returns
+        -------
+            dict | None: Version to metadata.
+        """
+        releases: dict | None = self.base_metadata.get("releases", None)
+        return releases
```
tromai, line 83: Some of the methods here I can see that they are using the base metadata obtained from the method call `_get_base_metadata`. From this comment https://github.com/oracle/macaron/pull/750/files#r1632579056, and also it's usually introduce more complexity by adding states to a class instance (the state in this case is represented by different `self.base_metadata`, which changes continuously depending on the package we are analyzing). Therefore, I think we should:
- Remove the use of storing the result of `base_metadata` as a class attribute.
- Because a lot of information can be extracted from the `base_metadata` dictionary, it's not good to keep performing the API again and again, let's change these methods signature so that it accepts the `response` as a parameter
```python
def get_releases(self, response: dict) -> dict | None:
    releases: dict | None = response.get("releases", None)
    return releases
```
Then within the check, we will keep the value of `response` and use it again and again
```
base_metadata = pypi_registry.get_base_metadata(package_name)
releases = pypi_registry.get_releases(base_metadata)
project_links = pypi_registry.get_project_links(base_metadata)
```

tromai: I additions, if you are working with a json like data structure (a dictionary that was loaded from a json content, i.e the API result), we have this utility function to extract values from it - https://github.com/oracle/macaron/blob/e214326697aee6e31f0c9e406b11344c8619581e/src/macaron/json_tools.py#L15.
For example, let's use this function for `get_releases`
```python
from macaron.json_tools import json_extract
...
def get_releases(self, base_metadata: dict) -> dict | None:
        """Get all releases.
        Returns
        -------
            dict | None: Version to metadata.
        """
        releases = json_extract(base_metadata, ["releases"], dict)
        return releases
```
We should consider changing other methods (e.g. `get_latest_version`, etc.) to use this `json_extract` method too).

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def get_project_links(self) -> dict | None:
+        """Get all project links.
+
+        Returns
+        -------
+            dict | None: All link's name to link.
+        """
+        project_urls: dict | None = self.base_metadata.get("info", None).get("project_urls", None)
```
tromai, line 92: This is not correct. Because if `self.base_metadata.get("info", None)` returns None, calling `get("project_urls", None)` on a NoneType value would raise error.
```python
>>> a = {}
>>> a.get("A").get("B")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'get'
```

tromai: However, if we are switch to https://github.com/oracle/macaron/blob/e214326697aee6e31f0c9e406b11344c8619581e/src/macaron/json_tools.py#L15, such issues can be prevented.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def get_package_page(self) -> str | None:
+        """Implement custom API to get package main page.
+
+        Returns
+        -------
+            str | None: Package main page.
+        """
+        url = os.path.join(self.base_url, "project", self.package)
+        response = send_get_http_raw(url, headers=self.headers)
```
tromai, line 137: @behnazh-w The `send_get_http_raw` function is performing the rate limit check using this function https://github.com/oracle/macaron/blob/31f1c87390d465d69bb24dbe5a500aafbc00e7f8/src/macaron/util.py#L115. However, the rate limit check logic here is specific for GitHub API. I think we need to do something about it :thinking: .
@Yao-Wen-Chang How does PyPI handles rate limit? Do they handle it differently to what GitHub API is doing? - https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -0,0 +1,198 @@
...
+
+        Returns
+        -------
+            list | None: Maintainers.
+        """
+        package_page: str | None = self.get_package_page()
+        if package_page is None:
+            return None
+        soup = BeautifulSoup(package_page, "html.parser")
+        maintainers = soup.find_all("span", class_="sidebar-section__user-gravatar-text")
```
tromai, line 154: Is it safe to annotate `maintainers: ResultSet[Tag]` here :thinking: ?

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
tromai: Furthermore, I have changed `base_url` (i.e `https://pypi.org`) to `hostname` (i.e. `pypi.org`).
Therefore, instead of joining the values to create the final url, you can leverage `urlunsplit` - https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlunsplit.
For example:
```def get_base_metadata(self, package_name: str) -> dict:
        """Get the metadata through PyPI official API.

        Returns
        -------
            dict | None: PyPI official metadata.
        """
        url = urlunsplit(
            SplitResult(
                scheme="https",
                netloc=self.hostname,
                path=f"/pypi/{package_name}.json",
                query="",
                fragment="",
            )
        )
        response: dict = send_get_http(url, headers=self.headers)
        return response
```

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -0,0 +1,198 @@
...
+            str | None: Package main page.
+        """
+        url = os.path.join(self.base_url, "project", self.package)
+        response = send_get_http_raw(url, headers=self.headers)
+        if response:
+            html_snippets = response.content.decode("utf-8")
+            return html_snippets
+        return None
+
+    def get_maintainer_of_package(self) -> list | None:
```
tromai, line 143: It's okay to assume that the returned list always contains elements of type `str`. If that's True, then we should annotate the return type of this method as `list[str] | None`.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -0,0 +1,198 @@
...
+            html_snippets = response.content.decode("utf-8")
+            return html_snippets
+        return None
+
+    def get_maintainer_join_date(self, maintainer: str) -> datetime | None:
+        """Implement custom API to get the maintainers join date.
+
+        Parameters
+        ----------
+            maintainer (_type_): Username.
```
tromai, line 180:
```suggestion
        maintainer : str
            Username.
```

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -0,0 +1,198 @@
...
+
+        Returns
+        -------
+            datetime | None: Maintainers join date. Only recent maintainer's data available.
```
tromai, line 184:
```suggestion
        datetime | None: 
            Maintainers join date. Only recent maintainer's data available.
```

tromai: It would be good to go through all the doc strings again when you have finalized the implementation following this guide - https://oracle.github.io/macaron/pages/developers_guide/style_guide.html

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -0,0 +1,192 @@
...
+            check_id=check_id,
+            description=description,
+        )
+
+    def _should_skip(self, results: dict, dependency_heuristic: tuple[HEURISTIC, RESULT]) -> bool:
+        if results.get(dependency_heuristic[0], None) is not dependency_heuristic[1]:
+            return True
+        return False
+
+    def _analyze(self, api_client: PyPIApiClient) -> dict:
```
tromai, line 125: How about renaming this method to `run_heuristics` instead?

src/macaron/slsa_analyzer/pypi_heuristics/analysis_result.py
```diff
@@ -0,0 +1,19 @@
...
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""This file defines the heuristic-analysis result type enum, and function to modify result."""
+
+from enum import Enum
+
+
+class RESULT(Enum):
```
tromai, line 9: I think `RESULT` is a bit too generic. How about "HeuristicResult" instead?
Also, the name of **all** Enum should follow the same convention as class name (i.e. `HeuristicResult` :heavy_check_mark: `HEURISTICRESULT` :x:

src/macaron/slsa_analyzer/pypi_heuristics/analysis_result.py
tromai: If we have removed the `Analysis` class completely, how about we rename this module into `heuristic_result.py`?

src/macaron/slsa_analyzer/pypi_heuristics/base_analyzer.py
tromai: It would be good to document what is involved in the `depends_on` list. For example, what each tuple contain, the type of each element in the tuple, etc.

src/macaron/slsa_analyzer/pypi_heuristics/base_analyzer.py
```diff
@@ -0,0 +1,25 @@
...
+    """The base analyzer initialization."""
+
+    def __init__(
+        self,
+        name: str = "",
+        heuristic: HEURISTIC | None = None,
+        depends_on: list[tuple] | None = None,
+    ) -> None:
+        self.name: str = name
+        self.heuristic: HEURISTIC | None = heuristic
```
tromai, line 20: When would there be a case where `heuristic` of an analyzer is None?

src/macaron/slsa_analyzer/pypi_heuristics/base_analyzer.py
```diff
@@ -0,0 +1,25 @@
...
+        self.depends_on: list[tuple] | None = depends_on
+
+    def analyze(self) -> tuple[RESULT, dict]:
```
tromai, line 23: I am not sure what `dict` in the return tuple contains? Could you please add some doc string here to explain:
- What are the keys and what are the values of this dictionary mean?
- The type for keys and values in this dictionary.

tromai: Furthermore, if I believe correctly, every heuristic implementation must implement this method. Hence we should make this method an abstract method:
```python
@abstractmethod
def analyze(self) -> tuple[RESULT, dict]:
    raise NotImplementedError
```

src/macaron/slsa_analyzer/pypi_heuristics/base_analyzer.py
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""Define and initialize the base analyzer."""
+
+from macaron.slsa_analyzer.pypi_heuristics.analysis_result import RESULT
+from macaron.slsa_analyzer.pypi_heuristics.heuristics import HEURISTIC
+
+
+class BaseAnalyzer:
```
tromai, line 10: How about renaming it to `BaseHeuristicAnalyzer`?

src/macaron/slsa_analyzer/pypi_heuristics/metadata/closer_release_join_date.py
```diff
@@ -0,0 +1,90 @@
...
+        -------
+            list[datetime] | None: Maintainers join date.
+
+        Note
+        ----
+            Each package might have multiple maintainers.
+        """
+        maintainers: list | None = self.api_client.get_maintainer_of_package()
+        if maintainers:
+            join_date: list[datetime | None] = [
```
tromai, line 44: Will None values in this list being used somehow within this heuristic?
If not, I think it would be better if we ignore any None value and keep `join_date` with only the valid `datetime` instances.

src/macaron/slsa_analyzer/pypi_heuristics/metadata/closer_release_join_date.py
```diff
@@ -0,0 +1,90 @@
...
+            tuple[RESULT, dict]: Result and confidence.
+        """
+        maintainers_join_date: list[datetime | None] | None = self._get_maintainers_join_date()
+        latest_release_date: datetime | None = self._get_latest_release_date()
+        detail_info = {
+            "maintainers_join_date": [date.strftime("%Y-%m-%d %H:%M:%S") for date in maintainers_join_date if date]
+            if maintainers_join_date
+            else [],
+            "latest_release_date": latest_release_date.strftime("%Y-%m-%d %H:%M:%S") if latest_release_date else "",
+        }
```
tromai, line 76: This part can be simplified if `self._get_maintainers_join_date` returns a list without None values instead.

src/macaron/slsa_analyzer/pypi_heuristics/metadata/closer_release_join_date.py
```diff
@@ -0,0 +1,90 @@
...
+    def _get_latest_release_date(self) -> datetime | None:
+        """Get package's latest release date.
+
+        Returns
+        -------
+            datetime | None: Package's latest release date.
+        """
+        upload_time: str | None = self.api_client.get_latest_release_upload_time()
+        if upload_time:
+            return datetime.strptime(upload_time, "%Y-%m-%dT%H:%M:%S")
```
tromai, line 59: `datetime.strptime(upload_time, "%Y-%m-%dT%H:%M:%S")` will raise `ValueError` if `upload_time` is not in the correct format.
```
>>> from datetime import datetime
>>> datetime.strptime("AA", "%Y-%m-%dT%H:%M:%S")
...
raise ValueError("time data %r does not match format %r" %
ValueError: time data 'AA' does not match format '%Y-%m-%dT%H:%M:%S'
```
It's better if we handle the case where upload time is not valid.
In general, it's not a safe thing to always assume external input to in a valid state (the same thing can be applied to user input). In this case, we are getting the upload time from API result, which could be anything.

src/macaron/slsa_analyzer/pypi_heuristics/metadata/closer_release_join_date.py
```diff
@@ -0,0 +1,90 @@
...
+        latest_release_date: datetime | None = self._get_latest_release_date()
+        detail_info = {
+            "maintainers_join_date": [date.strftime("%Y-%m-%d %H:%M:%S") for date in maintainers_join_date if date]
+            if maintainers_join_date
+            else [],
+            "latest_release_date": latest_release_date.strftime("%Y-%m-%d %H:%M:%S") if latest_release_date else "",
+        }
+
+        if maintainers_join_date is None or latest_release_date is None:
+            return RESULT.SKIP, detail_info
```
tromai, line 79: This if statement should go before:
```python
        detail_info = {
            "maintainers_join_date": [date.strftime("%Y-%m-%d %H:%M:%S") for date in maintainers_join_date if date]
            if maintainers_join_date
            else [],
            "latest_release_date": latest_release_date.strftime("%Y-%m-%d %H:%M:%S") if latest_release_date else "",
        }
```
By doing this, after this if statement, we can safely work with `latest_release_date` and `maintainer_join_date` without checking for nullness.

src/macaron/slsa_analyzer/pypi_heuristics/metadata/closer_release_join_date.py
```diff
@@ -0,0 +1,90 @@
...
+
+        for date in maintainers_join_date:
+            if date is None:
+                continue
```
tromai, line 83: This checking for nullness could be removed if `maintainers_join_date` is a list without any None values - as mentioned in https://github.com/oracle/macaron/pull/750/files#r1634135148

src/macaron/slsa_analyzer/pypi_heuristics/metadata/empty_project_link.py
```diff
@@ -0,0 +1,45 @@
...
+        """Get total number of links.
+
+        Returns
+        -------
+            int | None: Total number of links.
+        """
+        project_links: dict | None = self.api_client.get_project_links()
+        if project_links is None:
+            return None
+        return len(project_links), project_links
```
tromai, line 29: Why do we need to return `len(project_links)` when we have already returned `project_links`? I think we can remove `len(project_links)` to simplify the interface of this method and let the caller obtain the length later on.
```python
project_links = self._get_links_total()
total = len(project_links)
```

src/macaron/slsa_analyzer/pypi_heuristics/metadata/empty_project_link.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def analyze(self) -> tuple[RESULT, dict]:
+        """Check whether the package contains one link.
```
tromai, line 32: Hmmm I believe this heuristic is for checking whether a PyPI package has no project link. I think this doc string need to be updated.

src/macaron/slsa_analyzer/pypi_heuristics/metadata/high_release_frequency.py
```diff
@@ -0,0 +1,56 @@
...
+        """
+        version_to_releases: dict | None = self.api_client.get_releases()
+        if version_to_releases is None:
+            return RESULT.SKIP, {}
+        releases_amount = len(version_to_releases)
+        extract_data: dict = {
+            version: datetime.strptime(metadata[0].get("upload_time"), "%Y-%m-%dT%H:%M:%S")
+            for version, metadata in version_to_releases.items()
+            if metadata and "upload_time" in metadata[0]
+        }
```
tromai, line 44: There are a few points:
- `datetime.strptime` should be checked for exceptions (similar to https://github.com/oracle/macaron/pull/750/files#r1634141245).
- You can use the `json_extract` function https://github.com/oracle/macaron/blob/da182aaa537367c85625f37160da570d9952c975/src/macaron/json_tools.py#L15 to to exttract `upload_time` here.
- I would highly recommend using a for loop through `version_to_releases` and popluate the `extract_data` dictionary instead of a dictionary comprehension here.

Yao-Wen-Chang: Fix this in commit: 5afebddfa30e3564ebce36a0ba9b57b9e7c7f02f.

src/macaron/slsa_analyzer/pypi_heuristics/metadata/closer_release_join_date.py
Yao-Wen-Chang: The join date is empty when the users joined many years ago.

src/macaron/slsa_analyzer/pypi_heuristics/metadata/closer_release_join_date.py
Yao-Wen-Chang: The join date is empty when the users joined many years ago.

src/macaron/slsa_analyzer/pypi_heuristics/metadata/closer_release_join_date.py
tromai: Yep, but if the join date is empty (aka `self.api_client.get_maintainer_join_date(maintainer)` returns `None`), we are definitely not going to use this `None` value later because it contains no information. Therefore, it's better to simplify it by ensuring that `join_date` always contains valid datetime instances. In this way, later on you don't have to "unwrap" the value by performing nullness checking on each element of `join_date`:
```suggest
    def _get_maintainers_join_date(self) -> list[datetime] | None:
        """Get the join date of the maintainers.

        Returns
        -------
        list[datetime] | None
            Maintainers join date.

        Note
        -------
            Each package might have multiple maintainers.
        """
        maintainers: list | None = self.api_client.get_maintainer_of_package()
        if not maintainers:
            return None

        join_dates = [] # You don't actually need to annotate here, and let the type checker automatically pick up the type
        for maintainer in maintainers:
            maintainer_join_date = self.api_client.get_maintainer_join_date(maintainer)
            if maintainer_join_date is not None:
                join_dates.append(maintainer_join_date)
                
        return join_dates
```

src/macaron/slsa_analyzer/pypi_heuristics/metadata/one_release.py
```diff
@@ -0,0 +1,45 @@
...
+        -------
+            tuple[RESULT, dict]: Result and confidence.
+        """
+        result: tuple[int, dict] | None = self._get_releases_total()
+        if result is None:
+            return RESULT.SKIP, {"releases": {}}
+
+        if result[0] == 1:
+            return RESULT.FAIL, {"releases": result[1]}  # Higher false positive, so we keep it MEDIUM
+        return RESULT.PASS, {"releases": result[1]}
```
tromai, line 45: Nit: Because we only care about number of releases, you can just call `self.api_client.get_releases()` here and count the number of element in the list.
```suggestion
        result = self.api_client.get_releases()
        if result is None:
            return RESULT.SKIP, {"releases": {}}

        if len(result) == 1:
            return RESULT.FAIL, {"releases": result}  # Higher false positive, so we keep it MEDIUM
        return RESULT.PASS, {"releases": result}
```

src/macaron/slsa_analyzer/pypi_heuristics/metadata/unchanged_release.py
```diff
@@ -0,0 +1,58 @@
...
+class UnchangedReleaseAnalyzer(BaseAnalyzer):
+    """Analyzer checks heuristic."""
+
+    def __init__(self, api_client: PyPIApiClient) -> None:
+        super().__init__(
+            name="unchanged_release_analyzer",
+            heuristic=HEURISTIC.UNCHANGED_RELEASE,
+            depends_on=[(HEURISTIC.HIGH_RELEASE_FREQUENCY, RESULT.FAIL)],  # Analyzing when this heuristic fail
+        )
+        self.hash: str = "sha256"
```
tromai, line 23: nit:
```suggestion
        self.hash_algo: str = "sha256"
```

src/macaron/slsa_analyzer/pypi_heuristics/metadata/unchanged_release.py
```diff
@@ -0,0 +1,58 @@
...
+        self.api_client = api_client
+
+    def _get_digests(self) -> list | None:
```
tromai, line 26: We should also annotate the type of element in this list ( I believe it should be `list[str]`).

src/macaron/slsa_analyzer/pypi_heuristics/metadata/unchanged_release.py
```diff
@@ -0,0 +1,58 @@
...
+
+        Returns
+        -------
+            list | None: Digests.
+        """
+        releases: dict | None = self.api_client.get_releases()
+        if releases is None:
+            return None
+        digests: list = [
+            metadata[0].get("digests").get(self.hash)
```
tromai, line 37: Please use `json_extract` for extracting data from `releases` and use a normal for loop instead of list comprehension. 
I know that list comprehension works well and doesn't have any issue. However, it adds a bit of complexity to code maintaining and readability. Writing it in a traditional for loop make your logic explicit, which can be easily picked up by other people.

src/macaron/slsa_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
```diff
@@ -0,0 +1,118 @@
...
+
+                for root, _, files in os.walk(temp_dir):
+                    for file in files:
+                        if file == "setup.py":
+                            file_path = os.path.join(root, file)
+                            with open(file_path, encoding="utf-8") as py_file:
+                                return py_file.read()
+                return None
+            except requests.exceptions.RequestException:
+                return None
```
tromai, line 80: Can you please put only the relevant code that raises this error within the try-except block for `requests.exception.RequestException`? The outer-most try-except know contains code that do multiple different things, which can be irrelevant to this exception.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -0,0 +1,198 @@
...
+            pattern = r'Joined <time datetime="([^"]+)"'
+            match = re.search(pattern, user_page)
+            # Check if a match is found
+            if match:
+                # Extract the matched timestamp
+                datetime_str = match.group(1)
+                # Define the format of the datetime string
+                datetime_format = "%Y-%m-%dT%H:%M:%S%z"
+                # Return the parsed string to a datetime object
+                return datetime.strptime(datetime_str, datetime_format).replace(tzinfo=None)
```
tromai, line 197: `datetime.strptime` should be checked for exception when `datetime_str` is in an invalid state.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -0,0 +1,198 @@
...
+        ----------
+            maintainer (_type_): Username.
+
+        Returns
+        -------
+            datetime | None: Maintainers join date. Only recent maintainer's data available.
+        """
+        user_page: str | None = self.get_maintainer_profile_page(maintainer)
+        if user_page:
+            pattern = r'Joined <time datetime="([^"]+)"'
```
tromai, line 188: Why do we need to negate `^` in within this regex pattern? Is there a case where "^" exists within the datetime string :thinking: . If that is true, I think it would really help if we document that edge case here.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
Yao-Wen-Chang: They don't have strict rules for the rate limit. 
Checkout their document https://warehouse.pypa.io/api-reference/

src/macaron/slsa_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
Yao-Wen-Chang: On line 51, requests for the tarball might cause an error.

src/macaron/slsa_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
Yao-Wen-Chang: Fixed in 227dd058003abeaec2a476b5c35348b9d62869fa

src/macaron/slsa_analyzer/pypi_heuristics/base_analyzer.py
Yao-Wen-Chang: Fixed in 256082bcda803273d241a5c3ab5d287bd8f3d96e

src/macaron/slsa_analyzer/pypi_heuristics/analysis_result.py
Yao-Wen-Chang: Fixed this in 44e8ad65396270f1616d60f5833133d9087be151

src/macaron/slsa_analyzer/pypi_heuristics/base_analyzer.py
Yao-Wen-Chang: Fixed this in bcb627b6aa81b41663b39a0c62468c313505c469

src/macaron/slsa_analyzer/pypi_heuristics/base_analyzer.py
Yao-Wen-Chang: Fixed this in bcb627b6aa81b41663b39a0c62468c313505c469

src/macaron/slsa_analyzer/pypi_heuristics/base_analyzer.py
Yao-Wen-Chang: Fixed this in c965a2502a387321fc099f393652b1125c9cd47e

src/macaron/slsa_analyzer/pypi_heuristics/metadata/closer_release_join_date.py
Yao-Wen-Chang: Fixed this in 19d51b9916e4f33b13940681e58af4df9b991269

src/macaron/slsa_analyzer/pypi_heuristics/metadata/closer_release_join_date.py
Yao-Wen-Chang: Fixed this in ea48367e7b773cb2ac6cfae4b7609b71706179b1

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
Yao-Wen-Chang: Fixed this in ea48367e7b773cb2ac6cfae4b7609b71706179b1

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
Yao-Wen-Chang: Fixed this in f900dfef1bae0413a4474f9e8e9101db3c625af1
Modified regex to bs4.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
Yao-Wen-Chang: Fixed this in f900dfef1bae0413a4474f9e8e9101db3c625af1

tests/malware_analyzer/pypi/test_empty_project_link_analyzer.py
Yao-Wen-Chang: Fixed this in 
2505219163d42f3f08d11ff830fb654152703818
9943158edf9cc9e0ec10499096523a8fb974ab60
00f925762e739e41798e6b12351e3e68fec48820

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
Yao-Wen-Chang: Fixed this in f900dfef1bae0413a4474f9e8e9101db3c625af1

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
Yao-Wen-Chang: Fixed this in f900dfef1bae0413a4474f9e8e9101db3c625af1

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
Yao-Wen-Chang: Fixed this in f900dfef1bae0413a4474f9e8e9101db3c625af1

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
Yao-Wen-Chang: Fixed this in 26b8e140bf2de22b8edc57f66d5398fa150423cf

src/macaron/slsa_analyzer/pypi_heuristics/metadata/empty_project_link.py
Yao-Wen-Chang: Fixed this in 71f87ac0e2f6c6fed616a773e43a40bb642e990a

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
```diff
@@ -0,0 +1,117 @@
...
+
+                for root, _, files in os.walk(temp_dir):
+                    for file in files:
+                        if file == "setup.py":
+                            file_path = os.path.join(root, file)
+                            with open(file_path, encoding="utf-8") as py_file:
+                                return py_file.read()
+                return None
+            except requests.exceptions.RequestException:
+                return None
```
tromai, line 79: Could you please only put this try catch block for `requests.exceptions.RequestException` for the code that can raise this exception.
I think this is the only relevant line:
```
 response = requests.get(sourcecode_url, stream=True, timeout=40)
```

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
```diff
@@ -0,0 +1,117 @@
...
+            str | None: Source code.
+        """
+        # Create a temporary directory to store the downloaded source
+        with tempfile.TemporaryDirectory() as temp_dir:
+            try:
+                sourcecode_url = api_client.get_sourcecode_url()
+                if sourcecode_url is None:
+                    return None
+                response = requests.get(sourcecode_url, stream=True, timeout=40)
+                response.raise_for_status()
```
tromai, line 51: According to the documentation at https://requests.readthedocs.io/en/latest/api/#requests.Response.raise_for_status, `raise_for_status` can raise `HTTPError`.  It would be good if we add a try catch block for that exception of `response.raise_for_status()`.

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
```diff
@@ -0,0 +1,117 @@
...
+                filename = sourcecode_url.split("/")[-1]
```
tromai, line 52: Would there be a case where `sourcecode_url` is an empty string ? If `sourcecode_url` is an empty string, `sourcecode_url.split("/")[-1]` would raise `IndexError`.

src/macaron/malware_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -0,0 +1,250 @@
...
+        ----
+            results (dict[HEURISTIC, HeuristicResult]): Containing all heuristic results, where the key is the heuristic and the
+            value is the result associated with that heuristic.
+            depends_on (list[tuple[HEURISTIC, HeuristicResult]]): containing heuristics that the current heuristic depends on,
+            along with their expected results.
+
+        Returns
+        -------
+            bool: Returns True if any result of the dependency heuristic does not match the expected result.
+            Otherwise, returns False.
```
tromai, line 164:
```suggestion
        Parameters
        -----------------
        results : dict[HEURISTIC, HeuristicResult])
            Containing all heuristic results, where the key is the heuristic and the
            value is the result associated with that heuristic.
        depends_on : list[tuple[HEURISTIC, HeuristicResult]])
            Containing heuristics that the current heuristic depends on,
            along with their expected results.

        Returns
        -----------
        bool
            Returns True if any result of the dependency heuristic does not match the expected result.
            Otherwise, returns False.
```

src/macaron/malware_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -0,0 +1,250 @@
...
+    ) -> tuple[dict[HEURISTIC, HeuristicResult], dict[str, int | dict]]:
+        """Run the main logic of heuristics analysis.
+
+        Args
+        ----
+            api_client (PyPIRegistry): The PyPI API client object used to interact with the official PyPI API.
+
+        Returns
+        -------
+            tuple[dict[HEURISTIC, HeuristicResult], dict[str, Any]]: Containing the heuristic results and relevant metadata.
```
tromai, line 183:
```suggestion
        Parameters
        -----------------
        api_client: PyPIRegistry)
            The PyPI API client object used to interact with the official PyPI API.

        Returns
        -----------
        tuple[dict[HEURISTIC, HeuristicResult], dict[str, Any]]
            Containing the heuristic results and relevant metadata.
```

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -0,0 +1,292 @@
...
+from macaron.slsa_analyzer.build_tool import Pip, Poetry
+from macaron.slsa_analyzer.build_tool.base_build_tool import BaseBuildTool
+from macaron.slsa_analyzer.package_registry.package_registry import PackageRegistry
+from macaron.util import send_get_http_raw
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class PyPIRegistry(PackageRegistry):
+    """This class implements the npm package registry.
```
tromai, line 27:
```suggestion
    """This class implements the PyPI package registry.
```

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -0,0 +1,292 @@
...
+
+    There is no complete and up-to-date API documentation for the npm registry and the endpoints
+    are discovered by manual inspection of links on https://www.npmjs.com.
```
tromai, line 30: We should probably remove these 2 lines too.

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
Yao-Wen-Chang: I will change folder name to "tmp", so we don't need to handle this.

src/macaron/malware_analyzer/pypi_heuristics/base_analyzer.py
```diff
@@ -0,0 +1,38 @@
...
+        self,
+        name: str,
+        heuristic: HEURISTIC,
+        depends_on: list[tuple[HEURISTIC, HeuristicResult]] | None,
+    ) -> None:
+        self.name: str = name
+        self.heuristic: HEURISTIC = heuristic
+        self.depends_on: list[
+            tuple[HEURISTIC, HeuristicResult]
+        ] | None = depends_on  # Contains the dependent heuristics and the expected result of each heuristic
```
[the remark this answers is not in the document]
Yao-Wen-Chang: I will refine all the docstring in the next update, thanks for the suggestion.

pyproject.toml
Yao-Wen-Chang: Should I add both or just the `types-beautifulsoup4`?

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
Yao-Wen-Chang: Implemented this in ea48367e7b773cb2ac6cfae4b7609b71706179b1

src/macaron/slsa_analyzer/pypi_heuristics/analysis_result.py
Yao-Wen-Chang: Fixed this in 29004717395368e70f662f64dfabb95a7c920d3b

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
Yao-Wen-Chang: Fixed this in 26b8e140bf2de22b8edc57f66d5398fa150423cf

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
Yao-Wen-Chang: Fixed this in f900dfef1bae0413a4474f9e8e9101db3c625af1

src/macaron/slsa_analyzer/pypi_heuristics/metadata/closer_release_join_date.py
Yao-Wen-Chang: Fixed this in 19d51b9916e4f33b13940681e58af4df9b991269

src/macaron/slsa_analyzer/pypi_heuristics/metadata/empty_project_link.py
Yao-Wen-Chang: Fixed this in 54e5d3389622787e56127725a1ef7317f317a4e0

src/macaron/slsa_analyzer/pypi_heuristics/metadata/one_release.py
Yao-Wen-Chang: Fixed this in c965a2502a387321fc099f393652b1125c9cd47e

src/macaron/slsa_analyzer/pypi_heuristics/metadata/unchanged_release.py
Yao-Wen-Chang: Fixed this in 1d801c45b3f13c3cdb3e603935c314b1687cebc3

src/macaron/slsa_analyzer/pypi_heuristics/metadata/unchanged_release.py
Yao-Wen-Chang: Fixed this in 1d801c45b3f13c3cdb3e603935c314b1687cebc3

src/macaron/slsa_analyzer/pypi_heuristics/metadata/unchanged_release.py
Yao-Wen-Chang: Fixed this in 1d801c45b3f13c3cdb3e603935c314b1687cebc3

src/macaron/malware_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -0,0 +1,250 @@
...
+        description = "Check whether the features of package adhere to the heurisic."
+        super().__init__(
+            check_id=check_id,
+            description=description,
+        )
+
+    def _should_skip(
+        self, results: dict[HEURISTIC, HeuristicResult], depends_on: list[tuple[HEURISTIC, HeuristicResult]]
+    ) -> bool:
+        """Determine whether a particular heuristic result should be skipped based on the provided dependency heuristics.
```
tromai, line 152: It would be good to mention that the skip status of a heuristics is depended on the **status** of heuristics that it depends on.

src/macaron/malware_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -0,0 +1,250 @@
...
+        """Initialize a check instance."""
+        check_id = "mcn_detect_malicious_metadata_1"
+        description = "Check whether the features of package adhere to the heurisic."
+        super().__init__(
+            check_id=check_id,
+            description=description,
+        )
+
+    def _should_skip(
+        self, results: dict[HEURISTIC, HeuristicResult], depends_on: list[tuple[HEURISTIC, HeuristicResult]]
```
tromai, line 150: I think `depends_on` is ambiguous as it doesn't represents that it contains the results of "parent" heuristics. How about we rename this parameter to something better represent that? (e.g. `predecessor_results`). :thinking: 
@behnazh-w What do you think about it?

tests/malware_analyzer/pypi/test_empty_project_link_analyzer.py
tromai: Thanks for changing them. I will now look into the test content itself.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
tromai: I continue this discussion here - https://github.com/oracle/macaron/pull/750#discussion_r1642036458

src/macaron/malware_analyzer/pypi_heuristics/base_analyzer.py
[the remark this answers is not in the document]
tromai: Ah I see, so it's still pending work. No worries, I just put them here so that we don't forget (as already mentioned in https://github.com/oracle/macaron/pull/750#discussion_r1634024799).

src/macaron/slsa_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
tromai: I continue this discussion here - https://github.com/oracle/macaron/pull/750#discussion_r1642036458

pyproject.toml
tromai: Oh I meant you keep `beautifulsoup4` right where it currently is an **only** move `types-beautifulsoup4` to `[project.optional-dependencies].dev` in `pyproject.toml`. This is so that   `types-beautifulsoup4`  is only installed in the dev environment, and not in the final Macaron artifact that is installed within the Docker image.

src/macaron/malware_analyzer/pypi_heuristics/analysis_result.py
```diff
@@ -0,0 +1,19 @@
...
+
+from enum import Enum
+
+
+class HeuristicResult(Enum):
+    """Result type.
+
+    PASS: Not suspicious
+    FAIL: Suspicious
+    SKIP: Metadata miss
```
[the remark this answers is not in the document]
Yao-Wen-Chang: How about like this: 
```
    """Result type indicating the outcome of a heuristic.

    Attributes:
        PASS: Indicates that no suspicious activity was detected.
        FAIL: Indicates that suspicious activity was detected.
        SKIP: Indicates that the heuristic check could not be performed due to missing metadata.

    The `SKIP` result occurs when the necessary metadata is not available. This often happens 
    when fetching data through the PyPI API and the relevant data, such as the maintainer's 
    join date or release information, is missing or unavailable.
    """
```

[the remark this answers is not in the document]
tromai: Thanks! That sounds good to me.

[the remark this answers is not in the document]
Yao-Wen-Chang: Fixed this in ac73df0e695e9bfce3bf0cbbd05271ec5bfd4a88.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
Yao-Wen-Chang: Fixed this in 183aa83367ebb5be64cd2cb5bd27705258741058.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
Yao-Wen-Chang: Fixed this in 183aa83367ebb5be64cd2cb5bd27705258741058.

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
Yao-Wen-Chang: Fixed this in 3d825e9fa5a2556305833a92a60b3633899083b2.

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
Yao-Wen-Chang: Fixed this in d03b7d86bf804d5fa662ee6a1fb6bd53b0d62498.

src/macaron/malware_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -0,0 +1,271 @@
...
+    heuristic_result: Mapped[str] = mapped_column(
+        String, nullable=False, info={"justification": JustificationType.TEXT}
+    )
+
+    __mapper_args__ = {
+        "polymorphic_identity": "_detect_malicious_metadata_check",
+    }
+
+
+ANALYZERS: list = [
```
behnazh-w, line 57: Please add comments before each list and dictionary and explain what they contain.

Yao-Wen-Chang: Fixed this in b7e6ee15cf3d8b678a252f988f731fc5c173e296

tromai (approved):
LGTM. Thanks for the great contribution!

behnazh-w (approved):

PR #760, opened 2024-06-06
chore: convert expectation files to policies and use in integration tests

benmss:
This PR converts the integration test expectation files used for output and dependency comparison into policy files that can be verified using Macaron's `verify-policy` command. It also modifies the integration test script to use the new policy files.


tests/dependency_analyzer/expected_results/cyclonedx_apache_maven.dl
```diff
@@ -0,0 +1,43 @@
+/* Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved. */
+/* Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/. */
+
+#include "prelude.dl"
+
+Policy("check-dependencies", component_id, "Check the dependencies of pkg:github.com/apache/maven@3fc399318edef0d5ba593723a24fff64291d6f9b") :-
```
nathanwn, line 6: Is it worth converting this dependency resolution result to Souffle? I wonder what would happen if we wanted to create another test case of this type. Are we supposed to write this kind of result by hand?

nathanwn: Let me rephrase the question.
I think it is better if I start by asking: What do you think is being tested with this test case? If this test case was meant to test the exhaustive list of dependencies, then I'm not sure if converting it to Souffle is a good idea.

tests/dependency_analyzer/expected_results/skipdep_apache_maven.dl
```diff
@@ -0,0 +1,10 @@
+/* Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved. */
+/* Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/. */
+
+#include "prelude.dl"
+
+Policy("check-dependencies", component_id, "Check the dependencies of pkg:github.com/apache/maven@3fc399318edef0d5ba593723a24fff64291d6f9b") :-
+    .
```
nathanwn, line 7: Hmm why is this policy empty? Is this expected?

scripts/dev_scripts/integration_tests.sh
```diff
@@ -24,6 +25,11 @@ if [ $# -eq 3 ] && [ "$3" == "--update" ] ; then
     COMPARE_VSA="$COMPARE_VSA --update"
 fi
 
+function run_macaron() {
```
nathanwn, line 28: Nit: Can we name this something a bit more descriptive? How about `clean_run_macaron` or something similar?

tests/dependency_analyzer/expected_results/skipdep_apache_maven.dl
benmss: It seems the file this policy was created from was empty. Might have to manually create a policy for that case.

tests/dependency_analyzer/expected_results/cyclonedx_apache_maven.dl
benmss: All of the check dependency policies were converted from JSON files being passed into the `compare_dependencies.py` e2e script. As far as I'm aware, they exist to exhaustively test dependencies. If we want to change that, I think that would be best done in another PR. The same would apply to other possible optimisations/improvements that have been discussed.

nathanwn: What I'm checking here is whether we should convert dependency resolution tests to Souffle, or keep the old way with JSON output and compare with the `compare_dependencies.py` script. Since the expected output for dependency results is rather large and tedious to update by hand, my subjective opinion is keeping the JSON output may be still the better option for now, although others may argue otherwise. If using Souffle here provides no obvious upsides compared to the old way, then I think we may keep it as is. I'm not suggesting any further improvement to be done in this PR otherwise.

tests/dependency_analyzer/expected_results/skipdep_apache_maven.dl
nathanwn: This looks to be converted from a dependency result file. Can you elaborate on how you convert a dependency result file to a policy file? Thanks.

src/macaron/policy_engine/prelude/helper_rules.dl
```diff
@@ -9,6 +9,9 @@
 .decl check_passed(component_id: number, check_name: symbol)
 check_passed(component_id, check_name) :- check_result(_, check_name, 1, component_id).
 
+.decl check_failed(component_id: number, check_name: symbol)
+check_failed(component_id, check_name) :- check_result(_, check_name, 0, component_id).
```
nathanwn, line 13: Have you checked what the `check_result` fact would be, or if it is even produced in the first place, in case a check is skipped (e.g. due to the include/exclude check configuration)?

tests/dependency_analyzer/expected_results/apache_maven_with_sbom_provided.dl
```diff
@@ -0,0 +1,12 @@
+/* Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved. */
+/* Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/. */
+
+#include "prelude.dl"
+
+Policy("check-dependencies", component_id, "Check the dependencies of pkg:github.com/timyarkov/multibuild_test@a8b0efe24298bc81f63217aaa84776c3d48976c5") :-
+    transitive_dependency(component_id, dep_0),
+    is_component(dep_0, "pkg:maven/org.junit.jupiter/junit-jupiter-engine@5.8.1?type=jar"),transitive_dependency(component_id, dep_1),
+    is_component(dep_1, "pkg:maven/org.hamcrest/hamcrest-core@2.2?type=jar").
```
nicallen, line 9: I presume this corresponds to `apache_maven_with_sbom_provided.json`? I see that the original also checks for a bunch of other dependencies with `"available": "MISSING REPO URL"` that are not included in the policy version. What is the reason for this?

Also, this policy checks for the presence of the expected dependencies, but does not forbid additional unexpected dependencies (whereas the original test did).

nicallen: Also, it would be more readable if it were not formatted with the `transitive_dependency` for the next component on the same line as the last one.

src/macaron/policy_engine/prelude/helper_rules.dl
benmss: The policy verification will fail if any `check_passed` or `check_failed` checks are excluded. Although I had not checked until you mentioned it here.

tests/dependency_analyzer/expected_results/cyclonedx_apache_maven.dl
benmss: The script I used to create policies from expected dependency results could be cleaned up and made available. That would be another option for reducing tedium.

tests/dependency_analyzer/expected_results/skipdep_apache_maven.dl
benmss: The dependency result files were bulk converted using a script. It takes entries in the files and turns them into lines within the final policy. In this case the file was an empty list so no policy lines were created.

tests/dependency_analyzer/expected_results/apache_maven_with_sbom_provided.dl
benmss: Looks like dependency related policies will not be included and we will continue to use JSON comparisons for those tests.

src/macaron/policy_engine/prelude/helper_rules.dl
nathanwn: I see. So, excluded checks will not produce any facts. Thanks.

nicallen (approved):

nathanwn (approved):
Thanks for the PR.

src/macaron/policy_engine/prelude/helper_rules.dl
```diff
@@ -42,6 +45,12 @@ transitive_dependency(parent, dependency) :-
 .decl is_repo(repo_id: number, repo_complete_name: symbol, component_id: number)
  is_repo(repo_id, repo_complete_name, component_id) :- repository(repo_id, component_id, repo_complete_name, _, _, _, _, _, _, _, _, _, _).
 
+/**
+ *  Extract the id and PURL from the repository relation.
+ */
+.decl is_repo_purl(repo_id: number, purl: symbol)
+ is_repo_purl(repo_id, purl) :- repository(_, repo_id, _, _, _, _, _, purl, _, _, _, _, _).
```
behnazh-w, line 52: We don't store `purl` in the repository` table. We store the remote path, which is the repository URL.

```suggestion
/**
 *  Extract the id and URL from the repository relation.
 */
.decl is_repo_url(repo_id: number, url: symbol)
 is_repo_url(repo_id, url) :- repository(_, repo_id, _, _, _, _, _, url, _, _, _, _, _).
```

behnazh-w:
We are still comparing the JSON outputs in `integration_tests_docker.sh`. Can you please adjust these tests as well? I think we should be able to remove the JSON outputs from the repository in this PR.

behnazh-w (approved):

behnazh-w (approved):

PR #764, opened 2024-06-11
test: add new integration test utility

nathanwn:
This pull request adds a new integration test utility script that:
* Introduces a new YAML schema for specifying integration tests, which offers a few advantages over testing through bash. The script generally helps avoid certain kinds of mistakes observed previously while working with bash scripts, such as forgetting to enforce the exit code of a shell command.
* Streamlines the process of developing and running integration tests locally, which was suboptimal previously.
* Provides some level of fail-fast static validation for test cases to help save time during local development as well as CI in case unintentional mistakes happen when test cases are created or changed.
* Helps organize individual test cases into separate directories. This means the output of each test case, including the database, is separated from one another, which enables testing on specific states of the database.
* Allows testing not just the `macaron` Python package, but also any wrapper script of Macaron (including but not limited to the `run_macaron.sh` script).
* Allows selecting/deselecting a subset of test cases to run through the tagging mechanism. This is helpful for some purposes, such as sharing test cases between the Python package and the `run_macaron.sh` wrapper script, or disabling certain test cases when some services are unreachable.

This pull request also converts a large number of test cases from the old integration test setup to this new setup with the utility. Some conversions may conflict with some changes made in #748, but hopefully not be too hard to resolve; otherwise, we can selectively drop certain commits from this PR before merging and proceed with the conversion later.

# Changes

## The utility script and test data

The utility script and test data are under the `tests/integration` directory. Detailed instructions for the utility script can be found in the [`README.md`](https://github.com/oracle/macaron/blob/nathanwn/integration-test-utility/tests/integration/README.md) file under the same directory.

Most test cases specified in YAML config are converted from the old `integration_test.sh` and `integration_test_docker.sh` script. For each converted test case, I made it a separate commit for easy review. Please refer to the list of commits when you review the conversion.

## Changes to Makefile

Instead of completely replacing `scripts/dev_script/{integration_tests,integration_test_docker}.sh`, the new utility script is now called within both scripts, since this PR only converts parts of these scripts and not fully.

Since the `{integration_tests,integration_test_docker}.sh` are not fully replaced, no Makefile change is required.

## Changes to GitHub Actions Workflows

The integration test utility script requires two dependencies already available in the development virtual environment: `ruamel.yaml` for YAML parsing and `cfgv` for human-readable YAML validation errors.

The script also requires Python 3.11, although compatibility with other Python versions has not been tested. Another pending pull request, #748, is also adding Python to the `.github/workflows/_build_docker.yml` workflow file, and so our change is similar here.

## Changes to Documentation

Besides the [`./tests/integration/README.md`](https://github.com/oracle/macaron/blob/nathanwn/integration-test-utility/tests/integration/README.md) file, reference was made to the new integration test utility in the `Writing a New Check` section on the `Macaron Developer's Guide` website page.

## Commits

The first commit in this PR includes the integration test utility, GitHub Actions workflow changes, and documentation changes. Each later PR converts a test case from the previous setup to this new setup.


tests/integration/README.md
```diff
@@ -0,0 +1,219 @@
...
+### Create a new test case
+
+You create a new test case by creating a new directory, then a `test.yaml` within it. To schema-validate the `test.yaml` file, you can use the `check` command and point to the test case directory:
+
+```bash
+# Schema-validate the ./test/case/directory/test.yaml file.
+$ python3 ./tests/integration/run.py check ./test/case/directory
+```
+
+At this point, some expected result files does not exist yet, since you normally want to run `macaron` once, inspect the result files, then turn them to expected result files if they look good. To do this, you can run in **interactive** mode. In this mode, the utility stops at each steps and ask if you want to run or skip a step. For compare steps, the utility also ask if you want to "update" the expected result file instead of compare.
```
tromai, line 87: nit:
```suggestion
At this point, some expected result files does not exist yet, since you normally want to run `macaron` once, inspect the result files, then turn them to expected result files if they look good. To do this, you can run in **interactive** mode. In this mode, the utility stops at each steps and ask if you want to run or skip a step. For `compare` steps, the utility also ask if you want to "update" the expected result file instead of compare.
```

tests/integration/README.md
```diff
@@ -0,0 +1,219 @@
...
+```bash
+# Only run test cases not tagged with `npm`.
+$ python3 ./tests/integration/run.py run --exclude-tag npm ./all/cases/...
+```
+
+You can simply think of each `--include-tag`/`--exclude-tag` argument as adding an additional constraint that a selected test case must satisfy".
+
+### Debug utility script
+
+In case you want to debug the utility script itself, there is the verbose mode for all commands which can be enabled with the `-v/--verbose` flag.
```
tromai, line 171: nit: we can also mention that `-v/--verbose` flag is provided after the command name, not before.
- `python3 tests/integration/run.py vet -v ...` :heavy_check_mark: 
- `python3 tests/integration/run.py -v vet ...` :x:

tests/integration/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+
+## Test Config Reference
+
+### Test case Schema
+
+* `description` (`string`, required): The description of the test case.
+* `tags` (`array[string]`, optional, default is `[]`): The tags of the test case. When the `--tag <tag>` argument is passed with the `run` command , only run test cases having the tag `<tag>`. (The `--tag` argument is typically used in combination with test case discovery using the `...` wildcard).
```
tromai, line 179: I think `--tag` flag is not currently available for any command. Perhaps this hasn't been updated to the latest `--include-tag` and `--exclude-tag` flags.

tests/integration/README.md
```diff
@@ -0,0 +1,219 @@
...
+* `steps` (`array[string]`, required): The list of steps in a test case. Steps in a test case are executed sequentially. A test case stops execution and fails if any command fails.
```
tromai, line 180:
```suggestion
* `steps` (`array[maps]`, required): The list of steps in a test case. Steps in a test case are executed sequentially. A test case stops execution and fails if any command fails.
```

tests/integration/README.md
benmss:
```suggestion
At this point, some expected result files do not exist yet, since you normally want to run `macaron` once, inspect the result files, then turn them into expected result files if they look good enough. To do this, you can run in **interactive** mode. In this mode, the utility stops at each step and asks if you want to run or skip a step. For `compare` steps, the utility also asks if you want to "update" the expected result file instead of compare.
```

tests/integration/README.md
nathanwn: This is correct. I previously had `--include-tag` as `--tag`.

tests/integration/run.py
```diff
@@ -0,0 +1,984 @@
...
+    def _check(v: str) -> None:
+        filepath = os.path.join(cwd, v)
+        if not os.path.isfile(filepath):
+            raise cfgv.ValidationError(f"File {filepath} does not exist.")
+
+    return cast(Callable[[str], None], cfgv.check_and(cfgv.check_string, _check))
+
+
+def check_env(env: object) -> None:
+    """Check for a required file of a test case."""
```
tromai, line 97: I think this doc string needs to be updated.

tests/integration/run.py
```diff
@@ -0,0 +1,984 @@
...
+                        "shell",
+                        "compare",
+                        "analyze",
+                        "verify",
+                    ),
+                ),
+            ),
+            cfgv.ConditionalRecurse(
+                condition_key="kind",
+                condition_value="command",
```
tromai, line 476: I think this parameter should be `condition_value="shell"`.

tests/integration/run.py
```diff
@@ -0,0 +1,984 @@
...
+    return test_cases
+
+
+def do_check(test_case_dirs: list[str], check_expected_result_files: bool) -> int:
+    """Execute the check command."""
+    test_cases = load_test_cases(
+        test_case_dirs,
+        check_expected_result_files=check_expected_result_files,
+    )
+    if test_cases is None:
```
tromai, line 772: nit
```suggestion
    if test_cases is None:
        logger.error("Error encountered while loading test config.")
```

tests/integration/run.py
```diff
@@ -0,0 +1,984 @@
...
+    exclude_tags: list[str],
+) -> int:
+    """Execute the update command."""
+    test_cases = load_test_cases(
+        test_case_dirs,
+        check_expected_result_files=False,
+        include_tags=include_tags,
+        exclude_tags=exclude_tags,
+    )
+    if test_cases is None:
```
tromai, line 825: Same as https://github.com/oracle/macaron/pull/764/files#r1646827933

tests/integration/run.py
```diff
@@ -0,0 +1,984 @@
...
+
+COMPARE_SCRIPTS: dict[str, Sequence[str]] = {
+    "analysis_report": ["tests", "e2e", "compare_e2e_result.py"],
+    "policy_report": ["tests", "policy_engine", "compare_policy_reports.py"],
+    "deps_report": ["tests", "dependency_analyzer", "compare_dependencies.py"],
+    "vsa": ["tests", "vsa", "compare_vsa.py"],
+}
+
+
+def check_required_file(cwd: str) -> Callable[[str], None]:
```
tromai, line 85: Just a small observation, this function is not used for checking the existing of the yaml input config file `macaron -c <...yaml>`. Given that we are looking forward to remove the input yaml config file, it makes sense to ignore it from this utility.

tromai:
After our discussion, we have decided that we are going to wrap up this PR by fixing the merge conflicts and merge it to staging. Any pending comments and test cases that haven't been converted can be resolved in a subsequent PR.

tests/integration/README.md
nathanwn: I am probably not fixing this unless there is any other objection. The `--verbose` flag here is a subcommand-specific option and this is not a super abnormal behavior in general. Users should also utilize the help menu with `-h` in the command line for detailed usage, which already covers the available options under each subcommand.

tromai: Sounds good to me. We can leave it as it is.

tests/integration/run.py
nathanwn: This is correct.
I intentionally left it out since this use case involving `-c` is probably "deprecated" and we only maintain the related test cases temporarily until we resolve the related issues.

tests/integration/run.py
nathanwn: In 63829f3c6481f007d68a91632f92b3fe02fc4143, I have moved error-logging into the `load_test_cases` function. Consumers of this function should not need to add additional logging.

tests/integration/README.md
nathanwn: Fixed in a9fb833b679adc0d7ff797622561375ea451c692.

tests/integration/README.md
nathanwn: Thanks a lot. Fixed in 0a816d8dd918b77837b5c4e48cde4f689a37b630.

tests/integration/run.py
```diff
@@ -0,0 +1,984 @@
...
+
+class InvalidConfigError(cfgv.ValidationError):
+    """Error raised when there is a schema error in a test config file."""
+
+
+def load_config(
+    test_config_dir: str,
+    check_expected_result_files: bool,
+) -> CaseConfig:
+    """Test a ``test.yaml`` file."""
```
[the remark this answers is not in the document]
nathanwn: Fixed in 3662a504fc57a448b1cc70beec676e809e6ea71f.

tests/integration/README.md
nathanwn: Fixed in 04815dcdb782165e4912be6d7cd00687da348786.

tests/integration/run.py
nathanwn: Fixed in a11d365a6de339b78d91392be708900d87b9f1d4.

tests/integration/run.py
nathanwn: Good catch. Fixed in 46634c86ae3168ef7f997532bc0da0e5311bee44.

nathanwn:
Rebased to resolve conflicts with `staging` after all comments have been addressed.

tromai (approved):
LGTM. Thanks for the change.

behnazh-w:
> Rebased to resolve conflicts with `staging` after all comments have been addressed.

Thanks a lot :slightly_smiling_face:

behnazh-w (approved):

PR #768, opened 2024-06-13
fix: force docker to use linux/amd64 platform

### LINKED ISSUE #767: Make sure that docker images can run on ARM platforms
Currently we only have amd64 docker images which are not able to run on arm64 devices by default. As an immediate fix we can force docker (or podman) to use `linux/amd64` platform which makes them to use emulation (`Rosetta 2` under mac and `QEMU` under linux).

mabdollahpour-ol:
Using `--platform linux/amd64` makes it possible to run amd64 images on arm64 platforms through emulation

Closes #767

### DIFF
--- a/scripts/release_scripts/run_macaron.sh
+++ b/scripts/release_scripts/run_macaron.sh
@@ -584,6 +584,7 @@ else
     fi
 fi
 
+# Force docker to use linux/amd64 platform in order to make docker use emulation on ARM host platforms.
 docker run \
     --platform=linux/amd64 \
     --network=host \
--- a/scripts/release_scripts/run_macaron.sh
+++ b/scripts/release_scripts/run_macaron.sh
@@ -585,6 +585,7 @@ else
 fi
 
 docker run \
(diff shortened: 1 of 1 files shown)


scripts/release_scripts/run_macaron.sh
```diff
@@ -585,6 +585,7 @@ else
 fi
 
 docker run \
+    --platform=linux/amd64 \
```
behnazh-w, line 588: Can you please add a comment explaining why this option is added?

mabdollahpour-ol: It seems it's not possible to add comment in the middle of a command in bash. I have to put this flag in a separate variable and put the comment on top of that. Is that ok?

behnazh-w: You can just add the comment before the `docker run` command. No need to add a variable.

behnazh-w (approved):

tromai (approved):
LGTM. Thanks for the change.

PR #780, opened 2024-06-25
chore: improve token support in run_macaron.sh

tromai:
This Pull Request improves token file operations in `run_macaron.sh`.

- Moved the token file creation to the end of `run_macaron.sh` so that if the script exits early, the token file is not persisted within the file system - https://github.com/oracle/macaron/pull/780/commits/f88f00ab4bf3544e202a8e0e6b7a624473883fbc
- Made sure that `docker run` if returns a non-zero status code doesn't trigger early exit (enable by `set -e`). This enables us to perform the cleaning up before exiting `run_macaron.sh`. Furthermore, the return status code of `run_macaron.sh` follows that of `docker run` - https://github.com/oracle/macaron/pull/780/commits/af4173ecd83a764aed279c41c7a12b84e79aa111
- Added integration test cases to check for token file persistence - https://github.com/oracle/macaron/pull/780/commits/62b8098cfe235a8fc7bb75ff9217d140c1d66d06
- Added the token file to gitignore - https://github.com/oracle/macaron/pull/780/commits/c2b83a5770eb9d07e59077b3d1eee17f53183c45

### DIFF
--- a/scripts/dev_scripts/integration_tests.sh
+++ b/scripts/dev_scripts/integration_tests.sh
@@ -398,7 +398,7 @@ run_macaron_clean $ANALYZE -purl pkg:maven/io.github.behnazh-w.demo/example-mave
 check_or_update_expected_output $COMPARE_DEPS $DEP_RESULT $DEP_EXPECTED || log_fail
 
 python ./tests/integration/run.py run \
-    --exclude-tag token_file_clean_up \
+    --exclude-tag docker-only \
     ./tests/integration/cases/... || log_fail
 
 # Important: This should be at the end of the file
--- a/scripts/dev_scripts/integration_tests.sh
+++ b/scripts/dev_scripts/integration_tests.sh
@@ -398,6 +398,7 @@ run_macaron_clean $ANALYZE -purl pkg:maven/io.github.behnazh-w.demo/example-mave
 check_or_update_expected_output $COMPARE_DEPS $DEP_RESULT $DEP_EXPECTED || log_fail
 
 python ./tests/integration/run.py run \
+    --exclude-tag token_file_clean_up \
     ./tests/integration/cases/... || log_fail
 
 # Important: This should be at the end of the file
--- a/scripts/release_scripts/run_macaron.sh
+++ b/scripts/release_scripts/run_macaron.sh
@@ -259,16 +259,6 @@ function mount_file() {
     mounts+=("-v" "${file_on_host}:${file_in_container}:${mount_option}")
 }
 
-# Handle tokens.
-set +u
-echo "" > ${TOKEN_FILE}
-{
-    echo "GITHUB_TOKEN=${GITHUB_TOKEN}" >> ${TOKEN_FILE}
-    echo "MCN_GITLAB_TOKEN=${MCN_GITLAB_TOKEN}" >> ${TOKEN_FILE}
-    echo "MCN_SELF_HOSTED_GITLAB_TOKEN=${MCN_SELF_HOSTED_GITLAB_TOKEN}"
-} >> ${TOKEN_FILE}
-mount_file "macaron_env_file" ${TOKEN_FILE} ${MACARON_WORKSPACE}/${TOKEN_FILE} "rw,Z"
-set -u
 
 # Parse main arguments.
 while [[ $# -gt 0 ]]; do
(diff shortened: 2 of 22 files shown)


scripts/release_scripts/run_macaron.sh
```diff
@@ -598,7 +599,21 @@ else
...
 fi
 
+# Handle tokens.
+set +u
+echo "" > ${TOKEN_FILE}
+{
+    echo "GITHUB_TOKEN=${GITHUB_TOKEN}" >> ${TOKEN_FILE}
+    echo "MCN_GITLAB_TOKEN=${MCN_GITLAB_TOKEN}" >> ${TOKEN_FILE}
+    echo "MCN_SELF_HOSTED_GITLAB_TOKEN=${MCN_SELF_HOSTED_GITLAB_TOKEN}"
+} >> ${TOKEN_FILE}
```
nicallen, line 609: Why are the first two echoes redirected directly into the file, and then the last one is echoed via the containing block redirect? I mean, it works, but the inconsistency just looks a bit odd.

tromai: Interesting, I agree that we could echo the third one directly into the file :thinking: 
However, I merely move this block location within `run_macaron.sh` without changing its original content, therefore I am not sure of the reason behind it. 
This block was originally added in https://github.com/oracle/macaron/pull/605

scripts/dev_scripts/integration_tests.sh
```diff
@@ -398,6 +398,7 @@ run_macaron_clean $ANALYZE -purl pkg:maven/io.github.behnazh-w.demo/example-mave
 check_or_update_expected_output $COMPARE_DEPS $DEP_RESULT $DEP_EXPECTED || log_fail
 
 python3 ./tests/integration/run.py run \
+    --exclude-tag token_file_clean_up \
```
tromai, line 401: I am thinking instead of changing this tag to `docker-only` to make it more generic :thinking: Please let me know what you think about it.

behnazh-w: Do you mean to rename the test tag from `token_file_clean_up` to `docker-only`?

tromai: Yes, that is correct. I think that if we rename it to `docker-only`, in the future we don't need to add a new tag if we add any new docker-specific test.

behnazh-w: Thanks, sounds good to me.

scripts/release_scripts/run_macaron.sh
```diff
@@ -598,7 +599,21 @@ else
...
+mount_file "macaron_env_file" ${TOKEN_FILE} ${MACARON_WORKSPACE}/${TOKEN_FILE} "rw,Z"
+set -u
+
 # Force docker to use linux/amd64 platform in order to make docker use emulation on ARM host platforms.
+# We need the `docker run` command to be within `set +e` so that when it returns a non-zero status code,
+# it doesn't exit right away and still run the token file cleaning up command.
+set +e
```
nicallen, line 616: Are you certain that nothing can fail between when the token file is written, and when exit-on-error is disabled here? Should it be disabled earlier (perhaps even before writing the token file) to be safe? It seems to me that exit-on-error is really dangerous if you have cleanup that must be done.

tromai: Hmmm, that it's true, I agree that we should disable exit-on-error for the block that create the token file. I will look into addressing it.

tromai: Addressed in https://github.com/oracle/macaron/pull/780/commits/2f2b2cae99fcf096d47deb66943aef8837b7cd4d

scripts/dev_scripts/integration_tests.sh
tromai: Addressed in https://github.com/oracle/macaron/pull/780/commits/bb3d48f74110cb6cd1927af9e24bb5f89a61dc98 and https://github.com/oracle/macaron/pull/780/commits/ace58a760f964f348c35130bcb27e66ca7589a0d

scripts/release_scripts/run_macaron.sh
tromai: My solution in https://github.com/oracle/macaron/commit/2f2b2cae99fcf096d47deb66943aef8837b7cd4d ensures that exit-on-error is not enable for the whole scope of the existence of the token file. This will ensure that the clean up code will always be run after the life time of the token file.
There is an alternative solution, where we uses a "trap - https://www.gnu.org/software/bash/manual/html_node/Bourne-Shell-Builtins.html#index-trap. According to the documentation of `set -e` - https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html, "trap" can be use to run any command before the shell exits early (caused by `set -e`):
```base
trap \
    # $TOKEN_FILE here should be expanded now, instead of when the signal is trapped.
    "rm -rf '$TOKEN_FILE'" \
    ERR
``` 
I am fine with both, but I think I prefer the first solution as it doesn't rely on the exit on error behavior, which is not safe (when we have to deal with the token file) as pointed out by @nicallen .

tromai:
Rebased to update the changes from https://github.com/oracle/macaron/pull/783

scripts/release_scripts/run_macaron.sh
tromai: Addressed in https://github.com/oracle/macaron/pull/780/commits/a6a0d98f416f013000d8d14b210158d4b62a7f08

nicallen (approved):

behnazh-w (approved):

Issue #769, opened 2024-06-13
Add a test for the JSON output of the analyze command


behnazh-w:
With this [PR](https://github.com/oracle/macaron/pull/760) merged, we have started using Datalog policies in the e2e tests to check the `analyze` command results. But we no longer check the JSON outputs. We need to add a test for the JSON outputs to catch regressions.

tromai:
I wonder which test cases do you have in mind for checking the JSON outputs. 
What I think would work best is to convert the rest of integration test cases (leftover from #764 ). After that, we can go through all of them and pick specific test cases that we wan to check for JSON outputs, then I can update those test cases correspondingly.

behnazh-w:
> I wonder which test cases do you have in mind for checking the JSON outputs. What I think would work best is to convert the rest of integration test cases (leftover from #764 ). After that, we can go through all of them and pick specific test cases that we wan to check for JSON outputs, then I can update those test cases correspondingly.

Yes, sounds good. After converting the rest of the test cases, checking the JSON output for one of the test cases should be enough.

tromai:
@behnazh-w As #782 is merged, I am thinking of using this test case https://github.com/oracle/macaron/blob/429ba95abc6ee0b917d2115598838b07a4244caf/tests/integration/cases/micronaut-projects_micronaut-core/test.yaml for the JSON report checking. What do you think about it?

behnazh-w:
Let's use https://github.com/oracle/macaron/tree/429ba95abc6ee0b917d2115598838b07a4244caf/tests/integration/cases/micronaut-projects_micronaut-test  for the JSON report as it takes less time in CI.

tromai:
I can definitely use the test case you suggested. 
I believe the time taken won't be an issue because we are only adding an extra step in `test.yaml` to perform the JSON files comparison. Both tests will still be run during the CI :thinking:

behnazh-w:
> I can definitely use the test case you suggested. I believe the time taken won't be an issue because we are only adding an extra step in `test.yaml` to perform the JSON files comparison. Both tests will still be run during the CI 🤔

I see, in case you add an extra step to the existing test run, `micronaut-core` is also fine. We should also make sure the JSON output comparison does not check the list of checks, but focuses on the structure. For example, if a new check is added, we shouldn't need to change the expected JSON output.

PR #782, opened 2024-06-27
chore: wrap up integration tests conversion

### LINKED ISSUE #779: Final wrap up of the conversion to using the new utility integration test script
This PR follows #764 .

- [ ] Convert the remaining test cases in `integration_tests.sh` and `integration_tests_docker.sh`.
- [ ] Remove `integration_tests.sh` and `integration_tests_docker.sh` scripts and instead calling the new utility script directly.

tromai:
Closes #779.

- [x] Convert the remaining test cases.
- [x] Move gradle jackson databind tests to the new test setup but make sure they are disabled for now.
- [x] Address the issue of running update command of the test utility script (https://github.com/oracle/macaron/pull/786).
- [x] Remove `integration_tests.sh` and `integration _tests_docker.sh`
- [x] Update dev docs about instructions on the tags of integration tests cases.
- [x] Handle the NO_NPM environment variable in Makefile.
- [x] Remove incorrect npm-registry tag for `tests/integration/cases/gitlab_tinyMediaManager_purl/test.yaml`
- [x] Go through the `tests/e2e` directory and clean up remaining files.


tromai:
Rebased to update changes from https://github.com/oracle/macaron/pull/786

tests/integration/cases/apache_maven_local_paths_in_configuration_without_dep_resolution/config.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2024, Oracle and/or its affiliates. All rights reserved.
```
benmss, line 1: Looks to me like this file hasn't changed at all. Is it that fun little problem with copyright headers again?

tests/integration/cases/run_macaron_sh_script_unit_test/test_run_macaron_sh.py
```diff
@@ -41,12 +40,11 @@ def test_macaron_command() -> int:
         print(f"test_macaron_command[{name}]:", end=" ")
 
         result = subprocess.run(
-            [
-                "scripts/release_scripts/run_macaron.sh",
+            [  # nosec
```
benmss, line 43:
```suggestion
            [  # nosec B603
```

tests/integration/cases/apache_maven_local_paths_in_configuration_without_dep_resolution/config.yaml
tromai: That's true. The year shouldn't change. 
> Is it that fun little problem with copyright headers again?
Maybe :sweat_smile:. I honestly don't remember how it ended up like that.

tests/integration/cases/run_macaron_sh_script_unit_test/test_run_macaron_sh.py
tromai: Addressed in https://github.com/oracle/macaron/pull/782/commits/f3ecbf0b72a628ae8fa762959cd967185d8044ef

tests/integration/cases/run_macaron_sh_script_unit_test/test_run_macaron_sh.py
```diff
@@ -1,17 +1,16 @@
-# Copyright (c) 2023 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2023 - 2024, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """Tests for the ``run_macaron.sh`` script."""
 
 import os
-import subprocess
+import subprocess  # nosec
```
benmss, line 7:
```suggestion
import subprocess  # nosec B404
```

tromai: Addressed in https://github.com/oracle/macaron/pull/782/commits/df0a4a92b9acebccd96f8b0a34af251b324cdf59

tests/integration/cases/apache_maven_using_default_template_file_as_input_template/test.yaml
```diff
@@ -0,0 +1,25 @@
...
+  options:
+    command_args:
+    - -rp
+    - https://github.com/apache/maven
+    - -b
+    - master
+    - -d
+    - 3fc399318edef0d5ba593723a24fff64291d6f9b
+    - -g
+    - ../../../../src/macaron/output_reporter/templates/macaron.html
```
benmss, line 19: I think that if `../` is needed at all, then it might be better to alter the working directory of the tests instead. However, I only see 2 instances where this is used, so perhaps it is not an essential change. What do you think?

tromai: Hmm I don't think we should change the working directory of the tests as it defeats the purpose of isolating the test directories. However, if we want to remove `../` from this test configuration, I can definitely add a step to copy the target file into the test directory and clean it up afterward :thinking: .

tromai: Hmmm after having a look again at this test case, I think we should update it by creating a simple template file as input instead of relying on the default template of Macaron. This is because this case is testing an arbitrary template file as input, while we can reuse Macaron's default template file, it has already been tested as other cases. I would still prefer we simplify the test case by not relying on files that are outside of the test directories (as you have raised a good point). 

As for the second occurrence - https://github.com/oracle/macaron/blob/df0a4a92b9acebccd96f8b0a34af251b324cdf59/tests/integration/cases/run_macaron_sh_script_unit_test/test_run_macaron_sh.py#L42-L50
I can definitely follow my solution mentioned above.

tromai: Addressed in https://github.com/oracle/macaron/pull/782/commits/22c2e32c3c590a21e071228500bd850eb72e6d90

benmss (approved):

behnazh-w (commented):
There are still [configurations](https://github.com/oracle/macaron/tree/tromai/wrap-up-integration-tests-conversion/tests/e2e/configurations) and [expected_results](https://github.com/oracle/macaron/tree/tromai/wrap-up-integration-tests-conversion/tests/e2e/expected_results) under the `tests/e2e` directory, but I don't think they are used anywhere. Are these already converted?

tromai:
@behnazh-w I believe all of them can be safely removed. They are still there because I missed them when converting the test cases. I will also go through the e2e directory and clean up the remaining files.
> There are still [configurations](https://github.com/oracle/macaron/tree/tromai/wrap-up-integration-tests-conversion/tests/e2e/configurations)
1. `jackson_databind_config.yaml` was originally used for the jackson databind integration tests
- See the original usage - https://github.com/oracle/macaron/blob/516552a49631542997ba5cfa500f0df8bdae5c02/scripts/dev_scripts/integration_tests.sh#L320
- Within this commit from #508 - https://github.com/oracle/macaron/pull/508/commits/0373bc03480abd4476fef5ada8d7806f92c663f4 - The jackson databind tests were changed from using an input yaml config to using a PURL for triggering the repo and commit finder. Therefore, this `jackson_databind_config.yaml` is not being used anywhere. I believe we can safely remove it. 
-  As for the jackson databind tests, I have already converted the PURL version, hence I didn't notice this leftover input yaml config file.

2. `micronaut_test_config.yaml` was originally used for the micronaut-test tests
- See the original usage https://github.com/oracle/macaron/blob/9e90ab1c4af1a101a269841cabbf1dd2d8ef86c6/scripts/dev_scripts/integration_tests.sh#L237
- Within this commit from #508 - https://github.com/oracle/macaron/pull/508/commits/0373bc03480abd4476fef5ada8d7806f92c663f4 - The micronaut-test test was changed from using an input yaml config to using a PURL for triggering the repo and commit finder. 
- This `micronaut_test_config.yaml` was later converted to this test case https://github.com/oracle/macaron/blob/08183263a690dfc870ed750b53d93cf6a31a8096/tests/integration/cases/micronaut-projects_micronaut-test 
- Therefore, I think it's also safe to remove this `micronaut_test_config.yaml` here too.

> [expected_results](https://github.com/oracle/macaron/tree/tromai/wrap-up-integration-tests-conversion/tests/e2e/expected_results)
- `micronaut-test-branch-commit.dl` was originally created in https://github.com/oracle/macaron/commit/d9b56216cea2ee5deec919be714d92efe333b643 (#760), but I don't see where it was used. I believe it could be safely removed (@benmss Please let me know if my observation is correct).
- `urllib3.dl` was used for urllib3 tests and was already converted: https://github.com/oracle/macaron/blob/be50c335394ba56f16c97ffc1897473667e3bd15/tests/integration/cases/urllib3_expectation_file and https://github.com/oracle/macaron/blob/be50c335394ba56f16c97ffc1897473667e3bd15/tests/integration/cases/urllib3_expectation_dir

Makefile
```diff
@@ -298,16 +298,37 @@ test-go:
...
 
 .PHONY: integration-test-docker
 integration-test-docker:
-	scripts/dev_scripts/integration_tests_docker.sh $(REPO_PATH) scripts/release_scripts/run_macaron.sh
+	python ./tests/integration/run.py \
+		run \
+    	--macaron scripts/release_scripts/run_macaron.sh \
+		--include-tag docker-only \
+    	--include-tag shared-docker-python \
+    	./tests/integration/cases/...
```
tromai, line 324: There seems to be tab/spaces issues here. I will address them.

tromai: Addressed in https://github.com/oracle/macaron/pull/782/commits/0defbf663113b242a043603e0c8fd6e56b5797be

behnazh-w:
For the remaining `compare_e2e_result.py` under `tests/e2e`/, I think we can rename it to `tests/analyze_json_output/compare_analyze_json_output.py`? We can then use this comparison script and add a test case while addressing this [issue](https://github.com/oracle/macaron/issues/769).

tromai:
> For the remaining `compare_e2e_result.py` under `tests/e2e`/, I think we can rename it to `tests/analyze_json_output/compare_analyze_json_output.py`? We can then use this comparison script and add a test case while addressing this [issue](https://github.com/oracle/macaron/issues/769).

Yep I can definitely rename that script. Regarding issue #769 , do you prefer it to be addressed within this Pull Request too?

behnazh-w:
> Yep I can definitely rename that script. Regarding issue #769 , do you prefer it to be addressed within this Pull Request too?

I am happy either way :slightly_smiling_face:

tromai:
> > Yep I can definitely rename that script. Regarding issue #769 , do you prefer it to be addressed within this Pull Request too?
> 
> I am happy either way 🙂

Sure, I will address #769 in a different Pull Request. Thanks!

tromai:
@behnazh-w I have renamed the `compare_e2e_result.py` file in https://github.com/oracle/macaron/pull/782/commits/1c65b9a5f960336ac1eae735d558dc857eccf211

behnazh-w:
@tromai  The [Docker integration tests job](https://github.com/oracle/macaron/actions/runs/9886543118/job/27307062890) has finished in 41s and as far as I can see, none of the tests have run. Can you please check?

tromai:
> @tromai The [Docker integration tests job](https://github.com/oracle/macaron/actions/runs/9886543118/job/27307062890) has finished in 41s and as far as I can see, none of the tests have run. Can you please check?

Thanks for pointing it out. It was my mistake when setting up the tags. It should be resolved in https://github.com/oracle/macaron/pull/782/commits/838ef8665c28b45b897fd87e5747d58a496e5b06

benmss:
I don't think I agree with the current implementation of the include tag list. To me it would make more sense to run all tests with any of the tags in the include list. If there really is a case where we want to run only the tests with all of a list of include tags, which implies the intersecting subset of one or more tag groups, then the solution is to give those tests a new tag and use that.

tromai:
> I don't think I agree with the current implementation of the include tag list. To me it would make more sense to run all tests with any of the tags in the include list.

@benmss 
I agree that currently the set of tags we have can make the commands used in Makefile confusing. Therefore, I am suggesting this new setup of tags for all of our test cases:
- A test case with `python` tag will be run with the Python package in our CI.
- A test case with `docker` tag will be run with the Docker container in our CI.
- A test case without any tag will not be run in the CI (But it would still be automatically picked up and run locally with `python3 tests/integration/run.py run tests/integration/cases/...`).
- A test case with `npm-registry` tag will not run if `NO_NPM` is set to fail in our CI.

This new setup can still represent our use cases for subset of test cases while keeping the number of tag minimal:
- Tests that are shared between Python package and Docker container can use `docker` and `python`.
- Tests that are only run with Docker container will only contain `docker`.
- Tests that are disabled will have no tag. 

This set of tags would result in these commands for `Makefile`:
```
# Run the integration tests.
# Note: to disable npm tests set `NO_NPM` environment variable to `TRUE`.
.PHONY: integration-test
integration-test:
	if [ "${NO_NPM}" == "TRUE" ]; then \
    	echo "Note: NO_NPM environment variable is set to TRUE, so npm tests will be skipped."; \
		python ./tests/integration/run.py \
			run \
			--include-tag python \
			--exclude-tag npm-registry \
			./tests/integration/cases/...; \
	else \
		python ./tests/integration/run.py \
			run \
			--include-tag python \
			./tests/integration/cases/...; \
	fi

.PHONY: integration-test-docker
integration-test-docker:
	python ./tests/integration/run.py \
		run \
		--macaron scripts/release_scripts/run_macaron.sh \
		--include-tag docker \
		./tests/integration/cases/...
```

Am I understand the problem you mentioned correctly? :thinking:

benmss:
@tromai
I was advocating for changing the behaviour of the include list over something like that, but I suppose it's okay also.

I just meant that if you ran the tests with `include-list=[tag1, tag2]` then I'd expect all tests that have either `tag1` or `tag2` to run. Then for cases where you want only tests that have both `tag1` and `tag2`, that is something worth being tagged with a new tag, e.g. `tag1-tag2`.

tromai:
> @tromai I was advocating for changing the behaviour of the include list over something like that, but I suppose it's okay also.
> 
> I just meant that if you ran the tests with `include-list=[tag1, tag2]` then I'd expect all tests that have either `tag1` or `tag2` to run. Then for cases where you want only tests that have both `tag1` and `tag2`, that is something worth being tagged with a new tag, e.g. `tag1-tag2`.

Oh I see, yep I will change the behavior of `--include-tag`. @behnazh-w Please let me know if you are okay with this change too.

behnazh-w:
> > @tromai I was advocating for changing the behaviour of the include list over something like that, but I suppose it's okay also.
> > I just meant that if you ran the tests with `include-list=[tag1, tag2]` then I'd expect all tests that have either `tag1` or `tag2` to run. Then for cases where you want only tests that have both `tag1` and `tag2`, that is something worth being tagged with a new tag, e.g. `tag1-tag2`.
> 
> Oh I see, yep I will change the behavior of `--include-tag`. @behnazh-w Please let me know if you are okay with this change too.

I agree with the proposed changes. I suggest to improve the tag names though:

-  `macaron-python-package`
-  `macaron-docker-image`
-  `npm-registry-testcase`

tests/integration/README.md
```diff
@@ -133,58 +133,50 @@ $ python ./tests/integration/run.py run ./all/cases/...
...
-$ python ./tests/integration/run.py \
-      run \
-      --macaron scripts/release_scripts/run_macaron.sh \
-      --include-tag shared-docker-python \
-      ./tests/integration/cases/...
-```
-- Test cases marked with `skip` are not run.
-- Test cases marked with `npm-registry` are not run if the environment variable `NO_NPM` is set to `TRUE`. This only applies when you run the integration tests with:
+Instructions on how to tag a test case for our CI/CD pipeline:
+- If you want a test case to **only** run for the container image, using **only** `macaron-docker-image`.
```
benmss, line 174:
```suggestion
- If you want a test case to **only** run for the container image, use **only** `macaron-docker-image`.
```

tests/integration/README.md
```diff
@@ -133,58 +133,50 @@ $ python ./tests/integration/run.py run ./all/cases/...
...
+- If you want a test case to **only** run with the Macaron Python package **only** `macaron-python-package`.
```
benmss, line 175:
```suggestion
- If you want a test case to **only** run with the Macaron Python package, use **only** `macaron-python-package`.
```

tests/integration/README.md
```diff
@@ -133,58 +133,50 @@ $ python ./tests/integration/run.py run ./all/cases/...
...
+- To skip a test case, use `skip`. `skip` still has the same effect if it's used with other tags.
+- If you want to run a test case for both the Macaron Python package and the docker container, use `macaron-python-package` and `macaron-docker-image` tags.
+- If you want to run test cases that must contain a given set of tags (e.g. `['tag-a', 'tag-b']`), please create an additional tag for those test cases (e.g `tag-a-b`) and use it within `--include-tag`.
```
benmss, line 178:
```suggestion
- If you want to run test cases that must contain all of a given set of tags (e.g. `['tag-a', 'tag-b']`), please create an additional tag for those test cases (e.g `tag-a-b`) and use it within `--include-tag`.
```

tests/integration/run.py
```diff
@@ -744,17 +744,23 @@ def load_test_cases(
         else:
             # Each --include-tag/--exclude-tag argument adds an additional constraint
             # that a selected test case needs to satisfy, i.e. a selected test case must:
-            # - contains all tags specified with --include-tag
+            # - contains at least on tag specified with --include-tag
```
benmss, line 747:
```suggestion
            # - contains at least one tag specified with --include-tag
```

tromai: Addressed in https://github.com/oracle/macaron/pull/782/commits/874d7577e92969bc663cba56bd98b5fd435e3fa6

tests/integration/README.md
tromai: Addressed in https://github.com/oracle/macaron/pull/782/commits/874d7577e92969bc663cba56bd98b5fd435e3fa6

tests/integration/README.md
tromai: Addressed in https://github.com/oracle/macaron/pull/782/commits/874d7577e92969bc663cba56bd98b5fd435e3fa6

tests/integration/README.md
tromai: Addressed in https://github.com/oracle/macaron/pull/782/commits/874d7577e92969bc663cba56bd98b5fd435e3fa6

benmss:
@tromai 
It would be good to fix the three `urllib3` test cases that have invalid repo path values in their yaml files. E.g. `https://github.com/urllib3/urllib3/urllib3` -> `https://github.com/urllib3/urllib3`

tromai:
> @tromai It would be good to fix the three `urllib3` test cases that have invalid repo path values in their yaml files. E.g. `https://github.com/urllib3/urllib3/urllib3` -> `https://github.com/urllib3/urllib3`

Hmm interesting. I think the reason these tests still ran and passed was that the invalid repository path are "cleaned up" by Macaron before it's used to clone the repository. I will address those invalid paths.

tromai:
Rebased to obtain the changes in #795

behnazh-w (approved):

PR #783, opened 2024-06-28
test: use python instead of python3

behnazh-w:
We have made `Python 3.11` a requirement for our development environment and use `python` to run. This PR replaces `python3` with `python` in develoment scripts to be consistent.

### DIFF
--- a/scripts/dev_scripts/integration_tests_docker.sh
+++ b/scripts/dev_scripts/integration_tests_docker.sh
@@ -58,7 +58,7 @@ python $COMPARE_POLICIES $POLICY_RESULT $POLICY_EXPECTED || log_fail
 # Clean up and remove the virtual environment.
 rm -rf "$VIRTUAL_ENV_PATH"
 
-python3 ./tests/integration/run.py run \
+python ./tests/integration/run.py run \
     --macaron scripts/release_scripts/run_macaron.sh \
     --include-tag docker \
     ./tests/integration/cases/... || log_fail
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -165,7 +165,7 @@ repos:
   hooks:
   - id: integration-test-vet
     name: validate integration test cases
-    entry: python3
+    entry: python
     args:
     - ./tests/integration/run.py
     - vet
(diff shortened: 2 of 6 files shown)


tromai:
Hmmm I wonder if it's better to change everything to `python3` instead :thinking:. Not that `python` in this case is wrong, because we have indicated that we should run everything inside the dev environment created with Python 3.11. 
I think `python3` is still correct and would be more specific than `python`. When the reader/developer look through the snippet with `python3` in isolation, it would be more clear in my opinion.  Otherwise, everything looks good to me.

scripts/dev_scripts/integration_tests_docker.sh
```diff
@@ -58,7 +58,7 @@ python $COMPARE_POLICIES $POLICY_RESULT $POLICY_EXPECTED || log_fail
 # Clean up and remove the virtual environment.
 rm -rf "$VIRTUAL_ENV_PATH"
 
-python3 ./tests/integration/run.py run \
+python ./tests/integration/run.py run \
```
tromai, line 61: I think we would need to change the same for `integration_tests.sh` too.

behnazh-w:
> Hmmm I wonder if it's better to change everything to `python3` instead 🤔. Not that `python` in this case is wrong, because we have indicated that we should run everything inside the dev environment created with Python 3.11. I think `python3` is still correct and would be more specific than `python`. When the reader/developer look through the snippet with `python3` in isolation, it would be more clear in my opinion. Otherwise, everything looks good to me.

No, I don't think it's necessary anymore to use Python 3. Python 2.7 might be still supported in some places but it's rare and since we make it clear we use Python 3, it's fine to avoid repeating `python3` everywhere.

scripts/dev_scripts/integration_tests_docker.sh
behnazh-w: Not sure how I missed that :-) 
See commit [ba87342](https://github.com/oracle/macaron/pull/783/commits/ba87342e2dfc5bffd5011f879f08ade784f6175d)

tromai (approved):
LGTM. Thanks!

PR #786, opened 2024-07-04
chore: add exclude and include tag options to update command and add logging of each test case to update command

tromai:
# Description
Setup the dev environment and run the following command:
```
python ./tests/integration/run.py update ./tests/integration/cases/...
...
AttributeError: 'Namespace' object has no attribute 'include_tag'
```
This is because here
https://github.com/oracle/macaron/blob/a06ed41d306d21da54a7e3135049c630b686c720/tests/integration/run.py#L955-L964
we create option parsers  for `--excluded-tag` and `--include-tag` which are necessary for the update command of Macaron later 
https://github.com/oracle/macaron/blob/a06ed41d306d21da54a7e3135049c630b686c720/tests/integration/run.py#L1010-L1016

### DIFF
--- a/tests/integration/run.py
+++ b/tests/integration/run.py
@@ -573,6 +573,8 @@ class Case:
 
     def update(self, macaron_cmd: str) -> int:
         """Run the test case in update mode."""
+        logger.info("-" * 60)
+        logger.info("Case started: '%s'.", self.case_dir)
         for step in self.steps:
             if isinstance(step, CompareStep):
                 ret = step.update_result(cwd=self.case_dir)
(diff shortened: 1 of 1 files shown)


tests/integration/run.py
behnazh-w, line 1: What if a test case doesn't have any tags? Does it mean we can't update it?

tromai: By default, if we don't provide any `--include-tag` and `--exclude-tag`, the script will run all the test cases that it can find. Therefore, even if a test case has no tag, it can still be update if you don't provide any  `--include-tag` and `--exclude-tag`.

behnazh-w (approved):

PR #790, opened 2024-07-10
chore: restrict malware check to python projects; apply general fixes…

benmss:
… and improvements.

### DIFF
--- a/src/macaron/malware_analyzer/checks/detect_malicious_metadata_check.py
+++ b/src/macaron/malware_analyzer/checks/detect_malicious_metadata_check.py
@@ -201,7 +201,7 @@ class DetectMaliciousMetadataCheck(BaseCheck):
 
         Returns
         -------
-            bool
+        bool
             Returns True if any result of the dependency heuristic does not match the expected result.
             Otherwise, returns False.
         """
@@ -257,8 +257,9 @@ class DetectMaliciousMetadataCheck(BaseCheck):
         CheckResultData
             The result of the check.
         """
-        purl = ctx.component.purl
-        parsed_purl = PackageURL.from_string(purl)
+        parsed_purl = PackageURL.from_string(ctx.component.purl)
+        if parsed_purl.type != "pypi":
+            return CheckResultData(result_tables=[], result_type=CheckResultType.FAILED)
         package = parsed_purl.name
         result_tables: list[CheckFacts] = []
 
--- a/src/macaron/malware_analyzer/checks/detect_malicious_metadata_check.py
+++ b/src/macaron/malware_analyzer/checks/detect_malicious_metadata_check.py
@@ -260,7 +260,7 @@ class DetectMaliciousMetadataCheck(BaseCheck):
         """
         parsed_purl = PackageURL.from_string(ctx.component.purl)
         if parsed_purl.type != "pypi":
-            return CheckResultData(result_tables=[], result_type=CheckResultType.FAILED)
+            return CheckResultData(result_tables=[], result_type=CheckResultType.UNKNOWN)
         package = parsed_purl.name
         result_tables: list[CheckFacts] = []
 
--- a/src/macaron/malware_analyzer/checks/detect_malicious_metadata_check.py
+++ b/src/macaron/malware_analyzer/checks/detect_malicious_metadata_check.py
@@ -238,6 +238,7 @@ class DetectMaliciousMetadataCheck(BaseCheck):
                 if should_skip:
                     results[analyzer.heuristic] = HeuristicResult.SKIP
                     continue
+
             result, detail_info = analyzer.analyze(api_client)
             if analyzer.heuristic:
                 results[analyzer.heuristic] = result
--- a/src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
+++ b/src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
@@ -85,7 +85,7 @@ class SuspiciousSetupAnalyzer(BaseHeuristicAnalyzer):
 
             final_path = os.path.join(temp_dir, target_file)
             if not os.path.exists(final_path):
-                logger.debug("Failed to download source file.")
+                logger.debug("Source file could not be extracted from downloaded archive.")
                 return None
 
             with open(final_path, encoding="utf-8") as file:
--- a/src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
+++ b/src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
@@ -108,56 +108,11 @@ class SuspiciousSetupAnalyzer(BaseHeuristicAnalyzer):
         if content is None:
             return HeuristicResult.SKIP, {}
 
-        imports = set()
         # Catch the imported module.
         try:
-            tree = ast.parse(content)
-            for node in ast.walk(tree):
-                if isinstance(node, ast.Import):
-                    for alias in node.names:
-                        imports.add(alias.name)
-                elif isinstance(node, ast.ImportFrom):
-                    module = node.module
-                    if node.level > 0 and module:
-                        _module = "." * node.level + module
-                        for alias in node.names:
-                            if alias.name:
-                                import_name = _module + "." + alias.name
-                                imports.add(import_name)
+            imports = self.extract_from_ast(content)
         except SyntaxError:
-            pattern_import = r"(?:import\s+)(\w+(?:(?:\.{1}|(?:\s*,\s*))\w+)*)(?:(?:\s|#).*)?"
-            # Allows for a standard import statement.
-            # E.g.: import <module_name(s)> <other_text>
-            # Where <module_name(s)> consists of one or more <module_name>.
-            # Where <module_name> consists of one or more words (a-z or 0-9 or underscore) separated by periods.
-            # Where <other_text> allows any character(s) either after a single space or a hash (#).
-
-            pattern_from_import = r"(?:from\s+)(\w+(?:\.{1}\w+)*)(?:\s+import\s+\w+).*"
-            # Allows for a from import statement.
-            # E.g.: from <module_name> import <other_text>
-            # Where <module_name> is as above. (Note only a single module can be placed here.)
-            # Where <other_text> requires at least one space followed by one or more word characters, plus
-            # any other characters following on from that.
-
-            combined_pattern = f"^(?:{pattern_import})|(?:{pattern_from_import})$"
-            # The combined pattern creates two match groups:
-            # 1 - standard import statement.
-            # 2 - from import statement.
-
-            for line in content.splitlines():
-                line.strip()
-                match = re.match(combined_pattern, line)
-                if match:
-                    if match.group(1):
-                        # Standard import, handle commas if present.
-                        splits = match.group(1).split(",")
-                        for split in splits:
-                            split = split.strip()
-                            if split:
-                                imports.add(split)
-                    elif match.group(2):
-                        # From import
-                        imports.add(match.group(2))
+            imports = self.extract_from_lines(content)
 
         # Catch the install_requires packages
         pattern = r"install_requires\s*=\s*\[(.*?)\]"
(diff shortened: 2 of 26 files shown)


src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
```diff
@@ -32,86 +33,86 @@ def _get_setup_source_code(self, api_client: PyPIRegistry) -> str | None:
...
-                f.write(response.content)
+            if response.status_code != 200:
+                return None
 
-            files = os.listdir(temp_dir)
+            source_file = os.path.join(temp_dir, "tmp")
+            with open(source_file, "wb") as file:
+                try:
+                    for chunk in response.iter_content():
+                        file.write(chunk)
```
tromai, line 64: I wonder why we switched from `f.write(response.content)` to usin the `iter_content()` method :thinking: ?

src/macaron/malware_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -257,8 +258,9 @@ def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
         CheckResultData
             The result of the check.
         """
-        purl = ctx.component.purl
-        parsed_purl = PackageURL.from_string(purl)
+        parsed_purl = PackageURL.from_string(ctx.component.purl)
+        if parsed_purl.type != "pypi":
+            return CheckResultData(result_tables=[], result_type=CheckResultType.FAILED)
```
behnazh-w, line 263: Just to let you know, we want to use this check for other ecosystems in the future too. We can add this and I will improve this implementation in my next refactoring PR of this check.

Also, `CheckResultType.FAILED` means the check has found a malware. So the result type should be `CheckResultType.UNKNOWN` if the artifact is not a Python package.

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
```diff
@@ -32,86 +33,86 @@ def _get_setup_source_code(self, api_client: PyPIRegistry) -> str | None:
...
-            for root, _, files in os.walk(temp_dir):
-                for file in files:
-                    if file == "setup.py":
-                        file_path = os.path.join(root, file)
-                        with open(file_path, encoding="utf-8") as py_file:
-                            return py_file.read()
-            return None
+            final_path = os.path.join(temp_dir, target_file)
+            if not os.path.exists(final_path):
+                logger.debug("Failed to download source file.")
```
tromai, line 88: nit: I think it's better if we say we failed to find/extract `setup.py` from the source tarball as it had been downloaded earlier.

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
```diff
@@ -123,3 +124,88 @@ def analyze(self, api_client: PyPIRegistry) -> tuple[HeuristicResult, dict]:
...
+        tree = ast.parse(source_content)
+        for node in ast.walk(tree):
+            if isinstance(node, ast.Import):
+                for alias in node.names:
+                    imports.add(alias.name)
+            elif isinstance(node, ast.ImportFrom):
+                module = node.module
+                if node.level != 0 or not module:
+                    # Relative imports are ignored.
+                    continue
```
tromai, line 156: Why did we decide to ignore relative imports (where the original implementation accepted them)?

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
benmss: As `stream=True` was already set for the `request`, which was most likely a good idea as the source code might be quite large, using `iter_content` is the suggested way to handle the received data chunks as they come in.

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
```diff
@@ -123,3 +124,88 @@ def analyze(self, api_client: PyPIRegistry) -> tuple[HeuristicResult, dict]:
...
+                imports.add(module)
+
+        return imports
```
tromai, line 159: I made some comparison between this newer implementation and the old one. I noticed that there are some differences between them.
```
>>> line = "from .a.b.c import x, y, z"
>>> old(line)
{'.a.b.c.y', '.a.b.c.x', '.a.b.c.z'}
>>> new(line)
set()

>>> line =  "from a import x,y,z"
>>> old(line)
set()
>>> new(line)
{'a'}

>>> line = "from a.b.c import x,y,z"
>>> old(line)
set()
>>> new(line)
{'a.b.c'} 
```
Is it okay to say that these cases are as expected as we completely ignore any relative imports and only takes the absolute ones?
Hmm then my next question would be whether the behavior of the old implementation (i.e accept `from .a.b.c import x, y, z` and ignore `from a import x,y,z`) was incorrect, or was it intended? 
@behnazh-w Do you happen to remember what is the intended behavior of this logic :thinking: ?

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
```diff
@@ -123,3 +124,88 @@ def analyze(self, api_client: PyPIRegistry) -> tuple[HeuristicResult, dict]:
         if suspicious_setup:
             return HeuristicResult.FAIL, {"import_module": imports}
         return HeuristicResult.PASS, {"import_module": imports}
+
+    def extract_from_ast(self, source_content: str) -> set[str]:
+        """Extract imports from source code using the parsed AST.
```
tromai, line 129: We should also document the type of import are ignored here too (i.e the relative imports).

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
behnazh-w: We should support both relative and absolute imports. The idea here is to catch all possible imports. So, if the previous implementation does not handle `from a import x,y,z`, it could be a bug.

tromai: Oh okay, so then this method should return **all modules** imported regardless of its type.
- Absolute import `from A ...` or `from A.B.C`
- Relative import `from . ...` or `from .A.B.C`

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -31,9 +31,6 @@
     PURLNotFoundError,
     RepoCheckOutError,
 )
-from macaron.malware_analyzer.checks import (  # pylint: disable=unused-import # noqa: F401
-    detect_malicious_metadata_check,
-)
 from macaron.output_reporter.reporter import FileReporter
```
tromai, line 34: Is this import removed because it's not used within this file ?

benmss: Yes.

tromai: Oh right, pylint was disabled for this statement, I was wondering why the pre-commit hooks didn't mark it as error.

tromai: Look like this line was critical for the `detect_malicious_metadata_check` to run :sweat_smile: .
Apparently, the registering of `detect_malicious_metadata_check.py` here https://github.com/oracle/macaron/blob/739826e0ff285f452e66abc9b222c77008d853fc/src/macaron/malware_analyzer/checks/detect_malicious_metadata_check.py#L296 doesn't work because `detect_malicious_metadata_check.py` is not located in `src/macaron/slsa_analyzer/checks` where all checks modules are loaded here https://github.com/oracle/macaron/blob/84728190397faf40571b14a5a822e47c8fd7cdc7/src/macaron/slsa_analyzer/checks/__init__.py#L11
This also explains why an explicit import line for `detect_malicious_metadata_check` is put in this file. I guess we can consider move that `detect_malicious_metadata_check` module into its correct location. 
cc: @behnazh-w .

behnazh-w: Thanks @tromai for looking into this. Yes, we need to move the check under `src/macaron/slsa_analyzer/checks`. @benmss Please revert this change and I will fix it separately.

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
```diff
@@ -32,86 +33,86 @@ def _get_setup_source_code(self, api_client: PyPIRegistry) -> str | None:
...
+                    logger.debug("Error while streaming source file: %s", error)
+                    response.close()
+                    return None
+
+            target_file = "setup.py"
             try:
-                with tarfile.open(os.path.join(temp_dir, files[0]), "r:gz") as tar:
-                    tar.extractall(temp_dir)  # nosec B202
+                with tarfile.open(source_file, "r:gz") as tar:
+                    tar.extract(target_file)
```
behnazh-w, line 74: Is `extract` used instead of `extractall` for security issues? Note that `extractall` is more resilient to extraction issues and [recommended](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extract).

Maybe use:

```suggestion
                    tar.extractall(target_file, members=members_tarinfo, filter="data")  # nosec B202:tarfile_unsafe_members
```

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
```diff
@@ -32,86 +33,86 @@ def _get_setup_source_code(self, api_client: PyPIRegistry) -> str | None:
...
             except tarfile.ReadError as exception:
                 logger.debug("Error reading tar file: %s", exception)
 
                 try:
-                    with zipfile.ZipFile(os.path.join(temp_dir, files[0]), "r") as zip_ref:
-                        zip_ref.extractall(temp_dir)  # nosec B202
+                    with zipfile.ZipFile(source_file, "r") as zip_ref:
+                        zip_ref.extract(target_file)
```
behnazh-w, line 80: Same comment as above.

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
benmss: The previous implementation used `extractall` and extracted everything from the tar. I thought it best to only extract the single file of interest.

From what I can see in the documentation of the two methods, this recommendation of `extractall` being better is not explained in any further detail:

> The [extract()](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extract) method does not take care of several extraction issues.

Looking at the source code reveals only that directory extraction is handled slightly differently in the `extractall` function, so I don't see any clear benefit when we are not extracting directories.

One security issue here I see is that the `filter` argument was only added in Python 3.12. If we don't want to upgrade the version, then it is possible for symlinks to be extracted from the tar.

tromai: I think I agree that extracting a single file of interest (which is `setup.py` at the root of the source tar ball) is better than extracting the whole tar ball and iterating through the file to search for `setup.py`. 
I think the problem with the current usage of `extract(source_file)` is that if within the tar ball `source_file` is the name of a directory, you will only extract the directory (with empty content) to your file system. Let's say you have the following directory tree:
```
├── a.txt
├── boo
│   └── c.txt
├── b.txt
```
You create a tar file with:
```bash
$ tar cf test.tar a.txt b.txt boo
```
Try to extract `boo` from `test.tar`:
```python3
>>> import tarfile
>>> import os
>>> file = tarfile.open("./")
>>> file.extract("boo", path="./foo")
>>> os.listdir("./foo/boo")
[]
>>> os.path.isdir("./foo/boo")
True
```
This will mean that we cannot be sure the path we provide to `extract()` will finally result in a regular file (I haven't tested with symbolic link and others). This means that we must also validate if `setup.py` is actually a file with valid content after we extract it. 

However, we can achieve 2 goals: extracting a single file, making sure what we are extracting is a regular file by using `getmember()` - https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.getmember:
```python3
>>> import tarfile
>>> import os
>>> file = tarfile.open("./")
>>> boo_tar_info = file.getmember("boo")
>>> boo_tar_info.isfile()
False
```
If `boo_tar_info` is a file, we can then extract it using `extractall()` or `extract()`:
```
...
file.extractall(path=source_file, members=[boo_tar_info]) 

# or 

file.extract(boo_tar_info)
```
I don't have a strong opinion on whether `extractall()` or `extract()` should be used as long as we have perform the necessary validation.

behnazh-w: For now we can use `extract` or `extractfile` for that purpose because we just analyze `setup.py`. But in the future we might analyze more files. We can revisit `extractall` when we need it though.

Also, the current implementation is actually wrong. Test it on `macaron --verbose analyze -purl pkg:pypi/mortgage_calculations@0.1.1` (make sure the check is running be reverting the change that removes the imported check in this check) and you will see that the `target_file` path is `setup.py` instead of `mortgage_calculations-0.1.1/setup.py`, which results in a `KeyError` that is not handled neither.

benmss: > Also, the current implementation is actually wrong.

Interestingly, in my tests I was eventually using `TarInfo(setup.py)` and passing that to extract, which seems to work regardless of how many directories the file is buried under.

Checking the actual members first will eliminate this problem, but we will still have the danger of extracting a symbolic link unless we upgrade to Python `3.12` and use the `filter` parameter.

behnazh-w: >Interestingly, in my tests I was eventually using TarInfo(setup.py) and passing that to extract, which seems to work regardless of how many directories the file is buried under.

We actually need the top level `setup.py` because that's used to install.

benmss: > We actually need the top level setup.py because that's used to install.

Indeed. I was commenting on a seeming oddity with how `TarInfo` files can be created and nothing more. Apologies if that was not clear.

behnazh-w (approved):

tromai:
My approval still stands. Thanks for the changes!

tromai (approved):
LGTM. Thanks for the changes.

PR #791, opened 2024-07-10
chore: refactor provenance available check

benmss:
This PR refactors the provenance available check so that the provenance finding functions are all performed as part of analysis instead of within the check itself. The check is then free to reference the pre-discovered provenance, if any.


src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -504,45 +71,16 @@ def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
...
-            fallback=["intoto.jsonl"],
+        available = bool(ctx.dynamic_data["provenance"])
+        return CheckResultData(
+            result_tables=[
+                ProvenanceAvailableFacts(
+                    available=f"The provenance is {'available' if available else 'not available'}.",
+                    confidence=Confidence.HIGH,
+                )
+            ],
+            result_type=CheckResultType.PASSED if available else CheckResultType.FAILED,
```
tromai, line 82: I think the purpose of this `bool()` is to perform a null checking on `ctx.dynamic_data["provenance"]`. I think it would be better if we perform the null checking using an `if ... is None` statement instead. 
Although it may never happen in the current implementation, `bool()` can behave unexpectedly depending on the object we pass to it.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -324,7 +324,7 @@ def run_single(
                 status=SCMStatus.ANALYSIS_FAILED,
             )
 
-        if not provenance_payload and parsed_purl and not config.get_value("path"):
+        if not provenance_payload and parsed_purl:
```
tromai, line 327: I am a bit unclear on why we remove `not config.get_value("path")` here.

benmss: I think previously we wanted to prevent provenance discovery for runs where a repo path has been provided. In the current state, I could not see a good reason to keep it, while removing it allows for more provenance discovery in cases where there is otherwise none.

src/macaron/slsa_analyzer/checks/provenance_verified_check.py
```diff
@@ -92,8 +92,9 @@ def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
                 result_type=CheckResultType.FAILED,
             )
 
-        if not (build_type and build_type == "https://github.com/slsa-framework/slsa-github-generator/generic@v1"):
-            # Provenance is verified.
+        if not build_type == "https://github.com/slsa-framework/slsa-github-generator/generic@v1":
```
tromai, line 95: I wonder if the purpose of this change is for simplifying the if statement, or it's because it's failing the integration test :thinking:

benmss: It was intended as a simplification. Also improved in [348ebc7](https://github.com/oracle/macaron/pull/791/commits/348ebc7bb8385e77338bc7f57466ea3799b000d9)

tests/integration/cases/micronaut-projects_micronaut-test/micronaut-test.dl
```diff
@@ -7,9 +7,9 @@ Policy("test_policy", component_id, "") :-
     check_passed(component_id, "mcn_build_as_code_1"),
     check_passed(component_id, "mcn_build_script_1"),
     check_passed(component_id, "mcn_build_service_1"),
-    check_passed(component_id, "mcn_provenance_available_1"),
-    check_passed(component_id, "mcn_provenance_level_three_1"),
     check_passed(component_id, "mcn_version_control_system_1"),
+    check_failed(component_id, "mcn_provenance_available_1"),
```
tromai, line 11: Is it expected for these checks to fail due to the latest refactoring?
I wonder what was the reason for those checks to fail.

tromai: This comment also applies to these test cases where checks turns from pass to fail
- `tests/integration/cases/slsa-framework_slsa-verifier/policy.dl`
- `tests/integration/cases/urllib3_expectation_dir/policy.dl`
- `tests/integration/cases/urllib3_expectation_file/policy.dl`
- `tests/integration/cases/urllib3_invalid_expectation/policy.dl`

benmss: It is expected due to using the actual release instead of the latest. The commits provided for those test cases do not map to tags, and therefore there is no release to retrieve for any of them. 

We could change the commits to ones that have tags.

behnazh-w: Please fix the test case by providing PURL as input instead of commit  so that `mcn_provenance_available_1` passes when a provenance exists. You can also add an additional test case to assert `check_failed(component_id, "mcn_provenance_available_1"),` when the provided commit does not match a release tag.

src/macaron/repo_finder/provenance_finder.py
```diff
@@ -367,3 +375,154 @@ def find_gav_provenance(purl: PackageURL, registry: JFrogMavenRegistry) -> list[
...
+            digest = component.repository.commit_sha
+            tag = None
+            if git_obj:
+                # Use the software component commit to find the tag.
+                if not digest:
+                    logger.debug("Cannot retrieve asset provenance without commit digest.")
+                    return None
+                tags = git_obj.repo.tags
+                for _tag in tags:
+                    if _tag.commit and str(_tag.commit) == digest:
```
tromai, line 432: I think `_tag.commit` will raise ValueError if the tag is pointing to a tree or a blob - https://gitpython.readthedocs.io/en/stable/reference.html#git.refs.tag.TagReference.commit

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -83,414 +58,6 @@ def __init__(self) -> None:
         ]
         super().__init__(check_id=check_id, description=description, depends_on=depends_on, eval_reqs=eval_reqs)
 
-    def find_provenance_assets_on_package_registries(
```
tromai, line 86: Am I correct to say that this method is completely removed because we have functions that obtain provenances from NPM and Jfrog here?
- NPM - https://github.com/oracle/macaron/blob/927d8c3a2e59b0b323036ff5bd309fa765248899/src/macaron/repo_finder/provenance_finder.py#L143
- Jfrog - https://github.com/oracle/macaron/blob/927d8c3a2e59b0b323036ff5bd309fa765248899/src/macaron/repo_finder/provenance_finder.py#L282

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -433,6 +433,41 @@ def run_single(
         analyze_ctx.dynamic_data["expectation"] = self.expectations.get_expectation_for_target(
             analyze_ctx.component.purl.split("@")[0]
         )
+
+        git_service = self._determine_git_service(analyze_ctx)
+        self._determine_ci_services(analyze_ctx, git_service)
+        self._determine_build_tools(analyze_ctx, git_service)
+        self._determine_package_registries(analyze_ctx)
```
tromai, line 440: I am curious whether these operations are any different from what we have in https://github.com/oracle/macaron/blob/927d8c3a2e59b0b323036ff5bd309fa765248899/src/macaron/slsa_analyzer/analyzer.py#L1121

From what I see, `perform_checks()` is called at https://github.com/oracle/macaron/blob/927d8c3a2e59b0b323036ff5bd309fa765248899/src/macaron/slsa_analyzer/analyzer.py#L478
and still contains the logic for findings CI services, Git service, etc.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
benmss: Yes.

src/macaron/slsa_analyzer/analyzer.py
benmss: Indeed they are the same and `perform_checks` should be reduced/removed.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -1110,115 +1109,16 @@ def _determine_ci_services(self, analyze_ctx: AnalyzeContext, git_service: BaseG
...
         )
         for package_registry in PACKAGE_REGISTRIES:
             for build_tool in build_tools:
                 if package_registry.is_detected(build_tool):
                     analyze_ctx.dynamic_data["package_registries"].append(
-                        PackageRegistryInfo(
-                            build_tool=build_tool,
-                            package_registry=package_registry,
-                        )
+                        PackageRegistryInfo(build_tool=build_tool, package_registry=package_registry)
```
tromai, line 1117: nit: I wonder if we can add trailing comma to the list of parameters so that this diff can be removed? 
```suggestion
                        PackageRegistryInfo(
                            build_tool=build_tool,
                            package_registry=package_registry,
                        )
```

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -277,16 +271,6 @@ def check_if_input_repo_commit_provenance_conflict(
...
-    # Check the provenance commit against the input commit.
-    if digest_input and provenance_commit_digest and digest_input != provenance_commit_digest:
-        logger.debug(
-            "The commit digest from input does not match what exists in the provenance. "
-            "Input Commit: %s, Provenance Commit: %s.",
-            digest_input,
-            provenance_commit_digest,
-        )
-        return True
-
```
tromai, line 289: I think the reason we remove this validation is because if the user provide a commit that doesn't map to any  tag, then the provenance finder couldn't find any provenance for the analysis too. 
And even when the user provide a provenance as input, and also provide a commit as input, the [provenance_commit_check](https://github.com/oracle/macaron/blob/9c4444586afce02ca00fe62bae76aed9388684a1/src/macaron/slsa_analyzer/checks/provenance_commit_check.py) would be able to catch that inconsistency.
Is this correct?

benmss: That's right.

tromai (approved):
LGTM. Thanks for the change!

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -56,11 +34,8 @@ class ProvenanceAvailableFacts(CheckFacts):
     #: The primary key.
     id: Mapped[int] = mapped_column(ForeignKey("_check_facts.id"), primary_key=True)  # noqa: A003
 
-    #: The provenance asset name.
-    asset_name: Mapped[str] = mapped_column(String, nullable=True, info={"justification": JustificationType.TEXT})
-
-    #: The URL for the provenance asset.
-    asset_url: Mapped[str] = mapped_column(String, nullable=True, info={"justification": JustificationType.HREF})
```
behnazh-w, line 63: Please add these back as e need to store this info for provenances. Since these two columns are `nullable`, we can leave them unpopulated for now. See this [issue](https://github.com/oracle/macaron/issues/819) to add this metadata back.

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -56,11 +34,8 @@ class ProvenanceAvailableFacts(CheckFacts):
...
+    #: The availability of the provenance.
+    available: Mapped[str] = mapped_column(String, nullable=True, info={"justification": JustificationType.TEXT})
```
behnazh-w, line 38: I don't think this column is adding any valuable info. By the check result, we can already know whether a provenance is available or not and this column can be removed.

tests/integration/cases/urllib3_invalid_expectation/policy.dl
```diff
@@ -7,15 +7,15 @@ Policy("test_policy", component_id, "") :-
...
+    check_passed(component_id, "mcn_version_control_system_1"),
     check_passed(component_id, "mcn_provenance_available_1"),
+    check_passed(component_id, "mcn_provenance_derived_commit_1"),
+    check_passed(component_id, "mcn_provenance_derived_repo_1"),
     check_passed(component_id, "mcn_provenance_level_three_1"),
-    check_passed(component_id, "mcn_version_control_system_1"),
     check_failed(component_id, "mcn_infer_artifact_pipeline_1"),
-    check_failed(component_id, "mcn_provenance_derived_commit_1"),
-    check_failed(component_id, "mcn_provenance_derived_repo_1"),
     check_failed(component_id, "mcn_provenance_witness_level_one_1"),
```
behnazh-w, line 16: I think the `mcn_provenance_expectation_1` should fail, but somehow it is not included in the policy. Can you please add that?

tests/slsa_analyzer/checks/test_provenance_available_check.py
behnazh-w, line 1: We shouldn't remove this test completely. Instead, we should use the existing test provenances to test the check success and failures. Furthermore, we should move some of the tests to the provenance finder test modules.

behnazh-w (approved):

behnazh-w (approved):

PR #796, opened 2024-07-18
fix: use artifact filenames as keys for verifying jfrog assets in provenance_witness_l1_check

tromai:


tromai:
Rebase to obtain the changes from #798

src/macaron/slsa_analyzer/checks/provenance_witness_l1_check.py
```diff
@@ -73,13 +73,15 @@ def verify_artifact_assets(
     look_up: dict[str, dict[str, InTotoV01Subject]] = {}
 
     for subject in subjects:
-        if subject["name"] not in look_up:
-            look_up[subject["name"]] = {}
-        look_up[subject["name"]][subject["digest"]["sha256"]] = subject
+        # Get the artifact name, which should be the last part of the artifact subject value.
+        _, _, artifact_filename = subject["name"].rpartition("/")
```
nicallen, line 77: This code splitting this string is assuming that the name is of the form "https://witness.dev/attestations/product/v0.1/file:path/to/file". I know that this function is currently only called on a list that comes from `extract_build_artifacts_from_witness_subjects`, which currently only produces subjects of that form, but if that changes in future this code would silently break. I think it would be safer for this code to validate that it is a witness file subject before extracting the filename and  throw an exception if it is not (which currently should never happen).

tromai: That's a good point. I have added the validation here - https://github.com/oracle/macaron/pull/796/commits/c57936b1cb81d24905c000439ec02a9a1f0cda5d

nicallen (approved):

behnazh-w:
@tromai Not sure why `mcn_provenance_witness_level_one_1` was originally missing in this witness integration test: `tests/integration/cases/behnazh-w_example-maven-app/policy.dl`. Anyway, let's add it now.

tromai:
> @tromai Not sure why `mcn_provenance_witness_level_one_1` was originally missing in this witness integration test: `tests/integration/cases/behnazh-w_example-maven-app/policy.dl`. Anyway, let's add it now.

Thanks for the catch! I have added it here - https://github.com/oracle/macaron/pull/796/commits/fe787b28666038680f23c865474c6c7a19edd333

behnazh-w (approved):

PR #797, opened 2024-07-21
chore: improve the malicious metadata check

behnazh-w:
This PR refactors and improves the `_detect_malicious_metadata_check`:

- [x] Moves the check under `src/macaron/slsa_analyzer/checks/`
- [x] Refactors the implementation of the check to avoid storing the metadata in the `PyPIRegistry` object and uses the `AssetLocator` representation instead.
- [x] Uses DB `JSON` type to store the serialized metadata info instead of dumping it as a string value.
- [x] Adds a new unit test for the check and improves the other relevant tests.
- [x] Adds the check to the `django` integration test case and its dependencies.
- [x] Ensures that the source code retrieved by the `PyPIRegistry` API is the version that matches the artifact PURL.
- [x] Removes the heuristics that introduce too many FPs.


src/macaron/database/db_custom_types.py
```diff
@@ -60,3 +60,39 @@ def process_result_value(self, value: None | str, dialect: Any) -> None | dateti
...
+
+    def process_bind_param(self, value: None | dict, dialect: Any) -> None | dict:
+        """Process when storing a dict object to the SQLite db.
+
+        value: None | dict
+            The value being stored
+        """
+        if value is None:
+            return None
+        if not isinstance(value, dict):
```
tromai, line 83: Because the type for `value` is `None | dict`, would this check be unnecessary as we have checked for `None` before?

src/macaron/malware_analyzer/datetime_parser.py
```diff
@@ -12,14 +12,17 @@
...
+    ----------
+    datetime_str: str:
+        The datetime string to parse.
+    datetime_format str:
+        The format to use for parsing the datetime string.
 
     Returns
     -------
-        datetime: The parsed datetime object, or None if parsing failed.
+    datetime
```
tromai, line 24:
```suggestion
    datetime | None
```

src/macaron/malware_analyzer/pypi_heuristics/heuristics.py
```diff
@@ -6,51 +6,43 @@
...
     EMPTY_PROJECT_LINK = "empty_project_link"
+
+    #: Indicates that the package contains project links, but all of them are unreachable.
     UNREACHABLE_PROJECT_LINKS = "unreachable_project_links"
+
+    #: Indicates that the package contains only one release.
     ONE_RELEASE = "one_release"
+
+    #: The package has a high release frequency. The average release time (calculated as the sum of all release gaps
+    #: divided by the number of gaps) is below a set threshold, which defaults to 2 days.
```
tromai, line 22: nit: I think we shouldn't mention the default value in this Enum description but instead put it within the heuristic implementation itself or `defaults.ini`.

src/macaron/malware_analyzer/pypi_heuristics/heuristics.py
```diff
@@ -6,51 +6,43 @@
...
     HIGH_RELEASE_FREQUENCY = "high_release_frequency"
+
+    #: Indicates that all releases contain the same content (with identical digests).
     UNCHANGED_RELEASE = "unchanged_release"
-    CLOSER_RELEASE_JOIN_DATE = "closer_release_join_date"
-    SUSPICIOUS_SETUP = "suspicious_setup"
 
+    #: Refers to the gap between the date the maintainer registered their account and the date of the latest release.
```
tromai, line 28: nit: I think that this comment doesn't explicitly represent what is the malicious factor (compared to other descriptions). Perhaps we can change it to `Indicates that a gap between the date the maintainer registered their account and the date of the latest release is small`.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -305,3 +271,146 @@ def get_maintainer_join_date(self, username: str) -> datetime | None:
         res: datetime | None = parse_datetime(datetime_val, datetime_format)
 
         return res.replace(tzinfo=None) if res else None
+
+
+@dataclass
+class PyPIPackageJsonAsset:
```
tromai, line 277: I wonder if there is any reason why `PyPIPackageJsonAsset` doesn't inherit from `NamedTuple` compared to https://github.com/oracle/macaron/blob/da182aaa537367c85625f37160da570d9952c975/src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py#L26

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -305,3 +271,146 @@ def get_maintainer_join_date(self, username: str) -> datetime | None:
...
+    """The package JSON hosted on the PyPI registry.
+
+    This class matches the AssetLocator protocol.
+    """
+
+    #: The target pypi software component.
+    component: Component
```
tromai, line 284: I wonder what information of `component` are we using for this `PyPIPackageJsonAsset` class. 
I believe this class is storing the metadata of a PYPI package. Therefore we should only store the necessary data that is extracted from the original `Component` instance (e.g. the version, the package name) instead, which is how `JFrogMavenAsset` is doing - https://github.com/oracle/macaron/blob/da182aaa537367c85625f37160da570d9952c975/src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py#L26-L41
Unless there is a specific reason why we need to store a reference to the `component` instance that I missed.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -305,3 +271,146 @@ def get_maintainer_join_date(self, username: str) -> datetime | None:
...
+    #: protocol and is not used because pypi API registry does not provide it.
+    size_in_bytes: int
+
+    #: The asset content.
+    package_json: dict
+
+    @property
+    def name(self) -> str:
+        """Get the asset name."""
+        return "package_json"
```
tromai, line 299: I wonder if this is intended to be a constant for all instance of `PyPIPackageJsonAsset`?

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -305,3 +271,146 @@ def get_maintainer_join_date(self, username: str) -> datetime | None:
         res: datetime | None = parse_datetime(datetime_val, datetime_format)
 
         return res.replace(tzinfo=None) if res else None
+
+
+@dataclass
+class PyPIPackageJsonAsset:
+    """The package JSON hosted on the PyPI registry.
```
tromai, line 278: I think this comment gives me an impression that this class is representing a package called JSON published in PYPI. 
Am I correct to say that an instance of `PyPIPackageJsonAsset` represents the JSON content provided by PYPI's API for a package, instead of the actual artifact itself? 
This is different from `JfrogMavenAsset` where the actual artifact is being pointed to.

src/macaron/malware_analyzer/pypi_heuristics/base_analyzer.py
```diff
@@ -5,8 +5,9 @@
 
 from abc import abstractmethod
 
+from macaron.json_tools import JsonType
 from macaron.malware_analyzer.pypi_heuristics.heuristics import HeuristicResult, Heuristics
-from macaron.slsa_analyzer.package_registry.pypi_registry import PyPIRegistry
+from macaron.slsa_analyzer.package_registry.pypi_registry import PyPIPackageJsonAsset
 
 
 class BaseHeuristicAnalyzer:
```
benmss, line 13: It seems like this class could use the `dataclass` annotation.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -305,3 +271,146 @@ def get_maintainer_join_date(self, username: str) -> datetime | None:
...
+        Note: we assume that the path parameters used to construct the URL are sanitized already.
+
+        Returns
+        -------
+        str
+        """
+        json_endpoint = f"pypi/{self.component.name}/json"
+        return urllib.parse.urljoin(self.pypi_registry.registry_url, json_endpoint)
+
+    def download(self) -> bool:
```
tromai, line 314: I think this method is not compatible with the protocol member `download` of `AssetLocator` https://github.com/oracle/macaron/blob/db3231ffe0334c49f78582f6945ac2996087bc99/src/macaron/slsa_analyzer/asset/__init__.py#L27

I believe the reason why this error is not reported is that at call site, we are not annotating with `AssetLocator`. 
https://github.com/oracle/macaron/blob/13803f961e87fc6a6e4c6509f895a9956a70bb5d/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py#L234-L236 and https://github.com/oracle/macaron/blob/13803f961e87fc6a6e4c6509f895a9956a70bb5d/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py#L178-L180 and so on...

I guess in this implementation where `PyPIPackageJsonAsset` is not handled as an `AssetLocator` (the heuristic interface is using `PyPIPackageJsonAsset` in the type annotation), trying to implement `AssetLocator` protocol is not mandatory.  
I am fine with it being just a normal class for handling API data from PYPI's API (which is what it is essentially doing).

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
behnazh-w: `PyPIPackageJsonAsset` is not immutable because the `package_json` attribute will be set upon download. So, `NamedTuple` cannot be used.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -305,3 +271,146 @@ def get_maintainer_join_date(self, username: str) -> datetime | None:
...
+        """Download the package JSON metadata and stores it in the package_json attribute.
+
+        Returns
+        -------
+        bool
+            ``True`` if the asset is downloaded successfully; ``False`` if not.
+        """
+        try:
+            self.package_json = self.pypi_registry.download_package_json(self.url)
```
tromai, line 323: I wonder if instead of storing the API data as an attribute of this class instance (and introduce states into this class), can we return the downloaded JSON file as return value (which mean that we stay away from following the `AssetLocator` Protocol) ? 
I believe `AssetLocator` is supporting a different use case from what we are trying to achieve here (downloading an artifact vs performing an API call) and both `PyPIPackageJsonAsset` and `JfrogMavenAsset` are not handled at the same place, I think not following `AssetLocator` would be feasible.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
behnazh-w: I'm not sure why you are comparing against `JFrogMavenAsset` . Different assets can have different requirements and purposes. The Python JSON metdata asset requires to pass  PURL information, and in future enhancements of the malware detection check requires access to the repository associated with the artifact, which is accessible from the `component`.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
behnazh-w: yes

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
behnazh-w: The important thing here is to locate the data through a URI, which is also happening for `NPMAttestationAsset`: https://github.com/oracle/macaron/blob/behnazh/refactor-python-mal-check/src/macaron/slsa_analyzer/package_registry/npm_registry.py#L265.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
behnazh-w: Yes `download(self)` missed the `dest` argument and I will fix the type checking. However, we should be consistent across registry implementations. When we treat attestations in the npm registry as assets, I would like to follow the same pattern in pypi registry. Note that we are treating content stored in registries that can be located by a URL as assets, and it is not related to calling APIs.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
behnazh-w: See my explanation above for why it's good to use `AssetLocator` for Python metadata content.

The reason for avoiding returning the downloaded JSON as a separate dictionary is to keep the implementation of related APIs self-contained and to avoid storing state in separate containers. You can think of this asset as an object that allows downloading and processing its content.

src/macaron/malware_analyzer/pypi_heuristics/base_analyzer.py
behnazh-w: Do you see any particular benefit?

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
behnazh-w: See commit [7178e3d](https://github.com/oracle/macaron/pull/797/commits/7178e3d575a1086abf740cbab1025abc6c1dfbc7)

src/macaron/database/db_custom_types.py
behnazh-w: See commit [7178e3d](https://github.com/oracle/macaron/pull/797/commits/7178e3d575a1086abf740cbab1025abc6c1dfbc7)

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -305,3 +272,146 @@ def get_maintainer_join_date(self, username: str) -> datetime | None:
...
+                    configured_source_url = urllib.parse.ParseResult(
+                        scheme=self.pypi_registry.fileserver_url_scheme,
+                        netloc=self.pypi_registry.fileserver_url_netloc,
+                        path=parsed_url.path,
+                        params="",
+                        query="",
+                        fragment="",
+                    ).geturl()
+                    logger.debug("Found source URL: %s", configured_source_url)
+                    return configured_source_url
```
tromai, line 402: I have 2 questions about this implementation:
1. What was the reason that lead to the improvement 
```
if distribution.get("python_version") != "source":
    continue
```
to 
```
if distribution.get("packagetype") != "sdist":
    continue
```
2. Why do we intentionally assign `source_url` as an empty string if it's None. From what I understand:
```
if source_url:
    ...
```
would behave the same for None and an empty string (unless the value of an empty string is used elsewhere). If this is True, perhaps an if statement on the None value would be sufficient:
```
if source_url is None:
    continue

...
```

behnazh-w: > 1. What was the reason that lead to the improvement
> 
> 
> ```
> if distribution.get("python_version") != "source":
>     continue
> ```
> 
> to
> 
> ```
> if distribution.get("packagetype") != "sdist":
>     continue
> ```

Checking `packagetype` is more reliable.

behnazh-w: > 2\. Why do we intentionally assign `source_url` as an empty string if it's None.

That's because `source_url` has a `str` type. ` source_url: str = distribution.get("url", "")` is a common anti-pattern that allows the `url` attribute to be `None` but the expected type is `str` and not `None`. In this particular case the following line checks `if source_url is None:` in the following line can be removed.

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -22,31 +24,26 @@
...
-    detail_information: Mapped[str] = mapped_column(
-        String, nullable=False, info={"justification": JustificationType.TEXT}
-    )
+    detail_information: Mapped[dict[str, JsonType]] = mapped_column(DBJsonDict, nullable=False)
 
-    #: The result of heuristic analysis.
-    heuristic_result: Mapped[str] = mapped_column(
-        String, nullable=False, info={"justification": JustificationType.TEXT}
+    #: The result of analysis, which is of dict[Heuristics, HeuristicResult] type.
+    result: Mapped[dict[str, JsonType]] = mapped_column(
```
tromai, line 45: Could you let me know more about the reason why we don't use `dict[Heuristics, HeuristicResult]` directly as the annotation, but we have to instead use `dict[str, JsonType]` ?

behnazh-w: Actually,  I noticed that the comment should change to `dict[str, JsonType]`. And the reason for not using `dict[Heuristics, HeuristicResult]` is because `HeuristicResult` is not a DB supported type.

behnazh-w: Yes, I can also use `dict[Heuristics, HeuristicResult]` since `HeuristicResult` has a `str` type that is supported by the DB.

behnazh-w: See commit [882e4f0](https://github.com/oracle/macaron/pull/797/commits/882e4f0c8d07db37001375d6c2d8bf5b9a64ae3d)

src/macaron/malware_analyzer/pypi_heuristics/base_analyzer.py
benmss: It is already in use elsewhere for similarly simple classes. The same benefit applies here: generating the `__init__` function instead of explicitly adding one that has no non-standard behaviour.

tests/malware_analyzer/pypi/test_empty_project_link_analyzer.py
```diff
@@ -2,58 +2,69 @@
...
+from unittest.mock import MagicMock
+
 from macaron.malware_analyzer.pypi_heuristics.heuristics import HeuristicResult
+from macaron.malware_analyzer.pypi_heuristics.metadata.empty_project_link import EmptyProjectLinkAnalyzer
+from macaron.slsa_analyzer.package_registry.pypi_registry import PyPIPackageJsonAsset
 
 
-def test_analyze_no_links(setup_empty_project_link_analyzer: dict) -> None:
-    """Test for result failed.
+def setup_empty_project_link_analyzer() -> dict:
```
tromai, line 12: I wonder if it makes more sense to use a fixture for this just like the other heuristic tests?

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -274,25 +275,25 @@ def get_maintainer_join_date(self, username: str) -> datetime | None:
 
 
 @dataclass
-class PyPIPackageJsonAsset:
-    """The package JSON hosted on the PyPI registry.
-
-    This class matches the AssetLocator protocol.
-    """
+class PyPIPackageJsonAsset(AssetLocator):
```
benmss, line 278: What is the reasoning behind inheriting the `AssetLocator` protocol here?

tests/malware_analyzer/pypi/test_empty_project_link_analyzer.py
behnazh-w: The issue with adding a fixture in the same module is that we need to suppress pylint's `disable=redefined-outer-name`. Because the fixture is not meaningful outside this test, adding it in `conftest.py` also doesn't make sense. So, setting up the test using a normal function is cleaner.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
behnazh-w: That's used here to trigger type checking.

tests/slsa_analyzer/checks/test_detect_malicious_metadata_check.py
```diff
@@ -0,0 +1,82 @@
...
+from macaron.slsa_analyzer.specs.package_registry_spec import PackageRegistryInfo
+from tests.conftest import MockAnalyzeContext
+
+RESOURCE_PATH = Path(__file__).parent.joinpath("resources")
+
+
+@pytest.mark.parametrize(
+    ("purl", "expected"),
+    [
+        ("pkg:pypi/zlibxjson", CheckResultType.FAILED),
```
tromai, line 28: Do you think it would be possible to test for the result of each heuristic of this package? I believe it would help us in knowing how the heuristic should behave.

tromai: In additions, do you think it would make sense to also test for the scenario where the result is `CheckResultType.PASSED`?

tests/slsa_analyzer/checks/resources/pypi_files/zlibxjson-8.2.source
tromai, line 1: nit: I think it would be better if we check in the source as text files and create a tarball for it during the test. This could help us review the content of the package, and know which part of it is considered "Suspicious" in the test.

tests/malware_analyzer/pypi/test_empty_project_link_analyzer.py
tromai: I believe you can resolve this pylint error by defining a name for the fixture
```
@pytest.fixture(name"boo")
def _boo(...):
    ...
```
However, I don't have a strong opinion on whether we should use fixture or not. Therefore, the current implementation looks good to me.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
tromai: Hmmm I don't think sub-classing a protocol alone is meant to trigger the type checking.
The type checking should also be triggered at call site, where an instance of `PyPIPackageJsonAsset` is accessed.
```
def run_heuristics(asset: AssetLocator):
    ...

def run_check():
    asset_list = [PyPIPackageJsonAsset(...), OtherAsset(...)]
    invalid_one = SomeThingThatIsNotAnAsset(...)
    for asset in asset_list:
        run_heuristics(asset) # No type error.
    
    run_heuristics(invalid_one) # type error.
```
I think this serves the purpose of Protocol, where any classes that behave similar to `AssetLocator` should be allowed here. This enable us to freely extend to other type of asset without having to change the interface of `run_heuristics` (i.e change the type of the `asset` parameter).

tromai: In general I don't think we need to subclass a Protocol as described in this PEP document - https://typing.readthedocs.io/en/latest/spec/protocol.html#explicitly-declaring-implementation

> So while it’s possible to subclass a protocol explicitly, it’s not necessary to do so for the sake of type-checking.

tests/malware_analyzer/pypi/test_empty_project_link_analyzer.py
benmss: I would advocate for using a fixture, as above.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
behnazh-w: No, we don't need to subclass the `Protocol`, but as I said, I have added it to trigger type checking. Right now we don't store this asset like you have shown above, so we can't trigger the type checking at usage time.

tests/slsa_analyzer/checks/resources/pypi_files/zlibxjson.html
```diff
@@ -0,0 +1,968 @@
...
+            <h3>
+Source Distribution            </h3>
+
+                  <div class="file">
+      <div class="file__graphic">
+        <i class="far fa-file" aria-hidden="true"></i>
+      </div>
+
+      <div class="card file__card">
+        <a href="https://files.pythonhosted.org/packages/3e/1e/b1ecb05e7ca1eb74ca6257a7f43d052b90d2ac01feb28eb28ce677a871ab/zlibxjson-8.2.tar.gz">
```
tromai, line 574: This is not relevant to this PR, but I have some ideas for improvements:
1. Because we rely heavily on the HTML content (which is not guarantee to be stable), we need to have an integration test to test whether our logic for obtaining and processing the HTML content is not breaking (similar to what we have for repo finder https://github.com/oracle/macaron/blob/429ba95abc6ee0b917d2115598838b07a4244caf/tests/integration/cases/repo_finder_remote_calls.
2. I don't know if this would be possible. But we can change this HTML content into a template file (using Jinja2, which is already available) where the information that we are interested in (i.e. the source URL) into template string. Then the test will provide the necessary values and generate a HTML content from that template. We can then feel free to view and parameterize those values from within the test module. 

tests/slsa_analyzer/checks/test_detect_malicious_metadata_check.py
behnazh-w: The `PASSED` case is already tested in the integration tests. Testing for `PASSED` requires uploading the resource files to enable offline testing. I could add a test case as I said, but it's also covered in the integration tests any way.

tests/slsa_analyzer/checks/resources/pypi_files/zlibxjson-8.2.source
behnazh-w: That's beyond what this check is doing. This test is designed to exactly what happens in practice, i.e., analyzing a tarball.

tests/slsa_analyzer/checks/resources/pypi_files/zlibxjson.html
behnazh-w: I don't agree with mocking this behavior and that's why I am using the actual HTML content to exactly test HTML content processing. Note that this HTML content is the real one and has not been modified.

tests/slsa_analyzer/checks/test_detect_malicious_metadata_check.py
tromai: Oh right, I forgot that the integration tests for PYPI packages already cover this. Thanks for reminding me.

tests/slsa_analyzer/checks/resources/pypi_files/zlibxjson-8.2.source
tromai: Oh I don't think that my comment is about what the test is doing. It was more about how easier it is to maintain these testing source code files as plain text files compared to a tarball. The tarball creation should belong to the test initialization step and it's not part of the actual test logic.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
tromai: Right, I see.
If we are using `PyPIPackageJsonAsset` as the concrete typing at usage time (i.e. in the `run_heuristics` method), then I don't think we need to subclass Protocol to enable type checking. This is because the type checker already know the concrete type `PyPIPackageJsonAsset` and can provide the necessary type checking at call site. I think there is no issue with trying to make `PyPIPackageJsonAsset` follows `AssetLocator`'s interface, however, sub-classing `AssetLocator` won't help anything in our current use case and can be safely omitted here. 
In the future when we change `run_heuristic` 's interface to use `AssetLocator` (as I demonstrated above) the type checker should not raise any error because `PyPIPackageJsonAsset` has already followed `AssetLocator` interface.

I guess the reason why we didn't notice the incorrect interface for the `download` method is that at call site, we are using the concrete type of `PyPIPackageJsonAsset`. Therefore, the type check has nothing to complain about.

tests/slsa_analyzer/checks/resources/pypi_files/zlibxjson.html
tromai: Ah I see. Thanks Behnaz.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
behnazh-w: >however, sub-classing AssetLocator won't help anything in our current use case and can be safely omitted here.
That's not actually correct. By sub-classing `AssetLocator`, mypy does capture the incompatible signature of the `download` method.

tromai: Yes I agree, if we sub-class `AssetLocator`, that will lead to a constraint on the method interfaces of `PyPIPackageJsonAsset`. Mypy will ensure that `PyPIPackageJsonAsset` must exactly follows `AssetLocator` interface. 
If we don't sub-class `AssetLocator`, the code still work and mypy will only raise error if at call site we are using `download` differently than its definition in `PyPIPackageJsonAsset`. This means that our code is still type-correct. 
I guess if you want to make sure that the definition of `download` in `PyPIPackageJsonAsset` always follows that of `AssetLocator`, I am good with sub-classing `AssetLocator`. 
It's just that if we were sub-classing `AssetLocator` for the usecase of supporting multiple different asset types in `run_heuristics` then I'm afraid we need to change it because it's not recommended in the documentation. But as you say that we are not currently doing it that way, then I don't see any major issue with sub-classing `AssetLocator` to ensure `PyPIPackageJsonAsset` is in the right "shape".

behnazh-w: I noticed that the current implementation is recreating a new `PyPIRegistry` instead of using the detected one from `ctx.dynamic_data["package_registries"]`. This change fixes this issue and also stores the metadata in the `PackageRegistryInfo` object, which also enforces type checking. So, no need for sub-typing anymore.

See commit [a2b0821](https://github.com/oracle/macaron/pull/797/commits/a2b08218e5ac2e67c6d5d5095e3707690f743bce).

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -222,41 +224,44 @@ def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
...
+                    # Create an AssetLocator object for the PyPI package JSON object.
+                    pypi_package_json = PyPIPackageJsonAsset(
+                        component=ctx.component, pypi_registry=pypi_registry, package_json={}
                     )
 
-                    return CheckResultData(
-                        result_tables=result_tables,
-                        result_type=result_type,
-                    )
+                    pypi_registry_info.metadata.append(pypi_package_json)
```
tromai, line 241: Am I correct to say that the instance `pypi_registry_infor` exists solely for type-checking `pypi_package_json` and it's not current used any where below?

behnazh-w: `pypi_registry_info` is instantiated in the `analyzer` module and used here. But the `metadata` attribute is not used anywhere else. Apart from type checking, this implementation is now consistent with other registries and checks that depend on this current check can use the metadata without looking it up again.

src/macaron/malware_analyzer/pypi_heuristics/base_analyzer.py
behnazh-w: I don't have a very good answer for when to use a `dataclass` in general (and it could be opinionated at times), but I usually prefer to use them when there is a need to represent a data content (or collection of states), where `dataclass` can help with initialization, typing, or comparison. That's where we usually use  `dataclass`es in Macaron (there might be exceptions). In this case, `BaseHeuristicAnalyzer` isn't really representing a data component and the number of attributes that need to be set are fairly small. If we just use a `dataclass` to avoid initializing through the `__init__` method even if the number of attributes is small, then shouldn't we basically make all classes `dataclass`es?

tests/slsa_analyzer/checks/resources/pypi_files/zlibxjson-8.2.source
behnazh-w: I see, it would make sense for code reviews. However, adding source Python files would also require changing Macaron's configurations to turn off linting and type checking on these files. We can revisit this issue when we improve the source code analysis for this check.

tests/malware_analyzer/pypi/test_empty_project_link_analyzer.py
behnazh-w: See commit [8f76fc0](https://github.com/oracle/macaron/pull/797/commits/8f76fc04e6464408bdc28debf99adc079a567281)

src/macaron/malware_analyzer/pypi_heuristics/heuristics.py
behnazh-w: See commit [8f76fc0](https://github.com/oracle/macaron/pull/797/commits/8f76fc04e6464408bdc28debf99adc079a567281)

src/macaron/malware_analyzer/pypi_heuristics/heuristics.py
behnazh-w: See commit [8f76fc0](https://github.com/oracle/macaron/pull/797/commits/8f76fc04e6464408bdc28debf99adc079a567281)

src/macaron/malware_analyzer/datetime_parser.py
behnazh-w: See commit [8f76fc0](https://github.com/oracle/macaron/pull/797/commits/8f76fc04e6464408bdc28debf99adc079a567281)

tromai (approved):
LGTM. Thanks for the change.

src/macaron/malware_analyzer/pypi_heuristics/base_analyzer.py
benmss: Looking at other uses within Macaron, I have to agree that this is generally consistent with them. I will mark this as resolved.

benmss (approved):

PR #809, opened 2024-07-31
docs: update Installation Guide with Docker Desktop instructions

jenstroeger:
I tried running Macaron on my Mac and it failed because Docker wasn’t able to write to the host fs. This change adjusts the instructions with the fix.


behnazh-w (approved):
@jenstroeger Thank you for the contribution.

PR #810, opened 2024-07-31
chore: add CLOSER_RELEASE_JOIN_DATE heuristic as a dependency of SUSPICIOUS_SETUP

behnazh-w:
Right now the `CLOSER_RELEASE_JOIN_DATE` heuristic in the `mcn_detect_malicious_metadata_1` check does not depend on any other heuristic results. However, all the heuristic combinations require the `CLOSER_RELEASE_JOIN_DATE` heuristic to fail. This PR adds the `CLOSER_RELEASE_JOIN_DATE` heuristic as a dependency of `SUSPICIOUS_SETUP` to avoid running `SUSPICIOUS_SETUP` analyzer unnecessarily and improve performance.

### DIFF
--- a/src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
+++ b/src/macaron/malware_analyzer/pypi_heuristics/sourcecode/suspicious_setup.py
@@ -26,7 +26,11 @@ class SuspiciousSetupAnalyzer(BaseHeuristicAnalyzer):
     """Analyzer checks heuristic."""
 
     def __init__(self) -> None:
-        super().__init__(name="suspicious_setup_analyzer", heuristic=Heuristics.SUSPICIOUS_SETUP, depends_on=None)
+        super().__init__(
+            name="suspicious_setup_analyzer",
+            heuristic=Heuristics.SUSPICIOUS_SETUP,
+            depends_on=[(Heuristics.CLOSER_RELEASE_JOIN_DATE, HeuristicResult.FAIL)],
+        )
         self.blacklist: list = ["base64", "request"]
 
     def _get_setup_source_code(self, pypi_package_json: PyPIPackageJsonAsset) -> str | None:


tromai:
For this Heuristic - https://github.com/oracle/macaron/blob/5c2dbec369528e47fb2272a3f97852287ba75397/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py#L116-L126

If `SUSPICIOUS_SETUP` passes, it would mean that `CLOSER_RELEASE_JOIN_DATE` will be SKIPPED. This means that we will miss this heuristic combination.

behnazh-w:
> For this Heuristic -
> 
> https://github.com/oracle/macaron/blob/5c2dbec369528e47fb2272a3f97852287ba75397/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py#L116-L126
> 
> If `SUSPICIOUS_SETUP` passes, it would mean that `CLOSER_RELEASE_JOIN_DATE` will be SKIPPED. This means that we will miss this heuristic combination.

That's how the heuristics work and this scenario should not be possible. If heuristic `A` depends on heuristic `B` with result `R`, `A` will only run if `B` runs and returns `R`. With the change in this PR, `SUSPICIOUS_SETUP` will depend on `CLOSER_RELEASE_JOIN_DATE`. So, only if `CLOSER_RELEASE_JOIN_DATE` passes, `SUSPICIOUS_SETUP` will run, and will be skipped otherwise.

tromai:
Ah I see. Thanks for the clarification, I looked at the PR description 
> This PR adds the CLOSER_RELEASE_JOIN_DATE heuristic as a dependency of SUSPICIOUS_SETUP to avoid running SUSPICIOUS_SETUP analyzer unnecessarily and improve performance.
And I thought that we are planning to make `CLOSER_RELEASE_JOIN_DATE` a children of `SUSPICIOUS_SETUP`.

tromai (approved):

PR #812, opened 2024-08-05
chore(deps-dev): add test-docker dependencies to pyproject.toml

behnazh-w:
`cfgv` and `ruamel.yaml`, which are needed for Docker integration tests were missing from the `pyproject.toml`. because the GitHub Action job that runs the docker tests only needs these deps and not the rest, to improve CI runtime I have added a new `test-docker` set of optional dependencies which will by default be installed by the `setup` target in the `Makefile`.


mabdollahpour-ol (approved):
LGTM

PR #816, opened 2024-08-07
docs: add a tutorial for malware detection check

behnazh-w:
This PR adds a new tutorial to showcase the `mcn_detect_malicious_metadata_1` check and adjust a corresponding integration check to make sure the examples in the tutorial are continuously tested.

It also improves the [`Using Macaron`](https://oracle.github.io/macaron/pages/using.html) page to encourage users analyzing an artifact using `-purl` over a repository, and enhances the configuration instructions.

It adds two new helper relations for policies that let users add constraint for the confidence score of a check result:
*  `check_passed_with_confidence`
*  `check_failed_with_confidence`

Finally, it improves the rendering of justification column in the HTML report when the presented data is a dictionary.


src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -141,16 +141,15 @@ class MaliciousMetadataFacts(CheckFacts):
...
         check_id = "mcn_detect_malicious_metadata_1"
-        description = "Check whether the features of package adhere to the heuristics."
-        super().__init__(
-            check_id=check_id,
-            description=description,
-        )
+        description = """This check analyzes the metadata of a package based on reports malicious behavior.
+        Supported ecosystem: PyPI.
+        """
+        super().__init__(check_id=check_id, description=description, eval_reqs=[])
```
tromai, line 152: nit: we can add a trailing comma to the parameters list to simplify this diff.

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -0,0 +1,186 @@
...
+***********
+
+In this tutorial, we run Macaron on the ``django`` Python package as an example with and without its dependencies to check malicious behavior and apply a policy to fail if the corresponding check fails.
+
+
+'''''''''''''''''''''''''''''''''''''
+Analyzing django without dependencies
+'''''''''''''''''''''''''''''''''''''
+
+First, we need to run the ``analyze`` command of Macaron to run a number of :ref:`checks <checks>` on the ``django`` package. In this tutorial, we are interested in the results of the ``mcn_detect_malicious_metadata_1``. Check :ref:`this tutorial <include_exclude_checks>` if you would like to exclude other checks.
```
tromai, line 64:
```suggestion
First, we need to run the ``analyze`` command of Macaron to run a number of :ref:`checks <checks>` on the ``django`` package. In this tutorial, we are interested in the results of the ``mcn_detect_malicious_metadata_1`` check. Check :ref:`this tutorial <include_exclude_checks>` if you would like to exclude other checks.
```

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -0,0 +1,186 @@
...
+
+  open output/reports/pypi/django/django.html
+
+.. _fig_django-malware-check:
+
+.. figure:: ../../_static/images/tutorial_django_5.0.6_detect_malicious_metadata_check.png
+   :alt: Check ``mcn_detect_malicious_metadata_1`` result for ``django@5.0.6``
+   :align: center
+
+|
```
tromai, line 82: I wonder if this character is intended here.

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -0,0 +1,186 @@
...
+Now we can enforce the policy below to ensure that the ``mcn_detect_malicious_metadata_1`` check always passes on ``django`` and its dependencies and none of the dependencies have malicious behavior.
+
+.. code-block:: prolog
+
+  #include "prelude.dl"
+
+  Policy("check-dependencies", component_id, "Check the dependencies of django.") :-
+    is_component(component_id, _),
+    transitive_dependency(component_id, dependency),
+    check_passed(dependency, "mcn_detect_malicious_metadata_1").
```
tromai, line 165: I think we should have another rule for `component_id` passing `mcn_detect_malicious_metadata_1`.

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -0,0 +1,186 @@
...
+To learn more about changing configurations see :ref:`here <change-config>`.
+
+Now we can enforce the policy below to ensure that the ``mcn_detect_malicious_metadata_1`` check always passes on ``django`` and its dependencies and none of the dependencies have malicious behavior.
+
+.. code-block:: prolog
+
+  #include "prelude.dl"
+
+  Policy("check-dependencies", component_id, "Check the dependencies of django.") :-
+    is_component(component_id, _),
```
tromai, line 163: Why does we need `is_component(component_id, _)` while it's not needed in this policy file https://github.com/oracle/macaron/blob/24090493071ca4d9e3e5998338935f6410e9f41c/tests/integration/cases/django_with_dep_resolution_virtual_env_as_input/policy-recursive-deps.dl#L6-L9.
Or are they supposed to be different from each other ? :thinking:

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -0,0 +1,186 @@
...
+    transitive_dependency(component_id, dependency),
+    check_passed(dependency, "mcn_detect_malicious_metadata_1").
+
+  apply_policy_to("check-dependencies", component_id) :-
+      is_component(component_id, purl),
+      match("pkg:pypi.*", purl).
```
tromai, line 169: I think we should match the PURL to `django` instead. If we use `pkg:pypi,*` this policy will be run against all PYPI components which include the direct dependencies. 
If a component doesn't have any dependencies, it will fail the policy, which I don't think is desirable. 
For example, this could be the result of using this policy:
```
component_satisfies_policy
    ['1', 'pkg:pypi/django@5.0.6', 'check-dependencies']
passed_policies
component_violates_policy
    ['2', 'pkg:pypi/asgiref@3.8.1', 'check-dependencies']
    ['3', 'pkg:pypi/sqlparse@0.5.1', 'check-dependencies']
    ['4', 'pkg:pypi/pip@22.3.1', 'check-dependencies']
failed_policies
    ['check-dependencies']
```

tests/integration/cases/django_with_dep_resolution_virtual_env_as_input/policy-all-pypi.dl
```diff
@@ -0,0 +1,11 @@
+/* Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved. */
+/* Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/. */
+
+#include "prelude.dl"
+
+Policy("all-pypi", component_id, "Check al the Python packages.") :-
```
tromai, line 6:
```suggestion
Policy("all-pypi", component_id, "Check all the Python packages.") :-
```

tests/integration/cases/django_with_dep_resolution_virtual_env_as_input/test.yaml
```diff
@@ -36,6 +36,29 @@ steps:
     kind: policy_report
     result: output/policy_report.json
     expected: policy_report.json
+- name: Clean up the database.
+  kind: shell
+  options:
+    cmd: rm -f output/macaron.db
```
tromai, line 42: I wonder why we need to clean up the database here, is it because the policy we use after this step will fail for the component that we analyze in the previous steps ? :thinking:

tests/integration/cases/django_with_dep_resolution_virtual_env_as_input/test.yaml
```diff
@@ -36,6 +36,29 @@ steps:
...
+    - --python-venv
+    - ./django_venv
+- name: Run macaron verify-policy to check for all transitive dependencies.
+  kind: verify
+  options:
+    policy: policy-recursive-deps.dl
+- name: Run macaron verify-policy to check for all Python packages.
+  kind: verify
+  options:
+    policy: policy-all-pypi.dl
```
tromai, line 61: Do you think it's worth having steps to compare the policy result of these 2 verify steps ?

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -0,0 +1,186 @@
...
+    component_satisfies_policy
+        ['1', 'pkg:pypi/django@5.0.6', 'check-django']
+    failed_policies
+    component_violates_policy
+
+Note that the ``match`` constraint applies a regex and can be expanded to enforce the ``mcn_detect_malicious_metadata_1`` check to pass on all Python packages analyzed so far by Macaron:
+
+.. code-block:: prolog
+
+  apply_policy_to("check-django", component_id) :-
```
tromai, line 121: I think in this case, we should rename the policy to `check-all-pypi-packages` or something to prevent confusion:
```
.. code-block:: prolog
  #include "prelude.dl"
  Policy("check-all-pypi-packages", component_id, "Check django artifacts.") :-
      check_passed(component_id, "mcn_detect_malicious_metadata_1").
  apply_policy_to("check-all-pypi-packages", component_id) :-
      is_component(component_id, purl),
      match("pkg:pypi.*", purl).
```

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -0,0 +1,186 @@
...
+
+And pass that to the ``analyze`` command:
+
+.. code-block:: shell
+
+  ./run_macaron.sh --defaults-path configurations.ini analyze -purl pkg:pypi/django@5.0.6 --python-venv "/tmp/.django_venv"
+
+To learn more about changing configurations see :ref:`here <change-config>`.
+
+Now we can enforce the policy below to ensure that the ``mcn_detect_malicious_metadata_1`` check always passes on ``django`` and its dependencies and none of the dependencies have malicious behavior.
```
tromai, line 156:
```suggestion
Now we can enforce the policy below to ensure that the ``mcn_detect_malicious_metadata_1`` check always passes on ``django`` and its dependencies, indicating none of the dependencies have malicious behavior.
```
I think using "and" create a confusion that it's a third thing that the policy ensures, whereas it's more of a conclusion from the policy result (as least for me).

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -0,0 +1,186 @@
...
+  Policy("check-dependencies", component_id, "Check the dependencies of django.") :-
+    is_component(component_id, _),
+    transitive_dependency(component_id, dependency),
+    check_passed(dependency, "mcn_detect_malicious_metadata_1").
+
+  apply_policy_to("check-dependencies", component_id) :-
+      is_component(component_id, purl),
+      match("pkg:pypi.*", purl).
+
+As you can see below, the policy passes for ``django`` and all its transitive dependencies.
```
tromai, line 171:
```suggestion
As you can see below, the policy passes for ``django`` because it doesn't detect malicious behavior for `django` and all its transitive dependencies.
```
Because the target of the Policy is `django`, using `and all its transitive dependencies.` create an impression for me that each transitive dependency is also a target.

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -0,0 +1,186 @@
+.. Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved.
+.. Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+.. _detect-malicious-package:
+
+----------------------------
+Detecting malicious packages
+----------------------------
+
+In this tutorial we show how to use Macaron to find malicious packages. Imagine you have found a Python package and you would like to add this as a dependency to your project, but you are not sure whether you can trust the maintainers or not. You can run Macaron to see if it can detect any malicious behavior. Note that Macaron is an analysis tool and can either miss a malicious behavior or report a false positive. So, while we encourage you to use Macaron, you might still need to triage the results manually.
```
benmss, line 10:
```suggestion
In this tutorial we show how to use Macaron to find malicious packages. Imagine you have found a Python package that you would like to add as a dependency to your project, but whose maintainers you are unsure about. In this case, you can run Macaron to see if it can detect any malicious behavior. Note that Macaron is an analysis tool and can either miss malicious behavior or report a false positive. So while we encourage you to use Macaron, you might still need to triage the results manually.
```

I also think the "Note that Macaron..." part would be better as an actual note.

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -0,0 +1,186 @@
...
+   * - Python packages (PyPI)
+
+.. contents:: :local:
+
+
+**********
+Background
+**********
+
+Detecting malicious behavior in open-source software has been a focus for the `Open Source Security Foundation <https://github.com/ossf>`_ (OpenSSF) community in recent years. One significant initiative is :term:`SLSA`, which offers practical recommendations to enhance the integrity of software packages and infrastructure. Macaron is designed to detect poorly maintained or malicious packages by implementing checks inspired by the SLSA specification. However, some forms of attacks currently fall outside the scope of SLSA version 1—notably, SLSA doesn't address the issue of malicious maintainers. Our primary goal is to make it more difficult for malicious actors to compromise critical supply chains and infrastructure. To achieve this, we're developing new methods to detect when maintainers of open source projects are untrustworthy and deliberately spread malware.
```
benmss, line 26:
```suggestion
Detecting malicious behavior in open-source software has been a focus for the `Open Source Security Foundation <https://github.com/ossf>`_ (OpenSSF) community in recent years. One significant initiative is :term:`SLSA`, which offers practical recommendations to enhance the integrity of software packages and infrastructure. Macaron is designed to detect poorly maintained or malicious packages by implementing checks inspired by the SLSA specification. However, some forms of attacks currently fall outside the scope of SLSA version 1—notably, SLSA doesn't address the issue of malicious maintainers. Our primary goal is to make it more difficult for malicious actors to compromise critical supply chains and infrastructure. To achieve this, we're developing new methods to detect when maintainers of open source projects are untrustworthy and deliberately spreading malware.
```

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -0,0 +1,186 @@
...
+
+.. _fig_django-malware-check:
+
+.. figure:: ../../_static/images/tutorial_django_5.0.6_detect_malicious_metadata_check.png
+   :alt: Check ``mcn_detect_malicious_metadata_1`` result for ``django@5.0.6``
+   :align: center
+
+|
+
+The image above shows the result of the ``mcn_detect_malicious_metadata_1`` check for ``django@5.0.6``. The check has passed, which means this package is not malicious. If a package is malicious, this check fails, and if the ecosystem is not supported, the check returns ``UKNOWN``. You can also see the result of individual heuristics applied in this check under the ``Justification`` column.
```
benmss, line 84:
```suggestion
The image above shows the result of the ``mcn_detect_malicious_metadata_1`` check for ``django@5.0.6``. The check has passed, which means this package is not malicious. If a package is malicious, this check fails. If the ecosystem is not supported, the check returns ``UNKNOWN``. You can also see the result of individual heuristics applied in this check under the ``Justification`` column.
```

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -0,0 +1,186 @@
...
+
+Now we can write a policy to enforce that all versions of ``django`` pass the ``mcn_detect_malicious_metadata_1`` check. The policy will be enforced against the ``analyze`` command run outputs cached in the local database at ``output/macaron.db``.
```
benmss, line 86:
```suggestion
Now we can write a policy to ensure that all versions of ``django`` pass the ``mcn_detect_malicious_metadata_1`` check. The policy will be enforced against the output of the ``analyze`` command that is cached in the local database at ``output/macaron.db``.
```

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -0,0 +1,186 @@
...
+.. code-block:: javascript
+
+    passed_policies
+        ['check-django']
+    component_satisfies_policy
+        ['1', 'pkg:pypi/django@5.0.6', 'check-django']
+    failed_policies
+    component_violates_policy
+
+Note that the ``match`` constraint applies a regex and can be expanded to enforce the ``mcn_detect_malicious_metadata_1`` check to pass on all Python packages analyzed so far by Macaron:
```
benmss, line 117:
```suggestion
Note that the ``match`` constraint applies a regex pattern and can be expanded to ensure the ``mcn_detect_malicious_metadata_1`` check passes on all Python packages analyzed so far by Macaron:
```

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -0,0 +1,186 @@
...
+      is_component(component_id, purl),
+      match("pkg:pypi.*", purl).
+
+Note that if a policy fails to pass, Macaron returns a none-zero error code.
+
+''''''''''''''''''''''''''''''''''
+Analyzing django with dependencies
+''''''''''''''''''''''''''''''''''
+
+Macaron supports analyzing dependencies of a package and runs the same set of checks on the dependencies as the main target package. To analyze the dependencies of ``django@5.0.6`` Python package, you can either :ref:`generate an SBOM <python-sbom>` yourself or :ref:`point Macaron to a virtual environment <python-venv-deps>` where ``django`` is installed.
```
benmss, line 131:
```suggestion
Macaron supports analyzing dependencies of a package, and will run the same set of checks on the dependencies as for the main target package. To analyze the dependencies of the ``django@5.0.6`` Python package, you can either :ref:`generate an SBOM <python-sbom>` yourself or :ref:`point Macaron to a virtual environment <python-venv-deps>` where ``django`` is installed.
```

docs/source/pages/tutorials/detect_malicious_package.rst
benmss: I agree, except it should end with `or any of its transitive dependencies.`

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -0,0 +1,186 @@
...
+  component_satisfies_policy
+      ['1', 'pkg:pypi/django@5.0.6', 'check-dependencies']
+  failed_policies
+  component_violates_policy
+
+***********
+Future Work
+***********
+
+We are actively working on the malware detection analysis in Macaron to improve the precision and support more ecosystems. An upcoming feature is a more advanced source code analysis. Stay tuned and feel free to contribute to improve this check.
```
benmss, line 186:
```suggestion
We are actively working on the malware detection analysis check in Macaron — to improve precision, support more ecosystems, and in particular, perform more advanced source code analysis. Stay tuned and feel free to contribute to improve this check.
```

docs/source/pages/using.rst
```diff
@@ -13,6 +13,98 @@ Using Macaron
 
 .. contents:: :local:
 
+----------------------------------------
+Analyzing an artifact with a PURL string
+----------------------------------------
+
+Macaron can analyze an artifact (and it dependencies) to determine its supply chain security posture. To analyze an artifact, you need to provide the PURL identifier of the artifact:
```
benmss, line 20:
```suggestion
Macaron can analyze an artifact (and its dependencies) to determine its supply chain security posture. To analyze an artifact, you need to provide the PURL identifier of the artifact:
```

tromai (changes requested):

docs/source/pages/using.rst
```diff
@@ -13,6 +13,98 @@ Using Macaron
...
+   * - Maven (Java)
+     - ``pkg:maven/org.apache.xmlgraphics/batik-anim@1.9.1``
+   * - PyPi (Python)
+     - ``pkg:pypi/django@1.11.1``
+   * - Cargo (Rust)
+     - ``pkg:cargo/rand@0.7.2``
+   * - NuGet (.Net)
+     - ``pkg:nuget/EnterpriseLibrary.Common@6.0.1304``
+   * - NPM (NodeJS)
+     - ``pkg:npm/%40angular/animation@12.3.1``
```
benmss, line 43:
```suggestion
     - ``pkg:npm/@angular/animation@12.3.1``
```

docs/source/pages/using.rst
```diff
@@ -13,6 +13,98 @@ Using Macaron
...
+.. code-block:: shell
+
+  ./run_macaron.sh analyze -purl <artifact-purl>
+
+
+''''''''''''''''''''''''''''''''''''''
+Automated repository and commit finder
+''''''''''''''''''''''''''''''''''''''
+
+Macaron is capable of automatically determining the repository and exact commit that matches the given artifact. For repository URLs, this is achieved through examination of SCM meta data found within artifact POM files (for Java), or use of Google's Open Source Insights API (for other languages). For commits, Macaron will attempt to match repository tags with the artifact version being sought, thereby requiring that the repository supports and uses tags on commits that were used for releases.
```
benmss, line 59:
```suggestion
Macaron is capable of automatically determining the repository and exact commit that match a given artifact. For repository URLs, this is achieved through examination of SCM meta data found within artifact POM files (for Java), or use of Google's Open Source Insights API (for other languages). For commits, Macaron will attempt to match repository tags with the artifact version being sought, thereby requiring that the repository supports and uses tags on commits that were used for releases.
```

docs/source/pages/using.rst
```diff
@@ -13,6 +13,98 @@ Using Macaron
...
+
+Macaron will try to discover the corresponding repository of an artifact by default unless it is provided as input as shown `later <#analyze-repo>`_. This feature is enabled by default. To disable, or configure its behaviour in other ways, a custom ``.ini`` file should be passed to Macaron during execution. See `How to change the default configuration <#change-config>`_ for more details.
```
benmss, line 61:
```suggestion
By default, Macaron will try to discover the corresponding repository of an artifact unless it is already provided as input (as shown `later <#analyze-repo>`_). To disable or otherwise configure this behaviour, or others, a custom ``.ini`` file should be passed to Macaron during execution. See `How to change the default configuration <#change-config>`_ for more details.
```

docs/source/pages/using.rst
```diff
@@ -21,7 +113,7 @@ Analyzing a source code repository
 Analyzing a public GitHub repository
 ''''''''''''''''''''''''''''''''''''
 
-Macaron can analyze a GitHub public repository (and potentially the repositories of it dependencies) to determine its SLSA posture following the specification of `SLSA v0.1 <https://slsa.dev/spec/v0.1/>`_.
+Macaron can also analyze a GitHub public repository (and potentially the repositories of it dependencies).
```
benmss, line 116:
```suggestion
Macaron can also analyze a GitHub public repository (and potentially the repositories of its dependencies).
```

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -141,16 +141,15 @@ class MaliciousMetadataFacts(CheckFacts):
 
 
 class DetectMaliciousMetadataCheck(BaseCheck):
-    """This check analyzes the metadata of the pypi package based on seven heuristics."""
+    """This check analyzes the metadata of a package based on reports malicious behavior."""
```
benmss, line 144: > based on reports malicious behavior
Not sure what is supposed to be conveyed here.

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
behnazh-w: I think this is already very easy to read because the number of parameters is very small.

tests/integration/cases/django_with_dep_resolution_virtual_env_as_input/test.yaml
behnazh-w: No, the policy doesn't fail if we don't clean up the database, but I prefer to make sure the results from the previous step does not affect this step.

tests/integration/cases/django_with_dep_resolution_virtual_env_as_input/test.yaml
behnazh-w: The policy output may not be deterministic because the dependencies may have different versions in different runs. Plus, it doesn't add value because checking if the policy passes is the goal here and sufficient for our test.

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: This modification is suggested as a variation of the previous policy and the policy name should not change.

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
tromai: I think what I meant was that we are introducing extra diff when it's not necessary to the commit content (which may affect the ability to git blame and find the original commit/PR that introduce the semantic of this block). I think this is perfectly fine in terms of aesthetic. 
However, I don't have a strong opinion on this so it could be left as it is.

tests/integration/cases/django_with_dep_resolution_virtual_env_as_input/test.yaml
tromai: Hmm if we are expecting it to not fail, it failing unexpectedly would be a good sign for a potential issue/bug. Cleaning it here would prevent us from noticing such issue happening, especially if 3 analyze command are run sequentially similarly to the tutorial.

behnazh-w: The three `analyze` commands in the tutorial are not expected to run sequentially. 

On the contrary, the risk with not removing the cached `macaron.db` between steps is that a test might pass only because the previous step has populated facts accidentally. So, if `analyze` command has run again and a check is unexpectedly failed, depending on the policy, if the target PURL matches the previous run of `analyze` command, we might miss the error.

tromai: Hmm in this case, do you think it's worthwhile separating them into another test directory :thinking: ?

behnazh-w: I don't see any issue with keeping the tests together.

>In additions, if three commands in the tutorial are not expected to run sequentially, should we instruct the user to delete the database too?

Different use cases in this tutorial can run in any order and that doesn't matter. We definitely shouldn't assume they run the examples in order. Regardless of the order, users need to be careful with bugs in their policies if they use the same cached outputs between runs in general and that's independent of this tutorial.

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
behnazh-w: Black already has rules for handling such cases and let's avoid opinionated discussions. That's one one of the reasons we have so many linters in this project to let tools enforce conventions :slightly_smiling_face:

behnazh-w:
I have added two more subsections for the generated VSA and an example usage of confidence score in the policy: [925a5a4](https://github.com/oracle/macaron/pull/816/commits/925a5a46e17b1f6bb7e0730e77e8cd5f1d96b15f)

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: Good catch. I had forgotten to update the policy in docs after making changes in the test. See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: Yes, this was meant to be matching `django` as you can see in the related integration policy `[policy-recursive-deps.dl](https://github.com/oracle/macaron/blob/24090493071ca4d9e3e5998338935f6410e9f41c/tests/integration/cases/django_with_dep_resolution_virtual_env_as_input/policy-recursive-deps.dl#L6-L9)`.

See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

tests/integration/cases/django_with_dep_resolution_virtual_env_as_input/policy-all-pypi.dl
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: I think the note is fine as it is because this paragraph is at the top and is fairly short.

See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/using.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/using.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/using.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/using.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

docs/source/pages/using.rst
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
behnazh-w: See commit [728d753](https://github.com/oracle/macaron/pull/816/commits/728d753678ee948b303d9b2fb9c672f245ed7ee0)

src/macaron/policy_engine/prelude/helper_rules.dl
```diff
@@ -12,6 +12,38 @@ check_passed(component_id, check_name) :- check_result(_, check_name, 1, compone
...
+       The target software component id.
+   check_name: symbol
+       The check name that matches the ``^mcn_([a-z]+_)+([0-9]+)$`` regular expression.
+   confidence: float
+       The confidence score computed for the check result.
+ */
+.decl check_passed_with_confidence(component_id: number, check_name: symbol, confidence: float)
+check_passed_with_confidence(component_id, check_name, confidence) :-
+    check_result(check_id, check_name, 1, component_id),
+    check_facts(check_id, confidence, component_id,_,_).
```
tromai: I am not sure I understand the usage of `check_id` here. The first column of Check Result and Check Fact seems to be the unique ids for two separate tables:
- https://github.com/oracle/macaron/blob/5c2dbec369528e47fb2272a3f97852287ba75397/src/macaron/database/table_definitions.py#L421
- https://github.com/oracle/macaron/blob/5c2dbec369528e47fb2272a3f97852287ba75397/src/macaron/database/table_definitions.py#L389

behnazh-w: Good catch, I had used the wrong column. I renamed the column name to `check_result_id` to avoid mistakes. I also found a type mismatch in the SQLAlchemy mapping that SQLAlchemy/mypy had failed to catch.

See commit [2ff81c5](https://github.com/oracle/macaron/pull/816/commits/2ff81c530dc8336b9247a91e22e0e7d8bac970af)

tromai (approved):
LGTM. Thanks!

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -120,7 +120,46 @@ Note that the ``match`` constraint applies a regex pattern and can be expanded t
       is_component(component_id, purl),
       match("pkg:pypi.*", purl).
 
-Note that if a policy fails to pass, Macaron returns a none-zero error code.
++++++++++++++++++++++++++++++++++++++++
+Verification Summary Attestation report
++++++++++++++++++++++++++++++++++++++++
+
+Additionally, Macaron generates a Verification Summary Attestation (:term:`VSA`) report that contains the policy and information about the analyzed artifact. See :ref:`this page <vsa>` for more details. For instance, the VSA report for the ``check-django`` policy shown above can be viewed by running the command below:
```
benmss, line 127:
```suggestion
Additionally, Macaron generates a Verification Summary Attestation (:term:`VSA`) report that contains the policy, and information about the analyzed artifact. See :ref:`this page <vsa>` for more details. For instance, the VSA report for the ``check-django`` policy shown above can be viewed by running this command:
```

docs/source/pages/tutorials/detect_malicious_package.rst
```diff
@@ -177,6 +216,26 @@ As you can see below, the policy passes because Macaron doesn't detect malicious
   failed_policies
   component_violates_policy
 
+''''''''''''''''''''''''''''''''''''''''
+Require a confidence level in the policy
+''''''''''''''''''''''''''''''''''''''''
+
+Macaron also provides a confidence score for each check result, represented as a value between ``0`` and ``1`` inclusive, respectively. You can incorporate this score into your policy to ensure checks meet a required level of confidence. Currently, Macaron :class:`has these confidence levels <macaron.slsa_analyzer.checks.check_result.Confidence>`. For instance, you might adjust the :ref:`check-dependencies policy shown earlier <django_with_deps>` to require that the ``mcn_detect_malicious_metadata_1`` check passes with a high confidence, i.e., ``1``:
```
benmss, line 223:
```suggestion
Macaron also provides a confidence score for each check result, represented as a value ranging from ``0`` to ``1`` (inclusively). You can incorporate this score into your policy to ensure checks meet a required level of confidence. Currently, Macaron :class:`has these confidence levels <macaron.slsa_analyzer.checks.check_result.Confidence>`. For instance, you might adjust the :ref:`check-dependencies policy shown earlier <django_with_deps>` to require that the ``mcn_detect_malicious_metadata_1`` check passes with a high confidence, i.e., ``1``:
```

behnazh-w: See commit [003f728](https://github.com/oracle/macaron/pull/816/commits/003f7289e091ae385a59712b6711c3edd2f39c91)

docs/source/pages/tutorials/detect_malicious_package.rst
behnazh-w: See commit [003f728](https://github.com/oracle/macaron/pull/816/commits/003f7289e091ae385a59712b6711c3edd2f39c91)

benmss (approved):

PR #817, opened 2024-08-07
chore: refactor provenance level 3 check into analysis

benmss:
This PR refactors the provenance level 3 check so that its logic takes place as part of the analysis, and the results of which are stored in the database along with the provenance itself. This check then only needs to read from the analysis context to report the results. 

To better separate provenance related logic from the Repo / Commit Finder, provenance scripts are moved to a new sub module.


src/macaron/slsa_analyzer/analyze_context.py
```diff
@@ -52,6 +52,8 @@ class ChecksOutputs(TypedDict):
     """The commit digest extracted from provenance, if applicable."""
     provenance_verified: bool
     """True if the provenance exists and has been verified against a signed companion provenance."""
+    provenance_l3_verified: bool
+    """True if the provenance exists and has been verified to the SLSA level 3 standard."""
```
behnazh-w, line 56: I think the provenance-related metadata we are storing in the context object could use some abstraction. We can use (with adjustments if necessary) the `[Provenance` mappings](https://github.com/oracle/macaron/blob/5c6c9138fbbc01c19faaf8bcadf0e5cc09ba1149/src/macaron/database/table_definitions.py#L467-L607) in our data model. We can also associate a level for each provenance, and the `provenance_verified` would then indicate if the provenance has been verified together with the evidence.

https://github.com/oracle/macaron/blob/5c6c9138fbbc01c19faaf8bcadf0e5cc09ba1149/src/macaron/slsa_analyzer/analyze_context.py#L47-L54

We should also add back and store info about the assets: https://github.com/oracle/macaron/issues/819

src/macaron/provenance/provenance_extractor.py
```diff
@@ -318,18 +304,6 @@ def check_if_input_purl_provenance_conflict(
             )
             return True
 
-    # Check the PURL commit against the provenance.
```
tromai, line 321: Could you please remind me why we removed this commit validation ? 🤔

tromai: Ah nevermind, I think I remembered it now. This is because for some projects the commit hash recorded in their provenances are different from the tagged commit. 
This is due to how their pipelines are setup, where the provenances recorded the commits that the pipelines were triggered on. The version tagged commits can be different if they bumped the versions during the pipeline.

src/macaron/provenance/provenance_finder.py
```diff
@@ -476,56 +380,60 @@ def find_provenance_from_ci(analyze_ctx: AnalyzeContext, git_obj: Git | None) ->
     return None
 
 
-def download_provenances_from_github_actions_ci_service(ci_info: CIInfo) -> None:
+def download_provenances_from_ci_service(ci_info: CIInfo, download_path: str) -> None:
```
tromai, line 383: Hmm if we rename this function to `from_ci_service`, does it mean that we can use this for all supported ci services, and not just only GitHub Actions?.

benmss: That should be possible.

src/macaron/database/db_custom_types.py
```diff
@@ -1,13 +1,21 @@
-# Copyright (c) 2023 - 2024, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2023 - 2025, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """This module implements SQLAlchemy type for converting date format to RFC3339 string representation."""
```
nicallen, line 4: Should update this comment, RFC3339DateTime is not the only thing in this module (even before the addition from this change).

src/macaron/provenance/provenance_extractor.py
```diff
@@ -61,12 +56,40 @@ def extract_repo_and_commit_from_provenance(payload: InTotoPayload) -> tuple[str
...
+
+    Parameters
+    ----------
+    payload: InTotoPayload
+        The payload to extract from.
+
+    Returns
+    -------
+    str | None
+        The SLSA version, or None if .
```
nicallen, line 76: If what?

nicallen (approved):

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -42,6 +42,8 @@
 class MaliciousMetadataFacts(CheckFacts):
     """The ORM mapping for justifications in pypi heuristic check."""
 
+    # pylint: disable=unsubscriptable-object
```
behnazh-w, line 45: Could you also include the comment about temporarily silencing pylint here?

benmss: Oh, I was supposed to remove that one.

behnazh-w:
@benmss Please add the new `verify-provenance` cmd line option to the documentation and make sure the tutorials and documentation pages that require provenance verification are also updated : `docs/source/pages/cli_usage/index.rst`

tests/integration/cases/urllib3_expectation_dir/policy.dl
```diff
@@ -17,6 +17,9 @@ Policy("test_policy", component_id, "") :-
     build_tool_check(pip_id, "pip", "python"),
     check_facts(pip_id, _, component_id,_,_),
     check_failed(component_id, "mcn_find_artifact_pipeline_1"),
+    check_passed(component_id, "mcn_provenance_verified_1"),
+    provenance_verified_check(_, build_level, _),
```
behnazh-w, line 21: The `test.yaml` file for this test has not changed in this PR and `--verify-provenance` is not passed to the `analyze` command. How come this check passes? :thinking: 

I have the same comment for the other `urllib3` integration tests.

src/macaron/slsa_analyzer/checks/provenance_l3_check.py
behnazh-w, line 1: Do we still need to keep this check or can we combine it with `mcn_provenance_verified_1`?

behnazh-w:
For some reason the `tests/integration/cases/ossf_scorecard/test.yaml` and `tests/integration/cases/micronaut-projects_micronaut-test/test.yaml` tests, which verify the provenances, do not include `--verify-provenance` but still pass.

src/macaron/slsa_analyzer/checks/provenance_l3_check.py
benmss: I think we could remove it.

tests/integration/cases/urllib3_expectation_dir/policy.dl
benmss: I thought I had also disabled the level 3 verification with the command argument, but apparently not. That was overriding the base verification for these test cases.

PR #824, opened 2024-08-13
test: add analysis report JSON schema validation

### LINKED ISSUE #769: Add a test for the JSON output of the analyze command

tromai:
Closes #769 .

This Pull Request makes the following changes:
- https://github.com/oracle/macaron/pull/824/commits/93d23907f2bbbbb9524b2bba2bb4fc404a50c8b3
    - Add a script to validate the JSON schema of JSON reports generated by Macaron - `tests/analyze_report_json_schema/schema_validate.py`
    - Add a new compare step kind for this validation script in the integration test utility script - https://github.com/oracle/macaron/blob/93d23907f2bbbbb9524b2bba2bb4fc404a50c8b3/tests/integration/run.py#L78
    - Add a schema to the micronaut-test test case. 
- https://github.com/oracle/macaron/pull/824/commits/0dbcadcce8fb51ccded71701718749ff9d69f451
    - Add `jsonschema` to be installed in the docker integration test utility.


nicallen (approved):

tests/integration/cases/micronaut-projects_micronaut-test/report_schema.json
```diff
@@ -0,0 +1,110 @@
...
+    "type": "object",
+    "properties": {
+        "metadata": {
+            "type": "object"
+        },
+        "target": {
+            "type": "object",
+            "properties": {
+                "info": {
+                    "type": "object"
```
behnazh-w, line 14: The schema should include the nested properties of the `target.info` object.

tests/integration/cases/micronaut-projects_micronaut-test/report_schema.json
```diff
@@ -0,0 +1,110 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "macaron-json-report-schema",
+    "title": "Report",
+    "type": "object",
+    "properties": {
+        "metadata": {
+            "type": "object"
```
behnazh-w, line 8: Same comment regarding the nested properties.

tromai: I purposely left out the properties of this `metadata` and the `target.info` fields because we don't have any guarantee that its content is stable and these fields are somewhat ignored in the old JSON compare script. For example, the content of `target.info` can have null for the repository path because the PURL provided doesn't have any repository. 
I can definitely add the schema for those fields as we only use this schema specifically for this test case, where we have a deterministic set of parameters and expectation.

behnazh-w: > the content of `target.info` can have null for the repository path

Even if the `repository path` is null, the JSON output should still have this key, and regardless of whether we have been checking these properties in the original comparison script or not, we should check them now because users might be consuming them directly from the JSON outputs.

tromai: I have improved the schema in this commit - https://github.com/oracle/macaron/pull/824/commits/b80bc763e1b4b58a6f2ada5d5921a6475e4b4ccd

tromai:
Because of a need to share the output JSON schema, I have move the output JSON schema out of the test case directory.
I have also add a new Step in the integration test utility script - https://github.com/oracle/macaron/pull/824/commits/3d43defbb8fa89569f4422a25567239fa4ae43e4. 
This is because we cannot reuse the Compare script for schema validation as they have different semantics. For example, the update mode applies to the CompareStep, but not for schema validation.

tests/integration/README.md
```diff
@@ -218,6 +219,13 @@ In case you want to debug the utility script itself, there is the verbose mode f
 * `result` (`string`, required): The output file (a relative path from test case directory).
 * `expected` (`string`, required): The expected output file (a relative path from test case directory).
 
+### Vallidate Schema step options Schema
+
+* `kind` (`"json_schema"`, required): The kind of schema validation to perform. For now, only json-schema is supported.
+* `result` (`string`, required): The output file (a relative path from test case directory).
+* `schema_type` (`output_json_report`, required): The type of schema for the validation. These are the default schemas available for the integration test.
+* `custom_schema_path` (`string`, optional): The path to the custom schema (a relative path from test case directory). If it is provided, the validation will use this schema and ignore the schema corresponding with `schema_type`.
+
```
nicallen, line 228: I think `schema_type` should instead be `schema_name` or maybe just `schema`, since it refers to a particular schema rather than a "type" of schema (`schema_type` sounds more like what `kind` is for).

Also, there is a typo in `Vallidate` (-> `Validate`).

tromai: I have changed it to `schema` and fix the typo error in this commit - https://github.com/oracle/macaron/pull/824/commits/c56aabd62518f5fe2a27740f6fcb2f8f326ab7be

nicallen (approved):

behnazh-w (approved):

Issue #781, opened 2024-06-26
Expose the Commit Finder functionality via a command line argument


benmss:
The Commit Finder feature of Macaron could be made available via a command line argument for cases where a user wants to find the commit for their artifact but does not necessarily want to analyse it at the same time.

PR #827, opened 2024-08-15
feat: add command to run repo and commit finder without analysis

benmss:
This pull request adds a new command `find-source` that requires a PURL, and optionally accepts a repository path as input. If no repository path is provided, the command will call the Repo Finder and Commit Finder to find the repository and commit of the provided PURL. If a repository is provided as input, only the Commit Finder will be called.

Closes https://github.com/oracle/macaron/issues/781


src/macaron/repo_finder/repo_finder.py
```diff
@@ -148,3 +149,72 @@ def to_repo_path(purl: PackageURL, available_domains: list[str]) -> str | None:
...
+    git_logger.disabled = True
+
+    # Prepare the repo.
+    logger.debug("Preparing repo: %s", found_repo)
+    # Importing here to avoid cyclic import problem.
+    from macaron.slsa_analyzer.analyzer import Analyzer  # pylint: disable=import-outside-toplevel, cyclic-import
+
+    analyzer = Analyzer(global_config.output_path, global_config.build_log_path)
+    repo_dir = os.path.join(analyzer.output_path, analyzer.GIT_REPOS_DIR)
+    git_obj = analyzer.prepare_repo(repo_dir, found_repo, "", "", purl)
```
tromai, line 197: Looking again at the implementation of this `prepare_repo` function, it only uses `self` at a few locations to run static methods:
- https://github.com/oracle/macaron/blob/2d3a64a1b3af25d3dcc025f21d6018af0a15eb8a/src/macaron/slsa_analyzer/analyzer.py#L867
- https://github.com/oracle/macaron/blob/2d3a64a1b3af25d3dcc025f21d6018af0a15eb8a/src/macaron/slsa_analyzer/analyzer.py#L878
- https://github.com/oracle/macaron/blob/2d3a64a1b3af25d3dcc025f21d6018af0a15eb8a/src/macaron/slsa_analyzer/analyzer.py#L913
Do you think it would be worth while to refactor this method to another module to avoid having to silent the pylint `cyclic-import` error?

tromai:
I tried to run the `find-source` command against a Maven PURL that doesn't have a version:
```
$ macaron find-source -purl pkg:maven/apache/maven
2024-08-16 11:16:55,421 [INFO] Setting the output directory to .../output
2024-08-16 11:16:55,422 [INFO] The logs will be stored in debug.log
2024-08-16 11:16:55,424 [ERROR] Could not find repo for PURL: pkg:maven/apache/maven
```
I believe we don't support Maven PURL without a version now - https://github.com/oracle/macaron/blob/44dbf0acbd3dcb9f577db97b5c6876cfb8c87957/src/macaron/repo_finder/repo_finder_java.py#L53-L56

However, the debug message doesn't show up if I ran it without the verbose flag. I believe we should let the user of the `find-source` command know about this behavior by default.

tromai:
I also think it's great to add an integration test case for the new command too.

src/macaron/repo_finder/repo_finder.py
```diff
@@ -218,3 +213,158 @@ def find_source(purl_string: str, repo: str | None) -> bool:
     logger.info("%s/commit/%s", found_repo, digest)
 
     return True
+
+
+def prepare_repo(
```
tromai, line 218: Thanks for making the refactoring. However, I think `prepare_repo`, `get_git_service` and `get_local_repos_path` can be moved to a new package within `src/macaron/` (for example, `prepare_repo` is used both within `analyzer.py` and `repo_finder.py`).
We could name this new package something like "repo_utils" or something like that 🤔 
@behnazh-w I would love to know what you think about this.

src/macaron/repo_finder/repo_finder.py
```diff
@@ -148,3 +131,74 @@ def to_repo_path(purl: PackageURL, available_domains: list[str]) -> str | None:
...
+    # Disable other loggers for cleaner output.
+    analyzer_logger = logging.getLogger("macaron.slsa_analyzer.analyzer")
+    analyzer_logger.disabled = True
+    git_logger = logging.getLogger("macaron.slsa_analyzer.git_url")
+    git_logger.disabled = True
+
+    # Prepare the repo.
+    logger.debug("Preparing repo: %s", found_repo)
+    repo_dir = os.path.join(global_config.output_path, GIT_REPOS_DIR)
+    git_obj = prepare_repo(
```
behnazh-w, line 175: Instead of cloning the repo, I wonder if we can use the GitHub API to get the tags?
```
curl -s https://api.github.com/repos/npm/node-semver/tags | jq '.[] | {tag: .name, commit: .commit.sha}'
```

src/macaron/repo_finder/report.py
```diff
@@ -0,0 +1,106 @@
...
+        The commit hash to report.
+    repo: str
+        The repository to report.
+
+    Returns
+    -------
+    str
+        The JSON report as a string.
+    """
+    data = {"purl": purl, "commit": commit, "repo": repo}
```
behnazh-w, line 32: Can you please add the schema in a separate file and do a validation in the tests? I think it's also good to add `repo_validated`, and `commit_validate` fields which are `False` for now. We will add that in future PRs soon and without introducing a breaking change.

src/macaron/repo_finder/report_schema.py
behnazh-w, line 1: Sorry if I was not clear. By JSON schema, I was referring to something like [our report schema](https://github.com/oracle/macaron/blob/edfe06eb3f9337f86215b253f40f53ca1a0105c9/tests/schema_validation/report_schema.json), where each field is typed and marked as optional/required.

tests/integration/cases/find_source_avaje/test.yaml
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved.
```
tromai, line 1: I have two questions about this integration test:
- Should we add this test to the docker image test (using tag `macaron-docker-image`)?
- I notice that we only run the `find-source` commands in this test. Is this test case *only* intended to test whether the `find-source` commands exit with zero status code? If not, is it possible to validate the output files using the JSON schema file?

src/macaron/repo_finder/repo_utils.py
```diff
@@ -0,0 +1,279 @@
...
+    bool
+        True if the report was created. False otherwise.
+    """
+    report_json = create_report(purl, commit, repo)
+
+    try:
+        purl_object = PackageURL.from_string(purl)
+    except ValueError as error:
+        logger.debug("Failed to parse purl string as PURL: %s", error)
+        return False
```
tromai, line 84: nit: I think this block can be moved before `report_json = create_report(purl, commit, repo)`

src/macaron/repo_finder/repo_finder.py
```diff
@@ -148,3 +132,129 @@ def to_repo_path(purl: PackageURL, available_domains: list[str]) -> str | None:
...
+
+    for tag_line in tag_data.splitlines():
+        tag_line = tag_line.strip()
+        if not tag_line:
+            continue
+        split = tag_line.split("\t")
+        if len(split) != 2:
+            continue
+        possible_tag = split[1]
+        if "{" in possible_tag or "}" in possible_tag:
```
tromai, line 251: I think it would be good to document an example of this scenario. 
I'm curious why `{` or `}` appear in the result 🤔. Do you happen to know the reason behind it?

benmss: I have not found the reason for it. If we take `avaje-prism` as an example the list of reported tags includes:
```
d432a93941b98e4879a99fa294c5b98501c1bc54        refs/tags/avaje-prisms-parent-1.3
db60daeba004c446fbf8e98cacd7f062f47d3985        refs/tags/avaje-prisms-parent-1.3^{}
```
In this case the above snippet may actually be wrong, as if we look at the GitHub UI for tags we find that the commit of the tag `avaje-prism-parent-1.3` is the one of the tag ending with `^{}`. See https://github.com/avaje/avaje-prisms/tags?after=1.6

tests/integration/cases/find_source_avaje/test.yaml
benmss: - Probably makes sense to
- I think adding a validate step also makes sense

src/macaron/repo_finder/repo_finder.py
benmss: Further investigation reveals that tags ending with `^{}` are what the tag is pointing to. Also known as "peeled tags". If they exist, they should be chosen instead so I'll update the code to reflect that.

tromai: Right so from that example `d432a93941b98e4879a99fa294c5b98501c1bc54` is the hash of the annotated tag object. That tag objects references the commit object of `db60daeba004c446fbf8e98cacd7f062f47d3985`

tests/integration/cases/find_source_avaje/test.yaml
```diff
@@ -14,10 +14,20 @@ steps:
...
+    schema: find_source_json_report
+    result: output/reports/maven/io_avaje/avaje-prisms/avaje-prisms.source.json
 - name: Run macaron find source with cloning enabled
   kind: find-source
   options:
     ini: config.ini
     command_args:
     - -purl
     - pkg:maven/io.avaje/avaje-prisms@1.1
+- name: Check that the repository was cloned
```
tromai, line 31: I think I forgot another case: what happen if the step `- name: Run macaron find source` unexpectedly create a directory under `output/git_repos/github_com/avaje/avaje-prisms/` before `- name: Run macaron find source with cloning enabled` runs. 
I think to be on the safe side we can also validate that `output/git_repos/github_com/avaje/avaje-prisms/` doesn't exist after `- name: Run macaron find source` by using:
```yaml
- name: Check that the repository was NOT cloned
  kind: shell
  options:
    cmd: ls output/git_repos/github_com/avaje/avaje-prisms/
    expect_fail: true
```

tromai (approved):
LGTM. Thanks!

behnazh-w (approved):

behnazh-w (approved):

PR #840, opened 2024-08-21
feat!: allow specifying the dependency depth resolution through CLI and make dependency resolution off by default

### LINKED ISSUE #838: Make dependency resolution and analysis optional AND allow specifying the dep depth through CLI
# Description
We have agreed that it's better to have dependency resolution optional for various reasons:
- A lot of internal use cases don't require dependency analysis.
- Dependency resolution and analysis can take a long time to run. For users that want to try Macaron out, it's not ideal to lock them in a very long running process.

In additions, we also want to let the user, if they want to enable automatic dependency resolution, specify the depth of dependency resolution (right not we only support 1. direct dependencies and 2. all transitive dependencies).

# Solution
We will make dependency resolution and analysis off by default. If the user want to run dependency resolution, they must provide an additional flag. 
The `--skip-deps` flag will still be left in the command line interface of Macaron, however, enabling it will not do anything, except printing out a DEPRECATED message. This flag will be completely removed after the next Macaron release (v0.13.0).

The addition flag will be called `--deps-depth`. This flag accepts a value that specify the depth level of dependencies.:
- `1` means direct dependencies
- `0` means no dependency resolution
- `inf` means all transitive dependencies

In theory, we could accept any whole number as the value (e.g. `2`, `3`, etc.) if need arises in the future.

# Tasks
- [x] Make `--skip-deps` do nothing. 
- [x] Add deprecated message if `--skip-deps` is used.
- [x] Remove `--skip-deps` from existing test cases
- [x] Update the Sphinx documentation (e.g https://github.com/oracle/macaron/blob/ac8de708ed6f1740030e0999054fd2de3846579e/docs/source/pages/tutorials/detect_malicious_package.rst#L182-L187 or wherever we uses the recursive flag).
- [x] Update the Sphinx documentation where SBOM is provided as a CLI flag. Because right now we need to set `--deps-depth` for SBOM deps resolution to work.
- [x] Make sure that when `--deps-depth=0`, no dependency resolution is run even if an SBOM is provided,
- [x] Add `--deps-depth=1` to all test cases that require direct dependency resolution
- [x] Add `--deps-depth=inf` to all test cases that require transitive dependency resolution
- [x] Add a flag `--deps-depth`
- [x] Remove the `recursive` option in `[dependency.resolver]` in `defaults.ini`
- [x] Integration tests on invalid/valid input to --deps-depth

tromai:
Closes #838 .

In this Pull Request I made the following changes:
- https://github.com/oracle/macaron/pull/840/commits/63b981d982e829a0b5a8774357c4c6e2cc833ee7 I refactor the dependency resolver base class so that it accepts whether it should analyze all transitive dependencies through a method parameter instead of the `defaults.ini` configuration. 
```ini
[dependency.resolver] 
   recursive = True
```
- https://github.com/oracle/macaron/pull/840/commits/076318a7740af4de7f386a0789eb861849effa30 I modified the Macaron's CLI to make `--skip-deps` an no-op. I also added `--deps-depth` as a new flag for setting the dependency resolution behavior (it's disabled by default).
- https://github.com/oracle/macaron/pull/840/commits/a9dbacb8bcebcd146444a92e790f2e4bab3e14d4, https://github.com/oracle/macaron/pull/840/commits/41e6399a5e2041c3883c8ed2fbea5a881dd949df, https://github.com/oracle/macaron/pull/840/commits/a85e85521ae2e9613e2496f9e55ee83eb085c35e I modified the existing test cases given the changes in Macaron's analyze command CLI.
- https://github.com/oracle/macaron/pull/840/commits/d30215be8642447ebe978e94f6eefe7bd9326812 I added new integration test cases on invalid usage of `--deps-depth`:
    - Providing invalid values to `--deps-depth`
    - Not setting `--deps-depth` when providing SBOM or Python Virtual Environment as input.
- https://github.com/oracle/macaron/pull/840/commits/f2dde1f091b23099853ed62ff322422f4a6e3f08 Update the Sphinx documentation accordingly.

### DIFF
--- a/docs/source/pages/cli_usage/command_analyze.rst
+++ b/docs/source/pages/cli_usage/command_analyze.rst
@@ -35,7 +35,7 @@ Options
 
 .. option:: -sbom SBOM_PATH, --sbom-path SBOM_PATH
 
-    The path to the SBOM of the analysis target. If this is set, dependency resolution must be enabled with '--deps-depth'.
+    The path to the SBOM of the analysis target.
 
 .. option:: -purl PACKAGE_URL, --package-url PACKAGE_URL
 
@@ -77,7 +77,7 @@ Options
 
 .. option::  --python-venv PYTHON_VENV
 
-    The path to the Python virtual environment of the target software component. If this is set, dependency resolution must be enabled with '--deps-depth'.
+    The path to the Python virtual environment of the target software component.
 
 -----------
 Environment
--- a/docs/source/pages/cli_usage/command_analyze.rst
+++ b/docs/source/pages/cli_usage/command_analyze.rst
@@ -22,7 +22,8 @@ Usage
     usage: ./run_macaron.sh analyze
         [-h] [-sbom SBOM_PATH] [-purl PURL] [-rp REPO_PATH] [-b BRANCH]
         [-d DIGEST] [-pe PROVENANCE_EXPECTATION]
-        [--skip-deps] [-g TEMPLATE_PATH]
+        [--skip-deps] [--deps-depth DEPS_DEPTH] [-g TEMPLATE_PATH]
+        [--python-venv PYTHON_VENV]
 
 -------
 Options
@@ -34,7 +35,7 @@ Options
 
 .. option:: -sbom SBOM_PATH, --sbom-path SBOM_PATH
 
-    The path to the SBOM of the analysis target.
+    The path to the SBOM of the analysis target. If this is set, dependency resolution must be enabled with '--deps-depth'.
 
 .. option:: -purl PACKAGE_URL, --package-url PACKAGE_URL
 
@@ -64,11 +65,19 @@ Options
 
 .. option:: --skip-deps
 
-    Skip automatic dependency analysis.
+    DEPRECATED. Dependency resolution is off by default. This flag does nothing and will be removed in the next release.
+
+.. option:: --deps-depth DEPS_DEPTH
+
+    The depth of the dependency resolution. 0: disable, 1: direct dependencies, inf: all transitive dependencies. (Default: 0)
 
 .. option:: -g TEMPLATE_PATH, --template-path TEMPLATE_PATH
 
     The path to the Jinja2 html template (please make sure to use .html or .j2 extensions).
+
+.. option::  --python-venv PYTHON_VENV
+
+    The path to the Python virtual environment of the target software component.If this is set, dependency resolution must be enabled with '--deps-depth'.
 
 -----------
 Environment
--- a/docs/source/pages/cli_usage/command_analyze.rst
+++ b/docs/source/pages/cli_usage/command_analyze.rst
@@ -77,7 +77,7 @@ Options
 
 .. option::  --python-venv PYTHON_VENV
 
-    The path to the Python virtual environment of the target software component.If this is set, dependency resolution must be enabled with '--deps-depth'.
+    The path to the Python virtual environment of the target software component. If this is set, dependency resolution must be enabled with '--deps-depth'.
 
 -----------
 Environment
--- a/docs/source/pages/supported_technologies/index.rst
+++ b/docs/source/pages/supported_technologies/index.rst
@@ -107,6 +107,17 @@ Provenances
        * The provenance should be published on JFrog Artifactory
      - :doc:`page </pages/supported_technologies/jfrog>`
 
+.. _supported_automatic_deps_resolution:
+
+-------------------------------
+Automatic dependency resolution
+-------------------------------
+
+Currently, we support the following type of project for automatic dependency resolution.
+
+* Java Maven
+* Java gradle
+
 --------
 See also
 --------
--- a/docs/source/pages/supported_technologies/index.rst
+++ b/docs/source/pages/supported_technologies/index.rst
@@ -116,7 +116,7 @@ Automatic dependency resolution
 Currently, we support the following type of project for automatic dependency resolution.
 
 * Java Maven
-* Java gradle
+* Java Gradle
 
 --------
 See also
--- a/docs/source/pages/supported_technologies/index.rst
+++ b/docs/source/pages/supported_technologies/index.rst
@@ -117,6 +117,7 @@ Currently, we support the following type of project for automatic dependency res
 
 * Java Maven
 * Java Gradle
+* Python (with a Python virtual environment created and packages installed using Python3.11, see :ref:`providing Python virtual environment <python-venv-deps>`.)
 
 --------
 See also
--- a/docs/source/pages/using.rst
+++ b/docs/source/pages/using.rst
@@ -17,7 +17,7 @@ Using Macaron
 Analyzing an artifact with a PURL string
 ----------------------------------------
 
-Macaron can analyze an artifact (and its dependencies) to determine its supply chain security posture. To analyze an artifact, you need to provide the PURL identifier of the artifact:
+Macaron can analyze an artifact to determine its supply chain security posture. To analyze an artifact, you need to provide the PURL identifier of the artifact:
 
  .. code-block::
 
@@ -51,6 +51,7 @@ To run Macaron on an artifact, we use the following command:
 
   ./run_macaron.sh analyze -purl <artifact-purl>
 
+Macaron can also analyze the package's dependencies. Please see :ref:`automate-deps-resolution`.
 
 ''''''''''''''''''''''''''''''''''''''
 Automated repository and commit finder
@@ -79,6 +80,7 @@ Within the configuration file under the ``repofinder.java`` header, three option
 - ``repo_pom_paths`` (Values: List of POM tags) - Determines where to search for repository information in the POM files. E.g. scm.url.
 - ``find_parents`` (Values: True or False) - When enabled, the Repository Finding feature will also search for repository URLs in parents POM files of the current dependency.
 
+.. note:: Dependency related configurations like ``artifact_repositories`` or ``find_parents`` can affect  :ref:`Macaron automatic dependency resolution <automate-deps-resolution>`.
 
 .. note:: Finding repositories requires at least one remote call, adding some additional overhead to an analysis run.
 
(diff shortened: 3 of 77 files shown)


tests/integration/cases/apache_maven_sbom_with_no_deps_resolution/sbom.json
tromai, line 1: I add this full SBOM file because for this test case because of the following reason:

- I don't want to mock the content of an SBOM
- I want to prevent the case where the analyze command fails because of the SBOM content, not because of Macaron preventing the user from providing an SBOM without setting `--deps-depth` (this is what this case is testing for).

Please let me know if it's okay or should I look into creating a simplified version of this SBOM.

docs/source/pages/cli_usage/command_analyze.rst
```diff
@@ -68,12 +69,20 @@ Options
...
+
+    The depth of the dependency resolution. 0: disable, 1: direct dependencies, inf: all transitive dependencies. (Default: 0)
 
 .. option:: -g TEMPLATE_PATH, --template-path TEMPLATE_PATH
 
     The path to the Jinja2 html template (please make sure to use .html or .j2 extensions).
 
+.. option::  --python-venv PYTHON_VENV
+
+    The path to the Python virtual environment of the target software component.If this is set, dependency resolution must be enabled with '--deps-depth'.
```
nicallen, line 84: Missing space: `component.If`

docs/source/pages/supported_technologies/index.rst
```diff
@@ -107,6 +107,17 @@ Provenances
...
+.. _supported_automatic_deps_resolution:
+
+-------------------------------
+Automatic dependency resolution
+-------------------------------
+
+Currently, we support the following type of project for automatic dependency resolution.
+
+* Java Maven
+* Java gradle
```
nicallen, line 119: Inconsistent capitalisation: `gradle` -> `Gradle`

docs/source/pages/using.rst
```diff
@@ -283,10 +303,12 @@ To run the analysis against that SBOM, run this command:
 
 .. code-block:: shell
 
-  ./run_macaron.sh analyze -purl pkg:maven/org.apache.maven/maven@3.9.7?type=pom -sbom <path_to_sbom>
+  ./run_macaron.sh analyze -purl pkg:maven/org.apache.maven/maven@3.9.7?type=pom -sbom <path_to_sbom> --deps-depth=1
```
nicallen, line 306: Should we instead use `--deps-depth=inf` for the SBOM examples? I doesn't really matter either way, but I would think that probably most people who want to run over an SBOM want to run over everything.

src/macaron/__main__.py
```diff
@@ -147,10 +169,16 @@ def analyze_slsa_levels_single(analyzer_single_args: argparse.Namespace) -> None
             logger.error("Error while loading the input provenance file: %s", error)
             sys.exit(os.EX_DATAERR)
 
+    if analyzer_single_args.skip_deps:
+        logger.warning(
+            "The --skip-deps flag has been deprecated and WILL NOT do anything. "
+            + "Dependency resolution is off by default. This flag will be removed soon."
```
nicallen, line 175: There is a slight inconsistency between the warning message here and the message in the command line help:
```
Dependency resolution is off by default. This flag does nothing and will be removed in the next release.
```
They should probably just have the same text.

tromai: I agree. I changed one of them at one point and forgot to update the second one.

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -147,6 +149,8 @@ def collect_dependencies(self, dir_path: str, target_component: Component) -> di
             Local path to the target repo.
         target_component: Component
             The analyzed target software component.
+        recursive: bool
+            Set to False to get the direct dependencies only (default).
```
nicallen, line 153: Both `collect_dependencies` and `resolve_dependencies` have a `recursive` parameter that does the same thing, but for one the comment only says what happens when it is False, and the other only says what happens when it is True:
```
Set to False to get the direct dependencies only (default).
```
```
If True, perform transitive dependency resolution. Default: False.
```
For consistency, they should have the same description, which should ideally describe both behaviours, e.g.
```
Whether to get all transitive dependencies, otherwise only the direct dependencies will be returned (default: False).
```

docs/source/pages/using.rst
tromai: Yep. I don't see why not switch to `--deps-depth=inf`. 
The reason I chose `--deps-depth=1` is because in this section originally no `recursive` option is set in `defaults.ini`, which means that this example was configured to analyze direct dependencies.

docs/source/pages/cli_usage/command_analyze.rst
tromai: Fixed in https://github.com/oracle/macaron/pull/840/commits/58a8e11146ec8e6dabf2e660522fd264a55ed363

docs/source/pages/supported_technologies/index.rst
tromai: Fixed in https://github.com/oracle/macaron/pull/840/commits/92efd8d5b38ba1cb63d88cd6549e6bac10033178

docs/source/pages/using.rst
tromai: Fixed in https://github.com/oracle/macaron/pull/840/commits/f402af08f16013b234f4b5d5277464d318002cdc

src/macaron/__main__.py
tromai: Fixed in https://github.com/oracle/macaron/pull/840/commits/e880895f6f43889384d7a348754bfca2a93a2d82

src/macaron/dependency_analyzer/cyclonedx.py
tromai: Fixed in https://github.com/oracle/macaron/pull/840/commits/70ed39dcd26459b0fbc6d9e76fa9b87e8fdd53f5

src/macaron/dependency_analyzer/cyclonedx.py
```diff
@@ -728,6 +744,8 @@ def get_deps_from_sbom(self, sbom_path: str | Path, target_component: Component)
             The path to the SBOM file.
         target_component: Component
             The analyzed target software component.
+        recursive: bool
+            Set to False to get the direct dependencies only (default).
```
nicallen, line 748: Missed one.

tromai: Oops. I have fixed it here - https://github.com/oracle/macaron/pull/840/commits/eba8d7d674414df9898ac7ce8c1c7d7fbb1a98ee

nicallen (approved):

docs/source/pages/cli_usage/command_analyze.rst
```diff
@@ -68,12 +69,20 @@ Options
...
+
+    The depth of the dependency resolution. 0: disable, 1: direct dependencies, inf: all transitive dependencies. (Default: 0)
 
 .. option:: -g TEMPLATE_PATH, --template-path TEMPLATE_PATH
 
     The path to the Jinja2 html template (please make sure to use .html or .j2 extensions).
 
+.. option::  --python-venv PYTHON_VENV
+
+    The path to the Python virtual environment of the target software component. If this is set, dependency resolution must be enabled with '--deps-depth'.
```
behnazh-w, line 84: The Python virtual environment can be used for purposes other than dependency resolution, e.g., computing the hash of the package or analyzing its content. So, it should not be necessary to pass `--deps-depth` if the virtual environment path is provided.

tromai: Do you think we should do this to the SBOM path as well (i.e. the user can both provide a path to an SBOM using `--sbom-path` while disabling dependency resolution with `--deps-depth=0`) ?

tromai: I have modified the behavior to let SBOM and Python venv provided without having to enable dependency analysis - https://github.com/oracle/macaron/pull/840/commits/7d2ab75006b729bed17633a7740a2b7b8aa7fcd2

behnazh-w (changes requested):
The `--skip-deps` option is still mentioned in the usage documentation and various tutorials. Could you please remove these references and ensure that the remaining text remains clear and coherent?

docs/source/pages/supported_technologies/index.rst
```diff
@@ -107,6 +107,17 @@ Provenances
...
+.. _supported_automatic_deps_resolution:
+
+-------------------------------
+Automatic dependency resolution
+-------------------------------
+
+Currently, we support the following type of project for automatic dependency resolution.
+
+* Java Maven
+* Java Gradle
```
behnazh-w, line 119: We also support Python if the virtual environment path is provided and packages are built with Python 3.11.

tromai:
> The `--skip-deps` option is still mentioned in the usage documentation and various tutorials. Could you please remove these references and ensure that the remaining text remains clear and coherent?

I have removed the remaining `--skip-deps` in this commit https://github.com/oracle/macaron/pull/840/commits/8b6ceac98c2442da1701a238970ad1a301236e84.
The only places where `--skip-deps` exists are within `__main__.py` and `command_analyze.rst`. In all cases, there are DEPRECATED messages.

docs/source/pages/supported_technologies/index.rst
tromai:
```
* Python (if a Python virtual environment is created and packages are installed using Python3.11, see :ref:` providing Python virtual environment <python-venv-deps>.)
```
How do you think about it ?

behnazh-w: Sounds good.

tromai: Thanks. Added in https://github.com/oracle/macaron/pull/840/commits/87e7bb9f03e43260b91db2669a81008af5972fbc

behnazh-w (approved):

PR #853, opened 2024-09-08
fix: use gnu-sed on mac instead of the built-in sed command

### LINKED ISSUE #852: The sed command in copyright-checker.sh raises error on mac
Macos by default uses BSD implemenation of sed which has different API than gnu-sed. This results in "bad flag in substitute command:" error when there is a file with no copyright header and the pre-commit hook tries to run copyright-checker.sh to automatically fix it.

mabdollahpour-ol:
- Install gnu-sed as part of `setup` (`setup-binaries`). 
- Use `gsed` in `copyright-checker.sh` when running on mac.

Closes #852


behnazh-w:
Can you please rebase on `staging` to get the integration test fix for the CI to pass?

tromai (approved):

behnazh-w (approved):

PR #862, opened 2024-09-17
chore: improve commit finder accuracy

benmss:
This PR makes a number of improvements to the Commit Finder based on results from a much larger dataset than has been previously used for testing.

#### Prefixes

The base prefix case pattern has been made more accepting:
 - Can start with any character instead of requiring an alphabetic character
 - Can end with one or two numbers
 
A new prefix case, "prefix_2", has been added that accepts a string of any alphabetic characters with no prefix separator.

#### Suffixes

The logic for determining when suffix parts should be made optional has been improved. Requirements are:
- Version parts that are alphanumeric
- AND do not come before parts that are purely numeric
- OR version parts that come after a change in separator

Separator pattern logic between version parts has been changed to only allow differences when a version contains more than one separator, e.g. v1.2-3 vs. v1.2.3. Specific separators also become optional in cases where a version part was split from a single part, e.g. 1.3rc5 -> [1, 3, rc5]

#### Preventing Misaligned Matches

The old prefix patterns were designed to help prevent versions from matching incorrectly within a tag, such as part way through numbers, or version parts, e.g. 11.33 should not match 1.33, or 33. To address this a negative look behind has been added for the first version part instead, as well as a new function that performs realignment when a version part has been marked as a prefix by mistake.

#### Pre-step Evaluation

As a pre-step to the full evaluation, a vastly simplified regex has been added that checks if the tag matches the version closely enough, using: `<release_prefix>/<artifact_name>-<v><version>`, with only the version part being required.

#### Sorting Function

The compute similarity function has been extended to consider more information from the tag matches.
This includes preferring shorter prefixes and prefix separators; prefixes that are a superstring of the artifact name; and prefixes and suffixes that have release keywords in them, or use separators that match the version.

### Tests

More regression cases have been added for new tags of interest. Strange unicode tags that were previously rejected despite being valid Git tags are now accepted using the pre-step evaluation. 

The hypothesis test for pattern creation and evaluation has been removed. Creating a slightly less restrictive pattern that is still mostly correct has become far more difficult as the Commit Finder has evolved, and likely not worth the extra effort.

(More unit tests to come in a future PR.)

### DIFF
--- a/src/macaron/repo_finder/commit_finder.py
+++ b/src/macaron/repo_finder/commit_finder.py
@@ -19,10 +19,11 @@ from macaron.slsa_analyzer.git_service import GIT_SERVICES
 logger: logging.Logger = logging.getLogger(__name__)
 
 # An optional named capture group "prefix" that accepts one of the following:
-# - A string of any characters starting with an alphabetic character, ending with one of:
+# - A string of any characters that ends with one of:
 #   - One non-alphanumeric character, one alphabetic character, and one or more numbers.
 #   - One number and one alphabetic character.
 #   - Two alphabetic characters.
+#   - One or two numbers.
 # - OR
 # - Two alphabetic characters.
 # E.g.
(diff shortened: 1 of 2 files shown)


nicallen (approved):

src/macaron/repo_finder/commit_finder.py
```diff
@@ -407,7 +495,33 @@ def match_tags(tag_list: list[str], name: str, version: str) -> list[str]:
     list[str]
         The list of tags that matched the pattern.
     """
-    # Create the pattern for the passed version.
+    logger.debug("Tag Sample: %s", tag_list[:5])
+
+    # If any tag exactly matches the version, return it immediately.
+    # Also allow for an optional 'v' prefix, and tags of the form: <release_prefix>/<artifact_name>-<version>.
+    v_prefix = "(?:v)?" if not version.lower().startswith("v") else ""
```
behnazh-w, line 502: Can you please add one for this `v` prefix case?

tests/repo_finder/test_commit_finder.py
```diff
@@ -199,48 +197,3 @@ def test_pattern_generation(version: str) -> None:
...
-
-input_pattern = re.compile(r"[0-9]{1,3}(\.[0-9a-z]{1,3}){,5}([-+#][a-z0-9].+)?", flags=re.IGNORECASE)
-# These numbers should be kept low as the complex regex makes generation slow.
-VERSION_ITERATIONS = 50  # The number of times to iterate the test_version_to_tag_matching test.
-TAG_ITERATIONS = 1  # The number of tags to generate per version iteration.
-
-
-@given(data())
-@settings(max_examples=VERSION_ITERATIONS, deadline=None)
-def test_version_to_tag_matching(_data: DataObject) -> None:  # noqa: PT019
```
behnazh-w, line 212: I saw in the PR description that this hypothesis test has been removed because
>Creating a slightly less restrictive pattern that is still mostly correct has become far more difficult

But I'm not sure what that means. Was there any test case that failed this test and why is it safe to remove this test?

benmss: It did generate a case that failed: `Version: 1.0.0, Tag: C1:0;0`. The change in separator makes the final `0` an optional suffix, or would except it is fully numeric and therefore not accepted as a suffix.

src/macaron/repo_finder/commit_finder.py
benmss: Not sure what you mean here.

behnazh-w: Since `"(?:v)?"` is used when the version does not start with `v` (`if not version.lower().startswith("v")`), I was wondering about an example and a more elaborate comment to describe this. Because I was expecting the version to start with `v` and a check like below:
```
    v_prefix = "(?:v)?" if version.lower().startswith("v") else ""
```

benmss: Versions typically do not have a `v` prefix, while tags often do. That is why I've added an optional `v` if it's not present in the version. If it is present then it will be matched directly, and I'd expect it to also be present in the tag in that case.

tests/repo_finder/test_commit_finder.py
behnazh-w: Thank you! Since the commit finder performs very complex regex operations, it's important to maintain this hypothesis test to ensure it doesn't throw exceptions or encounter a ReDoS bug. Instead of keeping the `assert match` at the end, we can simply remove it and add a comment explaining why we're not performing that check.

src/macaron/repo_finder/commit_finder.py
behnazh-w: I see. Can you please add this explanation as a comment?

behnazh-w (approved):

PR #864, opened 2024-09-17
feat: obtain Java and Python artifacts from .m2 or Python virtual environment from input

### LINKED ISSUE #811: Use cached artifacts instead of downloading them when possible
We should use the cached artifacts in `.m2` or [Python virtual environment](https://github.com/oracle/macaron/blob/eb4936f905b29ea5cf95655cdcfd4712efd0d00c/src/macaron/config/global_config.py#L47) instead of downloading them both to improve performance and make sure we analyze the artifact that the end user is mostly likely using (which is very likely the cached one).

Here are the checks where we can avoid downloading the artifacts:

- `mcn_provenance_level_three_1`
- `mcn_detect_malicious_metadata_1`

tromai:
Closes #811 
# Description
This pull request adds the following features:
- The ability to provide Macaron with a [maven local repository](https://maven.apache.org/repositories/local.html)
- The ability to resolve the directories that contains artifacts corresponding to a PackageURL from the Maven local repository and/or the Python virtual environment.

## Providing a Maven local repository.
A new CLI parameter is added `--local-maven-repo` that takes a path to a maven local repo. This flag is optional.
If `--local-maven-repo` is not used, Macaron will load the path to the local maven repo (at `$HOME/.m2`) if that directory is available.
Macaron will treat this maven local repo as *READ-ONLY*.

### What directory to mount if Macaron is being run as a Docker container?
If `--local-maven-repo` is provided to `run_macaron.sh`, the directory provided with this flag is mount to `${MACARON_WORKSPACE}/analyze_local_maven_repo_readonly`.

If `--local-maven-repo` is NOT provided to `run_macaron.sh`, `run_macaron.sh` default behavior is consistent with Macaron running as a package, regarding loading the default location `$HOME/.m2` of the host:
- If `$HOME/.m2/` of the host is not available, an empty directory is created within the `output` dir and mounted as read only to `${MACARON_WORKSPACE}/analyze_local_maven_repo_readonly`. Then Macaron running within the Docker container will be provided with this empty directory as `--local-maven-repo`. We do so to avoid Macaron loading `$HOME/.m2` as maven local repo inside the container, as `$HOME/.m2` inside the container is being used by the cyclone-dx SBOM generator. 
- If `$HOME/.m2/` of the host is available, it will be mounted to `${MACARON_WORKSPACE}/analyze_local_maven_repo_readonly`.

## Providing Python virtual environment as input
This has been implemented in #748.

## Resolving artifacts from local repositories for a PackageURL

This is a high-level diagram of finding local artifacts for a PURL

```mermaid
graph LR
purl[Input - PackageURL] -->patt_gen{1 - Generating glob patterns} -->patts(artifact dir glob patterns)

lrps[Input - Local repository path]

patts -->fa{2 - Find artifact dis from local repo path}
lrps -->fa

fa -->result(Output - list of artifact directories existed within the local repo path.)
```

### Generating glob patterns
From a PackageURL, we can directly obtain the glob patterns representing the location of artifact dirs in a local repo. These glob patterns will be different between a maven PURL and a PYPI PURL.

### Glob patterns matching on the repo path corresponding to the input PURL's type.
In Macaron (the main's `Analyzer` instance), we will store mappings between a PURL type (e.g `maven`) to its corresponding local repo path (if provided by the user). Right now, we only have maximum 2 entries:
- `maven` maps to the maven local repo path
- `pypi` maps to the python virtual env path

If the user doesn't provide a local repo path, there will be no entry for it in these mappings.

Using the input PURL's type, we will find the corresponding local repo path from this mapping. 
Then we use the glob patterns obtained earlier to look for artifacts within this path. We don't raise error if we cannot find any artifact dir that matches any of the glob patterns.

## Storing the discovered artifacts paths

They will be stored in the analyze context instance's `dynamic_data` object as a list of string.

# Tasks
- [x] Add `--local-maven-repo` as CLI option.
- [x] Handle mounting .m2 depending on user provided `--local-maven-repo` into the container. 
- [x] Add integration tests for invalid usages of `--local-maven-repo`
- [x] Add unit tests for run_macaron.sh adding `--local-maven-repo` behavior
- [x] Fix and issue with `dumps-defaults` command not being handled in `run_macaron.sh`
- [x] Obtain directories and files corresponding to a Python software component from virtual environment.
    - [x] Add unit tests.
- [x] Obtain directories and files corresponding to a Java software component from the cached .m2.
    - [x] Add unit tests.
- [ ] Add tutorial - will be worked on later - https://github.com/oracle/macaron/issues/936

### DIFF
--- a/src/macaron/__main__.py
+++ b/src/macaron/__main__.py
@@ -91,7 +91,7 @@ def analyze_slsa_levels_single(analyzer_single_args: argparse.Namespace) -> None
         global_config.local_maven_repo = local_maven_repo
     else:
         user_provided_local_maven_repo = analyzer_single_args.local_maven_repo
-        if not os.path.exists(user_provided_local_maven_repo) or not os.path.isdir(user_provided_local_maven_repo):
+        if not os.path.isdir(user_provided_local_maven_repo):
             logger.error("The user provided local Maven repo at %s is not valid.", user_provided_local_maven_repo)
             sys.exit(os.EX_USAGE)
 
--- a/src/macaron/__main__.py
+++ b/src/macaron/__main__.py
@@ -74,7 +74,7 @@ def analyze_slsa_levels_single(analyzer_single_args: argparse.Namespace) -> None
             sys.exit(os.EX_OSFILE)
         global_config.load_python_venv(analyzer_single_args.python_venv)
 
-    # Set Python virtual environment path.
+    # Set local maven repo path.
     if analyzer_single_args.local_maven_repo is None:
         # Load the default user local .m2 directory.
         # Exit on error if $HOME is not set or empty.
--- a/src/macaron/__main__.py
+++ b/src/macaron/__main__.py
@@ -74,6 +74,29 @@ def analyze_slsa_levels_single(analyzer_single_args: argparse.Namespace) -> None
             sys.exit(os.EX_OSFILE)
         global_config.load_python_venv(analyzer_single_args.python_venv)
 
+    # Set Python virtual environment path.
+    if analyzer_single_args.local_maven_repo is None:
+        # Load the default user local .m2 directory.
+        # Exit on error if $HOME is not set or empty.
+        home_dir = os.getenv("HOME")
+        if not home_dir:
+            logger.critical("Environment variable HOME is not set.")
+            sys.exit(os.EX_USAGE)
+
+        local_maven_repo = os.path.join(home_dir, ".m2")
+        if not os.path.isdir(local_maven_repo):
+            logger.debug("The default local Maven repo at %s does not exist. Ignore ...")
+            global_config.local_maven_repo = None
+
+        global_config.local_maven_repo = local_maven_repo
+    else:
+        user_provided_local_maven_repo = analyzer_single_args.local_maven_repo
+        if not os.path.exists(user_provided_local_maven_repo) or not os.path.isdir(user_provided_local_maven_repo):
+            logger.error("The user provided local Maven repo at %s is not valid.", user_provided_local_maven_repo)
+            sys.exit(os.EX_USAGE)
+
+        global_config.local_maven_repo = user_provided_local_maven_repo
+
     analyzer = Analyzer(global_config.output_path, global_config.build_log_path)
 
     # Initiate reporters.
@@ -450,6 +473,14 @@ def main(argv: list[str] | None = None) -> None:
         help=(
             "The path to the Python virtual environment of the target software component. "
             + "If this is set, dependency resolution must be enabled with '--deps-depth'."
+        ),
+    )
+
+    single_analyze_parser.add_argument(
+        "--local-maven-repo",
+        required=False,
+        help=(
+            "The path to the local .m2 directory. If this option is not used, Macaron will use the default location at $HOME/.m2"
         ),
     )
 
--- a/src/macaron/artifact/local_artifact.py
+++ b/src/macaron/artifact/local_artifact.py
@@ -13,8 +13,8 @@ from macaron.artifact.maven import construct_maven_repository_path
 from macaron.errors import LocalArtifactFinderError
 
 
-def construct_local_artifact_paths_glob_pattern_maven_purl(maven_purl: PackageURL) -> list[str] | None:
-    """Return a list of glob pattern(s) representing maven artifacts in a local maven repository.
+def construct_local_artifact_dirs_glob_pattern_maven_purl(maven_purl: PackageURL) -> list[str] | None:
+    """Return a list of glob pattern(s) representing the directory that contains the local maven artifacts for ``maven_purl``.
 
     The glob pattern(s) can be used to search in `<...>/.m2/repository` directory.
 
@@ -48,8 +48,8 @@ def construct_local_artifact_paths_glob_pattern_maven_purl(maven_purl: PackageUR
     return [construct_maven_repository_path(group, artifact, version)]
 
 
-def construct_local_artifact_paths_glob_pattern_pypi_purl(pypi_purl: PackageURL) -> list[str] | None:
-    """Return a list of glob pattern(s) representing python artifacts in a virtual environment.
+def construct_local_artifact_dirs_glob_pattern_pypi_purl(pypi_purl: PackageURL) -> list[str] | None:
+    """Return a list of glob pattern(s) representing directories that contains the artifacts in a Python virtual environment.
 
     The glob pattern(s) can be used to search in `<...>/<python_venv>/lib/python3.x/site-packages`
     directory.
@@ -194,12 +194,16 @@ def get_local_artifact_paths(
     purl: PackageURL,
     local_artifact_repo_path: str,
 ) -> list[str]:
-    """Return the path to local artifacts for a PackageURL.
+    """Return the paths to directories that store local artifacts for a PackageURL.
 
-    We look for local artifacts of this PURL in ``local_artifact_repo_path``.
+    We look for local artifacts of ``purl`` in ``local_artifact_repo_path``.
 
     This function returns a list of paths (as strings), each has the format
-        ``local_artifact_repo_path``/path/to/artifact``
+        ``local_artifact_repo_path``/path/to/artifact_dir``
+
+    This will mean that no path to an artifact is returned. Therefore, it's the responsibility
+    of this function caller to inspect the artifact directory to obtain the required
+    artifact.
 
     We assume that ``local_artifact_repo_path`` exists.
 
@@ -223,7 +227,7 @@ def get_local_artifact_paths(
     purl_type = purl.type
 
     if purl_type == "maven":
-        maven_artifact_patterns = construct_local_artifact_paths_glob_pattern_maven_purl(purl)
+        maven_artifact_patterns = construct_local_artifact_dirs_glob_pattern_maven_purl(purl)
         if not maven_artifact_patterns:
             raise LocalArtifactFinderError(f"Cannot generate maven artifact patterns for {purl}")
 
@@ -233,7 +237,7 @@ def get_local_artifact_paths(
         )
 
     if purl_type == "pypi":
-        pypi_artifact_patterns = construct_local_artifact_paths_glob_pattern_pypi_purl(purl)
+        pypi_artifact_patterns = construct_local_artifact_dirs_glob_pattern_pypi_purl(purl)
         if not pypi_artifact_patterns:
             raise LocalArtifactFinderError(f"Cannot generate Python package patterns for {purl}")
 
--- a/src/macaron/artifact/local_artifact.py
+++ b/src/macaron/artifact/local_artifact.py
@@ -25,17 +25,17 @@ def get_local_artifact_repo_mapper() -> Mapping[str, str]:
     return local_artifact_mapper
 
 
-def construct_local_artifact_path_from_purl(
+def construct_local_artifact_paths_from_purl(
     build_purl_type: str,
     component_purl: PackageURL,
     local_artifact_repo_mapper: Mapping[str, str],
-) -> str | None:
+) -> list[str] | None:
     """Get B."""
     local_artifact_repo = local_artifact_repo_mapper.get(build_purl_type)
     if local_artifact_repo is None:
         return None
 
-    artifact_path = None
+    artifact_path = []
     match build_purl_type:
         case "maven":
             group = component_purl.namespace
@@ -45,10 +45,12 @@ def construct_local_artifact_path_from_purl(
             if group is None or version is None:
                 return None
 
-            artifact_path = os.path.join(
-                local_artifact_repo,
-                "repository",
-                construct_maven_repository_path(group, artifact, version),
+            artifact_path.append(
+                os.path.join(
+                    local_artifact_repo,
+                    "repository",
+                    construct_maven_repository_path(group, artifact, version),
+                )
             )
         case "pypi":
             # TODO: implement this.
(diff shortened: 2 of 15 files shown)


tests/artifact/test_maven.py
```diff
@@ -86,3 +86,78 @@ def test_to_maven_artifact_subject(
         )
         == provenance_payload.statement["subject"][subject_index]
     )
+
+
+@pytest.mark.parametrize(
```
tromai, line 91: These tests haven't been changed. They were just relocated from `tests/slsa_analyzer/package_registry/test_jfrog_maven_registry.py`

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -973,6 +987,30 @@ def _verify_repository_link(self, parsed_purl: PackageURL, analyze_ctx: AnalyzeC
...
+            site_packages_dir_pattern = os.path.join(
+                global_config.python_venv_path,
+                "lib",
+                "python3.*",
+                "site-packages",
+            )
+            site_packages_dirs = glob.glob(site_packages_dir_pattern)
+
+            if len(site_packages_dirs) == 1:
+                local_artifact_mapper["pypi"] = site_packages_dirs.pop()
```
tromai, line 1010: May be a printing out a logging message here if there are more than one `lib/python3.*` directories in the virtual environment would be good.

tests/artifact/test_local_artifact.py
```diff
@@ -0,0 +1,254 @@
...
+            "pkg:maven/com.google.guava/guava@33.2.1-jre",
+            [],
+            {},
+            id="A maven type PURL where no build tool is discovered",
+        ),
+        pytest.param(
+            "pkg:pypi/django@5.0.3",
+            [],
+            {},
+            id="A maven type PURL where no build tool is discovered",
```
tromai, line 163:
```suggestion
            id="A pypi type PURL where no build tool is discovered",
```

src/macaron/artifact/local_artifact.py
```diff
@@ -0,0 +1,275 @@
...
+
+def construct_local_artifact_paths_glob_pattern_pypi_purl(pypi_purl: PackageURL) -> list[str] | None:
+    """Return a list of glob pattern(s) representing python artifacts in a virtual environment.
+
+    The glob pattern(s) can be used to search in `<...>/<python_venv>/lib/python3.x/site-packages`
+    directory.
+
+    Parameters
+    ----------
+    maven_purl : PackageURL
```
nicallen, line 59:
```suggestion
    pypi_purl : PackageURL
```

src/macaron/__main__.py
```diff
@@ -74,6 +74,29 @@ def analyze_slsa_levels_single(analyzer_single_args: argparse.Namespace) -> None
...
+
+        local_maven_repo = os.path.join(home_dir, ".m2")
+        if not os.path.isdir(local_maven_repo):
+            logger.debug("The default local Maven repo at %s does not exist. Ignore ...")
+            global_config.local_maven_repo = None
+
+        global_config.local_maven_repo = local_maven_repo
+    else:
+        user_provided_local_maven_repo = analyzer_single_args.local_maven_repo
+        if not os.path.exists(user_provided_local_maven_repo) or not os.path.isdir(user_provided_local_maven_repo):
```
benmss, line 94: The `exists` check is not needed as it happens as part of `isdir`.

src/macaron/artifact/local_artifact.py
```diff
@@ -0,0 +1,275 @@
...
+        A list of glob patterns or None if an error happened.
+
+    Examples
+    --------
+    >>> from packageurl import PackageURL
+    >>> purl = PackageURL.from_string("pkg:maven/com.oracle.macaron/macaron@0.13.0")
+    >>> construct_local_artifact_paths_glob_pattern_maven_purl(purl)
+    ['com/oracle/macaron/macaron/0.13.0']
+    """
+    if not maven_purl.type == "maven":
```
benmss, line 38:
```suggestion
    if maven_purl.type != "maven":
```

src/macaron/artifact/local_artifact.py
```diff
@@ -0,0 +1,275 @@
...
+            root_dir=local_maven_repo,
+            pathname=pattern,
+        )
+
+        for found_path in found_paths:
+            full_path = os.path.join(local_maven_repo, found_path)
+            if os.path.isdir(full_path):
+                artifact_paths.append(full_path)
+
+    return artifact_paths
```
benmss, line 129:
```suggestion
    return artifact_paths or None
```

src/macaron/artifact/local_artifact.py
```diff
@@ -0,0 +1,275 @@
...
+        if os.path.isdir(entry_path):
+            all_package_dirs.append(entry)
+
+    for package_dir in all_package_dirs:
+        for pattern in glob_patterns:
+            if fnmatch.fnmatch(package_dir.lower(), pattern.lower()):
+                full_path = os.path.join(venv_path, package_dir)
+                artifact_paths.append(full_path)
+
+    return artifact_paths
```
benmss, line 175:
```suggestion
    return artifact_paths or None
```

src/macaron/artifact/local_artifact.py
```diff
@@ -0,0 +1,275 @@
...
+
+        artifact_paths = find_artifact_paths_from_python_venv(
+            venv_path=local_artifact_repo,
+            glob_patterns=pypi_artifact_patterns,
+        )
+
+        if artifact_paths:
+            return artifact_paths
+
+    return None
```
benmss, line 210:
```suggestion
    if build_tool_purl_type == "maven":
        maven_artifact_patterns = construct_local_artifact_paths_glob_pattern_maven_purl(purl)
        if not maven_artifact_patterns:
            return None

        return find_artifact_paths_from_local_maven_repo(
            local_maven_repo=local_artifact_repo,
            glob_patterns=maven_artifact_patterns,
        )
    elif build_tool_purl_type == "pypi":
        pypi_artifact_patterns = construct_local_artifact_paths_glob_pattern_pypi_purl(purl)
        if not pypi_artifact_patterns:
            return None

        return find_artifact_paths_from_python_venv(
            venv_path=local_artifact_repo,
            glob_patterns=pypi_artifact_patterns,
        )

    return None
```
A log message of some kind might also be nice for unsupported purl types.

src/macaron/slsa_analyzer/analyzer.py
benmss: We might also want to make some considerations on which version to choose in that case.

tests/artifact/test_local_artifact.py
```diff
@@ -0,0 +1,254 @@
...
+        get_local_artifact_paths(
+            purl=purl,
+            build_tool_purl_types=build_tool_purl_types,
+            local_artifact_repo_mapper=local_artifact_repo_mapper,
+        )
+        == expectation
+    )
+
+
+def test_get_local_artifact_paths_succeeded_maven(tmp_path: Path) -> None:
```
benmss, line 190: Why are PyPi related things included in this function? And the same question for Maven related things in the next function.

tests/integration/cases/docker_local_maven_repo_input_errors/test.yaml
```diff
@@ -0,0 +1,46 @@
...
+- name: Providing a file path to --local-maven-repo.
+  kind: analyze
+  options:
+    command_args:
+    - -purl
+    - pkg:maven/io.github.behnazh-w.demo/example-maven-app@1.0?type=jar
+    - --local-maven-repo
+    - ./test.txt
+  expect_fail: true
+- name: Clean up the test file.
```
benmss, line 43: Is cleanup not automatically done by the script? Regardless, I'm not against keeping it here for clarity even if that's true.

src/macaron/artifact/local_artifact.py
tromai: My original intention with this function is that:
- If it returns None, it means an error happened.
- If it returns a list, it means no error happened. Therefore, if an empty list is returned, everything went fine but we cannot find the local artifact for the input PURL. We shouldn't treat this as an error.
That's why at this line I don't return None because by doing so I could introduce a third behavior, which I don't think is ideal:
- A None value returned can indicate an error, or no local artifacts found.

I can think of 3 solutions:
1. I can use Exception for error cases, None value for when no local artifact is found, and a non-empty list to indicate some local artifacts are found. But I'm not sure if this is desired.
2. Use Exception instead of None to indicate error and remove the `None` usages in this function.
3. I can mention my original intention more clearly in the doc strings.

src/macaron/artifact/local_artifact.py
tromai: Please see https://github.com/oracle/macaron/pull/864#discussion_r1875343210

src/macaron/artifact/local_artifact.py
tromai: Good idea. I will add a log message for unsupported purl types.

Regarding the suggested change, this is also affected by https://github.com/oracle/macaron/pull/864#discussion_r1875343210. 

Overall, i think it's a good idea to do the return similar as 
```python
return find_artifact_paths_from_python_venv(
    venv_path=local_artifact_repo,
    glob_patterns=pypi_artifact_patterns,
)
```
So that we make sure that if errors happen in `find_artifact_paths_from_*`, this function will return None.

src/macaron/artifact/local_artifact.py
benmss: I think with how Python is, checking that something is actually None and not just falsey is unusual. If there is actually a case where we want to detect the error, maybe an exception is better.

I only really suggested this change because it enables the small simplification I suggested for the `_get_local_artifact_path_for_build_tool_purl_type` function. I don't think either change is essential.

src/macaron/artifact/local_artifact.py
benmss: I can agree with that. And in that case, the `elif` isn't really needed in my suggestion.

tests/artifact/test_local_artifact.py
tromai: This answer also applies to `test_get_local_artifact_paths_succeeded_pypi`.
The reason for having both of those bits originated from my implementation. I intended to support cases where for a single PURL, we could detect multiple build tools from its source code, hence multiple build tool PURL types. This follows what we have here - https://github.com/oracle/macaron/blob/4235041f706c7a811b5ddf369e26549bdea0206f/src/macaron/slsa_analyzer/analyzer.py#L885 where for a single target we could have multiple build tool purl types.
So my implementation ensures that for any PURL, it will be checked against all local repositories corresponding to build tools Macaron discovered for the PURL's source code. However, as we currently only support the type of input PURL to be either "maven" or "pypi", a PURL can only map to one type of local artifact ("maven" or "pypi" or not supported). 

My original intention was to make this as generic as possible. If it's a bit confusing, I can remove this for now and add them back if we think it's necessary 🤔

src/macaron/artifact/local_artifact.py
tromai: I agree. I will switch to use Exception for this one instead. Thanks.

tests/integration/cases/docker_local_maven_repo_input_errors/test.yaml
tromai: The integration test `run.py` script only cleans up the `output` directory. In this test case, I try to mimick the scenario where `test.txt` is provided by the user, hence it's not necessary to be in `output`.
Therefore, it needs to be cleaned up using a shell step here.
I will add a comment to clarify this.

tests/artifact/test_local_artifact.py
benmss: I see. I would still vote for not including them currently then, as my understanding is that they have no effect on the test behaviour.

src/macaron/slsa_analyzer/analyzer.py
tromai: At this stage, if there are multiple `lib/python3.*` directories, I consider that an error and don't add this python virtual environment to the local artifact mapper.
I think we can only choose the version if we decide to accept the case where there are multiple `lib/python3.*` directories.
In addition, we don't enforce the version that was used to created the Python virtual environment provided to Macaron as input, so it's difficult to only support a specific version. (There is a requirement for Python3.11, but it's only if you want to run Python dependency analyzer on it @behnazh-w I would love to know your input on this).

tests/artifact/test_local_artifact.py
tromai: Fixed in https://github.com/oracle/macaron/pull/864/commits/da0f514376752d38af26e2be396c6e31001a7a24

src/macaron/artifact/local_artifact.py
tromai: Fixed in https://github.com/oracle/macaron/pull/864/commits/da0f514376752d38af26e2be396c6e31001a7a24

src/macaron/artifact/local_artifact.py
```diff
@@ -0,0 +1,275 @@
...
+        A list of glob patterns or None if an error happened.
+
+    Examples
+    --------
+    >>> from packageurl import PackageURL
+    >>> purl = PackageURL.from_string("pkg:pypi/django@1.11.1")
+    >>> construct_local_artifact_paths_glob_pattern_pypi_purl(purl)
+    ['django', 'django-1.11.1.dist-info', 'django-1.11.1.data']
+    """
+    if not pypi_purl.type == "pypi":
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle/macaron/pull/864/commits/bd5eeaa78fde8fe25b550ec08ab74815aab3c04f

src/macaron/artifact/local_artifact.py
tromai: Fixed in https://github.com/oracle/macaron/pull/864/commits/bd5eeaa78fde8fe25b550ec08ab74815aab3c04f

src/macaron/__main__.py
tromai: Fixed in https://github.com/oracle/macaron/pull/864/commits/7f9fb29713687415c7ff7f935b128e75c575f3be

tests/artifact/test_local_artifact.py
tromai: I think the unit tests will need to be changed a bit. 
This is because if we remove the behavior of looking for artifacts in each build tool purl type, and only use the type of the input PURL, then the interface of `get_local_artifact_paths` will become
```
def get_local_artifact_paths(
    purl: PackageURL,
    local_artifact_repo_path: str,
) -> list[str]:
```

tests/integration/cases/docker_local_maven_repo_input_errors/test.yaml
tromai: Addressed in https://github.com/oracle/macaron/pull/864/commits/5c698ff880184c881e653f88330a9724c42e0f44

tests/artifact/test_local_artifact.py
tromai: Addressed in https://github.com/oracle/macaron/pull/864/commits/7adbee2cbd4ee972af55aefc50f7acd4fe6214c2

src/macaron/artifact/local_artifact.py
tromai: Addressed in https://github.com/oracle/macaron/pull/864/commits/7adbee2cbd4ee972af55aefc50f7acd4fe6214c2

src/macaron/artifact/local_artifact.py
tromai: Addressed in https://github.com/oracle/macaron/pull/864/commits/7adbee2cbd4ee972af55aefc50f7acd4fe6214c2

src/macaron/slsa_analyzer/analyzer.py
tromai: I have added a log message here - https://github.com/oracle/macaron/pull/864/commits/65c901433bb47b40e0764c6b2429460b8ea31fa5

benmss (approved):

nicallen (approved):

behnazh-w (approved):

PR #866, opened 2024-09-20
fix: prevent endless loop on 403 GitHub response

### LINKED ISSUE #865: Banned GitHub repositories cause endless attempts at GET retrieval
When using GitHub APIs, Macaron will check if the rate limit of the provided token has been reached and sleep if so. This is based on a 403 response status code from GitHub. If a GitHub repository is no longer accessible due to terms of use violation, a 403 response is also received, and in this case Macaron will get stuck in an endless loop.

benmss:
This PR prevents GET requests made by Macaron from looping endlessly when receiving a 403 error that is not related to GitHub's API rate limit. It also adds a configurable retry limit that applies to all request failures.

Closes #865

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -127,7 +127,7 @@ def test_get_http_partial_failure(httpserver: HTTPServer) -> None:
 
 
 def test_get_http_complete_failure(httpserver: HTTPServer) -> None:
-    """Test get http GET operations when too many errors are received and the request fails."""
+    """Test the http GET operation when too many errors are received and the request fails."""
     # Retrieve the allowed number of retries on a failed request.
     target_value = defaults.getint("requests", "error_retries", fallback=5)
 
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -1,14 +1,19 @@
-# Copyright (c) 2022 - 2022, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2024, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """
 This module test the Util methods
 """
-
+from collections.abc import Callable
 from unittest import TestCase
 from unittest.mock import call, patch
 
+from pytest_httpserver import HTTPServer
+from werkzeug import Request, Response
+
 from macaron import util
+from macaron.config.defaults import defaults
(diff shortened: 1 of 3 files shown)


behnazh-w:
To add a unit test, you can use the `pytest_httpserver` with `localhost` as the endpoint. You can then return error codes like `403` with the corresponding response object and make sure the request is attempted for the pre-configured number of times.

tests/test_util.py
```diff
@@ -78,3 +83,42 @@ def test_copy_file_bulk(self) -> None:
...
+        value += 1
+        response = Response()
+        response.status_code = 403 if value <= (target_value + 1) else 200
+        response.headers["X-VALUE"] = str(value)
+        return response
+
+    return generator
+
+
+def test_get_http_failure(httpserver: HTTPServer) -> None:
```
behnazh-w, line 104: To make the goal of this test more clear, it's better to split it to two test functions and describe the goal and expected behavior.

tests/test_util.py
```diff
@@ -78,3 +83,42 @@ def test_copy_file_bulk(self) -> None:
...
+    target_value = defaults.getint("requests", "error_retries", fallback=5)
+
+    # Create and assign the stateful handler.
+    handler = _response_generator(target_value)
+    httpserver.expect_request("").respond_with_handler(handler)
+
+    # Assert the request fails and returns nothing.
+    assert send_get_http_raw(mocked_url) is None
+
+    # Test for a correct response after the expected number of retries and other requests (including this one).
```
behnazh-w, line 119: I'm not sure I follow this part of the test. Doesn't it make more sense to test (in a separate function) that the request succeeds if the response does not have an error code within the allowed number of attempts?

tests/test_util.py
```diff
@@ -78,3 +83,55 @@ def test_copy_file_bulk(self) -> None:
...
+
+    # Test for a correct response after the expected number of retries.
+    response = send_get_http_raw(mocked_url)
+    assert response
+    assert "X-VALUE" in response.headers
+    assert response.headers["X-VALUE"] == str(target_value + 2)
+
+
+def test_get_http_complete_failure(httpserver: HTTPServer) -> None:
+    """Test get http GET operations when too many errors are received and the request fails."""
```
behnazh-w, line 130:
```suggestion
    """Test the http GET operations when too many errors are received and the request fails."""
```

behnazh-w (approved):

PR #872, opened 2024-09-27
refactor: accept provenance data in artifact pipeline check

behnazh-w:
## Refactoring the artifact pipeline detection check

- Renames `mcn_infer_artifact_pipeline_1` to `mcn_find_artifact_pipeline_1`. 
- This check can support all the package registries now.
- Modifies the check fact table schema by adding new columns and allowing some existing columns to be `nullable`. This change enables us to store the reasons for check failures, such as when a GitHub workflow run is deleted, which may result in some previous columns lacking values.
- Improve the heusristics, e.g., if an artifact is published before the corresponding code is committed, there cannot be a CI pipeline that triggered the publishing.
- This check depends on the deploy command identified by the ``mcn_build_as_code_1 check``. If a deploy command is detected, this check will attempt to locate a successful CI  pipeline that triggered the step containing the deploy command.
- When a verifiable provenance is found for an artifact, we use it to obtain the pipeline trigger. Otherwise, we use heuristics to find the triggering pipeline.

## Improvements to ``mcn_build_as_code_1``

- If a provenance is found, we obtain the workflow that has triggered the artifact release.
- Add support for Reusable GitHub Actions that perform automatic deployment. Since we do not analyze the external Reusable GitHub Actions, we use an allow list of approved Actions.
- A new function, ``infer_confidence_deploy_workflow`` is added to ``BaseBuildTool`` to infer the confidence for such Reusable workflows.

## The ``store_inferred_build_info_results`` function

- Renamed ``store_inferred_provenance`` to ``store_inferred_build_info_results``.
- To avoid confusion, we avoid using the term inferred provenance here and instead simplify store build related information in the context object provided to checks.
- Instead of using ``CIInfo["provenances"]`` for inferred build command analysis results, we use a new field: ``CIInfo["build_info_results"]``.

## Provenance Extractor

- New abstractions added to the provenance extractor to reuse the logic for extracting information such as ``ProvenanceBuildDefinition`` and ``ProvenancePredicate``. With these new abstractions, we don't need to hardcode the expected ``buildType`` value while processing a provenance.

## ``find_publish_timestamp``

- Added an API that can obtain the artifact timestamp for all the supported package registries.
- By default we use deps.dev to obtain the timestamp except for Maven artifacts because we have observed that Maven Central has more accurate results.
- Decoupled the Maven Central search API from the repository, making the hostname fully configurable to enable offline testing with a localhost server.

## Tutorial and integration tests

- Changed the ``Detecting a malicious Java dependency uploaded manually to Maven Central`` tutorial to ``Detecting Java dependencies manually uploaded to Maven Central``
- Used ``log4j-core`` artifact instead of ``guava``, which has an automated deployment workflow.
- Fixed the integration tests and added a new one for `log4j-core`.


src/macaron/slsa_analyzer/checks/infer_artifact_pipeline_check.py
```diff
@@ -131,66 +170,147 @@ def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
...
+            # If the artifact is published before the source code is committed, the check should fail.
+            if published_before_commit:
+                return CheckResultData(
+                    result_tables=[
+                        ArtifactPipelineFacts(
+                            from_provenance=bool(prov_workflow),
+                            run_deleted=ci_run_deleted,
+                            published_before_commit=published_before_commit,
+                            confidence=Confidence.HIGH,
+                        )
```
tromai, line 186: The `deploy_workflow` field of `ArtifactPipelineFacts` is not null-able and it's not set for the constructor of this  `ArtifactPipelineFacts` instance. 
If the code hit this location, a `sqlite3.IntegrityError` will be raised when Macaron tries to output these data into the SQLite database.
To reproduce this issue:
```
$ macaron analyze -purl pkg:maven/io.netty/netty-transport-native-epoll@5.0.0.Alpha2
...
Check mcn_find_artifact_pipeline_1: 
        Detects pipelines from which an artifact is published.
        When a verifiable provenance is found for an artifact, we use it to obtain the pipeline trigger.
        FAILED
...
2024-10-06 15:58:36,396 [CRITICAL] Database error (sqlite3.IntegrityError) NOT NULL constraint failed: _artifact_pipeline_check.deploy_workflow
[SQL: INSERT INTO _artifact_pipeline_check (id, deploy_workflow, deploy_job, deploy_step, run_url, from_provenance, run_deleted, published_before_commit) VALUES (?, ?, ?, ?, ?, ?, ?, ?)]
[parameters: (16, None, None, None, None, 0, 1, 1)]
(Background on this error at: https://sqlalche.me/e/20/gkpj)
2024-10-06 15:58:36,396 [CRITICAL] The main repository analysis failed. Cannot generate a report for it.
```

behnazh-w: See commit [1586789](https://github.com/oracle/macaron/pull/872/commits/1586789c2066849543b2d78cc6e7f5ed6a8897f5).

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -355,3 +358,354 @@ def check_if_repository_purl_and_url_match(url: str, repo_purl: PackageURL) -> b
...
+        tuple[str | None, str | None]
+            A tuple containing two elements:
+            - The first element is the build invocation entry point (e.g., workflow name), or None if not found.
+            - The second element is the invocation URL or identifier (e.g., job URL), or None if not found.
+        """
+        if statement["predicate"] is None:
+            return None, None
+
+        attestation_type = json_extract(statement["predicate"], ["attestations", "type"], str)
+        if not self.expected_attestation_type == attestation_type:
```
benmss, line 637:
```suggestion
        if self.expected_attestation_type != attestation_type:
```

src/macaron/slsa_analyzer/build_tool/base_build_tool.py
```diff
@@ -307,14 +307,63 @@ def match_cmd_args(self, cmd: list[str], tools: list[str], args: list[str]) -> b
...
+
+        Returns
+        -------
+        Confidence
+            The confidence level for the deploy command.
+        """
+        # Apply heuristics and assign weights and scores for the discovered evidence.
+        evidence_weight_map = EvidenceWeightMap(
+            [
+                Evidence(name="ci_workflow_deploy", found=False, weight=2),
```
benmss, line 329: The value of the weights here and elsewhere appear somewhat arbitrary. Is there any specific reasoning behind them?

behnazh-w: These weights are primarily based on my observations and experiments, so while they are not arbitrary, they may reflect some bias from the specific experiments I've conducted.

src/macaron/slsa_analyzer/ci_service/github_actions/github_actions_ci.py
```diff
@@ -243,14 +244,69 @@ def has_latest_run_passed(
...
+        Returns
+        -------
+        bool
+            True if the commit date is within the acceptable range from the publish start time,
+                False otherwise. Returns False in case of any errors during timestamp comparisons.
+        """
+        try:
+            if started_at < publish_date_time:
+                # Make sure the source-code commit date is also within acceptable range.
+                acceptable_range = time_range / 2
```
benmss, line 279: Can these lines raise errors?

src/macaron/slsa_analyzer/ci_service/github_actions/github_actions_ci.py
```diff
@@ -243,14 +244,69 @@ def has_latest_run_passed(
...
+                            " is not within %s seconds."
+                        ),
+                        started_at,
+                        commit_date_time,
+                        acceptable_range,
+                    )
+                    return False
+                return True
+
+        # Handle errors for calls to `fromisoformat()` and the time comparison.
```
benmss, line 293: I don't think `fromisoformat` is used here?

src/macaron/slsa_analyzer/package_registry/package_registry.py
```diff
@@ -40,3 +49,77 @@ def is_detected(self, build_tool: BaseBuildTool) -> bool:
...
+                    path=path_params,
+                    query="",
+                    fragment="",
+                )
+            )
+        except ValueError as error:
+            raise InvalidHTTPResponseError("Failed to construct the API URL.") from error
+
+        response = send_get_http_raw(url)
+        if response and response.text and response.status_code == 200:
```
benmss, line 106: The `send_get_http_raw` function will currently return a response with a status code of 200, or None.

tests/integration/cases/micronaut-projects_micronaut-core/config.ini
```diff
@@ -7,7 +7,7 @@ exclude =
     # temporarily because provenances have failed to publish due to an issue in `generator_generic_slsa3.yml@v1.9.0`:
     # https://github.com/slsa-framework/slsa-github-generator/issues/3350
     mcn_provenance_available_1
-    # Exclude `mcn_infer_artifact_pipeline_1`, due to a non-deterministic behavior in deploy command detection,
+    # Exclude `cn_find_artifact_pipeline_1`, due to a non-deterministic behavior in deploy command detection,
     # which will be fixed in PR #673.
-    mcn_infer_artifact_pipeline_1
+    cn_find_artifact_pipeline_1
```
benmss, line 12: What happened here?

src/macaron/slsa_analyzer/build_tool/base_build_tool.py
benmss: That's good enough for me.

src/macaron/repo_finder/provenance_extractor.py
behnazh-w: See commit [c9761dc](https://github.com/oracle/macaron/pull/872/commits/c9761dca9c853e65a196c033ebe45b53bbf3ac05).

src/macaron/slsa_analyzer/ci_service/github_actions/github_actions_ci.py
behnazh-w: Yes, this was left from another change:. See commit [c9761dc](https://github.com/oracle/macaron/pull/872/commits/c9761dca9c853e65a196c033ebe45b53bbf3ac05).

src/macaron/slsa_analyzer/package_registry/package_registry.py
behnazh-w: See commit [c9761dc](https://github.com/oracle/macaron/pull/872/commits/c9761dca9c853e65a196c033ebe45b53bbf3ac05).

tests/integration/cases/micronaut-projects_micronaut-core/config.ini
behnazh-w: Oops! We actually don't need to exclude this as the issue has been addressed. See commit [c9761dc](https://github.com/oracle/macaron/pull/872/commits/c9761dca9c853e65a196c033ebe45b53bbf3ac05).

src/macaron/slsa_analyzer/ci_service/github_actions/github_actions_ci.py
benmss: In that case I think it's better to move those lines out of the try block. For example, while also rearranging things a bit:
```python
if started_at >= publish_date_time:
    return False

acceptable_range = time_range / 2
try:
    if timedelta.total_seconds(abs(started_at - commit_date_time)) <= acceptable_range:
        return True
    logger.debug(
        (
            "The difference between GitHub Actions starting time %s and source commit time %s"
            " is not within %s seconds."
        ),
        started_at,
        commit_date_time,
        acceptable_range,
    )

except (ValueError, OverflowError, TypeError) as error:
    logger.debug(error)

return False
```

behnazh-w: The comparison `if started_at >= publish_date_time` can still raise errors. I have moved the assignment out of the try block. See commit [678abb1](https://github.com/oracle/macaron/pull/872/commits/678abb1330c7d667f2319ed7c88985daf0acb0f2).

benmss (approved):

src/macaron/config/defaults.ini
```diff
@@ -145,7 +145,12 @@ wrapper_files =
     mvnw
 
 [builder.maven.ci.build]
-github_actions = actions/setup-java
+github_actions =
+    actions/setup-java
+    # Parent project used in Maven-based projects of the Apache Logging Services.
+    apache/logging-parent/.github/workflows/build-reusable.yaml
+    # This action can be used to deploy artifacts to a JFrog artifactory server.
```
tromai, line 152: Should this entry belong to `builder.maven.ci.deploy` instead?

src/macaron/config/defaults.ini
```diff
@@ -494,7 +503,7 @@ artifact_extensions =
 # Package registries.
 [package_registry]
 # The allowed time range (in seconds) from a deploy workflow run start time to publish time.
-publish_time_range = 3600
+publish_time_range = 7200
```
tromai, line 506: I wonder why did we decide to increase the publish time range ?

src/macaron/slsa_analyzer/checks/infer_artifact_pipeline_check.py
tromai, line 1: Should we rename this file name to reflect the new name of the check ?

src/macaron/slsa_analyzer/package_registry/package_registry.py
```diff
@@ -40,3 +49,77 @@ def is_detected(self, build_tool: BaseBuildTool) -> bool:
...
         """
+
+    def find_publish_timestamp(self, purl: str, registry_url: str | None = None) -> datetime:
+        """Retrieve the publication timestamp for a package specified by its purl from the deps.dev repository by default.
+
+        This method constructs a request URL based on the provided purl, sends an HTTP GET
+        request to fetch metadata about the package, and extracts the publication timestamp
+        from the response.
+
+        Note: The method expects the response to include a ``version`` field with a ``publishedAt``
```
tromai, line 60: Does `deps.dev` have any reference about the format of this `version` field?. It would be good to include that here if that page is available.

src/macaron/slsa_analyzer/package_registry/package_registry.py
```diff
@@ -40,3 +49,77 @@ def is_detected(self, build_tool: BaseBuildTool) -> bool:
...
+            cannot be parsed correctly, or if the expected timestamp is missing or invalid.
+        NotImplementedError
+            If not implemented for a registry.
+        """
+        # TODO: To reduce redundant calls to deps.dev, store relevant parts of the response
+        # in the AnalyzeContext object retrieved by the Repo Finder. This step should be
+        # implemented at the beginning of the analyze command to ensure that the data
+        # is available for subsequent processing.
+
+        base_url_parsed = urllib.parse.urlparse(registry_url or "https://api.deps.dev")
```
tromai, line 90: If the default value of `registry_url` is `https://api.devs.dev`, could we make the type of this parameters `registry_url: str = "https://api.devs.dev"`?

src/macaron/slsa_analyzer/package_registry/package_registry.py
```diff
@@ -40,3 +49,77 @@ def is_detected(self, build_tool: BaseBuildTool) -> bool:
...
+        path_params = "/".join(["v3alpha", "purl", encode(purl).replace("/", "%2F")])
```
tromai, line 91: The `encode` function here allows you to specify the set of safe characters that will not be encoded (see [here](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote)). If we replace all `/` anyway, should we leverage this `safe` parameter?

src/macaron/slsa_analyzer/package_registry/package_registry.py
```diff
@@ -40,3 +49,77 @@ def is_detected(self, build_tool: BaseBuildTool) -> bool:
...
+                raise InvalidHTTPResponseError(f"Empty response returned by {url} .")
+
+            timestamp = json_extract(metadata, ["version", "publishedAt"], str)
+            if not timestamp:
+                raise InvalidHTTPResponseError(f"The timestamp is missing in the response returned by {url}.")
+
+            logger.debug("Found timestamp: %s.", timestamp)
+
+            try:
+                return datetime.fromisoformat(timestamp.replace("Z", "+00:00"))
```
tromai, line 121: If the timestamp value is already in ISO 8601 format, why do we need to perform extra modification on it before providing it to `datetime.fromisoformat`.

src/macaron/slsa_analyzer/package_registry/package_registry.py
```diff
@@ -40,3 +49,77 @@ def is_detected(self, build_tool: BaseBuildTool) -> bool:
...
+            except (OverflowError, OSError) as error:
```
tromai, line 122: I wonder in what scenario `OverflowError` and `OSError` happens. 
As far as I know, `datetime.fromisoformat` raises:
- `TypeError` if the input is not a string.
- `ValueError` if the date is in correct.

src/macaron/slsa_analyzer/package_registry/package_registry.py
```diff
@@ -40,3 +49,77 @@ def is_detected(self, build_tool: BaseBuildTool) -> bool:
...
+        datetime
+            A timezone-aware datetime object representing the publication timestamp
+            of the specified package.
+
+        Raises
+        ------
+        InvalidHTTPResponseError
+            If the URL construction fails, the HTTP response is invalid, or if the response
+            cannot be parsed correctly, or if the expected timestamp is missing or invalid.
+        NotImplementedError
```
tromai, line 82: nit: Should we document this exception if it's not raise in this implementation ? 🤔

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
tromai, line 1: I see that on line 249, we don't use the utility json extract method. We might need to consider replacing it. It doesn't need to be in this PR.

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
```diff
@@ -152,41 +182,45 @@ def is_detected(self, build_tool: BaseBuildTool) -> bool:
...
+            cannot be parsed correctly, or if the expected timestamp is missing or invalid.
         """
-        query_params = [f"q=g:{group_id}", f"a:{artifact_id}"]
-        if version:
-            query_params.append(f"v:{version}")
+        try:
+            purl_object = PackageURL.from_string(purl)
+        except ValueError as error:
+            logger.debug("Could not parse PURL: %s", error)
+        query_params = [f"q=g:{purl_object.namespace}", f"a:{purl_object.name}", f"v:{purl_object.version}"]
```
tromai, line 217: `purl_object.version` has type `str | None`. If it's none the last query param would be `'v:None'`. I don't think it has the same behavior as the old implementation. In the old implementation, if version is None, no extra query param is added. I wonder if this new behavior is intended?

src/macaron/slsa_analyzer/specs/ci_spec.py
```diff
@@ -36,3 +37,6 @@ class CIInfo(TypedDict):
 
     provenances: Sequence[DownloadedProvenanceData]
     """The provenances data."""
+
+    build_info_results: InTotoV01Payload
+    """The build information results computed for a build step. We use the in-toto 0.1 as the spec."""
```
tromai, line 42: It's mentioned in the PR description that `CIInfo["provenances"] is also renamed to CIInfo["build_info_results"]`. I think we should remove this `provenances` attribute here too

src/macaron/slsa_analyzer/checks/build_as_code_check.py
```diff
@@ -150,27 +162,30 @@ def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
...
                                 job = callee.caller
-                                if isinstance(job, GitHubJobNode):
-                                    job_id = job.parsed_obj.id
-                                    caller_path = job.source_path
 
+                                # We always expect the caller of the node that calls a third-party
+                                # or Reusable GitHub Action to be be a GitHubJobNode.
+                                if not isinstance(job, GitHubJobNode):
+                                    continue
+
```
tromai, line 175:
```suggestion
                                # or Reusable GitHub Action to be a GitHubJobNode.
                                if not isinstance(job, GitHubJobNode):
                                    continue

```

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -355,3 +358,354 @@ def check_if_repository_purl_and_url_match(url: str, repo_purl: PackageURL) -> b
...
+        str | None
+            The build type if found; otherwise, None.
+        """
+        if statement["predicate"] is None:
+            return None
+
+        if build_type := json_extract(statement["predicate"], ["buildType"], str):
+            return build_type
+
+        return json_extract(statement["predicate"], ["buildDefinition", "buildType"], str)
```
tromai, line 672: It might be good to document why we have two different way to extract the build type
- `json_extract(statement["predicate"], ["buildType"], str)`
- `json_extract(statement["predicate"], ["buildDefinition", "buildType"], str)`

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -355,3 +358,354 @@ def check_if_repository_purl_and_url_match(url: str, repo_purl: PackageURL) -> b
...
+            SLSAGCBBuildDefinitionV1(),
+            SLSAOCIBuildDefinitionV1(),
+            WitnessGitLabBuildDefinitionV01(),
+        ]
+
+        for build_def in build_defs:
+            if build_def.expected_build_type == build_type:
+                return build_def
+
+        raise ProvenanceError("Unable to find build definition in the provenance statement.")
```
tromai, line 711: A minor improvement could be that we differentiate between 2 cases that we raise an exception:
- There is no build definition, happens when `ProvenancePredicate.get_build_type(statement)` returns None.
- There is a build definition, but we don't have support for its value (the condition `build_def.expected_build_type == build_type` in the for loop never meet).

Right now we only raise one kind of exception message for both cases.

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -355,3 +358,354 @@ def check_if_repository_purl_and_url_match(url: str, repo_purl: PackageURL) -> b
         purl_path = f"{repo_purl.namespace}/{purl_path}"
     # Note that the urllib method includes the "/" before path while the PURL method does not.
     return f"{parsed_url.hostname}{parsed_url.path}".lower() == f"{expanded_purl_type or repo_purl.type}/{purl_path}"
+
+
+class ProvenanceBuildDefinition(ABC):
```
tromai, line 363: I wonder if these or some of these abstractions should be put within the package `src/macaron/slsa_analyzer/provenance` as they are closely related to the provenance format 🤔 ?

src/macaron/slsa_analyzer/checks/infer_artifact_pipeline_check.py
```diff
@@ -93,35 +116,55 @@ def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
...
-                        logger.debug(error)
+        for registry_info in ctx.dynamic_data["package_registries"]:
+            if registry_info.build_tool.purl_type == ctx.component.type:
+                try:
+                    artifact_published_date = registry_info.package_registry.find_publish_timestamp(ctx.component.purl)
+                    break
+                except InvalidHTTPResponseError as error:
+                    logger.debug(error)
+                except NotImplementedError:
+                    continue
```
tromai, line 133: I think catching `NotImplementedError` here has a strong indication that the interface of `find_publish_timestamp` from subtype of `PackageRegistry` will raise `NotImplementedError`. I believe this is not the case because only Jfrog Maven registry will raise `NotImplementedError`. A better way to communicate this would be to explicitly skip running `find_publish_timestamp` if `registry_info` is of type `JFrogMavenRegistry`.
```
        for registry_info in ctx.dynamic_data["package_registries"]:
            if isinstance(registry_info.package_registry, JfrogMavenRegistry):
                # We currently don't support this.
                continue 
            if registry_info.build_tool.purl_type == ctx.component.type:
                try:
                    artifact_published_date = registry_info.package_registry.find_publish_timestamp(ctx.component.purl)
                    break
                except InvalidHTTPResponseError as error:
                    logger.debug(error)
```
What do you think?

src/macaron/slsa_analyzer/checks/infer_artifact_pipeline_check.py
```diff
@@ -93,35 +116,55 @@ def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
...
-        # This check requires the artifact publish artifact to proceed. If the timestamp is not
-        # found, we return with a fail result.
         if not artifact_published_date:
+            logger.debug("Unable to find a publish date for the artifact.")
             return CheckResultData(result_tables=[], result_type=CheckResultType.FAILED)
 
+        # If an artifact is published before the corresponding code is committed, there cannot be
+        # a CI pipeline that triggered the publishing.
+        if published_before_commit := artifact_published_date < commit_date:
+            logger.debug("Publish date %s is earlier than commit date %s.", artifact_published_date, commit_date)
```
tromai, line 150: Should we return check status FAIL here ?

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -355,3 +358,354 @@ def check_if_repository_purl_and_url_match(url: str, repo_purl: PackageURL) -> b
...
+    """
+
+    #: Determines the expected ``buildType`` field in the provenance predicate.
+    expected_build_type = "https://github.com/slsa-framework/slsa-github-generator/generic@v1"
+
+    def get_build_invocation(self, statement: InTotoV01Statement | InTotoV1Statement) -> tuple[str | None, str | None]:
+        """Retrieve the build invocation information from the given statement.
+
+        This method is intended to be implemented by subclasses to extract
+        specific invocation details from a provenance statement.
```
tromai, line 416: Should we remove these 2 lines as this is a concrete method 🤔 ?
The same comment applies for `SLSAGithubActionsBuildDefinitionV1.get_build_invocation`

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -355,3 +358,354 @@ def check_if_repository_purl_and_url_match(url: str, repo_purl: PackageURL) -> b
...
+            return None, None
+        gha_workflow = json_extract(statement["predicate"], ["invocation", "configSource", "entryPoint"], str)
+        gh_run_id = json_extract(statement["predicate"], ["invocation", "environment", "github_run_id"], str)
+        repo_uri = json_extract(statement["predicate"], ["invocation", "configSource", "uri"], str)
+        repo = None
+        if repo_uri:
+            repo = _clean_spdx(repo_uri)
+        if repo is None:
+            return gha_workflow, repo
+        invocation_url = f"{repo}/" f"actions/runs/{gh_run_id}"
```
tromai, line 442: I wonder if it's intended to concatenate 2 formatted string? 
nit: we could use one single formatted string `f"{repo}/actions/runs/{gh_run_id}"`

src/macaron/repo_finder/provenance_extractor.py
```diff
@@ -355,3 +358,354 @@ def check_if_repository_purl_and_url_match(url: str, repo_purl: PackageURL) -> b
...
+            The provenance statement from which to extract the build invocation
+            details. This statement contains the metadata about the build process
+            and its associated artifacts.
+
+        Returns
+        -------
+        tuple[str | None, str | None]
+            A tuple containing two elements:
+            - The first element is the build invocation entry point (e.g., workflow name), or None if not found.
+            - The second element is the invocation URL or identifier (e.g., job URL), or None if not found.
```
tromai, line 474: I'm not clear about the different between `invocation URL` and `identifier`. Does it mean that if the second element of the returned tuple is not None, it could be an URL or an arbitrary string as the "identifier" of the workflow run?

src/macaron/slsa_analyzer/ci_service/github_actions/github_actions_ci.py
```diff
@@ -357,6 +427,35 @@ def workflow_run_in_date_time_range(
...
+        # we are using for GitHub Actions is in ISO format, which contains the offset
+        # from the UTC timezone. For example: 2022-04-10T14:10:01+07:00
+        # GitHub retains GitHub Actions pipeline data for 400 days. So, we cannot analyze the
+        # pipelines if artifacts are older than 400 days.
+        # https://docs.github.com/en/rest/guides/using-the-rest-api-to-interact-with-checks?
+        # apiVersion=2022-11-28#retention-of-checks-data
+        # TODO: change this check if this issue is resolved:
+        # https://github.com/orgs/community/discussions/138249
+        if datetime.now(timezone.utc) - timedelta(days=400) > timestamp:
+            logger.debug("Artifact published at %s is older than 410 days.", timestamp)
```
tromai, line 454:
```suggestion
            logger.debug("Artifact published at %s is older than 400 days.", timestamp)
```

tromai (approved):
I have finished my review. Thanks. 
Overall, there isn't any major changes needs. Most of my comments are for minor improvements/nit picking.

src/macaron/config/defaults.ini
behnazh-w: No, build and deploy GitHub Actions are obtained differently and from `builder.maven.ci` in `defaults.ini`.

src/macaron/config/defaults.ini
behnazh-w: Based on the artifacts I have experimented with, `publish_time_range` might need larger values, but this value can be adjusted by users as required.

src/macaron/slsa_analyzer/package_registry/package_registry.py
behnazh-w: This question doesn't seem relevant because versions don't have format in this context.

src/macaron/slsa_analyzer/package_registry/package_registry.py
behnazh-w: The `registry_url` is optional to allow unit testing of this function.

src/macaron/slsa_analyzer/package_registry/package_registry.py
behnazh-w: I have tried to be consistent with how repo finder constructs the request URL from a PURL. @benmss Was there any particular reason for not using `safe=""`?

benmss: I think we can safely leverage the `safe` parameter here, and for the repo finder too.

src/macaron/slsa_analyzer/package_registry/package_registry.py
behnazh-w: That's right. I don't remember why I had decided to replace `"Z"` with `"+00:00"`.

src/macaron/slsa_analyzer/package_registry/package_registry.py
behnazh-w: >[fromtimestamp()](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp) may raise [OverflowError](https://docs.python.org/3/library/exceptions.html#OverflowError), if the timestamp is out of the range of values supported by the platform C localtime() or gmtime() functions, and [OSError](https://docs.python.org/3/library/exceptions.html#OSError) on localtime() or gmtime() failure.

src/macaron/slsa_analyzer/package_registry/package_registry.py
behnazh-w: I decided to add this error in the parent class because it communicates that this method in a subclass might not be implemented.

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
behnazh-w: >On line 249, we don't use the utility json extract method. We might need to consider replacing it.

I made this decision on purpose to check for each expected field and log helpful error messages when any of the fields in the response is missing.

>On line 233, we can consider updating the if conditions similar to this discussion - https://github.com/oracle/macaron/pull/872#discussion_r1828709252
Thanks for spotting this, I'll fix it.

src/macaron/slsa_analyzer/specs/ci_spec.py
behnazh-w: My PR description was not accurate and I fixed it: Instead of using ``CIInfo["provenances"]`` for inferred build command analysis results, we use a new field: ``CIInfo["build_info_results"]``. `CIInfo["provenances"]` is still used for storing provenances.

src/macaron/repo_finder/provenance_extractor.py
behnazh-w: I leave further refactoring to @benmss based on the changes he is making for provenance processing and verification in other PRs.

src/macaron/slsa_analyzer/checks/infer_artifact_pipeline_check.py
behnazh-w: I prefer to catch `NotImplementedError` directly, rather than checking for specific subclass names. This approach is more general and avoids unnecessary dependencies on subclassing details.

src/macaron/repo_finder/provenance_extractor.py
behnazh-w: It just means that an identifier will be returned, which can be a URL.

src/macaron/slsa_analyzer/ci_service/github_actions/github_actions_ci.py
behnazh-w: See commit [17f3453](https://github.com/oracle/macaron/pull/872/commits/17f3453c4be75784f00a422b71ea829e3f04e4de).

src/macaron/repo_finder/provenance_extractor.py
behnazh-w: See commit [17f3453](https://github.com/oracle/macaron/pull/872/commits/17f3453c4be75784f00a422b71ea829e3f04e4de).

src/macaron/repo_finder/provenance_extractor.py
behnazh-w: See commit [17f3453](https://github.com/oracle/macaron/pull/872/commits/17f3453c4be75784f00a422b71ea829e3f04e4de).

src/macaron/repo_finder/provenance_extractor.py
behnazh-w: See commit [17f3453](https://github.com/oracle/macaron/pull/872/commits/17f3453c4be75784f00a422b71ea829e3f04e4de).

src/macaron/slsa_analyzer/checks/build_as_code_check.py
behnazh-w: See commit [17f3453](https://github.com/oracle/macaron/pull/872/commits/17f3453c4be75784f00a422b71ea829e3f04e4de).

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
behnazh-w: See commit [17f3453](https://github.com/oracle/macaron/pull/872/commits/17f3453c4be75784f00a422b71ea829e3f04e4de).

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
behnazh-w: See commit [17f3453](https://github.com/oracle/macaron/pull/872/commits/17f3453c4be75784f00a422b71ea829e3f04e4de).

src/macaron/slsa_analyzer/package_registry/package_registry.py
behnazh-w: See commit [17f3453](https://github.com/oracle/macaron/pull/872/commits/17f3453c4be75784f00a422b71ea829e3f04e4de).

src/macaron/slsa_analyzer/package_registry/package_registry.py
behnazh-w: See commit [17f3453](https://github.com/oracle/macaron/pull/872/commits/17f3453c4be75784f00a422b71ea829e3f04e4de).

src/macaron/slsa_analyzer/package_registry/package_registry.py
behnazh-w: See commit [17f3453](https://github.com/oracle/macaron/pull/872/commits/17f3453c4be75784f00a422b71ea829e3f04e4de).

src/macaron/repo_finder/provenance_extractor.py
tromai: Just for confirmation: `json_extract(statement["predicate"], ["buildDefinition", "buildType"], str)` is for `InTotoV1Statement` and `json_extract(statement["predicate"], ["buildType"], str)` is for `InTotoV01Statement` right?

tromai:
Except from a small question in https://github.com/oracle/macaron/pull/872#discussion_r1844910201, my approval still holds.

src/macaron/repo_finder/provenance_extractor.py
behnazh-w: Yes in this case, but a provenance, e.g., witness with a particular version of statement, e.g., `InTotoV01Statement` might contain `buildType` in a different access path. So, the statement version does not necessarily determine the location of `buildType` .

tromai (approved):

PR #873, opened 2024-09-29
feat: verify whether the reported repository can be linked back to the artifact

mabdollahpour-ol:
This version has initial support for maven and gradle build tools. 

The core part is added under `repo_verifier` directory. `analyzer` calls the `repo_verifier` and adds the info to `dynamic_data`. 

Also added a sample check (for maven) that shows how this data can be used.


mabdollahpour-ol:
The core part is added as "repo_verifier" under "repo_finder". 
"analyzer" calls the "repo_verifier" and adds the info to "dynamic_data".
I added a sample check (for maven) that shows how this data can be used.

src/macaron/repo_finder/repo_finder_java.py
```diff
@@ -301,3 +305,46 @@ def _resolve_properties(self, pom: Element, values: list[str]) -> list[str]:
...
+
+        info:
+          - dependentCount (how many other artifacts depend on this)
+          - repo_link (link to the repository)
+          - star_count (number of stars on the repository)
+          - fork_count (number of forks on the repository)
+        """
+        dd_info: dict[str, Any] = {}
+
+        dep_dev_url = f"https://deps.dev/_/s/maven/p/{group}%3A{artifact}/v/{version}"
```
behnazh-w, line 321: Please use `urllib.parse.ParseResult` to create the URL.

```python
urllib.parse.ParseResult(
            scheme="https",
            netloc="deps.dev"
            path="/".join(["_", "s", "maven", "p", f"{group}:{artifact}", "v", version]),
            params="",
            query="",
            fragment="",
        ).geturl()
```

src/macaron/repo_finder/repo_finder_java.py
```diff
@@ -301,3 +305,46 @@ def _resolve_properties(self, pom: Element, values: list[str]) -> list[str]:
...
+        session = requests.Session()
```
behnazh-w, line 322: Please use our internal [send_get_http](https://github.com/oracle/macaron/blob/edfe06eb3f9337f86215b253f40f53ca1a0105c9/src/macaron/util.py#L21) function instead. This central function ensures consistent error handling across our codebase.

behnazh-w:
> The core part is added as "repo_verifier" under "repo_finder". "analyzer" calls the "repo_verifier" and adds the info to "dynamic_data". I added a sample check (for maven) that shows how this data can be used.

Can you please add this information to the PR description?

mabdollahpour-ol:
@behnazh-w thanks for the comments! I'll apply the changes by EOD.

src/macaron/repo_finder/repo_finder_java.py
behnazh-w: @mabdollahpour-ol It's actually better to use [send_get_http_raw](https://github.com/oracle/macaron/blob/edfe06eb3f9337f86215b253f40f53ca1a0105c9/src/macaron/util.py#L56)  function, which has a few more options and can handle rate limits.

src/macaron/repo_finder/repo_verifier.py
```diff
@@ -0,0 +1,300 @@
...
+            if (current_dir / filename).exists():
+                return current_dir / filename
+
+            # ignore symlinks to prevent potential infinite loop
+            sub_dirs = [Path(it) for it in current_dir.iterdir() if it.is_dir() and not it.is_symlink()]
+            queue.extend(sub_dirs)
+
+        return None
+
+    def _verify_maven_git_ns(self) -> RepositoryVerificationResult:
```
behnazh-w, line 149: Wouldn't it make sense to add the Maven related functions to the [Maven build tool](https://github.com/oracle/macaron/blob/edfe06eb3f9337f86215b253f40f53ca1a0105c9/src/macaron/slsa_analyzer/build_tool/maven.py#L23) class and do the same later for the other build tools?

src/macaron/repo_finder/repo_verifier.py
```diff
@@ -0,0 +1,300 @@
...
+            current_dir = queue.popleft()
+
+            # don't look through non-main directories
+            if any(
+                keyword in current_dir.name.lower()
+                for keyword in ["test", "example", "sample", "doc", "demo", "spec", "mock"]
+            ):
+                continue
+
+            if (current_dir / filename).exists():
```
behnazh-w, line 140: We use `Path` to construct paths for better clarity and explicitness.

```suggestion
            if Path(current_dir, filename).exists():
```

src/macaron/repo_finder/repo_verifier.py
```diff
@@ -0,0 +1,300 @@
...
+            return False
+
+        for i in range(2):
+            if g1_parts[i] != g2_parts[i]:
+                return False
+
+        return True
+
+    @staticmethod
+    def _bfs_walk(root_dir: Path, filename: str) -> Path | None:
```
behnazh-w, line 124: You can probably use the [`get_build_dirs`](https://github.com/oracle/macaron/blob/edfe06eb3f9337f86215b253f40f53ca1a0105c9/src/macaron/slsa_analyzer/build_tool/base_build_tool.py#L190) function in build tools. You might just need to improve it to avoid looking into the `["test", "example", "sample", "doc", "demo", "spec", "mock"]` directories as implemented in this function.

src/macaron/repo_finder/repo_verifier.py
mabdollahpour-ol: Yeah actually I was thinking to have separate `RepoVerifier` classes for each build tool (probably all implementing a single abc or interface). But I thought that could be done when we actually have some implementation for other build tools. I can add that implementation now and see. Wdyt? 

However, I am not sure if it's a good idea to put repository verification logic directly inside a build tool class. I mean one would think it's too much of a (somewhat not related) responsibility for that class.

src/macaron/repo_finder/repo_verifier.py
mabdollahpour-ol: current_dir is actually of type `pathlib.Path`. `/` is a handy operator implemented by `Path` classes. but sure, I could change that to preserve consistency.

src/macaron/repo_finder/repo_verifier.py
mabdollahpour-ol: Interesting. Yes, I that's the correct funciton to be used here. Thanks!

src/macaron/repo_finder/repo_finder_java.py
benmss, line 321: I'm curious how this URL was formed. It does not match the typical API or browser usage.

However, I expect using the API is still preferable. The only issue there is that the information required above is only fetchable for repository PURLs: GitHub, etc. We currently use [v3alpha](https://docs.deps.dev/api/v3alpha/#getproject).

src/macaron/repo_finder/repo_verifier.py
```diff
@@ -0,0 +1,300 @@
...
+    """A class to verify a claimed repo."""
+
+    _known_maven_namespaces = {
+        "github",
+        "gitlab",
+        "bitbucket",
+        "gitee",
+    }
+
+    def __init__(
```
benmss, line 52: Please add doc comments for all functions, even when it seems obvious. For 'private' functions a single line is sufficient.

src/macaron/repo_finder/repo_finder_java.py
mabdollahpour-ol: It's not part of their official API but that's how they populate the project info (star and fork count) at the client side.

src/macaron/repo_finder/repo_verifier.py
```diff
@@ -0,0 +1,300 @@
...
+
+    def verify_claimed_repo(self) -> RepositoryVerificationResult:
+        """Verify whether the claimed repository links back to the artifact."""
+        default_res = RepositoryVerificationResult(
+            status=RepositoryVerificationStatus.UNKNOWN, reason="unsupported_type", build_tool=self.build_tool
+        )
+
+        match self.build_tool:
+            case Maven():
+                git_ns_ver = self._verify_maven_git_ns()
```
benmss, line 76: I would prefer it if variable names were not shortened to this degree. The `ns` part is probably acceptable given `fs` is used here and elsewhere, but `ver` should be the full word.

benmss: This also applies to other variables here, including `default_res`, `g1`, `g2`, etc.

src/macaron/repo_finder/repo_verifier.py
```diff
@@ -0,0 +1,300 @@
...
+            return False
+
+        if (g1_parts[0] in {"io", "com"} and g1_parts[1] in self._known_maven_namespaces) or (
+            g2_parts[0] in {"io", "com"} and g2_parts[1] in self._known_maven_namespaces
+        ):
+            if len(g1_parts) >= 3 and len(g2_parts) >= 3:
+                return all(g1_parts[i] == g2_parts[i] for i in range(3))
+            return False
+
+        for i in range(2):
```
benmss, line 117: Same variable length comment applies here. Please use `index` over `i`.

src/macaron/repo_finder/repo_verifier.py
```diff
@@ -0,0 +1,300 @@
...
+    def _bfs_walk(root_dir: Path, filename: str) -> Path | None:
+        if not os.path.exists(root_dir) or not os.path.isdir(root_dir):
+            return None
+
+        queue: deque[Path] = deque()
+        queue.append(Path(root_dir))
+        while queue:
+            current_dir = queue.popleft()
+
+            # don't look through non-main directories
```
benmss, line 133: Please start all comments with a capital letter (where applicable), and end them with a period.

src/macaron/repo_finder/repo_verifier.py
```diff
@@ -0,0 +1,300 @@
...
+
+            # ignore symlinks to prevent potential infinite loop
+            sub_dirs = [Path(it) for it in current_dir.iterdir() if it.is_dir() and not it.is_symlink()]
+            queue.extend(sub_dirs)
+
+        return None
+
+    def _verify_maven_git_ns(self) -> RepositoryVerificationResult:
+        parsed_url = urlparse(self.claimed_repo_url)
+        if parsed_url is None or parsed_url.hostname is None:
```
benmss, line 151:
```suggestion
        if not (parsed_url or parsed_url.hostname):
```
Applies to other uses of `is None` also, as that should be reserved for cases where it is strictly required.

src/macaron/repo_finder/repo_verifier.py
```diff
@@ -0,0 +1,300 @@
...
+        claimed_account = parsed_url.path.strip("/").split("/")[0]
+
+        group_parts = self.namespace.split(".")
+        for platform in self._known_maven_namespaces:
+            if (
+                group_parts[0].lower() in {"io", "com"}
+                and group_parts[1].lower() == platform.lower()
+                and group_parts[1].lower() == claimed_hostname.lower()
+                and group_parts[2].lower() == claimed_account.lower()
+            ):
```
benmss, line 167: What is happening here is not immediately clear to me. An example in a comment would be nice.

src/macaron/repo_finder/repo_verifier.py
```diff
@@ -0,0 +1,300 @@
...
+        pom_root = parse_pom_string(pom_content)
+
+        if not pom_root:
+            logger.debug("Could not parse pom.xml: %s", pom_file.as_posix())
+            return RepositoryVerificationResult(
+                status=RepositoryVerificationStatus.UNKNOWN, reason="not_parsed_pom", build_tool=self.build_tool
+            )
+
+        # find the group id in the pom (project/groupId)
+        pom_group_id_elem = next((ch for ch in pom_root if ch.tag.endswith("}groupId")), None)
```
benmss, line 199: Adding a comment as to why the closing curly brace is needed before the tag name would be good to have.

mabdollahpour-ol: Sure. That's the closing tag from the XML namespace. Although I haven't seen any `.+groupId` tag name, it just makes sure that it is what we actually want.

src/macaron/repo_finder/repo_verifier.py
```diff
@@ -0,0 +1,300 @@
...
+                status=RepositoryVerificationStatus.FAILED, reason="group_id_mismatch", build_tool=self.build_tool
+            )
+
+        return RepositoryVerificationResult(
+            status=RepositoryVerificationStatus.PASSED, reason="group_id_match", build_tool=self.build_tool
+        )
+
+    @staticmethod
+    def _is_valid_maven_group_id(group_id: str) -> bool:
+        pattern = r"^[a-zA-Z][a-zA-Z0-9-]*\.([a-zA-Z][a-zA-Z0-9-]*\.)*[a-zA-Z][a-zA-Z0-9-]*[a-zA-Z0-9]$"
```
benmss, line 219: A brief comment on what this pattern should match would be nice, including an example.

src/macaron/repo_finder/repo_verifier.py
```diff
@@ -0,0 +1,300 @@
...
+        return re.match(pattern, group_id) is not None
+
+    def _verify_gradle(self) -> RepositoryVerificationResult:
+        # check if repo contains gradle.properties
+        def _extract_group_id_from_properties() -> str | None:
```
benmss, line 224: This and the following two functions are very similar. I would suggest separating out the shared the code and reusing it.

src/macaron/repo_finder/repo_verifier.py
```diff
@@ -0,0 +1,300 @@
...
+                logger.debug("Could not find gradle.properties in the repository: %s", self.claimed_repo_url)
+                return None
+
+            properties_lines = gradle_properties.read_text().splitlines()
+            for line in properties_lines:
+                line_parts = list(filter(None, map(str.strip, line.strip().lower().split("="))))
+                if len(line_parts) != 2:
+                    continue
+                if line_parts[0] == "group":
+                    return line_parts[1]
```
benmss, line 237: The validity check isn't needed here?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -1134,6 +1137,33 @@ def _determine_package_registries(self, analyze_ctx: AnalyzeContext) -> None:
                         )
                     )
 
+    def _verify_repository_link(self, parsed_purl: PackageURL, analyze_ctx: AnalyzeContext) -> None:
+        """Verify whether the claimed repository links back to the artifact."""
+        if not analyze_ctx.component.repository:
+            logger.debug("The repository is not available. Skipping the repository verification.")
+            return
+
+        if parsed_purl.namespace is None or parsed_purl.name is None or parsed_purl.version is None:
```
benmss, line 1146: The `name` property of a `PackageURL` is required so you do not need to check if it is `None`

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
```diff
@@ -0,0 +1,107 @@
...
+    dd_dep_count: Mapped[int | None] = mapped_column(Integer, nullable=True)
+    dd_repo_link: Mapped[str | None] = mapped_column(String, nullable=True)
+
+    __mapper_args__ = {
+        "polymorphic_identity": "_maven_repo_verification_check",
+    }
+
+
+class MavenRepoVerificationCheck(BaseCheck):
+    """This check verifies that if a source repository claimed by a maven package is valid."""
```
benmss, line 48:
```suggestion
    """This check verifies whether the reported source repository of a maven package is valid."""
```
Furthermore, I feel that the others instances of `claimed` here are not grammatically correct. E.g. `claimed repository`, `repository claimed`, etc. Although the meaning is clear enough as is, I would advocate for using `reported` instead.

Pinging @nicallen for possible additional thoughts on this.

benmss:
Is there a plan to add unit testing and integration testing as part of this PR? @behnazh-w

behnazh-w:
> Is there a plan to add unit testing and integration testing as part of this PR? @behnazh-w

Yes for sure. The PR needs unit tests and integration tests.

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
nicallen: The phrasing of "claimed repository", "repository claimed by a maven package" etc. does sound a bit awkward, probably because of ambiguity with the "demand or take by right" definition of claim (that would usually use that form) rather than the "state or assert to be true" definition that is meant here.
I think the distinction of a "claim" vs "report" could be usefully meaningful, though. To me, calling the statement that a maven package comes from a particular repo a "claim" raises the question of who is making the claim (in this case, the maven package itself) and whether they can be trusted or have would have reason to lie (which is relevant in the context of potentially malicious packages), whereas "report" seems more neutral, like it could be inaccurate but we would not be presuming malice.
I'd also take issue with the use of "valid", which is unclear exactly as to what is meant.
Maybe the purpose of this check could be more precisely stated as "to determine whether the claims of source repository provenance that are made by a maven package can be independently verified (or at least, independently corroborated)".

src/macaron/repo_finder/repo_verifier.py
mabdollahpour-ol: Good catch! I'll merge them as you suggested and put the validation in the shared part

mabdollahpour-ol:
> > Is there a plan to add unit testing and integration testing as part of this PR? @behnazh-w
> 
> Yes for sure. The PR needs unit tests and integration tests.

Yes tests are on the way.

src/macaron/repo_finder/repo_verifier.py
```diff
@@ -0,0 +1,300 @@
...
+            case NPM():
+                return default_res
+            case Yarn():
+                return default_res
+            case Go():
+                return default_res
+            case _:
+                raise NotImplementedError(f"Unsupported build tool: {self.build_tool}")
+
+    def _same_group(self, g1: str, g2: str) -> bool:
```
mabdollahpour-ol, line 101: @benmss @behnazh-w I'm thinking to move this and `is_valid_maven_group_id` to `build_tool.Maven` class as public static methods. What do you think?

behnazh-w: > @benmss @behnazh-w I'm thinking to move this and `is_valid_maven_group_id` to `build_tool.Maven` class as public static methods. What do you think?

Sounds good to me. Although as a general note, there is no real public and private function in Python as they can be called from anywhere with no restriction. I try to avoid "private" functions as much as possible. Also our configurations doesn't mandate doctring for private functions that start with `_`. So there is a chance of missing lots of documentation in using these private functions.

mabdollahpour-ol: Or should they be in `MavenCentralRegistry` ?

behnazh-w: We have three related abstractions to consider:

- [Maven artifact](https://github.com/oracle/macaron/blob/staging/src/macaron/artifact/maven.py)
- Maven related package registries: [Maven Central Registry](https://github.com/oracle/macaron/blob/staging/src/macaron/slsa_analyzer/package_registry/maven_central_registry.py) and [JFrog](https://github.com/oracle/macaron/blob/staging/src/macaron/slsa_analyzer/package_registry/jfrog_maven_registry.py)
- [Maven build tool](https://github.com/oracle/macaron/blob/staging/src/macaron/slsa_analyzer/build_tool/maven.py)

Thinking some more, as a static method, the appropriate abstraction may be to use the Maven artifact class :thinking:

mabdollahpour-ol: You mean putting a module level function in `artifact.maven`?

mabdollahpour-ol: The thing is that, `same_group` function relies on certain recognized code services which only makes sense in the context of maven central (a registry)

behnazh-w: > You mean putting them as module level functions in `artifact.maven`?

Yes, since it doesn't depend on the state of any object we don't need to add it to a class and `artifact.maven` could be a suitable module for it.

mabdollahpour-ol: Sure. I don't have any strong preference for them to be static methods or module level functions. But they could also be in the same module that defines maven central.

mabdollahpour-ol: But for `is_valid_maven_group_id` I agree that artifact is a better place

behnazh-w: Agreed. Because the logic of this function depends on the way a repository handles group IDs and domains, it can be added to the Maven Central registry abstraction.

src/macaron/repo_finder/repo_verifier.py
mabdollahpour-ol: > Applies to other uses of `is None` also, as that should be reserved for cases where it is strictly required.

Mypy is not happy. It complains that None doesn't have `split` for `parsed_url.hostname.split`

mabdollahpour-ol: Strangely, `parsed_url is None or not parsed_url.hostname` works. Not sure how the type inference works there.

src/macaron/repo_finder/repo_verifier.py
mabdollahpour-ol: Everything should have comment now. Let me know if I missed anything

src/macaron/repo_finder/repo_finder_java.py
mabdollahpour-ol: Added a method to `DepsDepRepoFinder` to get project info from the official API.

src/macaron/repo_finder/repo_verifier.py
mabdollahpour-ol: I decided to keep this for now (with another name `find_file_in_repo` under `repo_verifier_base`) because there amount of refactoring needed to make use of those functions for this purpose seems to be quite high.

src/macaron/repo_finder/repo_verifier.py
mabdollahpour-ol: Now each repo verifier has its own class. Let me know what you think

src/macaron/repo_finder/repo_verifier.py
behnazh-w: OK, no worries. Can you please add a TODO comment to refactor and use [get_build_dirs](https://github.com/oracle/macaron/blob/edfe06eb3f9337f86215b253f40f53ca1a0105c9/src/macaron/slsa_analyzer/build_tool/base_build_tool.py#L190) later?

src/macaron/artifact/maven.py
```diff
@@ -140,3 +140,10 @@ def create_maven_purl_from_artifact_filename(
             )
 
     return None
+
+
+def is_valid_maven_group_id(group_id: str) -> bool:
+    """Check if the provided string is a valid maven group id."""
```
behnazh-w, line 146: Please add `Parameters` and `Returns` sections to the docstrings. Please have a look at the other docstrings in the codebase.

tromai: For more information on our doc strings guide, you can have a look at: https://github.com/oracle/macaron/blob/main/docs/source/pages/developers_guide/style_guide.rst#docstrings

behnazh-w: I had forgotten about the guide, thanks :slightly_smiling_face:

src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -71,6 +73,41 @@ def find_repo(self, purl: PackageURL) -> str:
...
+        clean_repo_url = clean_url(project_url)
+        if clean_repo_url is None or clean_repo_url.hostname is None:
+            logger.debug("Invalid project url format: %s", project_url)
+            return None
+
+        project_key = clean_repo_url.hostname + clean_repo_url.path
+
+        request_url = f"https://api.deps.dev/v3alpha/projects/{encode(project_key, safe='')}"
+        response = send_get_http_raw(request_url)
+        if response is None or response.status_code != 200 or not response.text:
```
benmss, line 99: With the current implementation of `send_get_http_raw` if a response object is returned (i.e. not `None`), it will have a status code of 200. So, no need to explicitly check that here. See the comment of that function for more information.

src/macaron/repo_verifier/repo_verifier_base.py
```diff
@@ -0,0 +1,139 @@
...
+        The path to the file if it exists, otherwise
+    """
+    # TODO: Consider using BaseBuildTool.get_build_dirs.
+    #   + Refactor 'get_build_dirs' to skip certain directories
+    #   that are most likely not part of the main codebase (e.g., sample).
+    #   + Need to find a way to look for other
+    #   files (e.g., gradle.properties) for the purpose of repo verification
+    #   without breaking the current logic of finding build directories.
+    #   + Add the capability to return the content/path of the file.
+    if not os.path.exists(root_dir) or not os.path.isdir(root_dir):
```
benmss, line 42: `isdir` includes a check for `exists` so no need to call both.

benmss (approved):
Looks good to me.

behnazh-w:
> > The core part is added as "repo_verifier" under "repo_finder". "analyzer" calls the "repo_verifier" and adds the info to "dynamic_data". I added a sample check (for maven) that shows how this data can be used.
> 
> Can you please add this information to the PR description?

Looks like you forgot to change the PR description. Could you please fix that?

behnazh-w:
Please add the new check to the check table in the [documentation](https://oracle.github.io/macaron/#current-checks-in-macaron).

behnazh-w:
Please run `make docs-full` to generate the RST templates for the new modules on the documentation website.

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
```diff
@@ -0,0 +1,132 @@
...
+from macaron.repo_verifier.repo_verifier_base import RepositoryVerificationStatus
+from macaron.slsa_analyzer.analyze_context import AnalyzeContext
+from macaron.slsa_analyzer.checks.base_check import BaseCheck
+from macaron.slsa_analyzer.checks.check_result import CheckResultData, CheckResultType, Confidence
+from macaron.slsa_analyzer.registry import registry
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class MavenRepoVerificationFacts(CheckFacts):
```
behnazh-w, line 23: This check should not be specific to Maven. Please clarify in the documentation that it currently supports only Maven, but avoid including `Maven/maven` in the names of the module, fact table, and check. In the implementation, you can verify that the artifact type is Maven and report `UNKNOWN` for all other ecosystems.

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
```diff
@@ -0,0 +1,132 @@
...
+    """The ORM mapping for justifications in maven source repo check."""
+
+    __tablename__ = "_maven_repo_verification_check"
+
+    #: The primary key.
+    id: Mapped[int] = mapped_column(ForeignKey("_check_facts.id"), primary_key=True)  # noqa: A003
+
+    group: Mapped[str] = mapped_column(String, nullable=False)
+    artifact: Mapped[str] = mapped_column(String, nullable=False)
+    version: Mapped[str] = mapped_column(String, nullable=False)
```
behnazh-w, line 33: Is there any particular reason for adding these columns? We can obtain them from the `component` table too.

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # Repository link identified by Macaron's repo finder.
+    repo_link: Mapped[str] = mapped_column(String, nullable=True)
```
behnazh-w, line 36: It would be good to make these visible in the HTML report. You just need to add the following property to the `mapped_column`.
```suggestion
    repo_link: Mapped[str] = mapped_column(String, nullable=True, info={"justification": JustificationType.HREF})
```

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # Repository link identified by deps.dev.
+    deps_dev_repo_link: Mapped[str | None] = mapped_column(String, nullable=True)
```
behnazh-w, line 39:
```suggestion
    deps_dev_repo_link: Mapped[str | None] = mapped_column(String, nullable=True), info={"justification": JustificationType.HREF}
```

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
```diff
@@ -0,0 +1,132 @@
...
+    __tablename__ = "_maven_repo_verification_check"
+
+    #: The primary key.
+    id: Mapped[int] = mapped_column(ForeignKey("_check_facts.id"), primary_key=True)  # noqa: A003
+
+    group: Mapped[str] = mapped_column(String, nullable=False)
+    artifact: Mapped[str] = mapped_column(String, nullable=False)
+    version: Mapped[str] = mapped_column(String, nullable=False)
+
+    # Repository link identified by Macaron's repo finder.
```
behnazh-w, line 35: The comments should be in `#:` format to ensure the docstring renders correctly.

This comment applies for all the properties.

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
```diff
@@ -0,0 +1,132 @@
...
+    repo_link: Mapped[str] = mapped_column(String, nullable=True)
+
+    # Repository link identified by deps.dev.
+    deps_dev_repo_link: Mapped[str | None] = mapped_column(String, nullable=True)
+
+    # Number of stars on the repository identified by deps.dev.
+    deps_dev_stars_count: Mapped[int | None] = mapped_column(Integer, nullable=True)
```
behnazh-w, line 42: This property should be rendered as `TEXT` so, it's slightly different:
```suggestion
    deps_dev_stars_count: Mapped[int | None] = mapped_column(Integer, nullable=True, info={"justification": JustificationType.TEXT})
```

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # Number of forks on the repository identified by deps.dev.
+    deps_dev_fork_count: Mapped[int | None] = mapped_column(Integer, nullable=True)
```
behnazh-w, line 45: Same comment as above regarding adding `info={"justification": JustificationType.TEXT}`.

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # The status of the check: passed, failed, or unknown.
+    status: Mapped[str] = mapped_column(String, nullable=False)
```
behnazh-w, line 48: Same comment as above regarding adding `info={"justification": JustificationType.TEXT}`.

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # The reason for the status.
+    reason: Mapped[str] = mapped_column(String, nullable=False)
```
behnazh-w, line 51: Same comment as above regarding adding `info={"justification": JustificationType.TEXT}`.

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # The build tool used to build the package.
+    build_tool: Mapped[str] = mapped_column(String, nullable=False)
```
behnazh-w, line 54: Same comment as above regarding adding `info={"justification": JustificationType.TEXT}`.

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
mabdollahpour-ol: Not really, just for the ease of analysis. I can remove them.

mabdollahpour-ol:
> > > The core part is added as "repo_verifier" under "repo_finder". "analyzer" calls the "repo_verifier" and adds the info to "dynamic_data". I added a sample check (for maven) that shows how this data can be used.
> > 
> > 
> > Can you please add this information to the PR description?
> 
> Looks like you forgot to change the PR description. Could you please fix that?

I moved `repo_verifier` to its own directory. I changed the PR description (the first comment in this thread to reflect the change). I didn't add this info the body of the squashed commit which I will in the next merge. Does that sound good?

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
mabdollahpour-ol: Yeah, specifically if we don't want it to be Maven-specific, they shouldn't be there.

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
mabdollahpour-ol: Where would be a good place to mention that it only supports Maven packages at this moment? The `description` field is the most visible one for the end users but technically under `returns` section of the docstring for `run_check` would be the proper place. What do you think? @behnazh-w

mabdollahpour-ol:
> Please add the new check to the check table in the [documentation](https://oracle.github.io/macaron/#current-checks-in-macaron) at `docs/source/index.rst`.

The SLSA requirement is "Source - [Version controlled](https://slsa.dev/spec/v0.1/requirements#version-controlled)", right?

behnazh-w:
> > Please add the new check to the check table in the [documentation](https://oracle.github.io/macaron/#current-checks-in-macaron) at `docs/source/index.rst`.
> 
> The SLSA requirement is "Source - [Version controlled](https://slsa.dev/spec/v0.1/requirements#version-controlled)", right?
I've added [my suggestion](https://github.com/oracle/macaron/pull/913#discussion_r1821977708) to the PR. This check is related to `Source - Version controlled`, but it also includes a validation aspect. Generally, while the SLSA requirement column may not use the exact terminology from slsa.dev, there is a relationship.

src/macaron/slsa_analyzer/checks/maven_repo_verification_check.py
behnazh-w: We can mention in in both places.

PR #878, opened 2024-09-29
chore: improve the indentation consistency of souffle Makefile goal

behnazh-w:
This PR makes the following improvements to the `souffle` goal of the `Makefile`:

- A new comment that explains the behavior when Souffle cannot be installed automatically.
- Ensured consistent indentation and spacing for readability based on recommendations in https://github.com/oracle/macaron/pull/877
- Prints a useful message when the OS distribution is not supported.

cc @jenstroeger

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,10 +166,10 @@ souffle:
 	        echo "Unable to install Souffle. Please install it manually." && exit 0; \
 	      fi;; \
 	    *) \
-	        echo "Unsupported OS distribution: $(OS_DISTRO)"; exit 0;; \
+	      echo "Unsupported OS distribution: $(OS_DISTRO)"; exit 0;; \
 	  esac; \
 	fi && \
-	command -v souffle || true
+	command -v souffle
 
 # Install gnu-sed on mac using homebrew
 .PHONY: gnu-sed
--- a/Makefile
+++ b/Makefile
@@ -143,32 +143,33 @@ else
     OS_DISTRO := "$(shell grep '^NAME=' /etc/os-release | sed 's/^NAME=//' | sed 's/"//g')"
   endif
 endif
+# If Souffle cannot be installed, we advise the user to install it manually
+# and return status code 0, which is not considered a failure.
 .PHONY: souffle
 souffle:
 	if ! command -v souffle; then \
-		echo "Installing system dependency: souffle" && \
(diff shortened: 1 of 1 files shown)


tromai (approved):

Makefile
```diff
@@ -143,32 +143,33 @@ else
...
+	      sudo apt update; \
+	      sudo apt install souffle;; \
+	    "Darwin") \
+	      if command -v brew; then \
+	        brew install --HEAD souffle-lang/souffle/souffle; \
+	      else \
+	        echo "Unable to install Souffle. Please install it manually." && exit 0; \
+	      fi;; \
+	    *) \
+	        echo "Unsupported OS distribution: $(OS_DISTRO)"; exit 0;; \
```
jenstroeger, line 169: I think this is indented two spaces too far?

Makefile
```diff
@@ -143,32 +143,33 @@ else
...
+	  esac; \
 	fi && \
-    command -v souffle || true
+	command -v souffle || true
```
jenstroeger, line 172: Hmm, so the `souffle` rule is basically
```bash
if ! command -v souffle; then
  ...
fi && command -v souffle || true
```
I don’t quite understand the purpose of the second check here: is that to ensure that `souffle` was indeed installed by the first `if`? But even if the second check fails, the `|| true` would mask that… 🤔

jenstroeger (approved):
Thank you for formatting @behnazh-w, reads much easier now 😁

Makefile
behnazh-w: Nice catch. It was a tab instead of two spaces. See commit [cd12b4b](https://github.com/oracle/macaron/pull/878/commits/cd12b4be6ca7e8b1c26e9fab89a3a63ef8755a42).

Makefile
behnazh-w: See commit [cd12b4b](https://github.com/oracle/macaron/pull/878/commits/cd12b4be6ca7e8b1c26e9fab89a3a63ef8755a42).

jenstroeger (approved):
The `exit 0` for unsupported host OS is intended, right?

behnazh-w:
> The `exit 0` for unsupported host OS is intended, right?

Yes, and I added a comment to clarify.

PR #880, opened 2024-09-30
test: remove unnecessary integration tests for micronaut-test-junit5

behnazh-w:
I noticed that the integration test `tests/integration/cases/micronaut-projects_micronaut-test` still includes `slf4j` and `caffeine`, which were originally added for the deprecated `-c` option that allowed manually adding dependencies. Since this is no longer possible, we can reduce the integration test runtime by removing these two cases. Additionally, I updated the `micronaut-projects_micronaut-test` test specification to use `purl` instead of the repository path.


tromai:
We should remove `caffeine.dl` and `slf4j.dl` files too.

behnazh-w:
> We should remove `caffeine.dl` and `slf4j.dl` files in that test case directory too.

See commit [ff53dfb](https://github.com/oracle/macaron/pull/880/commits/ff53dfb87dbd84710e8e944a4962d9afe9ae77d2).

tromai (approved):

PR #882, opened 2024-09-30
chore: fix pylint message use-set-for-membership

jenstroeger:
This change is part of issue #876 and addresses the [`use-set-for-membership`](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/use-set-for-membership.html) check messages.

I’ll add the appropriate configuration to pyproject.toml in a separate PR later.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -213,6 +213,7 @@ fail-under = 10.0
 suggestion-mode = true  # Remove this setting when pylint v4 is released.
 load-plugins = [
     "pylint.extensions.for_any_all",
+    "pylint.extensions.set_membership",
 ]
 disable = [
     "fixme",
--- a/src/macaron/repo_finder/provenance_finder.py
+++ b/src/macaron/repo_finder/provenance_finder.py
@@ -69,7 +69,7 @@ class ProvenanceFinder:
             discovery_functions = [partial(find_npm_provenance, purl, self.npm_registry)]
             return self._find_provenance(discovery_functions)
 
-        if purl.type in ["gradle", "maven"]:
+        if purl.type in {"gradle", "maven"}:
             # TODO add support for Maven Central provenance.
             if not self.jfrog_registry:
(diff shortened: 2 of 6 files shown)


behnazh-w (approved):

tromai:
I wonder if this check flags an error on cases where you are performing a membership test in cases like this?
```
a: dict[str, str] = get_some_dictionary(...)
for key in a:
    ...
```

```
import glob

for filename in glob.iglob("*.txt"):
    print(filename)
```

```
a: list[str] = get_some_list(...)
for ele in a:
    print(ele)
```

```
a: list[str] = get_some_list(...)
for ele in a:
    print(ele)
```
Or does it only flag cases where we perform membership test on "statically-defined" lists
```
fruit in ["foo", "boo"]
```

jenstroeger:
> I wonder if […]

Why don’t you just try — see PR https://github.com/oracle/macaron/pull/890 on how to add the plugin ? 😉

behnazh-w (approved):

tromai:
> however, the code in your examples above are not membership tests: a for statement [iterates](https://docs.python.org/3/library/stdtypes.html#iterator-types) over a collection, it does not check whether a given item exists in the collection.

Ah, thanks for pointing it out. I made some mistakes when coming up with the examples, please see the revised version here:

```python
a: dict[str, str] = get_some_dictionary(...)
return if "key" in a.keys()
# OR: return if "key" in a
```

```python
import glob

return "boo.txt" in glob.iglob("*.txt"):
```

```python
a: list[str] = get_some_list(...)
return "boo" in a
```

```python
b = [1,2]
c: list[int] = get_int_list(...)
d = [4,3]
a = [b, c, d]

c in a
```

> Note also that these are not “errors” per se; the optional pylint checks are mostly recommendations for and improvements of code quality and performance.

I should have been a bit more explicit in my wordings. I meant whether pylint would exit with an error/non-zero status code for those cases, but not about the content of the checks.

PR #892, opened 2024-10-03
feat: add Repo Finder and Commit Finder outcomes to database

benmss:
This PR adds a new table `RepoFinderMetadata` that exists to record the various possible outcomes of the Repo Finder and Commit Finder. Possible outcomes include successes as well as all types of failures.


src/macaron/database/table_definitions.py
```diff
@@ -177,7 +178,14 @@ class Component(PackageURLMixin, ORMBase):
         lazy="immediate",
     )
 
-    def __init__(self, purl: str, analysis: Analysis, repository: "Repository | None"):
+    #: The one-to-one relationship with Repo Finder metadata.
+    repo_finder_metadata: Mapped["RepoFinderMetadata"] = relationship(
+        uselist=False, back_populates="component", lazy="immediate"
```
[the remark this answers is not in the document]
benmss: It seems like it is not needed. Can't say for sure why I originally added it at this point.

src/macaron/repo_finder/repo_finder_java.py
```diff
@@ -141,18 +178,26 @@ def _retrieve_pom(self, url: str) -> str:
...
-            The retrieved file data or an empty string.
+        tuple[str, RepoFinderOutcome] :
+            The retrieved file data or an empty string, and the outcome to report.
         """
-        response = send_get_http_raw(url, {})
+        response = send_get_http_raw(url, always_return_response=True)
 
         if not response:
-            return ""
+            return "", RepoFinderOutcome.HTTP_INVALID
```
art1f1c3R, line 187: does `send_get_http_raw` return a `Response` object if the status code isn't `200`? It seems to return `None` when it reaches the max retries and hasn't got a `200 OK` response.

src/macaron/repo_finder/repo_finder_enums.py
```diff
@@ -0,0 +1,70 @@
...
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""This module contains Enums used to represent the outcome of Repo Finder or Commit Finder executions."""
+from enum import Enum
+
+
+class RepoFinderOutcome(Enum):
+    """An Enum of all outcomes of the Repo Finder being run for a software component."""
+
+    # States that relate to problems with user input.
```
behnazh-w, line 11: Can you please add a comment using `#:` for each `Enum` item so that Sphinx can render? Also, more detailed explanation would be helpful to understand the differences and purposes. e.g., in `NO_MAVEN_HOST_PROVIDED`, what is a host and how is it provided, through the POM file? The more information the better.

src/macaron/repo_finder/repo_finder.py
```diff
@@ -407,8 +406,9 @@ def prepare_repo(
 
     Returns
     -------
-    Git | None
-        The pydriller.Git object of the repository or None if error.
+    tuple[Git | None, CommitFinderOutcome, str]
+            The pydriller.Git object of the repository or None if error; the outcome of the Commit Finder; and the final
```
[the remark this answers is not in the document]
benmss: If there is an error relating to the commit finder it will be reported as an outcome. For unrelated errors, no outcome will be reported.

Generally it is not recommended to use `tuple | None` due to how it impacts the unpacking of values.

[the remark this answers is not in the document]
behnazh-w: I believe that even before these changes, the `prepare_repo` function’s interface was somewhat unclear, and these updates make it even harder to follow.

The function’s intent is to return a Git object based on a provided digest, or, if the digest is an empty string, it should return the digest found by the `find_commit` function. If an error occurs, it will not return a Git object. To reduce confusion, I suggest only returning one type, i.e., `GitRepoOutcome`. This type could also hold a reference to the Git object, the `digest` provided as input, and `CommitFinderOutcome`. In cases where the `GitRepoOutcome` cannot be returned, an exception could be thrown.

What do you think about this approach?

src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -39,7 +40,7 @@ class DepsDevRepoFinder(BaseRepoFinder):
     # See https://docs.deps.dev/api/v3alpha/
     BASE_URL = "https://api.deps.dev/v3alpha/purl/"
 
-    def find_repo(self, purl: PackageURL) -> str:
+    def find_repo(self, purl: PackageURL) -> tuple[str, RepoFinderOutcome]:
```
behnazh-w, line 43: I have a similar comment about this interface. Instead of a `tuple`, why not including the repo URL in the `RepoFinderOutcome` object?

src/macaron/repo_finder/repo_finder_enums.py
```diff
@@ -0,0 +1,116 @@
...
+
+"""This module contains Enums used to represent the outcome of Repo Finder or Commit Finder executions."""
+from enum import Enum
+
+
+class RepoFinderOutcome(Enum):
+    """An Enum of all outcomes of the Repo Finder being run for a software component."""
+
+    #: Reported if the URL that serves requested Maven packages is not provided by the user in Macaron's config.
+    # E.g. Maven central.
```
behnazh-w, line 12: All the comment lines before the attribute should start with `#:` otherwise Sphinx won't render it. This applies to all the attributes in both `RepoFinderOutcome` and `CommitFinderOutcome`.

```suggestion
    #: Reported if the URL that serves requested Maven packages is not provided by the user in Macaron's config.
    #: E.g. Maven central.
```

src/macaron/repo_finder/repo_finder_enums.py
```diff
@@ -0,0 +1,116 @@
...
+
+    #: Default value. Reported if the Repo Finder was not called. E.g. Because the repository URL was already present.
+    NOT_USED = "Not used"
+
+
+class CommitFinderOutcome(Enum):
+    """An Enum of all outcomes of the Commit Finder being run for a software component."""
+
+    #: Reported if the user does not provide a version for the Repo Finder via the command line, or allow the version to
+    # be automatically discovered.
```
behnazh-w, line 83:
```suggestion
    #: Reported if the user does not provide a version for the Source Finder via the command line, or allow the version to
    #: be automatically discovered.
```

src/macaron/repo_finder/repo_finder_enums.py
```diff
@@ -0,0 +1,116 @@
...
+
+    #: Reported if the user provides an unsupported type in the PURL command line argument.
+    UNSUPPORTED_PURL_TYPE = "Unsupported PURL type"
+
+    #: Reported if the user provided a repository type PURL with the tag or commit in the version, but neither were
+    # valid.
+    REPO_PURL_FAILURE = "Repository PURL failure"
+
+    #: Reported if the repository has no tags.
+    NO_TAGS = "No tags"
```
behnazh-w, line 94:
```suggestion
    NO_TAGS = "No Git tags"
```

src/macaron/repo_finder/repo_finder_enums.py
```diff
@@ -0,0 +1,116 @@
...
+
+    #: Reported if the repository has no tags with associated commits.
+    NO_TAGS_WITH_COMMITS = "No tags with commits"
```
behnazh-w, line 97:
```suggestion
    NO_TAGS_WITH_COMMITS = "No Git tags with commits"
```

src/macaron/repo_finder/repo_finder_enums.py
```diff
@@ -0,0 +1,116 @@
...
+
+    #: Reported if the tag selected from the repository fails to resolve to a commit despite having one associated with
+    # it.
+    NO_TAG_COMMIT = "No tag commit"
```
behnazh-w, line 101:
```suggestion
    NO_TAG_COMMIT = "No valid commit found for Git tag"
```

src/macaron/repo_finder/repo_finder_enums.py
```diff
@@ -0,0 +1,116 @@
...
+
+    #: Reported if the version part of the user provided PURL is invalid.
```
behnazh-w, line 103: The comment only says the `version` part. Is that intended?

src/macaron/repo_finder/repo_finder_enums.py
```diff
@@ -0,0 +1,116 @@
...
+    INVALID_PURL = "No valid parts"
+
+    #: Reported if the Regex pattern to be created from parts of the user provided PURL fails to compile.
```
behnazh-w, line 106: I'm not sure I understand this comment. Can you please give an example?

src/macaron/repo_finder/repo_finder_enums.py
```diff
@@ -0,0 +1,116 @@
+# Copyright (c) 2024 - 2025, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""This module contains Enums used to represent the outcome of Repo Finder or Commit Finder executions."""
+from enum import Enum
+
+
+class RepoFinderOutcome(Enum):
```
behnazh-w, line 8: I think this `enum` is not really the repo finder outcome, as in it doesn't contain the repo URL. 

```suggestion
class RepoFinderMetadata(Enum):
```

src/macaron/repo_finder/repo_finder_enums.py
```diff
@@ -0,0 +1,116 @@
...
+    FOUND_FROM_PARENT = "Found from parent"
+
+    #: Reported when a repository is found from a more recent version than was provided by the user.
+    FOUND_FROM_LATEST = "Found form latest"
+
+    #: Default value. Reported if the Repo Finder was not called. E.g. Because the repository URL was already present.
+    NOT_USED = "Not used"
+
+
+class CommitFinderOutcome(Enum):
```
behnazh-w, line 79: Same comment regarding the name of the class.

```suggestion
class CommitFinderMetadata(Enum):
```

src/macaron/repo_finder/repo_finder.py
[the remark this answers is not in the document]
behnazh-w: Why is the digest included in the return values? Since we can obtain the commit from the Git object, is there any scenario where retrieving the commit this way wouldn’t work?

behnazh-w:
Please update the [ER Diagram](https://github.com/oracle/macaron/blob/7cfc8392af87f6b7b412f23fa4315dc1c0b5791a/docs/source/assets/er-diagram.svg).

You can use the `eralchemy2` Python library:
```bash
eralchemy2 -i 'sqlite:///<path>/output/macaron.db' -o er-diagram.svg
```

src/macaron/util.py
```diff
@@ -126,7 +126,11 @@ def send_head_http_raw(
 
 
 def send_get_http_raw(
-    url: str, headers: dict | None = None, timeout: int | None = None, allow_redirects: bool = True
+    url: str,
+    headers: dict | None = None,
+    timeout: int | None = None,
+    allow_redirects: bool = True,
+    always_return_response: bool = False,
```
behnazh-w, line 133: The docstring misses the new parameter and the return value description needs to be adjusted.

Also, it seems more intuitive to change the parameter to 
```suggestion
   check_response_fail: bool = True,
```

tests/integration/cases/repo_finder_remote_calls/repo_finder.py
```diff
@@ -73,17 +80,18 @@ def test_repo_finder() -> int:
 
     # Test Java package whose SCM metadata only points to the repo in later versions than is provided here.
     purl = PackageURL.from_string("pkg:maven/io.vertx/vertx-auth-common@3.8.0")
-    repo = find_repo(purl)
+    repo, _ = find_repo(purl)
```
behnazh-w, line 83: It would be good to check the expected metadata here as well.

tests/integration/cases/repo_finder_remote_calls/repo_finder.py
```diff
@@ -73,17 +80,18 @@ def test_repo_finder() -> int:
...
     if repo == "https://github.com/eclipse-vertx/vertx-auth":
         return os.EX_UNAVAILABLE
-    latest_purl = DepsDevRepoFinder().get_latest_version(purl)
+    latest_purl, _ = DepsDevRepoFinder().get_latest_version(purl)
```
behnazh-w, line 86: Same comment regarding metadata checking.

tests/integration/cases/repo_finder_remote_calls/repo_finder.py
```diff
@@ -73,17 +80,18 @@ def test_repo_finder() -> int:
...
     assert latest_purl
-    repo = find_repo(latest_purl)
+    repo, _ = find_repo(latest_purl)
```
behnazh-w, line 88: Same comment regarding metadata checking.

src/macaron/repo_finder/repo_finder_enums.py
benmss: It is the Repo Finder that this relates to. The Source Finding operation does not make use of this enum.

src/macaron/repo_finder/repo_finder_enums.py
benmss: Yes. I've changed it to `INVALID_VERSION`

src/macaron/repo_finder/repo_finder_enums.py
benmss: The Commit Finder uses the version parts from the PURL provided as input to create a regex pattern. If this pattern creation fails, this will be reported. At the moment I do not have an example of this failure (it would be in a test), but it is theoretically possible.

docs/source/assets/er-diagram.svg
behnazh-w, line 1: Looks like the `repo_finder_metadata` relationship is still missing in the new diagram.

behnazh-w (approved):

behnazh-w (approved):

PR #914, opened 2024-11-01
feat: add a new check to report the build tool

behnazh-w:
This pull request adds a new check that identifies whether a supported build tool configuration exists in the associated source code repository of a software component. It also includes relevant unit and integration tests.


src/macaron/slsa_analyzer/registry.py
```diff
@@ -147,10 +147,13 @@ def _validate_check(check: Any) -> bool:
             True if check is valid, else False.
         """
         if not isinstance(check, BaseCheck):
-            logger.error(
-                "The registered Check is of type %s. Please register a child class of BaseCheck.",
-                type(check).__name__,
-            )
+            if hasattr(check, __name__):
```
tromai, line 150:
```suggestion
            if hasattr(type(check), "__name__"):
```

src/macaron/slsa_analyzer/registry.py
```diff
@@ -147,10 +147,13 @@ def _validate_check(check: Any) -> bool:
...
+                logger.error(
+                    "The registered Check %s is not a valid instance of BaseCheck.",
+                    check.__name__,
```
tromai, line 153:
```suggestion
                    type(check).__name__,
```

behnazh-w: `type(check).__name__` resolves to the string value `type` instead of the check name, and that's actually what is fixed here.

behnazh-w: This [263a4e2](https://github.com/oracle/macaron/pull/914/commits/263a4e2e5668f9272fc8a55a675d9eb535d0bbbc) also addresses the scenario in which `check` is an instance object.

behnazh-w: Printing the name of the class that was supposed to be the check is still helpful to the user for debugging purposes. I noticed that I had accidentally removed `return False` in this block, which I fixed [10f011a](https://github.com/oracle/macaron/pull/914/commits/10f011a8da5fa33fedaecebe715c2415918e4bdd).

tromai (approved):

nicallen (approved):

PR #918, opened 2024-11-13
fix: block terminal prompts in find source

benmss:
Applies a fix for the standalone commit finder feature: preventing login related terminal prompts from interrupting the process.

### DIFF
--- a/src/macaron/repo_finder/repo_finder.py
+++ b/src/macaron/repo_finder/repo_finder.py
@@ -235,7 +235,9 @@ def get_tags_via_git_remote(repo: str) -> dict[str, str] | None:
     """
     tags = {}
     try:
-        tag_data = git.cmd.Git().ls_remote("--tags", repo)
+        command = git.cmd.Git()
+        command.update_environment(GIT_TERMINAL_PROMPT="")
+        tag_data = command.ls_remote("--tags", repo)
     except git.exc.GitCommandError as error:
         logger.debug("Failed to retrieve tags: %s", error)
         return None
--- a/src/macaron/repo_finder/repo_finder.py
+++ b/src/macaron/repo_finder/repo_finder.py
@@ -236,13 +236,6 @@ def get_tags_via_git_remote(repo: str) -> dict[str, str] | None:
     if not tag_data:
         return None
     tags = {}
-    # try:
-    #    command = git.cmd.Git()
-    # command.update_environment(GIT_TERMINAL_PROMPT="")
-    #    tag_data = command.ls_remote("--tags", repo)
-    # except git.exc.GitCommandError as error:
-    #    logger.debug("Failed to retrieve tags: %s", error)
-    #    return None
 
     for tag_line in tag_data.splitlines():
         tag_line = tag_line.strip()
--- a/src/macaron/repo_finder/repo_finder.py
+++ b/src/macaron/repo_finder/repo_finder.py
@@ -36,7 +36,6 @@ import logging
 import os
 from urllib.parse import ParseResult, urlunparse
 
-import git
 from packageurl import PackageURL
 
 from macaron.config.defaults import defaults
@@ -47,7 +46,7 @@ from macaron.repo_finder.repo_finder_base import BaseRepoFinder
 from macaron.repo_finder.repo_finder_deps_dev import DepsDevRepoFinder
 from macaron.repo_finder.repo_finder_java import JavaRepoFinder
 from macaron.repo_finder.repo_utils import generate_report, prepare_repo
-from macaron.slsa_analyzer.git_url import GIT_REPOS_DIR
+from macaron.slsa_analyzer.git_url import GIT_REPOS_DIR, list_remote_references
 
 logger: logging.Logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


src/macaron/repo_finder/repo_finder.py
```diff
@@ -235,7 +235,9 @@ def get_tags_via_git_remote(repo: str) -> dict[str, str] | None:
     """
     tags = {}
     try:
-        tag_data = git.cmd.Git().ls_remote("--tags", repo)
+        command = git.cmd.Git()
+        command.update_environment(GIT_TERMINAL_PROMPT="")
+        tag_data = command.ls_remote("--tags", repo)
```
behnazh-w, line 240: We already have a [function](https://github.com/oracle/macaron/blob/edfe06eb3f9337f86215b253f40f53ca1a0105c9/src/macaron/environment_variables.py#L10) that manages the environment variables.

It's also better to be consistent with the implementation of [clone_remote_repo](https://github.com/oracle/macaron/blob/edfe06eb3f9337f86215b253f40f53ca1a0105c9/src/macaron/slsa_analyzer/git_url.py#L294) by calling the `git` process. I think the functionality can be implemented in the `git_url.py` module to be reused elsewhere. 

Also, please add a unit test.

tromai: I wonder what logic should we unit test here 
If we decide to use [`get_patched_env`](https://github.com/oracle/macaron/blob/edfe06eb3f9337f86215b253f40f53ca1a0105c9/src/macaron/environment_variables.py#L10) then there are existing unit tests for env patching [here](https://github.com/oracle/macaron/blob/ec4e1907e0acf213d80ff7c8420e073d9d37494d/tests/test_environment_variables.py#L43)

But if we want to test that calling `git ls-remote` doesn't prompt for user input, then I'm not sure if it is suitable for a unit test because `git ls-remote` accesses external network.

src/macaron/repo_finder/repo_finder.py
```diff
@@ -233,12 +232,17 @@ def get_tags_via_git_remote(repo: str) -> dict[str, str] | None:
...
+    if not tag_data:
         return None
+    tags = {}
+    # try:
+    #    command = git.cmd.Git()
+    # command.update_environment(GIT_TERMINAL_PROMPT="")
+    #    tag_data = command.ls_remote("--tags", repo)
+    # except git.exc.GitCommandError as error:
+    #    logger.debug("Failed to retrieve tags: %s", error)
+    #    return None
```
behnazh-w, line 245: Are these commented lines left intentionally?

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -376,6 +377,45 @@ def clone_remote_repo(clone_dir: str, url: str) -> Repo | None:
...
+            cwd=global_config.output_path,
+            check=False,
+        )
+    except (subprocess.CalledProcessError, OSError):
+        return None
+
+    if result.returncode != 0:
+        error_string = result.stderr.decode("utf-8").strip()
+        if error_string.startswith("fatal: could not read Username"):
+            # Occurs when a repository cannot be accesses either because it does not exist, or it requires a login
```
behnazh-w, line 411:
```suggestion
            # Occurs when a repository cannot be accessed either because it does not exist, or it requires a login
```

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -376,6 +377,45 @@ def clone_remote_repo(clone_dir: str, url: str) -> Repo | None:
...
+            # that is blocked.
+            logger.error("Could not access repository: %s", repo)
+        return None
```
behnazh-w, line 414: It would be good to also print a general error message when the error message does not start with `fatal: could not read Username`.

tests/integration/cases/find_source_fail/test.yaml
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+description: |
+  Analyzing the find source command on the a non-existent repository.
```
behnazh-w, line 5:
```suggestion
  Analyzing the find source command on a non-existent repository.
```

tests/integration/cases/find_source_fail/test.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+steps:
+- name: Run macaron find source on private repository
+  kind: find-source
+  options:
+    command_args:
+    - -purl
+    - pkg:maven/com.example/example@1.0.0
+    - -rp
+    - https://github.com/oracle/hopefully-this-repository-will-never-exist-0
```
behnazh-w, line 19: If a repository does not exist, do we receive the same terminal prompt as when it exists but is private?

benmss: Yes. This is to prevent people discovering repositories that exist but they don't have access to.

behnazh-w (approved):

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -376,6 +377,47 @@ def clone_remote_repo(clone_dir: str, url: str) -> Repo | None:
...
+    str
+        The result of the command.
+    """
+    try:
+        result = subprocess.run(  # nosec B603
+            args=["git", "ls-remote"] + arguments + [repo],
+            capture_output=True,
+            # By setting stdin to /dev/null and using a new session, we prevent all possible user input prompts.
+            stdin=subprocess.DEVNULL,
+            start_new_session=True,
```
tromai, line 401: Just to confirm: we will use this to resolve the issue instead of relying to patching the env variable `GIT_TERMINAL_PROMPT`.
This is a different approach to https://github.com/oracle/macaron/blob/c1c4eea36fafd35f72b9409c9d4df119cbb07451/src/macaron/slsa_analyzer/git_url.py#L351-L365

benmss: It is a different approach. Possibly it is more robust as it prevents any requests for input instead of just a specific type, but it may well be that the specific type is the only one we need to worry about.

tromai (approved):

behnazh-w (approved):

PR #922, opened 2024-11-19
feat: report known malware for all ecosystems

behnazh-w:
If a package is already known to be malicious, this PR reports it as part of the `mcn_detect_malicious_metadata_1` check. Additionally, two new integration tests for known Python and npm malware have been added.

### DIFF
--- a/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
+++ b/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
@@ -5,12 +5,13 @@
 
 import logging
 
-from sqlalchemy import ForeignKey
+import requests
+from sqlalchemy import ForeignKey, String
 from sqlalchemy.orm import Mapped, mapped_column
 
 from macaron.database.db_custom_types import DBJsonDict
 from macaron.database.table_definitions import CheckFacts
-from macaron.json_tools import JsonType
+from macaron.json_tools import JsonType, json_extract
 from macaron.malware_analyzer.pypi_heuristics.base_analyzer import BaseHeuristicAnalyzer
 from macaron.malware_analyzer.pypi_heuristics.heuristics import HeuristicResult, Heuristics
 from macaron.malware_analyzer.pypi_heuristics.metadata.closer_release_join_date import CloserReleaseJoinDateAnalyzer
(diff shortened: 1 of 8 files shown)


art1f1c3R (approved):

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -40,13 +42,16 @@ class MaliciousMetadataFacts(CheckFacts):
...
+
     #: Detailed information about the analysis.
     detail_information: Mapped[dict[str, JsonType]] = mapped_column(DBJsonDict, nullable=False)
 
-    #: The result of analysis, which is of dict[Heuristics, HeuristicResult] type.
-    result: Mapped[dict[Heuristics, HeuristicResult]] = mapped_column(
-        DBJsonDict, nullable=False, info={"justification": JustificationType.TEXT}
-    )
+    #: The result of analysis, which can be an empty dictionary.
+    result: Mapped[dict] = mapped_column(DBJsonDict, nullable=False, info={"justification": JustificationType.TEXT})
```
nicallen, line 54: Why change `dict[Heuristics, HeuristicResult]]` to `dict`? In the new case it is empty, but that is still consistent with the old type.

src/macaron/util.py
```diff
@@ -125,6 +125,80 @@ def send_get_http_raw(
     return response
 
 
+def send_post_http_raw(
```
nicallen, line 128: There is quite a bit of duplication between `send_post_http_raw`, `send_get_http_raw` and `send_get_http` (each has an implementation of error handling and retry logic), it would be good to refactor that at some point (but it doesn't have to be done as part of this change).

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
behnazh-w: That's right. I don't need to change `dict[Heuristics, HeuristicResult]]`.

behnazh-w: See commit [a61bb94](https://github.com/oracle/macaron/pull/922/commits/a61bb94fb8403268e0931445dc256b641f35e4d2).

nicallen (approved):

src/macaron/util.py
```diff
@@ -125,6 +125,80 @@ def send_get_http_raw(
...
+    while response.status_code != 200:
+        logger.debug(
+            "Receiving error code %s from server.",
+            response.status_code,
+        )
+        if retry_counter <= 0:
+            logger.debug("Maximum retries reached: %s", error_retries)
+            return None
+        if response.status_code == 403:
+            check_rate_limit(response)
```
tromai, line 187: This could be a part of another PR. The logic for `check_rate_limit` was originally developed to handle limit exceeding in querying GithubAPI. Therefore going forward we need to decide / test if it's applicable to be used in this generic function call.

tromai (approved):

PR #923, opened 2024-11-22
test(integration): handle missing steps info in GitHub API response

behnazh-w:
The GitHub API for some reason does not anymore return the steps information of the job that has published `pkg:maven/io.micronaut.test/micronaut-test-junit5@4.5.0` even though it was published in Aug 2024, which is much earlier than [the 400 retention policy](https://docs.github.com/en/rest/guides/using-the-rest-api-to-interact-with-checks?apiVersion=2022-11-28#retention-of-checks-data). This PR raises a new exception to handle this case and allows the corresponding integration test to fail.

### DIFF
--- a/src/macaron/slsa_analyzer/ci_service/github_actions/github_actions_ci.py
+++ b/src/macaron/slsa_analyzer/ci_service/github_actions/github_actions_ci.py
@@ -13,7 +13,7 @@ from datetime import datetime, timedelta, timezone
 from macaron.code_analyzer.call_graph import BaseNode, CallGraph
 from macaron.config.defaults import defaults
 from macaron.config.global_config import global_config
-from macaron.errors import CallGraphError, ParseError
+from macaron.errors import CallGraphError, GitHubActionsValueError, ParseError
 from macaron.parsers.bashparser import BashNode, BashScriptType
 from macaron.slsa_analyzer.build_tool.base_build_tool import BaseBuildTool, BuildToolCommand
 from macaron.slsa_analyzer.ci_service.base_ci_service import BaseCIService
@@ -333,6 +333,11 @@ class GitHubActions(BaseCIService):
         -------
         set[str]
             The set of URLs found for the workflow within the time range.
+
+        Raises
+        ------
+        GitHubActionsValueError
+            This error is raised when the GitHub Action workflow run misses values.
         """
         logger.debug(
             "Getting the latest workflow run of %s at publishing time %s and source commit date %s within time range %s.",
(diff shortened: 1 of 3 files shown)


src/macaron/slsa_analyzer/ci_service/github_actions/github_actions_ci.py
```diff
@@ -377,6 +382,8 @@ def workflow_run_in_date_time_range(
 
                     # Find the matching step and check its `conclusion` and `started_at` attributes.
                     html_url = None
+                    if not run_jobs["jobs"]:
```
tromai, line 385: I believe at this point `run_jobs` is a dictionary. This if condition will only covers cases where the key `jobs` is available but its value is None or an empty dictionary. This will not cover cases where `jobs` key is not available in `run_jobs`. I wonder if this is intended ?

src/macaron/slsa_analyzer/ci_service/github_actions/github_actions_ci.py
```diff
@@ -393,6 +400,11 @@ def workflow_run_in_date_time_range(
                                 html_url = item["html_url"]
                                 break
 
+                        if not job["steps"]:
```
tromai, line 403: Similar comment as https://github.com/oracle/macaron/pull/923/files#r1853266727

src/macaron/slsa_analyzer/ci_service/github_actions/github_actions_ci.py
tromai: I'm also not clear on the new behavior of GitHub API? Does it return an empty list now or does it completely remove the `jobs` field?

tromai (commented):

src/macaron/slsa_analyzer/ci_service/github_actions/github_actions_ci.py
behnazh-w: There are multiple scenarios. GitHub removes data gradually, e.g., first the `steps` data is removed, then `jobs` data. Finally, the whole run data disappears. But the keys are not removed, instead they contain empty lists, until its parent object is removed.

src/macaron/slsa_analyzer/ci_service/github_actions/github_actions_ci.py
behnazh-w: Same comment as above.

src/macaron/slsa_analyzer/ci_service/github_actions/github_actions_ci.py
tromai: Right so we have the assumption that `jobs` and `steps` fields are always available at this point of execution.

tromai (approved):

PR #930, opened 2024-11-27
feat: check PyPI packages for presence of a wheel file in malware metadata analysis

art1f1c3R:
A new heuristic has been added to the PyPI package malware analyzer heuristics that checks if the specified package and version is available as a wheel (`.whl`) file. The rationale behind this is that, if a malicious package has a suspicious setup file (`setup.py`), then it will omit a wheel file in the package so that when installed, the setup file will be run automatically. This new heuristic passes if there is a wheel file available, and fails if there is not. It has been added to the suspicious combinations (`SUSPICIOUS_COMBO`) such that, when the suspicious setup heuristic fails, this heuristic must fail. When the suspicious setup heuristic passes, this heuristic is indifferent and has no effect.
New files added to the project are:

- `src/macaron/malware_analyzer/pypi_heuristics/metadata/wheel_presence.py`: the new heuristic and its analyze functionality
- `tests/malware_analyzer/pypi/test_wheel_presence.py`: a test file for the new heuristic.


src/macaron/malware_analyzer/pypi_heuristics/heuristics.py
```diff
@@ -31,6 +31,9 @@ class Heuristics(str, Enum):
     #: Indicates that the setup.py file contains suspicious imports, such as base64 and requests.
     SUSPICIOUS_SETUP = "suspicious_setup"
 
+    #: Indicates that the package does not include a .whl file
+    WHEEL_PRESENCE = "wheel_presence"
```
nicallen, line 35: Would it be better to call this "WHEEL_ABSENCE"? All the others appear to be named based on the negative signal they represent when they fail, whereas this one passes when the wheel is present.

tests/malware_analyzer/pypi/test_wheel_presence.py
```diff
@@ -0,0 +1,185 @@
...
+    """Test for when there is no release information, so error"""
+    analyzer = WheelPresenceAnalyzer()
+
+    pypi_package_json.get_releases.return_value = None
+
+    try:
+        analyzer.analyze(pypi_package_json)
+        pytest.fail("Analyzer did not through a HeuristicAnalyzerValueError")
+    except HeuristicAnalyzerValueError:
+        pass
```
behnazh-w, line 24:
```suggestion
    with pytest.raises(HeuristicAnalyzerValueError):
        analyzer.analyze(pypi_package_json)
```

tests/malware_analyzer/pypi/test_wheel_presence.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def test_analyze_tar_present(pypi_package_json: MagicMock) -> None:
+    """Test for when only .tar.gz is present, so failed"""
```
behnazh-w, line 28: I'm not sure understand the purpose of this test. Could you clarify?

tests/malware_analyzer/pypi/test_wheel_presence.py
```diff
@@ -0,0 +1,185 @@
...
+
+    pypi_package_json.get_releases.return_value = release
+    pypi_package_json.get_latest_version.return_value = version
+    pypi_package_json.component.version = None
+    expected_result: tuple[HeuristicResult, dict] = (HeuristicResult.FAIL, {version: [filename]})
+
+    try:
+        actual_result = analyzer.analyze(pypi_package_json)
+    except HeuristicAnalyzerValueError:
+        pytest.fail("Analyzer threw a HeuristicAnalyzerValueError")
```
behnazh-w, line 68:
```suggestion
    with pytest.raises(HeuristicAnalyzerValueError):
        analyzer.analyze(pypi_package_json)
```

tests/malware_analyzer/pypi/test_wheel_presence.py
```diff
@@ -0,0 +1,185 @@
...
+
+    assert actual_result == expected_result
```
behnazh-w, line 70: Is this assertion needed?

tests/malware_analyzer/pypi/test_wheel_presence.py
```diff
@@ -0,0 +1,185 @@
...
+            }
+        ]
+    }
+
+    pypi_package_json.get_releases.return_value = release
+    pypi_package_json.component.version = version
+    expected_result: tuple[HeuristicResult, dict] = (HeuristicResult.PASS, {version: [filename]})
+
+    try:
+        actual_result = analyzer.analyze(pypi_package_json)
```
behnazh-w, line 111: Are you expecting an exception to be raised here? If not, the `try except` should be removed.

tests/malware_analyzer/pypi/test_wheel_presence.py
```diff
@@ -0,0 +1,185 @@
...
+
+    pypi_package_json.get_releases.return_value = release
+    pypi_package_json.component.version = version
+    expected_result: tuple[HeuristicResult, dict] = (
+        HeuristicResult.PASS,
+        {version: [f"{file_prefix}.whl", f"{file_prefix}.tar.gz"]},
+    )
+
+    try:
+        actual_result = analyzer.analyze(pypi_package_json)
```
behnazh-w, line 181: Same comment here.

tests/malware_analyzer/pypi/test_wheel_presence.py
art1f1c3R: In most scenarios where the heuristic should fail, only a tarball is distributed and a wheel file is not. It's a scenario where there is valid release information but the wheel file is not present.

tests/malware_analyzer/pypi/test_wheel_presence.py
art1f1c3R: I think so. The release information is valid so it should now throw an error, and then the following assertion checks that it returned a fail with the filename as data.

tests/malware_analyzer/pypi/test_wheel_presence.py
art1f1c3R: I'm not expecting an exception, but if the `HeuristicAnalyzerValueError` is thrown the test should fail. I can remove it and just let the error be thrown, which would also indicate a fail too?

tests/malware_analyzer/pypi/test_wheel_presence.py
art1f1c3R: I've added this to `test_analyze_no_information` to make it cleaner, but since this test (`test_analyze_tar_present`) does not expect an error I haven't made the change there.

src/macaron/malware_analyzer/pypi_heuristics/heuristics.py
art1f1c3R: That's fair, I can change this, the file-names, and the heuristic names to that since, for example `EmptyProjectLinkAnalyzer` passes when there is no empty project link.

tests/malware_analyzer/pypi/test_wheel_presence.py
behnazh-w: I see. Since an error is not expected, we could remove the `try except` block to catch unexpected errors. But using `pytest.fail` is also fine.

tests/malware_analyzer/pypi/test_wheel_presence.py
behnazh-w: Makes sense. I was confused by the previous `try except` block. Since, we don't expect an error, this assertion indeed makes sense.

tests/malware_analyzer/pypi/test_wheel_presence.py
behnazh-w: Yes, you can remove it because if an error is thrown, the test will fail and indicate a bug.

behnazh-w (approved):

nicallen (approved):

PR #931, opened 2024-11-28
feat: find repo from latest artifact when provided artifact has none

benmss:
This PR improves the Repo Finder by making it check for repositories using the latest version of an artifact if the provided version does not contain that information. This latest version finding behaviour is also added for the case where a Java PURL was provided without a version, and is added for the `find-source` command.

### DIFF
--- a/src/macaron/repo_finder/repo_finder.py
+++ b/src/macaron/repo_finder/repo_finder.py
@@ -36,28 +36,48 @@ import logging
 import os
 from urllib.parse import ParseResult, urlunparse
 
+from git import InvalidGitRepositoryError
 from packageurl import PackageURL
+from pydriller import Git
 
 from macaron.config.defaults import defaults
 from macaron.config.global_config import global_config
+from macaron.errors import CloneError, RepoCheckOutError
 from macaron.repo_finder import to_domain_from_known_purl_types
-from macaron.repo_finder.commit_finder import match_tags
+from macaron.repo_finder.commit_finder import find_commit, match_tags
 from macaron.repo_finder.repo_finder_base import BaseRepoFinder
 from macaron.repo_finder.repo_finder_deps_dev import DepsDevRepoFinder
 from macaron.repo_finder.repo_finder_java import JavaRepoFinder
-from macaron.repo_finder.repo_utils import check_repo_urls_are_equivalent, generate_report, prepare_repo
-from macaron.slsa_analyzer.git_url import GIT_REPOS_DIR, list_remote_references
+from macaron.repo_finder.repo_utils import (
+    check_repo_urls_are_equivalent,
+    generate_report,
+    get_git_service,
+    get_local_repos_path,
+)
+from macaron.slsa_analyzer.git_url import (
+    GIT_REPOS_DIR,
+    check_out_repo_target,
+    get_remote_origin_of_local_repo,
+    get_remote_vcs_url,
+    get_repo_dir_name,
+    is_empty_repo,
+    is_remote_repo,
+    list_remote_references,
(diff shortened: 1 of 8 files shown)


src/macaron/repo_finder/repo_finder.py
```diff
@@ -133,7 +133,7 @@ def to_repo_path(purl: PackageURL, available_domains: list[str]) -> str | None:
     )
 
 
-def find_source(purl_string: str, input_repo: str | None) -> bool:
+def find_source(purl_string: str, input_repo: str | None, latest_version_fallback: bool = True) -> bool:
```
behnazh-w, line 136: It seems that the repo finder used in the [analyzer](https://github.com/oracle/macaron/blob/a2b6ba929778f97e3cb6ac1453230e78a89fbce4/src/macaron/slsa_analyzer/analyzer.py#L783) module is inconsistent with this implementation, as it doesn't utilize the same fallback for the latest version?

benmss: It does use the same mechanism just through a different path. The repository is potentially re-found during the [prepare_repo](https://github.com/oracle/macaron/blob/a2b6ba929778f97e3cb6ac1453230e78a89fbce4/src/macaron/repo_finder/repo_utils.py#L129) function call in `repo_utils.py`

src/macaron/repo_finder/repo_finder.py
```diff
@@ -113,8 +148,7 @@ def to_repo_path(purl: PackageURL, available_domains: list[str]) -> str | None:
     domain = to_domain_from_known_purl_types(purl.type) or (purl.type if purl.type in available_domains else None)
     if not domain:
         logger.info("The PURL type of %s is not valid as a repository type. Trying to find the repository...", purl)
-        # Try to find the repository
-        return find_repo(purl)
+        return None
```
tromai, line 151: I wonder why we decided to return None here instead of `find_repo(purl)`

benmss: It was redundant. The only callsite of this function performs the removed function call if None is returned.

behnazh-w:
Can you please add an integration test for one of the artifacts that required this feature to find the correct repository?

benmss:
One is included [here](https://github.com/oracle/macaron/pull/931/files#diff-716ef03f4db35ce956c78525fefde4d5f92eca1262216ffb3540d2090525c25a).

behnazh-w:
> One is included [here](https://github.com/oracle/macaron/pull/931/files#diff-716ef03f4db35ce956c78525fefde4d5f92eca1262216ffb3540d2090525c25a).

I meant to add an integration test with two steps for the `analyze` and `find-source` commands to make sure they both have a similar behavior.

tests/integration/cases/latest_repo_comparison/test.yaml
```diff
@@ -0,0 +1,36 @@
...
+- name: Check that the repository was not cloned
+  kind: shell
+  options:
+    cmd: ls output/git_repos/github_com/avaje/avaje-prisms/
+  expect_fail: true
+- name: Check the report contents
+  kind: shell
+  options:
+    cmd: ./check_output.sh
+- name: Run macaron analyze
```
behnazh-w, line 27: I think we should clean the `output/` directory to ensure the cached repo from the previous step is not used by this step?

benmss: The previous step checks that the repository directory does not exist, so there is nothing to clean up at this point.

behnazh-w (approved):

PR #932, opened 2024-12-02
feat: add a new setup.py related heuristic in the pypi malware analyzer

art1f1c3R:
A new heuristic has been added to the PyPI package malware analyzer heuristics that checks if the specified package and version is available as a wheel (.whl) file. The rationale behind this is that, if a malicious package has a suspicious setup file (setup.py), then it will omit a wheel file in the package so that when installed, the setup file will be run automatically. This new heuristic passes if there is a wheel file available, and fails if there is not. It has been added to the suspicious combinations (SUSPICIOUS_COMBO) such that, when the suspicious setup heuristic fails, this heuristic must fail. When the suspicious setup heuristic passes, this heuristic is indifferent and has no effect.
New files added to the project are:

- `src/macaron/malware_analyzer/pypi_heuristics/metadata/wheel_absence.py`: the new heuristic and its analyze functionality

- `tests/malware_analyzer/pypi/test_wheel_absence.py`: a test file for the new heuristic.

This PR was previously approved as #930, but there were issues with commit signing, so a new branch and PR has been created to resolve those issues.

### DIFF
--- a/src/macaron/database/table_definitions.py
+++ b/src/macaron/database/table_definitions.py
@@ -103,7 +103,7 @@ class PackageURLMixin:
     name: Mapped[str] = mapped_column(String(100), nullable=False, comment="Name of the package.")
 
     #: Version of the package.
-    version: Mapped[str] = mapped_column(String(100), nullable=True, comment="Version of the package.")
+    version: Mapped[str | None] = mapped_column(String(100), nullable=True, comment="Version of the package.")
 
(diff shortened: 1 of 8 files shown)


behnazh-w (approved):
This changeset was reviewed earlier in [this PR](https://github.com/oracle/macaron/pull/930) and had to be recreated due to a commit signing issue.

PR #935, opened 2024-12-03
feat: include inspector package urls as part of the malicious metadata facts for pypi packages

art1f1c3R:
To allow for retention of packages when they are taken off PyPI, this new feature includes the inspector.pypi.io URL for the distribution files as part of the `MaliciousMetadataFacts` `detail_information` field. This has been done by modifying the `wheel_absence.py` heuristic to instead of returning filenames, return the python hosted URL and corresponding pypi inspector URL. The unit test for this, `test_wheel_absence.py` has been updated to reflect these changes accordingly.

### DIFF
--- a/src/macaron/malware_analyzer/pypi_heuristics/metadata/wheel_absence.py
+++ b/src/macaron/malware_analyzer/pypi_heuristics/metadata/wheel_absence.py
@@ -5,8 +5,10 @@
 
 import logging
 
+import requests
+
 from macaron.errors import HeuristicAnalyzerValueError
-from macaron.json_tools import JsonType
+from macaron.json_tools import JsonType, json_extract
 from macaron.malware_analyzer.pypi_heuristics.base_analyzer import BaseHeuristicAnalyzer
 from macaron.malware_analyzer.pypi_heuristics.heuristics import HeuristicResult, Heuristics
 from macaron.slsa_analyzer.package_registry.pypi_registry import PyPIPackageJsonAsset
@@ -23,6 +25,8 @@ class WheelAbsenceAnalyzer(BaseHeuristicAnalyzer):
     """
 
     WHEEL: str = "bdist_wheel"
+    INSPECTOR_PREFIX = "https://inspector.pypi.io/project/"
+    PYPI_PREFIX = "https://files.pythonhosted.org/"
 
     def __init__(self) -> None:
         super().__init__(
@@ -72,7 +76,23 @@ class WheelAbsenceAnalyzer(BaseHeuristicAnalyzer):
                 if release_metadata["packagetype"] == self.WHEEL:
                     wheel_present = True
 
-                release_files.append(release_metadata["filename"])
+                name = json_extract(pypi_package_json.package_json, ["info", "name"], str)
+                if name is None:
+                    error_msg = "There is no 'name' field for this package."
+                    logger.debug(error_msg)
+                    raise HeuristicAnalyzerValueError(error_msg)
+
+                # include the pypi inspector link, which uses the same suffix of
+                # packages/{blake2b_256}/file_name
+                inspector_prefix = f"{self.INSPECTOR_PREFIX}{name.lower()}/{version}/"
+                inspector_link = release_metadata["url"].replace(self.PYPI_PREFIX, inspector_prefix)
+
+                if not self._valid_url(inspector_link, pypi_package_json.pypi_registry.request_timeout):
+                    inspector_link = ""
+
+                release_files.append(release_metadata["url"])
+                release_files.append(inspector_link)
+
         except KeyError as error:
             error_msg = f"The version {version} is not available as a release."
             logger.debug(error_msg)
@@ -82,3 +102,10 @@ class WheelAbsenceAnalyzer(BaseHeuristicAnalyzer):
             return HeuristicResult.PASS, {version: release_files}
 
         return HeuristicResult.FAIL, {version: release_files}
+
+    def _valid_url(self, url: str, timeout: int) -> bool:
+        try:
+            response = requests.head(url, allow_redirects=True, timeout=timeout)
+            return response.status_code == 200
+        except requests.exceptions.RequestException:
+            return False
--- a/src/macaron/malware_analyzer/pypi_heuristics/metadata/wheel_absence.py
+++ b/src/macaron/malware_analyzer/pypi_heuristics/metadata/wheel_absence.py
@@ -5,13 +5,12 @@
 
 import logging
 
-import requests
-
 from macaron.errors import HeuristicAnalyzerValueError
 from macaron.json_tools import JsonType, json_extract
 from macaron.malware_analyzer.pypi_heuristics.base_analyzer import BaseHeuristicAnalyzer
 from macaron.malware_analyzer.pypi_heuristics.heuristics import HeuristicResult, Heuristics
 from macaron.slsa_analyzer.package_registry.pypi_registry import PyPIPackageJsonAsset
+from macaron.util import send_head_http_raw
 
 logger: logging.Logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


art1f1c3R:
One change I have made I am unsure about is, in `wheel_absence.py` if `_valid_url` returns `False`, then currently I add an empty string. I am not sure what action should be taken if the constructed inspector link is not accessible.

src/macaron/malware_analyzer/pypi_heuristics/metadata/wheel_absence.py
```diff
@@ -72,7 +75,24 @@ def analyze(self, pypi_package_json: PyPIPackageJsonAsset) -> tuple[HeuristicRes
...
+                # include the pypi inspector link, which uses the same suffix of
+                # packages/{blake2b_256}/file_name
+                inspector_prefix = f"{self.INSPECTOR_PREFIX}{name.lower()}/{version}/"
+                inspector_link = release_metadata["url"].replace(self.PYPI_PREFIX, inspector_prefix)
+
+                # use a head request because we don't care about the response contents
+                if send_head_http_raw(inspector_link) is None:
+                    inspector_link = ""
+
+                release_files.append(release_metadata["url"])
```
behnazh-w, line 93: Can you please add separate fields in a `dict` for the inspector and file_server instead of adding all to a `list`?

Also, we could add a separate field for version instead of using the version as the key for the return value of this heuristic.

behnazh-w:
> One change I have made that I am unsure about is, in `wheel_absence.py:91` if `_valid_url` returns `False`, then currently I add an empty string. I am not sure what action should be taken if the constructed inspector link is not accessible.
> 
> The reason the check for a 200 OK response exists is due to the way I have constructed the pypi inspector link. I observe that the format of the pypi inspector link was the same as `files.pythonhosted.org` except replacing this prefix with the inspector URL, followed by the package name and then version. Everything else after was the same (using `blake2b_256` and the filename), only this was only an observation and I haven't managed to find this documented anywhere, so I added in that check.

I have added a suggestion in this [comment](https://github.com/oracle/macaron/pull/935#discussion_r1868681797) to improve the return value. If the inspector HEAD request fails, the value corresponding to the inspector could be `None`.

src/macaron/malware_analyzer/pypi_heuristics/metadata/wheel_absence.py
```diff
@@ -72,7 +75,24 @@ def analyze(self, pypi_package_json: PyPIPackageJsonAsset) -> tuple[HeuristicRes
...
+                name = json_extract(pypi_package_json.package_json, ["info", "name"], str)
+                if name is None:
+                    error_msg = "There is no 'name' field for this package."
+                    logger.debug(error_msg)
+                    raise HeuristicAnalyzerValueError(error_msg)
+
+                # include the pypi inspector link, which uses the same suffix of
+                # packages/{blake2b_256}/file_name
+                inspector_prefix = f"{self.INSPECTOR_PREFIX}{name.lower()}/{version}/"
+                inspector_link = release_metadata["url"].replace(self.PYPI_PREFIX, inspector_prefix)
```
behnazh-w, line 87: If I understand the code correctly, if a project doesn't publish the wheel or tarball, we won't send a HEAD request to check the inspector link, right?

art1f1c3R: Yes, since we iterate over the release data for that version, each item in that is for a separate way of distributing it (`.tar.gz`, `.whl`) so if one of them doesn't exist it won't exist to iterate over, so the HEAD request won't get sent.

src/macaron/malware_analyzer/pypi_heuristics/metadata/wheel_absence.py
art1f1c3R: I've done three fields, `version`, `file_server_links`, and `inspector_links`. I agree that makes much more sense, and allows me to resolve an invalid pypi inspector URL much easier.

src/macaron/malware_analyzer/pypi_heuristics/metadata/wheel_absence.py
```diff
@@ -64,21 +67,45 @@ def analyze(self, pypi_package_json: PyPIPackageJsonAsset) -> tuple[HeuristicRes
...
```
tromai, line 88: Although it may never happen but would it be better to ensure that we always replace `self.PYPI_PREFIX` as a prefix, but not as a substring (i.e in the middle of `release_metadata["url"]`)?

src/macaron/malware_analyzer/pypi_heuristics/metadata/wheel_absence.py
```diff
@@ -64,21 +67,45 @@ def analyze(self, pypi_package_json: PyPIPackageJsonAsset) -> tuple[HeuristicRes
...
+                inspector_links.append(inspector_link)
+
         except KeyError as error:
             error_msg = f"The version {version} is not available as a release."
             logger.debug(error_msg)
             raise HeuristicAnalyzerValueError(error_msg) from error
 
+        detail_info: dict[str, JsonType] = {
+            "version": version,
+            "file_server_links": file_server_links,
```
behnazh-w, line 104: I checked the results of the other heuristics. We already store the version and `file_server_links` as part of the JSON metadata returned by PyPI, so we can just keep `inspector_links`.

art1f1c3R:
Took a look at the source code for inspector, and I have confirmed the way it organises its URLs is what is used in this. At `https://github.com/pypi/inspector/blob/main/inspector/main.py` line 125:
```
@app.route(
    "/project/<project_name>/<version>/packages/<first>/<second>/<rest>/<distname>/"
)
```

src/macaron/malware_analyzer/pypi_heuristics/metadata/wheel_absence.py
art1f1c3R: I've refactored it instead so that it directly conforms to pypi inspector's source code instead of replacing strings now. See my comment below for this.

behnazh-w (approved):

src/macaron/malware_analyzer/pypi_heuristics/metadata/wheel_absence.py
```diff
@@ -64,21 +69,46 @@ def analyze(self, pypi_package_json: PyPIPackageJsonAsset) -> tuple[HeuristicRes
...
                     wheel_present = True
 
-                release_files.append(release_metadata["filename"])
+                name = json_extract(pypi_package_json.package_json, ["info", "name"], str)
+                if name is None:
+                    error_msg = "There is no 'name' field for this package."
+                    logger.debug(error_msg)
+                    raise HeuristicAnalyzerValueError(error_msg)
+
+                blake2b_256 = release_metadata["digests"]["blake2b_256"]
```
tromai, line 86: Should we use the `json_extract` method for extracting this value and validate it's a string that has more than 4 characters ?

art1f1c3R: This information will exist and be generated by pypi if we get to this line, so a scenario where the hash doesn't exist or is malformed should never happen if the release information is actually there (which is checked earlier). If for whatever reason this does crash, it would cause an exception and the result to be UNKNOWN, which is what we would throw for a custom exception anyway. I think this is safe to leave for now. 

`json_extract` doesn't really need to be used on this since it would need to start from the top (i.e. `["releases"]["version"]["digests"]["blake2b_256"]).

art1f1c3R: I've refactored it to use `json_extract` on all accesses and raise exceptions now, making the code a little cleaner and addressing your issue as well.

tromai (approved):
LGTM. Thanks.

behnazh-w (approved):

PR #939, opened 2024-12-09
fix: disable gpg signing for commit finder tests

### LINKED ISSUE #938: The test_commit_finder.py unit test hangs and terminate if we set gpg signing on tag
# Reproduce
Ensure that the user's global `$HOME/.gitconfig` has the following settings. Note that we have generated registered the GPG signing key using the instructions here - https://docs.github.com/en/authentication/managing-commit-signature-verification/generating-a-new-gpg-key and https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key
```
...
[user]
        signingkey = ...
        email = ...
        name = ...
[commit]
        gpgsign = true
[tag]
        gpgsign = true
[gpg]
        program = gpg
...
```
Prepare the development environment with
```
make setup
```
Then run the `test_commit_finder.py` unit test
```
pytest --no-cov tests/repo_finder/test_commit_finder.py
```

The unit test will hang for a very long time and terminate with the following error
```
client_loop: send disconnect: Broken pipe
```

If you remove these lines in the `.gitconfig` file, that unit test will work again
```
[tag]
        gpgsign = true
```

# Some thoughts
It could be that because we create new tags in this test case https://github.com/oracle/macaron/blob/5c490f9ef575eea4ac78508f6cc8dafdf8e2cb64/tests/repo_finder/test_commit_finder.py#L90

Those extra lines in the `.gitconfig` tell git to ask for gpg signing when a new tag is created - https://git-scm.com/docs/git-config#Documentation/git-config.txt-taggpgSign. Because it couldn't sign it, it timed out after a while and printed out the error message.

I think this is not a serious issue at the moment and it only happens because of a specific git config, but overall this should be resolved because it's not a good idea to make the unit test nondeterministic. I haven't thought of a solution for this yet.

benmss:
This PR prevents the commit finder tests from hanging if the user's Git is configured to require gpg signing for tags, as noted by #938. This is achieved by overriding the config value for the temporary repository created by the test.

Closes #938

### DIFF
--- a/src/macaron/database/table_definitions.py
+++ b/src/macaron/database/table_definitions.py
@@ -103,7 +103,7 @@ class PackageURLMixin:
     name: Mapped[str] = mapped_column(String(100), nullable=False, comment="Name of the package.")
 
     #: Version of the package.
-    version: Mapped[str] = mapped_column(String(100), nullable=True, comment="Version of the package.")
+    version: Mapped[str | None] = mapped_column(String(100), nullable=True, comment="Version of the package.")
 
     #: Extra qualifying data for a package such as the name of an OS.
     qualifiers: Mapped[str] = mapped_column(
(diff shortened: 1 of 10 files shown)


benmss:
See #940.

PR #944, opened 2024-12-13
feat: add in new metadata-based heuristic to pypi malware analyzer

art1f1c3R:
This PR adds in a new heuristic to the PyPI malware analyzer that focuses on identifying anomalistic version numbers. An anomalistic version number is classified as one that is unrealistically high for a single release. This heuristic depends on the single-release heuristic, and currently implements thresholds for determining suspicious version numbers on the epoch, major, and minor version components. Versioning numbers must adhere to the PyPI standards (PEP 440, as per the `packaging` module), otherwise they will not be analysed.

This heuristic attempts to identify the versioning pattern to reduce false-positives. Calendar versioning is defined as a versioning pattern where the major is in the form of YYYY or YY, the minor is in the form of MM or M, and the micro in the form of DD or D, with no other release components (i.e. in the form YYYY.MM.DD only). These values must correspond to the upload time with a 48-hour window for time differences to be classified as calendar versioning. 

Calendar-semantic versioning is defined as a versioning pattern where the major is in the form of YYYY or YY, but all other components are not detected as calendar versioning. All other versioning patterns are classed as semantic versioning.

Outstanding tasks for this PR:

- [x] run Macaron with this new feature to determine suitable default thresholds
- [x] make the thresholds configurable through `defaults.ini`

### DIFF
--- a/src/macaron/config/defaults.ini
+++ b/src/macaron/config/defaults.ini
@@ -584,12 +584,13 @@ include = *
 [heuristic.pypi]
 releases_frequency_threshold = 2
 # The gap threshold.
-# The timedelta indicate the gap between the date maintainer registers their pypi's account and the date of latest release.
+# The timedelta represents the gap between when the date maintainer registers their pypi account, and the
+# date of the latest release.
 timedelta_threshold_of_join_release = 5
 
-# Any major version above this value is detected as anomalistic and marked as suspicious
+# Any major version above this value is detected as anomalous and marked as suspicious.
 major_threshold = 20
-# Any epoch number avove this value is detected as anomalistic and marked as suspicious
+# Any epoch number avove this value is detected as anomalous and marked as suspicious.
 epoch_threshold = 3
-# The number of days +/- the day of publish the calendar versioning day may be
+# The number of days +/- the day of publish the calendar versioning day may be.
 day_publish_error = 4
--- a/src/macaron/config/defaults.ini
+++ b/src/macaron/config/defaults.ini
@@ -586,3 +586,10 @@ releases_frequency_threshold = 2
 # The gap threshold.
 # The timedelta indicate the gap between the date maintainer registers their pypi's account and the date of latest release.
 timedelta_threshold_of_join_release = 5
+
+# Any major version above this value is detected as anomalistic and marked as suspicious
+major_threshold = 20
+# Any epoch number avove this value is detected as anomalistic and marked as suspicious
+epoch_threshold = 3
+# The number of days +/- the day of publish the calendar versioning day may be
+day_publish_error = 4
--- a/src/macaron/malware_analyzer/pypi_heuristics/heuristics.py
+++ b/src/macaron/malware_analyzer/pypi_heuristics/heuristics.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2024 - 2025, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """Define the heuristic enum."""
@@ -33,6 +33,9 @@ class Heuristics(str, Enum):
 
     #: Indicates that the package does not include a .whl file
     WHEEL_ABSENCE = "wheel_absence"
+
+    #: Indicates that the package has an unusually large version number for a single release
+    ANOMALISTIC_VERSION = "anomalistic_version"
 
 
 class HeuristicResult(str, Enum):
--- a/src/macaron/malware_analyzer/pypi_heuristics/heuristics.py
+++ b/src/macaron/malware_analyzer/pypi_heuristics/heuristics.py
@@ -31,11 +31,11 @@ class Heuristics(str, Enum):
     #: Indicates that the setup.py file contains suspicious imports, such as base64 and requests.
     SUSPICIOUS_SETUP = "suspicious_setup"
 
-    #: Indicates that the package does not include a .whl file
+    #: Indicates that the package does not include a .whl file.
     WHEEL_ABSENCE = "wheel_absence"
 
-    #: Indicates that the package has an unusually large version number for a single release
-    ANOMALISTIC_VERSION = "anomalistic_version"
+    #: Indicates that the package has an unusually large version number for a single release.
+    ANOMALOUS_VERSION = "anomalous_version"
 
 
 class HeuristicResult(str, Enum):
--- a/src/macaron/malware_analyzer/pypi_heuristics/metadata/anomalistic_version.py
+++ b/src/macaron/malware_analyzer/pypi_heuristics/metadata/anomalistic_version.py
@@ -1,0 +1,223 @@
+# Copyright (c) 2024 - 2025, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""The heuristic analyzer to check for an anomalistic package version."""
+
+import logging
+from enum import Enum
+
+from packaging.version import InvalidVersion, parse
+
+from macaron.config.defaults import defaults
+from macaron.errors import HeuristicAnalyzerValueError
+from macaron.json_tools import JsonType, json_extract
+from macaron.malware_analyzer.datetime_parser import parse_datetime
+from macaron.malware_analyzer.pypi_heuristics.base_analyzer import BaseHeuristicAnalyzer
+from macaron.malware_analyzer.pypi_heuristics.heuristics import HeuristicResult, Heuristics
+from macaron.slsa_analyzer.package_registry.pypi_registry import PyPIPackageJsonAsset
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class AnomalisticVersionAnalyzer(BaseHeuristicAnalyzer):
+    """
+    Analyze the version number (if there is only a single release) to detect if it is anomalistic.
+
+    A version number is anomalistic if it is above the thresholds for an epoch, major, or minor value.
+    If the version does not adhere to PyPI standards (PEP 440, as per the 'packaging' module), this heuristic
+    cannot analyze it.
+
+    Calendar versioning is detected as version numbers with the year, month and day present in the following combinations:
+    (using the example 11th October 2016)
+    - YYYY.MM.DD, e.g. 2016.10.11
+    - YYYY.DD.MM, e.g. 2016.11.10
+    - YY.DD.MM, e.g. 16.11.10
+    - YY.MM.DD, e.g. 16.10.11
+    - MM.DD.YYYY, e.g. 10.11.2016
+    - DD.MM.YYYY, e.g. 11.10.2016
+    - DD.MM.YY, e.g. 11.10.16
+    - MM.DD.YY, e.g. 10.11.16
+    - YYYYMMDD, e.g. 20161011
+    - YYYYDDMM, e.g. 20161110
+    - YYDDMM, e.g. 161110
+    - YYMMDD, e.g. 161011
+    - MMDDYYYY, e.g. 10112016
+    - DDMMYYYY, e.g. 11102016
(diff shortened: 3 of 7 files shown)


tests/malware_analyzer/pypi/test_anomalistic_version.py
```diff
@@ -0,0 +1,563 @@
...
+    """Test for when there is no release information, so error"""
+    analyzer = AnomalisticVersionAnalyzer()
+
+    pypi_package_json.get_releases.return_value = None
+
+    with pytest.raises(HeuristicAnalyzerValueError):
+        analyzer.analyze(pypi_package_json)
+
+
+def test_invalid_version(pypi_package_json: MagicMock) -> None:
```
tromai, line 24: From my understanding, some of the unit tests in this module have similar workflows
- Version and Upload date as input
- Generate a release data dictionary
- Some mock patching
- Run the heuristic and compare against the expected value
Therefore, I think some of the unit tests can be refactored using pytest parametrize [feature](https://docs.pytest.org/en/stable/how-to/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions). 
This is an over-simplified example that I'm thinking of
```
@pytest.mark.parametrize(
    ("version", "upload_date", "expect"),
    [
        pytest.param(
            "boo",
            "foo",
            (HeuristicResult.SKIP, {"versioning": "invalid"}),
            id="Some description",
        ),
    ],
)
def test_something(version: str, upload_date: str, expect: tuple[HeuristicResult, dict[str,str]]) -> None:
...
```
However, it doesn't mean that all test functions in this module need to be converted to this format (e.g. the one where we expect it to raise exceptions) so you might need to decide how to restructure it.

art1f1c3R:
5537 existing PyPI packages with a single release were analyzed for their epoch, major, minor, and micro version numbers. From the results, all packages used an Epoch number of 0 (the implied epoch number as none supplied one). The data for major, minor, and micro version numbers included too many anomalies to plot nicely. From the data, the following changes will be made:

- Minor and Micro versions are not worth analyzing as they can contain many obscure values that are intended to be interpreted in a different way (e.g. `aggregateGithubCommits` with version 3.20200806 where the minor is actually a date). They are a less likely attack vector as to guarantee being prioritized over another package.
- The epoch threshold will be lowered to 3. The purpose of using an epoch is generally to ensure changing versioning systems prioritizes the new versioning system. It will still be included as it is a realistic attack vector, since most packages do not use an epoch at all.
- The calendar versioning will be updated to account for different ordering (e.g. YYYY.MM.DD or YYYY.MM.DD or DD.MM.YYYY), with a motivating example of `abbrfix` with version 26.2.2024 published on February 27 2024
- The threshold on the number of days of error allowed for calendar versioning will be increased to 4. A motivating example is `aiocmdline` version 2024.1.25 released on January 28 2024.
- The major version number may be checked to represent a date itself (e.g. `andronnlib` version 160122 published Jan 18 2022)
- The restriction for calendar versioning to only have 3 release components will be removed. A motivating example is `aigc` with version 2023.2.15.14.18.59 published February 15 2023. Checking the implied major version from this (14) is unnecessary as the date would be prioritized over this.
- The major version threshold of 20 will be kept. Many version numbers around this range are often calendar-serial or calendar versioning for the year (e.g. `adwords-client` version 17.7 published July 13 2017), so any major version number not corresponding to the calendar year here and above this threshold is sufficiently large to be considered suspicious.

There were some anomalies to these rules (e.g. `almavik` version 362 published June 24 2023, or `abqpy2016` to `abqpy2025` all versions 20xx.7.9, all published Dec 8 2024), but not enough to warrant relaxing the threshold.

art1f1c3R:
I am going to request review on this as I don't believe an integration test may be appropriate for this feature, since there is no guarantee that any given package may be assumed to always have one release, so if I use a package for the test if it releases a new version the test no longer makes sense.

src/macaron/malware_analyzer/pypi_heuristics/metadata/anomalistic_version.py
```diff
@@ -0,0 +1,202 @@
...
+        publish_days = []
+
+        for distribution in release_metadata:
+            upload_time = json_extract(distribution, ["upload_time"], str)
+            if upload_time is None:
+                error_msg = "Missing upload time from release information"
+                logger.debug(error_msg)
+                raise HeuristicAnalyzerValueError(error_msg)
+
+            parsed_time = parse_datetime(upload_time)
```
tromai, line 125: I wonder if we document the format of `upload_time` anywhere? Or are we assuming that `upload_time` has the same default format for `parse_datetime`, which is `%Y-%m-%dT%H:%M:%S`

src/macaron/malware_analyzer/pypi_heuristics/metadata/anomalistic_version.py
```diff
@@ -0,0 +1,202 @@
...
+
+class AnomalisticVersionAnalyzer(BaseHeuristicAnalyzer):
+    """
+    Analyze the version number (if there is only a single release) to detect if it is anomalistic.
+
+    A version number is anomalistic if it is above the thresholds for an epoch, major, or minor value.
+    If the version does not adhere to PyPI standards (PEP 440, as per the 'packaging' module), this heuristic
+    cannot analyze it.
+
+    Calendar versioning is detected as version numbers with the major value as the year (either yyyy or yy),
```
tromai, line 30: nit: it would be good if we have an example for each type
- calendar versioning
- calendar-semantic versioning

src/macaron/malware_analyzer/pypi_heuristics/metadata/anomalistic_version.py
```diff
@@ -0,0 +1,202 @@
...
+        except InvalidVersion:
+            return HeuristicResult.SKIP, {self.DETAIL_INFO_KEY: Versioning.INVALID.value}
+
+        years = []
+        months = []
+        publish_days = []
+
+        for distribution in release_metadata:
+            upload_time = json_extract(distribution, ["upload_time"], str)
+            if upload_time is None:
```
tromai, line 120: I wonder if it's possible to have unit tests for cases where `upload_time` is missing and `upload_time` is not in the PYPI date time format?

src/macaron/malware_analyzer/pypi_heuristics/metadata/anomalistic_version.py
art1f1c3R: The format of the release JSON is always `%Y-%m-%dT%H:%M:%S`, which in some other heuristics is currently hardcoded (e.g. in `closer_release_join_date.py`). I just noticed that it was the default value for the `datetime_format` argument, so I omitted the argument.

src/macaron/config/defaults.ini
```diff
@@ -586,3 +586,10 @@ releases_frequency_threshold = 2
 # The gap threshold.
 # The timedelta indicate the gap between the date maintainer registers their pypi's account and the date of latest release.
 timedelta_threshold_of_join_release = 5
+
+# Any major version above this value is detected as anomalistic and marked as suspicious
+major_threshold = 20
+# Any epoch number avove this value is detected as anomalistic and marked as suspicious
```
benmss, line 592:
```suggestion
# Any epoch number above this value is detected as anomalous and marked as suspicious.
```

src/macaron/config/defaults.ini
```diff
@@ -586,3 +586,10 @@ releases_frequency_threshold = 2
 # The gap threshold.
 # The timedelta indicate the gap between the date maintainer registers their pypi's account and the date of latest release.
 timedelta_threshold_of_join_release = 5
+
+# Any major version above this value is detected as anomalistic and marked as suspicious
```
benmss, line 590:
```suggestion
# Any major version above this value is detected as anomalous, and marked as suspicious.
```

src/macaron/config/defaults.ini
```diff
@@ -586,3 +586,10 @@ releases_frequency_threshold = 2
 # The gap threshold.
 # The timedelta indicate the gap between the date maintainer registers their pypi's account and the date of latest release.
```
benmss, line 587:
```suggestion
# The timedelta represents the gap between when the date maintainer registers their pypi account, and the date of the latest release.
```

src/macaron/config/defaults.ini
```diff
@@ -586,3 +586,10 @@ releases_frequency_threshold = 2
...
 timedelta_threshold_of_join_release = 5
+
+# Any major version above this value is detected as anomalistic and marked as suspicious
+major_threshold = 20
+# Any epoch number avove this value is detected as anomalistic and marked as suspicious
+epoch_threshold = 3
+# The number of days +/- the day of publish the calendar versioning day may be
```
benmss, line 594:
```suggestion
# The number of days +/- the day of publish the calendar versioning day may be.
```

src/macaron/malware_analyzer/pypi_heuristics/heuristics.py
```diff
@@ -34,6 +34,9 @@ class Heuristics(str, Enum):
     #: Indicates that the package does not include a .whl file
```
benmss, line 34:
```suggestion
    #: Indicates that the package does not include a .whl file.
```

src/macaron/malware_analyzer/pypi_heuristics/heuristics.py
```diff
@@ -34,6 +34,9 @@ class Heuristics(str, Enum):
...
     WHEEL_ABSENCE = "wheel_absence"
 
+    #: Indicates that the package has an unusually large version number for a single release
```
benmss, line 37:
```suggestion
    #: Indicates that the package has an unusually large version number for a single release.
```

src/macaron/malware_analyzer/pypi_heuristics/metadata/anomalistic_version.py
```diff
@@ -0,0 +1,223 @@
...
+from macaron.malware_analyzer.pypi_heuristics.base_analyzer import BaseHeuristicAnalyzer
+from macaron.malware_analyzer.pypi_heuristics.heuristics import HeuristicResult, Heuristics
+from macaron.slsa_analyzer.package_registry.pypi_registry import PyPIPackageJsonAsset
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class AnomalisticVersionAnalyzer(BaseHeuristicAnalyzer):
+    """
+    Analyze the version number (if there is only a single release) to detect if it is anomalistic.
```
benmss, line 24:
```suggestion
    Analyze the version number (if there is only a single release) to detect if it is anomalous.
```

src/macaron/malware_analyzer/pypi_heuristics/metadata/anomalistic_version.py
```diff
@@ -0,0 +1,223 @@
...
+
+    A version number is anomalistic if it is above the thresholds for an epoch, major, or minor value.
```
benmss, line 26:
```suggestion
    A version number is anomalous if any of its values are greater than the epoch, major, or minor threshold values.
```

src/macaron/malware_analyzer/pypi_heuristics/metadata/anomalous_version.py
```diff
@@ -0,0 +1,223 @@
...
+        detail_info = {self.DETAIL_INFO_KEY: Versioning.SEMANTIC.value}
+
+        if version.epoch > self.epoch_threshold:
+            return HeuristicResult.FAIL, detail_info
+        if version.major > self.major_threshold:
+            return HeuristicResult.FAIL, detail_info
+
+        return HeuristicResult.PASS, detail_info
+
+    def __integer_date(self, value: int, years: list[int], months: list[int], days: list[int]) -> bool:
```
benmss, line 206: Does this function need name mangling (double underscore prefix)? Also a short comment description would be nice.

art1f1c3R: I can add in a docstring. The purpose of the double underscore prefix was to indicate it was a private function, but I just checked other examples and it seems a single underscore prefix is what we use, so I'll change it to that.

tromai (approved):
LGTM. Thanks!

benmss (approved):

PR #949, opened 2024-12-19
fix: update already present repositories

### LINKED ISSUE #948: Macaron does not update existing repository directories
During analysis, if the directory a repository should be cloned into already exists, Macaron will not attempt to update the cloned contents of that repository, therefore making it unable to analyse any newer changes made since the previous clone took place.

benmss:
This fixes the issue with previously cloned repositories not updating when newer changes are present at their remote origins, by using `git fetch` when the target directory of the clone is already present. This comes with a small performance hit for any analysis performed on a previously cloned target.

Closes #948

### DIFF
--- a/src/macaron/slsa_analyzer/git_url.py
+++ b/src/macaron/slsa_analyzer/git_url.py
@@ -124,10 +124,6 @@ def check_out_repo_target(
 
     This function assumes that a remote "origin" exist and checkout from that remote ONLY.
 
-    If ``offline_mode`` is False, this function will fetch new changes from origin remote. The fetching operation
-    will prune and update all references (e.g. tags, branches) to make sure that the local repository is up-to-date
-    with the repository specified by origin remote.
-
     If ``offline_mode`` is True and neither ``branch_name`` nor commit are provided, this function will not do anything
     and the HEAD commit will be analyzed. If there are uncommitted local changes, the HEAD commit will
     appear in the report but the repo with local changes will be analyzed. We leave it up to the user to decide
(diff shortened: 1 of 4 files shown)


src/macaron/slsa_analyzer/git_url.py
```diff
@@ -337,11 +337,27 @@ def clone_remote_repo(clone_dir: str, url: str) -> Repo | None:
             os.rmdir(clone_dir)
             logger.debug("The clone dir %s is empty. It has been deleted for cloning the repo.", clone_dir)
         except OSError:
-            logger.debug(
-                "The clone dir %s is not empty. Cloning will not be proceeded.",
-                clone_dir,
-            )
-            return None
+            # If the repository has already been cloned, we attempt to update using `git pull`.
```
tromai, line 340: There is an existing logic to fetch the latest changes from origin remote. You can found it [here](https://github.com/oracle/macaron/blob/b65f0dba4826d32fe79adc1694a7d0f57451654c/src/macaron/slsa_analyzer/git_url.py#L169-L197).
The usual workflow is that after this function is run, the [`checkout_repo_target`](https://github.com/oracle/macaron/blob/b65f0dba4826d32fe79adc1694a7d0f57451654c/src/macaron/slsa_analyzer/git_url.py#L117) function will be run. That `checkout_repo_target` will attempt to fetch latest changes, and make sure our local repository is synced with the origin remote (e.g tags, branches, etc.). 

I wonder for which project did this bug happened?

benmss: Yeah I see that. The problem was that the `checkout_repo_target` function was being called after the Commit Finder, so in cases where a new tag has been added, the fetch wouldn't happen. I've moved the fetch operation now, replacing the pull I added previously.

benmss: This will happen for any project that has added new tags when compared to a previously cloned local version.

src/macaron/slsa_analyzer/git_url.py
```diff
@@ -166,36 +166,13 @@ def check_out_repo_target(
     bool
         True if succeed else False.
     """
-    if not offline_mode:
```
tromai, line 169: Could you please also update the doc strings for `check_out_repo_target` and `clone_remote_repo` to reflect the changes?

tromai (approved):

behnazh-w (approved):

PR #956, opened 2025-01-05
fix: report known malware even when not labeled

behnazh-w:
Some of the malicious packages are not clearly labeled in the OSV knowledge base. This PR improves the known malware detection in `detect_malicious_metadata_check`. It also adds a new abstraction for the deps.dev service to improve code reusability.

### DIFF
--- a/src/macaron/config/defaults.ini
+++ b/src/macaron/config/defaults.ini
@@ -543,7 +543,7 @@ inspector_url_scheme = https
 [deps_dev]
 url_netloc = api.deps.dev
 url_scheme = https
-v3alpha_purl_endpoint = v3alpha/purl
+purl_endpoint = v3alpha/purl
 
 # Configuration options for selecting the checks to run.
 # Both the exclude and include are defined as list of strings:
--- a/src/macaron/config/defaults.ini
+++ b/src/macaron/config/defaults.ini
@@ -540,6 +540,11 @@ fileserver_url_scheme = https
 inspector_url_netloc = inspector.pypi.io
 inspector_url_scheme = https
 
+[deps_dev]
+url_netloc = api.deps.dev
+url_scheme = https
+v3alpha_purl_endpoint = v3alpha/purl
+
 # Configuration options for selecting the checks to run.
 # Both the exclude and include are defined as list of strings:
 #   - The exclude list is used to specify the checks that will not run.
--- a/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
+++ b/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
@@ -11,7 +11,7 @@ from sqlalchemy.orm import Mapped, mapped_column
 
 from macaron.database.db_custom_types import DBJsonDict
 from macaron.database.table_definitions import CheckFacts
-from macaron.errors import HeuristicAnalyzerValueError
+from macaron.errors import HeuristicAnalyzerValueError, InvalidHTTPResponseError
 from macaron.json_tools import JsonType, json_extract
 from macaron.malware_analyzer.pypi_heuristics.base_analyzer import BaseHeuristicAnalyzer
 from macaron.malware_analyzer.pypi_heuristics.heuristics import HeuristicResult, Heuristics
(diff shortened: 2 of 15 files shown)


src/macaron/slsa_analyzer/package_registry/deps_dev.py
```diff
@@ -0,0 +1,82 @@
...
+            If a network error happens or unexpected response is returned by the API.
+        """
+        section_name = "deps_dev"
+        if not defaults.has_section(section_name):
+            return None
+        section = defaults[section_name]
+
+        url_netloc = section.get("url_netloc")
+        if not url_netloc:
+            raise ConfigurationError(
```
tromai, line 48: This parts raises `ConfigurationError` but the doc string of this method doesn't include this as a possibly raised exception.

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -259,21 +260,29 @@ def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
...
-        res_obj = None
-        if response:
-            try:
-                res_obj = response.json()
-            except requests.exceptions.JSONDecodeError as error:
-                logger.debug("Unable to get a valid response from %s: %s", url, error)
-        if res_obj:
-            for vuln in res_obj.get("vulns", {}):
-                v_id = json_extract(vuln, ["id"], str)
-                if v_id and v_id.startswith("MAL-"):
```
benmss, line 276: I think this conditional check should still exist.

src/macaron/config/defaults.ini
```diff
@@ -538,6 +538,11 @@ registry_url_scheme = https
 fileserver_url_netloc = files.pythonhosted.org
 fileserver_url_scheme = https
 
+[deps_dev]
+url_netloc = api.deps.dev
+url_scheme = https
+v3alpha_purl_endpoint = v3alpha/purl
```
benmss, line 544: The name of this config option should probably be `purl_endpoint`.

src/macaron/slsa_analyzer/package_registry/deps_dev.py
```diff
@@ -0,0 +1,82 @@
...
+                urllib.parse.SplitResult(
+                    scheme=url_scheme,
+                    netloc=url_netloc,
+                    path=path_params,
+                    query="",
+                    fragment="",
+                )
+            )
+        except ValueError as error:
+            raise InvalidHTTPResponseError("Failed to construct the API URL.") from error
```
art1f1c3R, line 70: Should this be an `InvalidPURLError` or `PURLNotFoundError` or some other URL-related error? The http response hasn't actually been sent yet at this stage, this is just constructing the link.

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
behnazh-w: That's actually the reason I have created this PR. Sometimes the malware are not labeled with `MAL-`. For example, the malicious versions of `ultralytics` are labeled with `PYSEC`: https://osv.dev/vulnerability/PYSEC-2024-154. So this check is not reliable.

src/macaron/slsa_analyzer/package_registry/deps_dev.py
behnazh-w: To address this concern with the error name, I have created a more generic error, `APIAccessError` to account for different cases. Raising different errors in a function in Python increases the risk of missing exception handling and I would like to avoid that.

See commit [20155c3](https://github.com/oracle/macaron/pull/956/commits/20155c37d9d4709f64a7a26eb2ca1e231f41feaa).

src/macaron/slsa_analyzer/package_registry/deps_dev.py
behnazh-w: I have created a more generic error, `APIAccessError` to account for different cases. 

See commit [20155c3](https://github.com/oracle/macaron/pull/956/commits/20155c37d9d4709f64a7a26eb2ca1e231f41feaa).

tromai (approved):
LGTM. Thanks!

art1f1c3R (approved):

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
benmss: The first part of the check still seems to be applicable, otherwise `None` can end up in the produced URL. If we are certain that can never happen due to the nature of the data, a comment to that effect would be nice.

behnazh-w: Ah, I had misunderstood which check you were talking about. Sure, fixed in commit [60d5ec2](https://github.com/oracle/macaron/pull/956/commits/60d5ec2ecb004c39b935c5104a5a11485238716e).

Issue #962, opened 2025-01-17
Souffle dependency not available in Ubuntu 24.04


benmss:
`Souffle 2.4.1` has `libffi7 (>= 3.3~20180313)` as a dependency, which is no longer available in the GitHub runners using `ubuntu-latest` after the [recent upgrade to Ubuntu 24.04](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md). Until Souffle updates their dependency, an alternative solution must be found or else Macaron will fail to pass most of the pipeline checks.

benmss:
The Souffle team has released a new version, 2.5.0, that fixes the above issue. According to a comment from one of their contributors, this new version may not make it to the Ubuntu package depot, leaving manual installation as our best option to used the fixed release for the foreseeable future.

PR #963, opened 2025-01-17
fix: set problem workflows' OS to ubuntu 22.04

benmss:
This PR sets the version of ubuntu used by the workflows to `22.04` instead of `latest`, as that has been updated to `24.04` and does not provide the required `libffi7` dependency.
See issue #962 and alternative solution PR #964


benmss:
The Souffle team will release an update that fixes this issue.

PR #964, opened 2025-01-17
build: install libffi7 manually

benmss:
An alternative solution to the problem with Souffle and Ubuntu 24.04: Installing the `libffi7` library manually. See #962

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,6 +145,8 @@ else
 endif
 # If Souffle cannot be installed, we advise the user to install it manually
 # and return status code 0, which is not considered a failure.
+# Souffle depends upon the libffiX library, where X is the current version it requires. Depending on the version of Ubuntu being used, the exact library
+# may not be present. In this script, we manually download and install version 7 on Ubuntu 24.04.
 .PHONY: souffle
 souffle:
 	if ! command -v souffle; then \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -158,6 +158,9 @@ souffle:
 	      sudo wget https://souffle-lang.github.io/ppa/souffle-key.public -O /usr/share/keyrings/souffle-archive-keyring.gpg; \
 	      echo "deb [signed-by=/usr/share/keyrings/souffle-archive-keyring.gpg] https://souffle-lang.github.io/ppa/ubuntu/ stable main" | sudo tee /etc/apt/sources.list.d/souffle.list; \
 	      sudo apt update; \
+	      sudo wget http://archive.ubuntu.com/ubuntu/pool/main/libf/libffi/libffi7_3.3-4_amd64.deb; \
```
behnazh-w, line 161: Can you please add a comment to clarify why installing this dependency like this is needed?

behnazh-w (dismissed):

Makefile
```diff
@@ -145,6 +145,8 @@ else
 endif
 # If Souffle cannot be installed, we advise the user to install it manually
 # and return status code 0, which is not considered a failure.
+# Souffle depends upon the libffiX library, where X is the current version it requires. Depending on the version of Ubuntu being used, the exact library
+# may not be present. In this script, we manually download and install version 7 on Ubuntu 24.04.
```
behnazh-w, line 149: The change in this PR doesn't seem to check the Ubuntu version.
```suggestion
# may not be present. In this script, we manually download and install version 7 on Ubuntu operating systems.
```

behnazh-w (approved):

benmss:
The Souffle team will release an update that fixes this issue.

benmss:
This will be merged as a temporary fix -- to be reverted once a new version of Souffle is released.

PR #851, opened 2024-09-05
chore: implement method to validate suspicious packages for malicious…

Yao-Wen-Chang:
This PR refers to issue #810.

This PR implements the validator to confirm malware on PyPI. We analyze the data flow by walking through the AST and finding out the actual value of the variable.

For example:
```python
requests.get(malicious_endpoint)
malicious_endpoint = "https://malicious.com"
```
The new method should be able to detect the `https://malicious.com`. Furthermore, we will analyze the historical malware data to define the suspicious pattern as a .yaml.

The suspicious_setup heuristic will be removed since it overlaps with our new method.

Following are the tasks for implementing this method:

- [x] Analyze the historical malware and define the malicious pattern in .yaml
- [x] Implement the method to analyze the data flow
- [ ] Provide unit tests
- [ ] Test the method on the latest packages on PyPI to ensure the detector is more accurate

PR #965, opened 2025-01-17
refactor: improve experimental source code pattern analysis of pypi packages

art1f1c3R:
This pull request builds upon the work completed in #851, refactoring and refining the first step of source code analysis, matching suspicious code patterns. The intended behavior of this functionality is for the user to be able to specify pattern categories and subcategories (but no deeper) in a `.yaml` file, where this file path is provided in `defaults.ini`; a default file exists. These patterns will then be used to scan all python source code files for suspicious behavior. This suspicious pattern scanning will be the first phase of a broader feature for full source code analysis that includes dataflow analysis.


src/macaron/config/defaults.ini
```diff
@@ -596,5 +596,5 @@ epoch_threshold = 3
 day_publish_error = 4
 
 # yaml configuration file containing suspicious patterns. Can be full path or relative to
-# folder where macaron is installed
+# folder where macaron is installed. This will be normalised to the OS path type.
```
behnazh-w, line 599: We can move this pattern file to a folder under the [resources](https://github.com/oracle/macaron/tree/staging/src/macaron/resources), and access them through code by default, i.e., the file path is not read from `defaults.ini`. Alternatively, the user can provide a path to a file that contains pattern. What do you think?

art1f1c3R: That sounds like a good way to do it. I am hoping to provide the user-configurable part in a way that means the rules are the union of the default and user-supplied patterns. The user can then provide an argument or boolean in `defaults.ini` to disable the default rules? I'm currently using the similar `.yaml` patterns we had from the previous work, but am debating whether the user-supplied rules are just `semgrep` rules, rather than opening the patterns file and writing `semgrep` rules programmatically from that?

.semgrepignore
art1f1c3R, line 1: This file has been added to ensure `semgrep` does not ignore any files during its scan, so packages can not bypass this analyzer by naming malicious files in the same manner that is ignored by the default `.semgrepignore`. See the development notes in the PR description for a more in-depth description.

scripts/dev_scripts/samples_permissions_checker.sh
art1f1c3R, line 1: This has been run through Shellcheck, no issues are detected.

.pre-commit-config.yaml
```diff
@@ -197,6 +203,18 @@ repos:
     always_run: true
     pass_filenames: false
 
+# Checks that tests/malware_analyzer/pypi/resources/sourcecode_samples files do not have executable permissions
```
tromai, line 206: Would this hook run for all files within `tests/malware_analyzer/pypi/resources/sourcecode_samples` even though they are not tracked by git ?
For example
```
touch tests/malware_analyzer/pypi/resources/sourcecode_samples/obfuscation/test.py
chmod +x tests/malware_analyzer/pypi/resources/sourcecode_samples/obfuscation/test.py
pre-commit run --all-files sourcecode-sample-permissions # Would this raise any error about test.py?
```

art1f1c3R: I just tried running this and yes it does detect all files, even those not tracked by git. Is this an issue or should I change it to only run on git-tracked files?

tromai: It could be an issue when we are testing with multiple files in that directory, then create a commit with files that are elsewhere, the hook would fail and stop us from committing even though we are not checking in wrong files.
I don't see this problem will pop up that often, but I think it's better if we only check what is necessary (similar to how our copyright checking hook works).

art1f1c3R: At the very least, I can see having the hook consider untracked files an inconvenience when testing, and there would be no risk to not checking untracked files. I've updated it to only check files that are added and scheduled to be updated.

tromai: Thanks

art1f1c3R:
Semgrep rules have been refined and updated based on the results of scanning the ICSE25 dataset as was provided to Hercule. A summary of what was observed and what was changed as a result:
- The `obfuscation_default-assigning` rule triggered many false positives, as packages often perform these operations for custom typing purposes, as well as other unknown but benign behaviors. The rule did not, by itself, detect many malware within the dataset, and was often detected in conjunction with some other rule. For this reason, it has been removed.
- The rules `obfuscation_decode-and-execute` and `exfiltration_remote-exfiltration` have been rewritten to make them more context aware. What is meant by this is, many false positives were being generated for patterns such as `$BYTES.decode(...)`, as since `$BYTES` is not type checked to actually be a bytes object, anything with a `.decode(...)` method was triggered. This had a heavy impact on the socket patterns and file read/write patterns. These have been rewritten to ensure that any file read/write, or any socket action, is taken when a relevant object is available and detected. This has significantly reduced false positives.
- The `ast.literal_eval()` function has been removed, as it does not, by itself, offer any real malicious functionality (only parsing types and other strings, not executing code). 
- The `obfuscation_inline-imports` rule has been rewritten to detect hexadecimal and octal import after observing malicious behaviors similar in some of the malware datasets, as well as only detecting certain, hardcoded suspicious imports. This is to avoid the many false positives it triggered for trusted packages which use `__import__(...)` for configuration purposes, and use this function as opposed to the suggested `importlib.import_module(...)`.

scripts/dev_scripts/samples_permissions_checker.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+# Copyright (c) 2022 - 2025, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+#
+# Checks if the files in tests/malware_analyzer/pypi/resources/sourcecode_samples have executable permissions,
+# failing if any do.
+#
+
```
tromai, line 10: Would it be good to add `set -euo pipefail` in this script?
What we really want to prevent is commands failing silently, leading to an incorrect `SAMPLES_PATH` value, which will make the result of this script wrong.

art1f1c3R: Yes I agree. I'll do that now.

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/pypi_sourcecode_analyzer.py
```diff
@@ -0,0 +1,206 @@
...
+            Defaults to None
+
+        Raises
+        ------
+        ConfigurationError
+            If the default rule path is invalid, the heuristic.pypi entry is not present, or if the semgrep
+            validation of the custom rule path failed.
+        """
+        super().__init__(
+            name="anomalous_version_analyzer",
```
behnazh-w, line 54:
```suggestion
            name="suspicious_patterns_analyzer",
```

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
behnazh-w, line 1: I have noticed an increase in the number of FPs with the Problog rules. Here is a modified version that seems to improve the precision:
```
    malware_rules_problog_model = f"""
    % Heuristic groupings
    % These are common combinations of heuristics that are used in many of the rules, thus themselves representing
    % certain behaviors. When changing or adding rules here, if there are frequent combinations of particular
    % heuristics, group them together here.

    % Maintainer has recently joined, publishing an undetailed page with no links.
    quickUndetailed :- not {Heuristics.EMPTY_PROJECT_LINK.value}, not {Heuristics.CLOSER_RELEASE_JOIN_DATE.value}.

    % Maintainer releases a suspicious setup.py and forces it to run by omitting a .whl file.
    forceSetup :- not {Heuristics.SUSPICIOUS_SETUP.value}, not {Heuristics.WHEEL_ABSENCE.value}.

    % Suspicious Combinations

    % Package released recently with little detail, forcing the setup.py to run.
    {Confidence.HIGH.value}::high :- quickUndetailed, forceSetup, not {Heuristics.ONE_RELEASE.value}.
    {Confidence.HIGH.value}::high :- quickUndetailed, forceSetup, not {Heuristics.HIGH_RELEASE_FREQUENCY.value}.

    % Package released recently with little detail, with some more refined trust markers introduced: project links,
    % multiple different releases, but there is no source code repository matching it and the setup is suspicious.
    {Confidence.HIGH.value}::high :- not {Heuristics.SOURCE_CODE_REPO.value},
        not {Heuristics.HIGH_RELEASE_FREQUENCY.value},
        not {Heuristics.CLOSER_RELEASE_JOIN_DATE.value},
    %    not {Heuristics.UNCHANGED_RELEASE.value},
        forceSetup.

   % Package released recently with little detail and an anomalous version number for a single-release package.
   {Confidence.MEDIUM.value}::medium :- quickUndetailed,
        not {Heuristics.ONE_RELEASE.value},
        not {Heuristics.ANOMALOUS_VERSION.value}.

    {problog_result_access} :- high.
    {problog_result_access} :- medium.

    query({problog_result_access}).
    """

We should also check if the skipped heuristics indeed disable a Problog rule.

**Review by behnazh-w — COMMENTED** [review]
Since the source code analysis has become more mature, we can now run it by default.

File: src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
Comment (art1f1c3R) [reply src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py:1]: Looking into this, see opened issue #1027 here.

File: src/macaron/malware_analyzer/pypi_heuristics/sourcecode/pypi_sourcecode_analyzer.py [hunk 5]
```diff
@@ -0,0 +1,206 @@
...
+        Raises
+        ------
+        ConfigurationError
+            If the default rule path is invalid, the heuristic.pypi entry is not present, or if the semgrep
+            validation of the custom rule path failed.
+        """
+        super().__init__(
+            name="suspicious_patterns_analyzer",
+            heuristic=Heuristics.SUSPICIOUS_PATTERNS,
+            depends_on=[(Heuristics.EMPTY_PROJECT_LINK, HeuristicResult.FAIL)],
```
behnazh-w, line 56:
```suggestion
            depends_on=[(Heuristics.SOURCE_CODE_REPO, HeuristicResult.FAIL)],
```

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
art1f1c3R: This issue was closed from #1031 and should now be resolved.

tests/malware_analyzer/pypi/test_pypi_sourcecode_analyzer.py
```diff
@@ -114,3 +144,116 @@ def test_rules(
...
+
+    mock_defaults.has_section.side_effect = lambda section: section == "heuristic.pypi"
+    mock_defaults.__getitem__.side_effect = lambda section: sub_section if section == "heuristic.pypi" else None
+
+    analyzer = PyPISourcecodeAnalyzer(resources_path=RESOURCES_PATH)
+    pypi_package_json.package_sourcecode_path = sample_path
+
+    result, analysis = analyzer.analyze(pypi_package_json)
+
+    print(analysis)
```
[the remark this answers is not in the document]
art1f1c3R: removed them [here](https://github.com/oracle/macaron/pull/965/commits/8dad8117801948e35850fd05fb9e86a6644e8d97)

behnazh-w (dismissed):

src/macaron/config/defaults.ini
```diff
@@ -600,3 +600,17 @@ major_threshold = 20
 epoch_threshold = 3
 # The number of days +/- the day of publish the calendar versioning day may be.
 day_publish_error = 4
+
+# disable default semgrep rulesets here (i.e. all rule IDs in a Semgrep .yaml file) using comma separated ruleset names. Use the name
+# without the .yaml prefix here. Currently, we disable the exfiltration rulesets by default due to a high false positive rate.
+disabled_default_rulesets = exfiltration
```
tromai, line 606: Could you please walk me through the differences between the default rules and the individual rules ? Are they from two separate set, and doesn't have any interaction with each other ?

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/pypi_sourcecode_analyzer.py
```diff
@@ -0,0 +1,319 @@
...
+                raise ConfigurationError(error_msg)
+
+            logger.debug("Including custom ruleset from %s.", custom_rule_path)
+
+        disabled_rule_ids = set()
+
+        configuration_name = "disabled_default_rulesets"
+        disabled_default_rulesets_string = section.get(configuration_name)
+        if disabled_default_rulesets_string:
+            target_files = {f"{name.strip()}.yaml" for name in disabled_default_rulesets_string.split(",") if name}
```
tromai, line 143: We have an utility function to extract list of elements from a default section, see here - https://github.com/oracle/macaron/blob/aebf8791630f5c3c716d8483585b22697f001f8b/src/macaron/config/defaults.py#L18 . I wonder if we could use that instead ? 
Also, I wonder if there is a specific reason why we decided to go with comma delimiter instead of new lines like other places in `defaults.ini` 🤔 . For example:
- https://github.com/oracle/macaron/blob/4b20c182068537b7305e9350897bacffb3b7d7bc/src/macaron/config/defaults.ini#L206

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/pypi_sourcecode_analyzer.py
```diff
@@ -0,0 +1,319 @@
...
+            A set of unique Semgrep rule IDs extracted from all provided target files.
+
+        Raises
+        ------
+        ConfigurationError
+            If any Semgrep rule file could not be safely loaded, or if their format was not in the expected Semgrep
+            format, or if there were any files in 'target_files' not found when searching in 'path'.
+        """
+        rule_ids = set()
+        for root, _, files in os.walk(path):
```
tromai, line 191: May be we could use glob.glob to find the files in the nested directories first, then raise and error if not all of `target_files` cannot be found (so we can skip loading the yaml files if not necessary) ?

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/pypi_sourcecode_analyzer.py
```diff
@@ -0,0 +1,319 @@
...
+        Raises
+        ------
+        HeuristicAnalyzerValueError
+            if there is no source code available.
+        """
+        analysis_result: dict = {}
+        disabled_results: dict = (
+            {}
+        )  # since we have to run them anyway, return disabled rule findings for debug information
+        # only run semgrep open-source features, and disable 'nosemgrep' ignoring so this does not bypass our scan
```
tromai, line 246: I'm a bit lost on the `disable nosemgrep ignoring`. So by default, what does it ignore ?

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/pypi_sourcecode_analyzer.py
```diff
@@ -0,0 +1,319 @@
...
+        semgrep_commands: list[str] = ["semgrep", "scan", "--oss-only", "--disable-nosem"]
+        result: HeuristicResult = HeuristicResult.PASS
+
+        source_code_path = pypi_package_json.package_sourcecode_path
+        if not source_code_path:
+            error_msg = "Unable to retrieve PyPI package source code path"
+            logger.debug(error_msg)
+            raise HeuristicAnalyzerValueError(error_msg)
+
+        semgrep_commands.extend(["--config", self.default_rule_path])
```
tromai, line 256: Just to double check, `semgrep` allows you specify `--config` multiple times, and it will merge them all into a big "configuration"?

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/pypi_sourcecode_analyzer.py
```diff
@@ -0,0 +1,319 @@
...
+                )
+                logger.debug(error_msg)
+                raise HeuristicAnalyzerValueError(error_msg) from semgrep_error
+
+            if process.returncode != 0:
+                error_msg = f"Error running semgrep on {source_code_path} with arguments" f" {process.args}"
+                logger.debug(error_msg)
+                raise HeuristicAnalyzerValueError(error_msg)
+
+            semgrep_output = json.loads(output_json_file.read())
```
tromai, line 278: We don't catch the exception raised by `json.loads` here, I wonder if this is intended ?

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/pypi_sourcecode_analyzer.py
art1f1c3R: I've changed it to use the same convention as the utility function [here](https://github.com/oracle/macaron/pull/965/commits/aa21abc953fb1f7484904d52d1b87328c28586b7).

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/pypi_sourcecode_analyzer.py
art1f1c3R: A more extensive explanation can be found on Semgrep's website [here](https://semgrep.dev/docs/ignoring-files-folders-code). Essentially, by default Semgrep will look to `.semgrepignore` and `.gitignore` for things to exclude from its scan. Semgrep may also ignore lines of code where there is a comment like `# nosemgrep`. Since we are analyzing lines of code for the purpose of determining if it malicious or not, we don't want an evasion technique to be adding `# nosemgrep` comments to malware source code, so we disable this feature as a precaution against this.

src/macaron/config/defaults.ini
```diff
@@ -600,3 +600,16 @@ major_threshold = 20
...
+# without the .yaml prefix. Currently, we disable the exfiltration rulesets by default due to a high false positive rate.
+disabled_default_rulesets = exfiltration
+# disable individual rules here (i.e. individual rule IDs inside a Semgrep .yaml file) using rule IDs. You may also
+# provide the IDs of your custom semgrep rules here too, as all Semgrep rule IDs must be unique.
+disabled_rules =
+# absolute path to a directory where a custom set of semgrep rules for source code analysis are stored. These will be included
+# with Macaron's default rules. The path will be normalised to the OS path type.
+custom_semgrep_rules =
+# disable custom semgrep rulesets here (i.e. all rule IDs in a Semgrep .yaml file) using ruleset names, the name without the
+# .yaml prefix. Note, this will be ignored if a path to custom semgrep rules is not provided. These file names must be unique.
```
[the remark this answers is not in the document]
tromai: Could I know more about the reason why we enforce unique on those elements ? Or it's just easier for us to load the values.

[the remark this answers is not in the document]
art1f1c3R: Yes I do. I'll add some clarification for this.

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/pypi_sourcecode_analyzer.py
tromai: Ah I see. Thanks. Could we have that link added to this comment line?

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/pypi_sourcecode_analyzer.py
art1f1c3R: This error should be caught. Resolved [here](https://github.com/oracle/macaron/pull/965/commits/0df29bab288cc79e0d1f6cbd96fce0673b53bb3f).

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/pypi_sourcecode_analyzer.py
art1f1c3R: That's a good idea. I've implemented that [here](https://github.com/oracle/macaron/pull/965/commits/0df29bab288cc79e0d1f6cbd96fce0673b53bb3f), and also added a unit test to ensure it works as expected.

tromai: I think instead of `glob.glob(os.path.join(path, "**", "*")`, may be you can use ` glob.glob(os.path.join(path, "**", file_name)` which `file_name` being an element in `target_files` to find the path to all file with that name in `path`. This way, you wouldn't need to traverse through the directories to find that file again.

src/macaron/malware_analyzer/pypi_heuristics/sourcecode/pypi_sourcecode_analyzer.py
```diff
@@ -0,0 +1,322 @@
...
+            start = json_extract(finding, ["start", "line"], int)
+            end = json_extract(finding, ["end", "line"], int)
+            message = json_extract(finding, ["extra", "message"], str)
+
+            # We manually filter out disabled rule IDs, as Semgrep's command line argument `--exclude-rule` appears to
+            # only work if `--experimental` is also supplied to enable experimental features, which we do not use.
+            # Semgrep provides a relative path separated by '.' to the rule ID, where the rule ID is always the
+            # final element in that path, so we use that to match our rule IDs.
+            if rule_id.split(".")[-1] in self.disabled_rule_ids:
+                if rule_id not in self.disabled_rule_ids:
```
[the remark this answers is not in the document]
art1f1c3R: `self.disabled_rule_ids` do not contain the "." separated path correct, so contain only the rule ID string (e.g. "obfuscation_decode-and-execute" from `obfuscatoin.yaml`). the extracted `rule_id` from the JSON output however does contain the "." separated path, hence the split operation.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -159,77 +166,102 @@ def download_package_json(self, url: str) -> dict:
...
-
-        # Create a temporary directory to store the downloaded source.
-        with tempfile.TemporaryDirectory() as temp_dir:
+        _, _, file_name = url.rpartition("/")
+        package_name = re.sub(r"\.tar\.gz$", "", file_name)
+
+        # temporary directory to unzip and read all source files
+        temp_dir = tempfile.mkdtemp(prefix=f"{package_name}_")
+        response = send_get_http_raw(url, stream=True)
+        if response is None:
```
tromai, line 194: I think you can use https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory if you are looking for a way to create a temporary directory and clean it up afterward.

art1f1c3R: This was a while ago, but I believe the rationale was that the `TemporaryDirectory` object made it difficult to store the extracted path persistent across the entire run. I want this capability as I would like other heuristics and potentially other components of Macaron to use this feature once this is merged, to prevent us downloading the tarball repeatedly. Currently, since this PR is isolated to my feature, I implement and use a context manager to download, extract, and clean it up in `DetectMaliciousMetadataCheck.analyze_source`.

tromai (approved):
LGTM. Thanks!

behnazh-w (approved):

PR #975, opened 2025-02-03
chore: store provenance asset info

### LINKED ISSUE #819: Store information about the provenance asset
With [this PR](https://github.com/oracle/macaron/pull/791), we have more accurate provenance discovery and support various places that provenances may get stored, including GitHub release assets. With this refactoring, the information about the the provenance, [such as the name and asset URL is not retained](https://github.com/oracle/macaron/pull/791/files#diff-4c6ee6828dbcdff0830647d11b4088836cc2636dfbf61882808ca8756c078de2L59-L63), however.

We need to refactor the `AnalyzeContext.dynamic_data["provenance"]` to retain the information related to the provenance.

benmss:
This PR adds provenance asset info to the database alongside the provenance it identifies. This was previously removed during refactoring.

Closes #819


src/macaron/database/table_definitions.py
```diff
@@ -511,6 +511,12 @@ class Provenance(ORMBase):
     #: The provenance payload.
     provenance_payload: Mapped[InTotoPayload] = mapped_column(ProvenancePayload, nullable=False)
 
+    #: The name of the provenance asset.
+    provenance_asset_name: Mapped[str] = mapped_column(String, nullable=True)
+
+    #: The URL of the provenance asset.
+    provenance_asset_url: Mapped[str] = mapped_column(String, nullable=True)
+
```
behnazh-w, line 519: Since you are adding these two columns, I was thinking to revisit this ORM mapping. Do you think 
```python
    #: The release tag commit sha.
    release_commit_sha: Mapped[str] = mapped_column(String, nullable=True)
    #: The release tag.
    release_tag: Mapped[str] = mapped_column(String, nullable=True)
```
could be redundant since the `commit_sha: Mapped[str] = mapped_column(String, nullable=True)` should be the same as release tag?

src/macaron/slsa_analyzer/checks/provenance_available_check.py
```diff
@@ -74,18 +74,31 @@ def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
...
+        provenance_info = None
+        inferred = False
+        if ctx.dynamic_data["provenance_info"]:
+            provenance_info = ctx.dynamic_data["provenance_info"]
+            inferred = ctx.dynamic_data["is_inferred_prov"]
+
+        if not provenance_info or not provenance_info.provenance_payload or inferred:
+            return CheckResultData(
+                result_tables=[
+                    ProvenanceAvailableFacts(
```
behnazh-w, line 86: Looks like we are not storing any useful info, so I think we can skip this? Or am I missing something here? :thinking:

behnazh-w (commented):
It would be good to add an integration test that checks the values stored in the `Provenance` and `ProvenanceAvailableFacts` tables.

behnazh-w (approved):

PR #977, opened 2025-02-06
feat: add callgraph and build cmd detection for Jenkins

behnazh-w:
### Summary:

This PR introduces the construction of callgraphs for bash scripts inlined within `Jenkinsfile` configurations. Additionally, it creates `BuildToolCommand` objects, which will be utilized in the build-related checks. With this update, Jenkins is no longer categorized as an unparsed CI service, although parsing support remains partial.

### Tests:
- An integration test has been added for the `pkg:maven/org.apache.dubbo/dubbo-rpc-memcached@2.7.7` artifact, which uses Jenkins for builds.
- The expected result in `tests/integration/cases/jenkinsci_plotplugin` has been updated to reflect the absence of a build service command.

### TODO:
Similar to GitHub Actions, we aim to support recursive calls to bash scripts within `Jenkinsfile`. The implementation should be straightforward, but it is recommended to first refactor the `bashparser` to decouple it from GitHub Actions, making it more general: https://github.com/oracle/macaron/issues/978

### DIFF
--- a/src/macaron/config/defaults.ini
+++ b/src/macaron/config/defaults.ini
@@ -158,8 +158,6 @@ jenkins =
     withMaven
     buildPlugin
     asfMavenTlpStdBuild
-    ./mvnw
-    ./mvn
 
 [builder.maven.ci.deploy]
 github_actions =
@@ -232,14 +230,12 @@ wrapper_files =
 [builder.gradle.ci.build]
 github_actions = actions/setup-java
 travis_ci =
-    jdk
-    ./gradlew
+    gradle
 circle_ci =
-    ./gradlew
+    gradle
 gitlab_ci =
-    ./gradlew
+    gradle
 jenkins =
-    ./gradlew
 
 [builder.gradle.ci.deploy]
 github_actions =
@@ -247,24 +243,28 @@ github_actions =
     spring-io/artifactory-deploy-action
 travis_ci =
     artifactoryPublish
+    gradle publish
     ./gradlew publish
     publishToSonatype
     gradle-git-publish
     gitPublishPush
 circle_ci =
     artifactoryPublish
+    gradle publish
     ./gradlew publish
     publishToSonatype
     gradle-git-publish
     gitPublishPush
 gitlab_ci =
     artifactoryPublish
+    gradle publish
     ./gradlew publish
     publishToSonatype
     gradle-git-publish
     gitPublishPush
 jenkins =
     artifactoryPublish
+    gradle publish
     ./gradlew publish
     publishToSonatype
     gradle-git-publish
--- a/src/macaron/slsa_analyzer/ci_service/jenkins.py
+++ b/src/macaron/slsa_analyzer/ci_service/jenkins.py
@@ -1,11 +1,26 @@
-# Copyright (c) 2022 - 2024, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2025, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """This module analyzes Jenkins CI."""
 
+import glob
+import logging
+import os
+import re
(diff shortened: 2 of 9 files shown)


src/macaron/config/defaults.ini
```diff
@@ -159,8 +159,6 @@ jenkins =
     withMaven
     buildPlugin
     asfMavenTlpStdBuild
-    ./mvnw
```
tromai, line 162: Is the reason why we remove these two lines is because `mvnw` and `mvn` has been covered by the `[builder.maven]` section in `defaults.ini`?
These two seems to be left over from an older version of Macaron.

src/macaron/config/defaults.ini
```diff
@@ -233,39 +231,41 @@ wrapper_files =
...
-    ./gradlew
+    gradle
 circle_ci =
-    ./gradlew
+    gradle
 gitlab_ci =
-    ./gradlew
+    gradle
 jenkins =
-    ./gradlew
```
tromai, line 243: I wonder why we don't have `gradle` for jenkins ?

tests/integration/cases/jenkinsci_plotplugin/policy.dl
```diff
@@ -1,12 +1,12 @@
...
 /* Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/. */
 
 #include "prelude.dl"
 
 Policy("test_policy", component_id, "") :-
-    check_passed(component_id, "mcn_build_script_1"),
-    check_passed(component_id, "mcn_build_service_1"),
     check_passed(component_id, "mcn_version_control_system_1"),
+    check_passed(component_id, "mcn_build_script_1"),
+    check_failed(component_id, "mcn_build_service_1"),
```
tromai, line 9: I wonder why `mcn_build_service_1` is changed from PASSED to FAILED ?

src/macaron/slsa_analyzer/ci_service/jenkins.py
```diff
@@ -56,7 +81,111 @@ def build_call_graph(self, repo_path: str, macaron_path: str = "") -> CallGraph:
...
+                name=workflow_name,
+                node_type=JenkinsNodeType.INTERNAL,
+                source_path=workflow_path,
+                caller=root,
+            )
+            root.add_callee(workflow_node)
+
+            # Find matching lines.
+            for line in lines:
+                match = re.match(pattern, line)
```
tromai, line 115: If we are matching the Jenkinsfile one line at a time, would we miss multi-lines cases? For example
```
sh '''
    echo "Multiline shell steps works too"
    ls -lah
'''
```
Example taken from https://www.jenkins.io/doc/pipeline/tour/running-multiple-steps/#linux-bsd-and-mac-os.

Or are we not supporting multi-line cases at the moment.

tromai (commented):
I have finished the first round of review. I left some questions about the PR. Thank you.

src/macaron/config/defaults.ini
behnazh-w: Yes, these are not needed anymore. They also hardcode the path, which is not desired.

src/macaron/config/defaults.ini
behnazh-w: Not sure why `./gradle` was not used, but searching for these as keywords is not needed anymore.

tests/integration/cases/jenkinsci_plotplugin/policy.dl
behnazh-w: The previous `PASSED` result was a false positive.

src/macaron/slsa_analyzer/ci_service/jenkins.py
behnazh-w: No, the regex only supports a single-line shell scripts. I will add a TODO to support multi-line scripts.

tromai (approved):

tromai (approved):

PR #1011, opened 2025-03-11
chore: add config option to disable latest PURL retrieval

### LINKED ISSUE #1010: Maven artefact update includes valid URL in SCM, breaks integration test
The latest update to Maven artefacts, version `4.0.0-rc-3`, now includes a URL in the SCM of the POM that can be parsed by the Repo Finder, resolving to `https://github.com/apache/maven`. The integration test `apache_maven_sbom` assumes that the dependencies of `maven` have no discoverable repository URL. This new version causes them to have one, because the sought version `4.0.0-alpha-1-SNAPSHOT` does not have a POM, and the find latest PURL feature retrieves the newest instead, which now does.

benmss:
This PR adds a configuration option to disable the latest PURL retrieval feature, and disables the feature in the `apache_maven_sbom` integration test so that the results of that test are static.

Closes #1010

### DIFF
--- a/src/macaron/config/defaults.ini
+++ b/src/macaron/config/defaults.ini
@@ -60,6 +60,7 @@ redirect_urls =
     gitbox.apache.org
     git-wip-us.apache.org
 find_source_should_clone = False
+try_latest_purl = True
 
 [repofinder.java]
 # The list of maven-like repositories to attempt to retrieve artifact POMs from.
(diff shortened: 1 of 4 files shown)


src/macaron/config/defaults.ini
```diff
@@ -60,6 +60,7 @@ redirect_urls =
     gitbox.apache.org
     git-wip-us.apache.org
 find_source_should_clone = False
+try_latest_purl = True
```
behnazh-w, line 63: Can you please add a comment to explain the option (and while at it for `find_source_should_clone` as well)?

behnazh-w (approved):

PR #1014, opened 2025-03-12
chore: remove usages of hard-coded provenance build types

benmss:


src/macaron/provenance/provenance_extractor.py
behnazh-w, line 1: You can also replace `build_def = json_extract(predicate, ["buildDefinition"], dict)` with `ProvenancePredicate.find_build_def(prov_payload.statement)` throughout this module.

behnazh-w: Also, we might be able to use the `get_build_invocation` instead of  `json_extract(predicate, ["invocation"]).`

behnazh-w (approved):

PR #1019, opened 2025-03-18
fix: use 'isDefault' version from deps dev api

### LINKED ISSUE #1018: Versions returned by deps.dev API may need further sorting
When trying to find the latest version of an artefact from deps.dev, the ordering of the results may not reflect the expected ordering of the versions. Currently, it is assumed that the last entry in the returned list is the latest version. However, in some cases it seems this is not true. For example `https://api.deps.dev/v3alpha/purl/pkg%3Amaven%2Fcom.google.guava%2Fguava` shows `r09` as the final entry in the list, when the real latest version is `33.4.0-jre`

benmss:
When retrieving versions of artefacts from `deps.dev` API, one of the entries will have the `isDefault` property set to `True`. It is this version that is considered to be the latest version in the list, not the last entry as was previously assumed.

Closes #1018

### DIFF
--- a/src/macaron/repo_finder/repo_finder_deps_dev.py
+++ b/src/macaron/repo_finder/repo_finder_deps_dev.py
@@ -148,12 +148,14 @@ class DepsDevRepoFinder(BaseRepoFinder):
             return None, RepoFinderInfo.DDEV_JSON_INVALID
 
         latest_version = None
-        for index, version_result in enumerate(versions):
-            if version_result["isDefault"] or (index == (len(versions) - 1) and not latest_version):
+        for version_result in reversed(versions):
+            if version_result["isDefault"]:
+                # Accept the version as the latest if it is marked with the "isDefault" property.
                 latest_version = json_extract(version_result, ["versionKey", "version"], str)
                 break
 
         if not latest_version:
+            logger.debug("No latest version found in version list: %s", len(versions))
             return None, RepoFinderInfo.DDEV_JSON_INVALID
 
         namespace = purl.namespace + "/" if purl.namespace else ""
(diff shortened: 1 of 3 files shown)


src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -146,7 +146,13 @@ def get_latest_version(purl: PackageURL) -> tuple[PackageURL | None, RepoFinderI
         versions = json_extract(metadata, versions_keys, list)
         if not versions:
             return None, RepoFinderInfo.DDEV_JSON_INVALID
-        latest_version = json_extract(versions[-1], ["versionKey", "version"], str)
+
+        latest_version = None
+        for index, version_result in enumerate(versions):
+            if version_result["isDefault"] or (index == (len(versions) - 1) and not latest_version):
```
behnazh-w, line 152: Can `latest_version` not be `None` and reach here?

behnazh-w: Please add a comment in the code to explain the logic here.

src/macaron/repo_finder/repo_finder_deps_dev.py
behnazh-w, line 1: Let's add an integration test for the `find-source` command to find the repo for `pkg:maven/com.google.guava/guava@14.0.1?type=jar`.

src/macaron/repo_finder/repo_finder_deps_dev.py
benmss: It is intended to accept the version that has the `isDefault` property set to `True`, or the final entry in the list in case one isn't found. However, if none are set to default, maybe it's better to log an error and return nothing as that represents some issue with `deps.dev` or our expectations.

behnazh-w: > It is intended to accept the version that has the `isDefault` property set to `True`, or the final entry in the list in case one isn't found. However, if none are set to default, maybe it's better to log an error and return nothing as that represents some issue with `deps.dev` or our expectations.

Thanks, I meant to just add that in code as a comment :slightly_smiling_face:

behnazh-w (approved):

PR #1020, opened 2025-03-19
feat: add GitHub attestation discovery

### LINKED ISSUE #915: Discover GitHub Attestations
We currently [support GitHub Attestations](https://github.com/oracle/macaron/pull/732) when they are provided as input, but we do not yet implement automatic discovery of these attestations.

For example, here is a Java library that generates GitHub Attestations:
- `pkg:maven/io.liftwizard/liftwizard-reladomo-operation-compiler@2.1.21`
- GitHub Attestation: https://github.com/liftwizard/liftwizard/attestations/2827088

benmss:
This PR adds support for GitHub attestation discovery. To access GitHub attestations, the SHA256 hash of the repositories artefacts must be generated and submitted via the API. Artefacts may be found in local caches, or downloaded from remote repositories.

- [x] Support for Maven artefacts
- [x] Support for PyPI artefacts

Closes #915

### DIFF
--- a/src/macaron/artifact/local_artifact.py
+++ b/src/macaron/artifact/local_artifact.py
@@ -1,16 +1,21 @@
-# Copyright (c) 2024 - 2024, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2024 - 2025, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """This module declares types and utilities for handling local artifacts."""
 
 import fnmatch
 import glob
+import hashlib
+import logging
 import os
 
 from packageurl import PackageURL
 
 from macaron.artifact.maven import construct_maven_repository_path
 from macaron.errors import LocalArtifactFinderError
+from macaron.slsa_analyzer.package_registry import MavenCentralRegistry
+
+logger: logging.Logger = logging.getLogger(__name__)
 
 
 def construct_local_artifact_dirs_glob_pattern_maven_purl(maven_purl: PackageURL) -> list[str] | None:
@@ -247,3 +252,53 @@ def get_local_artifact_dirs(
         )
 
     raise LocalArtifactFinderError(f"Unsupported PURL type {purl_type}")
+
+
+def get_local_artifact_hash(purl: PackageURL, artifact_dirs: list[str], hash_algorithm_name: str) -> str | None:
+    """Compute the hash of the local artifact.
+
+    Parameters
+    ----------
+    purl: PackageURL
+        The PURL of the artifact being sought.
+    artifact_dirs: list[str]
+        The possible locations of the artifact.
+    hash_algorithm_name: str
+        The hash algorithm to use.
+
+    Returns
+    -------
+    str | None
+        The hash, or None if not found.
+    """
+    if not artifact_dirs:
+        logger.debug("No artifact directories provided.")
+        return None
+
+    if not purl.version:
+        logger.debug("PURL is missing version.")
+        return None
+
+    artifact_target = None
+    if purl.type == "maven":
+        artifact_target = MavenCentralRegistry.get_artifact_file_name(purl)
+
+    if not artifact_target:
+        logger.debug("PURL type not supported: %s", purl.type)
+        return None
+
+    for artifact_dir in artifact_dirs:
+        full_path = os.path.join(artifact_dir, artifact_target)
+        if not os.path.exists(full_path):
+            continue
+
+        with open(full_path, "rb") as file:
+            try:
+                hash_result = hashlib.file_digest(file, hash_algorithm_name)
+            except ValueError as error:
+                logger.debug("Error while hashing file: %s", error)
+                continue
+
+            return hash_result.hexdigest()
+
+    return None
--- a/src/macaron/artifact/local_artifact.py
+++ b/src/macaron/artifact/local_artifact.py
@@ -11,9 +11,8 @@ import os
 
 from packageurl import PackageURL
 
-from macaron.artifact.maven import construct_maven_repository_path
+from macaron.artifact.maven import construct_maven_repository_path, construct_primary_jar_file_name
 from macaron.errors import LocalArtifactFinderError
-from macaron.slsa_analyzer.package_registry import MavenCentralRegistry
 
 logger: logging.Logger = logging.getLogger(__name__)
 
@@ -254,7 +253,7 @@ def get_local_artifact_dirs(
     raise LocalArtifactFinderError(f"Unsupported PURL type {purl_type}")
 
 
-def get_local_artifact_hash(purl: PackageURL, artifact_dirs: list[str], hash_algorithm_name: str) -> str | None:
+def get_local_artifact_hash(purl: PackageURL, artifact_dirs: list[str]) -> str | None:
     """Compute the hash of the local artifact.
 
     Parameters
@@ -262,9 +261,7 @@ def get_local_artifact_hash(purl: PackageURL, artifact_dirs: list[str], hash_alg
     purl: PackageURL
         The PURL of the artifact being sought.
     artifact_dirs: list[str]
-        The possible locations of the artifact.
-    hash_algorithm_name: str
-        The hash algorithm to use.
+        The list of directories that may contain the artifact file.
 
     Returns
     -------
@@ -281,7 +278,9 @@ def get_local_artifact_hash(purl: PackageURL, artifact_dirs: list[str], hash_alg
 
     artifact_target = None
     if purl.type == "maven":
-        artifact_target = MavenCentralRegistry.get_artifact_file_name(purl)
+        artifact_target = construct_primary_jar_file_name(purl)
+
+    # TODO add support for other PURL types here.
 
     if not artifact_target:
         logger.debug("PURL type not supported: %s", purl.type)
@@ -294,7 +293,7 @@ def get_local_artifact_hash(purl: PackageURL, artifact_dirs: list[str], hash_alg
 
         with open(full_path, "rb") as file:
             try:
-                hash_result = hashlib.file_digest(file, hash_algorithm_name)
+                hash_result = hashlib.file_digest(file, "sha256")
             except ValueError as error:
                 logger.debug("Error while hashing file: %s", error)
                 continue
--- a/src/macaron/artifact/local_artifact.py
+++ b/src/macaron/artifact/local_artifact.py
@@ -281,6 +281,8 @@ def get_local_artifact_hash(purl: PackageURL, artifact_dirs: list[str]) -> str |
         artifact_target = construct_primary_jar_file_name(purl)
 
     # TODO add support for other PURL types here.
+    # Other purl types can be easily supported if user provided artifacts are accepted from the command line.
+    # See https://github.com/oracle/macaron/issues/498.
 
     if not artifact_target:
         logger.debug("PURL type not supported: %s", purl.type)
--- a/src/macaron/repo_finder/repo_finder.py
+++ b/src/macaron/repo_finder/repo_finder.py
@@ -111,8 +111,6 @@ def find_repo(
     logger.debug("Analyzing %s with Repo Finder: %s", purl, type(repo_finder))
     found_repo, outcome = repo_finder.find_repo(purl)
 
-    print(package_registries_info)
-
     if not found_repo:
         found_repo, outcome = find_repo_alternative(purl, outcome, package_registries_info)
 
--- a/src/macaron/repo_finder/repo_finder.py
+++ b/src/macaron/repo_finder/repo_finder.py
@@ -111,6 +111,8 @@ def find_repo(
     logger.debug("Analyzing %s with Repo Finder: %s", purl, type(repo_finder))
     found_repo, outcome = repo_finder.find_repo(purl)
 
+    print(package_registries_info)
+
     if not found_repo:
         found_repo, outcome = find_repo_alternative(purl, outcome, package_registries_info)
 
@@ -166,7 +168,12 @@ def find_repo_alternative(
         found_repo, outcome = repo_finder_pypi.find_repo(purl, package_registries_info)
 
     if not found_repo:
-        logger.debug("Could not find repository using type specific (%s) methods for PURL: %s", purl.type, purl)
+        logger.debug(
+            "Could not find repository using type specific (%s) methods for PURL %s. Outcome: %s",
+            purl.type,
+            purl,
+            outcome,
+        )
 
     return found_repo, outcome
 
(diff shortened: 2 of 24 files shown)


src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
```diff
@@ -236,3 +239,84 @@ def find_publish_timestamp(self, purl: str) -> datetime:
                 raise InvalidHTTPResponseError(f"The timestamp returned by {url} is invalid") from error
 
         raise InvalidHTTPResponseError(f"Invalid response from Maven central for {url}.")
+
+    @staticmethod
+    def get_artifact_file_name(purl: PackageURL) -> str | None:
```
tromai, line 244: I think it's better if we rename this function and update the doc string to explicitly communicate that we only care about the "main" Maven artifact corresponding to a Maven type PackageURL. This is because for a GAV directory on Central repository for example, there can be many type of artifacts with different extensions and classifiers.
In additions, because the "main" Maven artifact doesn't have to always ends with ".jar". For example - https://repo1.maven.org/maven2/org/apache/batchee/batchee-webapp/1.0.3/, the "main" artifacts ends with ".war" instead. I wonder if we are going to support this case in the future ?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -482,6 +503,39 @@ def run_single(
...
+                            with tempfile.TemporaryDirectory() as temp_dir:
+                                attestation_file = os.path.join(temp_dir, "attestation")
+                                with open(attestation_file, "w", encoding="UTF-8") as file:
+                                    json.dump(git_attestation, file)
+
+                                try:
+                                    payload = load_provenance_payload(attestation_file)
+                                    provenance_payload = payload
+                                except LoadIntotoAttestationError as error:
+                                    logger.debug("Failed to load provenance payload: %s", error)
```
nicallen, line 537: Is it really necessary to dump it to a temp file then read it back in? Can't we just process the `git_attestation` directly as a provenance in-memory?

src/macaron/artifact/local_artifact.py
```diff
@@ -247,3 +252,53 @@ def get_local_artifact_dirs(
         )
 
     raise LocalArtifactFinderError(f"Unsupported PURL type {purl_type}")
+
+
+def get_local_artifact_hash(purl: PackageURL, artifact_dirs: list[str], hash_algorithm_name: str) -> str | None:
```
tromai, line 257: Is this function intended to only work with maven PackageURL or we are going to support Python PackagePURL in the future ?

behnazh-w: We plan to support other ecosystems in the future as well. GitHub attestation can be used for any type of package.

behnazh-w: We plan to support other ecosystems in the future as well. GitHub attestation can be used for any type of package.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -942,6 +988,99 @@ def create_analyze_ctx(self, component: Component) -> AnalyzeContext:
 
         return analyze_ctx
 
+    def get_artifact_hash(
+        self,
+        purl: PackageURL,
+        cached_artifacts: list[str] | None,
+        hash_algorithm: Any,
```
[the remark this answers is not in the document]
benmss: I do not think so. There's some discussion on it here: https://github.com/python/typeshed/issues/2928

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
benmss: I think it would make sense to specify the `jar` artefact as our actual target, currently.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -942,6 +988,99 @@ def create_analyze_ctx(self, component: Component) -> AnalyzeContext:
...
+        package_registries_info: list[PackageRegistryInfo],
+    ) -> str | None:
+        """Get the hash of the artifact found from the passed PURL using local or remote files.
```
tromai, line 998: Could we add more description on the behaviors of this function regarding the local or remote files. For example
- When a remote files are fetched ? Are they stored persistently in memory ?
- The function returns early if a local artifact is discovered.

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
```diff
@@ -236,3 +239,84 @@ def find_publish_timestamp(self, purl: str) -> datetime:
...
+        str | None
+            The artifact file name, or None if invalid.
+        """
+        if not purl.version:
+            return None
+
+        return purl.name + "-" + purl.version + ".jar"
+
+    def get_artifact_hash(self, purl: PackageURL, hash_algorithm: Any) -> str | None:
+        """Return the hash of the artifact found by the passed purl relevant to the registry's URL.
```
tromai, line 263: Could I have some more description in the doc string about the `relevant to the registry's URL` part ? Some examples would be highly appreciated too.

src/macaron/slsa_analyzer/analyzer.py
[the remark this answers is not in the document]
tromai: Oh okay. Then I think it's good enough to add some comment, or doc string describing the expected type, and why we have to annotate it `Any` here.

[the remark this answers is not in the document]
tromai: An additional thing that I have been thinking about is, the hash object we pass to this parameter is not used anywhere outside of this function, see - 
```
artifact_hash = self.get_artifact_hash(
    parsed_purl, local_artifact_dirs, hashlib.sha256(), package_registries_info
)
```
Therefore, I think it would be good to only create the object `hashlib.sha256()` in the function where it's being used to calculate the sha256 hash. For example, within `MavenCentralRegistry.get_artifact_path`(not passed as a parameter):
```
def get_artifact_hash(self, purl: PackageURL, hash_algorithm: str) -> str | None:
    ... = hashlib.new('sha256')
```
The reason is hashlib.sha256() object can be mutated with `update()` so I wouldn't want it to be passed through a lot of places, which can risk accidental mutation.

tests/integration/cases/github_pypi_attestation/test.yaml
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2024 - 2025, Oracle and/or its affiliates. All rights reserved.
```
tromai, line 1: I'm not sure if I run this test correctly, but apparently for this case, the code for computing artifact hash and obtaining attestation from github is not run. 
This is because at https://github.com/oracle/macaron/blob/b6e3ea36244d87c89ca045158e4fcce4e114764a/src/macaron/slsa_analyzer/analyzer.py#L509

`provenance_payload` is not None. Adding a break point at that place reveals the content
```
(Pdb) pp(provenance_payload)
InTotoV1Payload(statement={'_type': 'https://in-toto.io/Statement/v1',
                           'predicate': {'buildType': 'pypi_certificate',
                                         'invocationUrl': 'https://github.com/beeware/toga/actions/runs/13847909380',
                                         'sourceDigest': 'e38f09f2624a35067226c6b7490f8344e3ff7c36',
                                         'sourceUri': 'https://github.com/beeware/toga',
                                         'workflow': '.github/workflows/publish.yml'},
                           'predicateType': 'https://docs.pypi.org/attestations/publish/v1',
                           'subject': [{'digest': {'sha256': '061bb843da0225e19b5911609502986df9ff361aaa18b91541fbb11d6d5bbe32'},
                                        'name': 'toga-0.5.0-py3-none-any.whl'}]},
                verified=True)
(Pdb) 
```

tests/integration/cases/github_maven_attestation_local/test.yaml
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) 2024 - 2025, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+description: |
+  Discovering GitHub attestation of a local Maven artifact but failing because the artifact is wrong
+
+tags:
+- macaron-python-package
```
tromai, line 8: I wonder if we could also run this test as part of the Docker integration test. There is extra handling of the local artifact directories for the Docker container so we want to make sure it's working fine too.

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
```diff
@@ -236,3 +239,84 @@ def find_publish_timestamp(self, purl: str) -> datetime:
...
+            response = requests.get(artifact_url, stream=True, timeout=40)
+            response.raise_for_status()
+        except requests.exceptions.HTTPError as http_err:
+            logger.debug("HTTP error occurred: %s", http_err)
+            return None
+
+        if response.status_code != 200:
+            return None
+
+        # Download file and compute hash as chunks are received.
```
tromai, line 306: Just a curious question: was there a specific reason why we have to download chunk by chunk at this location?

tromai (changes requested):
I have finished my first round of review. Thank you.

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
```diff
@@ -236,3 +239,84 @@ def find_publish_timestamp(self, purl: str) -> datetime:
...
+
+        Returns
+        -------
+        str | None
+            The hash of the artifact, or None if not found.
+        """
+        if not purl.namespace:
+            return None
+
+        file_name = MavenCentralRegistry.get_artifact_file_name(purl)
```
behnazh-w, line 280: Since we just want to get the "main" artifact name from a Maven PURL, this method should not be specific to Maven Central. We also have a [`maven.py` ](https://github.com/oracle/macaron/blob/915-discover-github-attestations/src/macaron/artifact/maven.py)module that might be more suitable for this method.

src/macaron/artifact/local_artifact.py
```diff
@@ -247,3 +252,53 @@ def get_local_artifact_dirs(
...
+    if not artifact_dirs:
+        logger.debug("No artifact directories provided.")
+        return None
+
+    if not purl.version:
+        logger.debug("PURL is missing version.")
+        return None
+
+    artifact_target = None
+    if purl.type == "maven":
```
behnazh-w, line 283: Why don't we support `PyPI` packages here?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -482,6 +503,39 @@ def run_single(
...
+
+        # Try to find an attestation from GitHub, if applicable.
+        if parsed_purl and not provenance_payload and analysis_target.repo_path and isinstance(git_service, GitHub):
+            # Try to discover GitHub attestation for the target software component.
+            url = None
+            try:
+                url = urllib.parse.urlparse(analysis_target.repo_path)
+            except TypeError as error:
+                logger.debug("Failed to parse repository path as URL: %s", error)
+            if url and url.hostname == "github.com":
```
tromai, line 515: I wonder if `url` is being used anywhere else in this function ? If not, would `isinstance(git_service, GitHub)` be sufficient to check if the git service is Github?

src/macaron/slsa_analyzer/analyzer.py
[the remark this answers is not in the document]
benmss: I think that makes sense. I've made the algorithm instantiations local.

src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -531,3 +576,59 @@ def get_sourcecode(self) -> dict[str, str] | None:
...
+            artifact_hash = json_extract(self.package_json, ["urls", 0, "digests", "sha256"], str)
+        else:
+            artifact_hash = json_extract(
+                self.package_json, ["releases", self.component_version, 0, "digests", "sha256"], str
+            )
+        logger.debug("Found sha256 hash: %s", artifact_hash)
+        return artifact_hash
+
+
+def find_or_create_pypi_asset(
```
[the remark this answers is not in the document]
benmss: I disagree with this suggestion. It's true that the function can be split into smaller pieces, however, in the 2 places that it is used, all split pieces would still need to be called sequentially. E.g. If we are about to download an asset we must first check if it already has been downloaded, then if not, create the asset, add it to the registry, etc. Ultimately, this means identical blocks of code calling all 2 to 3 sub functions.

Perhaps we can improve the name and/or documentation instead?

(There is one more place this function should be called. I will add it there instead of the almost identical solution that exists there currently.)

tests/integration/cases/github_pypi_attestation/test.yaml
benmss: Good catch. This is because the test target provides both PyPI and GitHub attestation, and the recently added PyPI discovery happens first. I will fix this by testing on an older version that only has GitHub attestation.

src/macaron/slsa_analyzer/package_registry/maven_central_registry.py
benmss: This is the preferred way to download large files. Whether we expect the artefact files to be "large" could be worth discussing.

src/macaron/slsa_analyzer/analyzer.py
benmss: I think you may be right. I'll remove it.

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -482,6 +503,39 @@ def run_single(
...
+                    parsed_purl, local_artifact_dirs, hashlib.sha256(), package_registries_info
+                )
+                if artifact_hash:
+                    git_attestation_dict = git_service.api_client.get_attestation(
+                        analyze_ctx.component.repository.full_name, artifact_hash
+                    )
+                    if git_attestation_dict:
+                        git_attestation_list = json_extract(git_attestation_dict, ["attestations"], list)
+                        if git_attestation_list:
+                            git_attestation = git_attestation_list[0]
```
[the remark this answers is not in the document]
benmss: We are assuming that the first attestation is the one we are looking for. The alternative is to decode the payload of each and compare against the target artifact name. I'm unsure how common it is to find multiple attestations for a given artifact hash.

For the schema, there is an example given in the API. I will add this to the doc comment.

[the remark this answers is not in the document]
tromai: Thanks. Updating the doc comment sounds good to me.

src/macaron/slsa_analyzer/analyzer.py
benmss: This will be addressed as part of a separate issue #1095

src/macaron/repo_finder/repo_finder.py
```diff
@@ -111,6 +111,8 @@ def find_repo(
     logger.debug("Analyzing %s with Repo Finder: %s", purl, type(repo_finder))
     found_repo, outcome = repo_finder.find_repo(purl)
 
+    print(package_registries_info)
```
tromai, line 114: This print statement could be removed or turned into a log message.

behnazh-w (approved):

tromai (approved):

Issue #1027, opened 2025-03-26
pypi malware reporting false positives due to incorrect skip result evaluation


art1f1c3R:
The `detect_malicious_metadata_check.py` report is generating false positives due the `HeuristicResult.SKIP` values being evaluated incorrectly in the internal ProbLog model. The value of `0.0` assigned to a heuristic causes it to be evaluated as `false`. This is what is generating the false positives.

PR #1031, opened 2025-03-27
fix: fix incorrect skip result evaluation causing false positives in PyPI malware reporting"

art1f1c3R:
Addressing issue identified in #1027, where skips were being evaluated as false. This PR introduces wrappers `passed()` and `failed()` into the ProbLog model that use `try_call()` statements. Skipped heuristics are no longer defined in the ProbLog model, which is why this `try_call()` statement is used. This means that evaluating `failed(heuristic)` will be false if the heuristic passed, or if it was not defined (i.e. was skipped). Similarly, for evaluating `passed()`, this will be false if the heuristic failed, or if it was not defined. This should handle situations where skips should not cause rules they are part of to trigger. This method was the easiest way to keep as much of the ProbLog model in a static string as possible, without having to perform extensive string operations.

Rule IDs have also been added for debugging purposes, and a method to extract them, so that it is evident what rule was triggered.

### DIFF
--- a/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
+++ b/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
@@ -336,9 +336,7 @@ class DetectMaliciousMetadataCheck(BaseCheck):
     % ----- Heuristic groupings -----
     % These are common combinations of heuristics that are used in many of the rules, thus themselves representing
     % certain behaviors. When changing or adding rules here, if there are frequent combinations of particular
-    % heuristics, group them together here. Note, these should only be used to check if a grouping statement
-    % is true. Evaluating 'not quickUndetailed' would be true if empty project link and closer release join
-    % date passed, or if they were both skipped, which is not desired behaviour.
+    % heuristics, group them together here.
 
     % Maintainer has recently joined, publishing an undetailed page with no links.
     quickUndetailed :- failed({Heuristics.EMPTY_PROJECT_LINK.value}), failed({Heuristics.CLOSER_RELEASE_JOIN_DATE.value}).
--- a/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
+++ b/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
@@ -140,11 +140,18 @@ class DetectMaliciousMetadataCheck(BaseCheck):
 
         Returns
         -------
-        tuple[float, JsonType]
+        tuple[float, list[str]]
             Returns the confidence associated with the detected malicious combination, and associated rule IDs detailing
             what rules were triggered.
         """
         facts_list: list[str] = []
+        triggered_rules = []
+        # confidence is calculated using the probability of the package being benign, so the negation of the confidence values
+        # in the problog model. Multiplying these probabilities together on several triggers will further decrease the probability
+        # of the package being benign. This is then negated after calculation to get the probability of the package being malicious.
+        # If no rules are triggered, this will simply result in 1.0 - 1.0 = 0.0.
+        confidence: float = 1.0
+
         for heuristic, result in heuristic_results.items():
             if result == HeuristicResult.PASS:
                 facts_list.append(f"{heuristic.value} :- true.")
@@ -159,10 +166,11 @@ class DetectMaliciousMetadataCheck(BaseCheck):
         problog_model = PrologString(problog_code)
         problog_results: dict[Term, float] = get_evaluatable().create_from(problog_model).evaluate()
 
-        confidence = sum(conf for conf in problog_results.values() if conf is not None)
-        triggered_rules: JsonType = ["No malicious rules triggered"]
-        if confidence > 0:
-            triggered_rules = [term.args[0] for term in problog_results]
+        for term, conf in problog_results.items():
+            if conf is not None and conf > 0:
+                confidence *= 1.0 - conf  # decrease the probability of the package being benign
+                triggered_rules.append(term.args[0])
+        confidence = round(1.0 - confidence, 2)  # 2 decimal places
 
         return confidence, triggered_rules
 
--- a/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
+++ b/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
@@ -140,9 +140,9 @@ class DetectMaliciousMetadataCheck(BaseCheck):
 
         Returns
         -------
-        tuple[float, list[str]]
+        tuple[float, JsonType]
             Returns the confidence associated with the detected malicious combination, and associated rule IDs detailing
-            what rules were triggered.
+            what rules were triggered and their confidence as a dict[str, float] type.
         """
         facts_list: list[str] = []
         triggered_rules: dict[str, JsonType] = {}
@@ -328,6 +328,7 @@ class DetectMaliciousMetadataCheck(BaseCheck):
         AnomalousVersionAnalyzer,
     ]
 
+    # name used to query the result of all problog rules, so it can be accessed outside the model.
     problog_result_access = "result"
 
     malware_rules_problog_model = f"""
--- a/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
+++ b/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
@@ -347,14 +347,14 @@ class DetectMaliciousMetadataCheck(BaseCheck):
     % ----- Suspicious Combinations -----
 
     % Package released recently with little detail, forcing the setup.py to run.
-    {Confidence.HIGH.value}::result("high_confidence_1") :-
+    {Confidence.HIGH.value}::result("malware_high_confidence_1") :-
         quickUndetailed, forceSetup, failed({Heuristics.ONE_RELEASE.value}).
-    {Confidence.HIGH.value}::result("high_confidence_2") :-
+    {Confidence.HIGH.value}::result("malware_high_confidence_2") :-
         quickUndetailed, forceSetup, failed({Heuristics.HIGH_RELEASE_FREQUENCY.value}).
 
     % Package released recently with little detail, with some more refined trust markers introduced: project links,
     % multiple different releases, but there is no source code repository matching it and the setup is suspicious.
-    {Confidence.HIGH.value}::result("high_confidence_3") :-
+    {Confidence.HIGH.value}::result("malware_high_confidence_3") :-
         failed({Heuristics.SOURCE_CODE_REPO.value}),
         failed({Heuristics.HIGH_RELEASE_FREQUENCY.value}),
         passed({Heuristics.UNCHANGED_RELEASE.value}),
@@ -363,14 +363,14 @@ class DetectMaliciousMetadataCheck(BaseCheck):
 
     % Package released recently with little detail, with multiple releases as a trust marker, but frequent and with
     % the same code.
-    {Confidence.MEDIUM.value}::result("medium_confidence_1") :-
+    {Confidence.MEDIUM.value}::result("malware_medium_confidence_1") :-
         quickUndetailed,
         failed({Heuristics.HIGH_RELEASE_FREQUENCY.value}),
         failed({Heuristics.UNCHANGED_RELEASE.value}),
         passed({Heuristics.SUSPICIOUS_SETUP.value}).
 
     % Package released recently with little detail and an anomalous version number for a single-release package.
-    {Confidence.MEDIUM.value}::result("medium_confidence_2") :-
+    {Confidence.MEDIUM.value}::result("malware_medium_confidence_2") :-
         quickUndetailed,
         failed({Heuristics.ONE_RELEASE.value}),
         passed({Heuristics.WHEEL_ABSENCE.value}),
--- a/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
+++ b/src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
@@ -128,7 +128,9 @@ class DetectMaliciousMetadataCheck(BaseCheck):
         is_malware, detail_info = sourcecode_analyzer.analyze()
         return is_malware, detail_info
 
-    def evaluate_heuristic_results(self, heuristic_results: dict[Heuristics, HeuristicResult]) -> float | None:
+    def evaluate_heuristic_results(
+        self, heuristic_results: dict[Heuristics, HeuristicResult]
+    ) -> tuple[float, JsonType]:
         """Analyse the heuristic results to determine the maliciousness of the package.
 
         Parameters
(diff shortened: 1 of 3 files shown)


src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -158,10 +159,12 @@ def evaluate_heuristic_results(self, heuristic_results: dict[Heuristics, Heurist
         problog_model = PrologString(problog_code)
         problog_results: dict[Term, float] = get_evaluatable().create_from(problog_model).evaluate()
 
-        confidence: float | None = problog_results.get(Term(self.problog_result_access))
-        if confidence == 0.0:
-            return None  # no rules were triggered
-        return confidence
+        confidence = sum(conf for conf in problog_results.values() if conf is not None)
```
behnazh-w, line 162: Can you please add a comment in the code why `sum` is used here, and maybe provide an example.

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -321,51 +325,59 @@ def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
...
-        not {Heuristics.ANOMALOUS_VERSION.value}.
-
-    {problog_result_access} :- high.
-    {problog_result_access} :- medium.
-
-    query({problog_result_access}).
+    {Confidence.MEDIUM.value}::result("medium_confidence_2") :-
+        quickUndetailed,
+        failed({Heuristics.ONE_RELEASE.value}),
+        passed({Heuristics.WHEEL_ABSENCE.value}),
```
behnazh-w, line 376: Why is`Heuristics.WHEEL_ABSENCE.value` passed here?

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -321,51 +325,59 @@ def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
...
-    forceSetup :- not {Heuristics.SUSPICIOUS_SETUP.value}, not {Heuristics.WHEEL_ABSENCE.value}.
+    forceSetup :- failed({Heuristics.SUSPICIOUS_SETUP.value}), failed({Heuristics.WHEEL_ABSENCE.value}).
 
-    % Suspicious Combinations
+    % ----- Suspicious Combinations -----
 
     % Package released recently with little detail, forcing the setup.py to run.
-    {Confidence.HIGH.value}::high :- quickUndetailed, forceSetup, not {Heuristics.ONE_RELEASE.value}.
-    {Confidence.HIGH.value}::high :- quickUndetailed, forceSetup, not {Heuristics.HIGH_RELEASE_FREQUENCY.value}.
+    {Confidence.HIGH.value}::result("high_confidence_1") :-
```
behnazh-w, line 350: How about `malware_high_confidence_1`, and similar pattern for others?

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -321,51 +325,59 @@ def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
...
-    {problog_result_access} :- medium.
-
-    query({problog_result_access}).
+    {Confidence.MEDIUM.value}::result("medium_confidence_2") :-
+        quickUndetailed,
+        failed({Heuristics.ONE_RELEASE.value}),
+        passed({Heuristics.WHEEL_ABSENCE.value}),
+        failed({Heuristics.ANOMALOUS_VERSION.value}).
+
+    % ----- Evaluation -----
```
behnazh-w, line 379: Can you please add a comment in the code to explain what happens here? I.e., which rules will be selected and how they are combined, e.g., is it enough if one of them is `True`? What if more than one rule is evaluated to `True`?

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
art1f1c3R: That's a result of the translation from these combinations:
```
    (
        HeuristicResult.FAIL,  # Empty Project
        HeuristicResult.SKIP,  # Source Code Repo
        HeuristicResult.FAIL,  # One Release
        HeuristicResult.SKIP,  # High Release Frequency
        HeuristicResult.SKIP,  # Unchanged Release
        HeuristicResult.FAIL,  # Closer Release Join Date
        HeuristicResult.PASS,  # Suspicious Setup
        HeuristicResult.PASS,  # Wheel Absence
        HeuristicResult.FAIL,  # Anomalous Version
        # No project link, only one release, and the maintainer released it shortly
        # after account registration.
        # The setup.py file has no effect and .whl file is present.
        # The version number is anomalous.
    ): Confidence.MEDIUM,
    (
        HeuristicResult.FAIL,  # Empty Project
        HeuristicResult.SKIP,  # Source Code Repo
        HeuristicResult.FAIL,  # One Release
        HeuristicResult.SKIP,  # High Release Frequency
        HeuristicResult.SKIP,  # Unchanged Release
        HeuristicResult.FAIL,  # Closer Release Join Date
        HeuristicResult.FAIL,  # Suspicious Setup
        HeuristicResult.PASS,  # Wheel Absence
        HeuristicResult.FAIL,  # Anomalous Version
        # No project link, only one release, and the maintainer released it shortly
        # after account registration.
        # The setup.py file has no effect and .whl file is present.
        # The version number is anomalous.
    ): Confidence.MEDIUM,
    (
        HeuristicResult.FAIL,  # Empty Project
        HeuristicResult.SKIP,  # Source Code Repo
        HeuristicResult.FAIL,  # One Release
        HeuristicResult.SKIP,  # High Release Frequency
        HeuristicResult.SKIP,  # Unchanged Release
        HeuristicResult.FAIL,  # Closer Release Join Date
        HeuristicResult.SKIP,  # Suspicious Setup
        HeuristicResult.PASS,  # Wheel Absence
        HeuristicResult.FAIL,  # Anomalous Version
        # No project link, only one release, and the maintainer released it shortly
        # after account registration.
        # The setup.py file has no effect and .whl file is present.
        # The version number is anomalous.
    ): Confidence.MEDIUM,
```

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
art1f1c3R: Commit [442b4b4](https://github.com/oracle/macaron/pull/1031/commits/442b4b48bcc888fd09c97862b9e032e087925393) should now handle this, with a comment description in the `evaluate_heuristic_results` function body. 

The confidence supplied to the model is the probability of maliciousness. The code accumulates this by starting with a probability of the package being benign at `1.0`, and then multiplies the negation of the model probabilities that are triggered, resulting in the probability of the package being benign being reduced with each rule triggered. This accumulation is then negated again to get the probability of the package being malicious. This should handle when more than one rule is triggered, and ensure that the confidence in the package being malicious is higher when more rules are triggered.

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
art1f1c3R: This has been changed, see commit [442b4b4](https://github.com/oracle/macaron/pull/1031/commits/442b4b48bcc888fd09c97862b9e032e087925393) and the related comment below.

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
behnazh-w: I think in the previous implementation, we wanted to make sure if wheel absence does not fail, we still report as a malicious patterns as long as anomalicious version fails, there is one release, and quickUndetailed matches. With the new declarative approach, do we still need explicitly require `passed({Heuristics.WHEEL_ABSENCE.value})`?

art1f1c3R: | Month    | Savings |
| -------- | ------- |
| January  | $250    |
| February | $80     |
| March    | $420    |

art1f1c3R: I don't think the previous implementation would have caught that, as what we would have done was made two copies of the rule, one with wheel absence failing and one with it passing, but the previous implementation only had a copy of the rule for wheel absence passing. This may be a bug in the previous implementation though, as I do agree what should be happening is wheel absence has no effect, unless the rule had some other intention for not include the fail case. ProbLog won't require us to explicitly consider wheel absence for the rule if it has no effect on its outcome, so I'm happy to remove `passed({Heuristics.WHEEL_ABSENCE.value})` if you think it's more appropriate?

behnazh-w: > so I'm happy to remove `passed({Heuristics.WHEEL_ABSENCE.value})` if you think it's more appropriate?

Yes, that would make more sense. I think Problog now allows us to more effectively express our intent, so we can improve the rules regardless of the original version.

behnazh-w (dismissed):

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -138,18 +140,19 @@ def evaluate_heuristic_results(self, heuristic_results: dict[Heuristics, Heurist
 
         Returns
         -------
-        float | None
-            Returns the confidence associated with the detected malicious combination, otherwise None if no associated
-            malicious combination was triggered.
+        tuple[float, list[str]]
```
tromai, line 143: I think this should be `tuple[float, JsonType]` ?

tests/slsa_analyzer/checks/test_detect_malicious_metadata_check.py
```diff
@@ -98,3 +99,33 @@ def test_detect_malicious_metadata(
...
+            },
+            id="test_skipped_evaluation",
+        )
+    ],
+)
+def test_evaluations(combination: dict[Heuristics, HeuristicResult]) -> None:
+    """Test heuristic combinations to ensure they evaluate as expected."""
+    check = DetectMaliciousMetadataCheck()
+
+    confidence, _ = check.evaluate_heuristic_results(combination)
```
tromai, line 129: Is it possible to check for the value of triggered rules here, instead of ignoring it ?

tests/slsa_analyzer/checks/test_detect_malicious_metadata_check.py
```diff
@@ -98,3 +99,33 @@ def test_detect_malicious_metadata(
...
     assert check.run_check(ctx).result_type == expected
+
+
+@pytest.mark.parametrize(
+    ("combination"),
+    [
+        pytest.param(
+            {
+                # similar to rule ID malware_high_confidence_1, but SUSPICIOUS_SETUP is skipped since the file does not exist,
+                # so the rule should not trigger.
```
tromai, line 110: I'm not so clear what rule should not trigger here ?

tests/slsa_analyzer/checks/test_detect_malicious_metadata_check.py
art1f1c3R: Yes I can check to make sure it is empty.

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
```diff
@@ -321,51 +328,79 @@ def run_check(self, ctx: AnalyzeContext) -> CheckResultData:
...
-    % Heuristic groupings
+    % ----- Wrappers ------
+    % When a heuristic is skipped, it is ommitted from the problog model facts definition. This means that references in this
+    % static model must account for when they are not existent. These wrappers perform this function using the inbuilt try_call
+    % problog function. It will try to evaluate the provided logic, and return false if it encounters an error, such as the fact
+    % not being defined. For example, you are expecting A to pass, so we do:
+    %
+    % passed(A)
+    %
+    % If A was 'true', then this will return true, as A did pass. If A was 'false', then this will return false, as A did not pass.
```
tromai, line 343: I wonder if it would be good to have a minimal example (if it's not available yet) to demonstrate how SKIP, PASS and FAILED handle a rule (e.g. quickUndetailed) status differently 🤔

tests/slsa_analyzer/checks/test_detect_malicious_metadata_check.py
art1f1c3R: No rule should trigger, it is using a combination that is similar to malware_high_confidence_1 but has replaced one of the failed results with a skip, to ensure that skips are not handled as fails (which is what the bug was).

src/macaron/slsa_analyzer/checks/detect_malicious_metadata_check.py
art1f1c3R: So an example for `quickUndetailed` would be:

- `EMPTY_PROJECT_LINK = FAIL` and `CLOSER_RELEASE_JOIN_DATE = FAIL` means `not H` can be evaluated for both. `not false` would return true for both, meaning both failed as expected. This would cause `quickUndetailed` to evaluate to true.
-  `EMPTY_PROJECT_LINK = SKIP` and `CLOSER_RELEASE_JOIN_DATE = FAIL` means `not H` can be evaluated for `CLOSER_RELEASE_JOIN_DATE` and would return true, but it cannot be evaluated for `EMPTY_PROJECT_LINK` so it would return false. This would cause `quickUndetailed` to evaluate to false.
-  `EMPTY_PROJECT_LINK = TRUE` and `CLOSER_RELEASE_JOIN_DATE = FAIL` means means `not H` can be evaluated for both, but would return false for `EMPTY_PROJECT_LINK` and true for `CLOSER_RELEASE_JOIN_DATE`. This would cause `quickUndetailed` to evaluate to false.
- `EMPTY_PROJECT_LINK = FAIL` and `CLOSER_RELEASE_JOIN_DATE = SKIP` means `not H` can be evaluated for `EMPTY_PROJECT_LINK` and would return true, but it cannot be evaluated for `CLOSER_RELEASE_JOIN_DATE` so it would return false. This would cause `quickUndetailed` to evaluate to false.
- `EMPTY_PROJECT_LINK = FAIL` and `CLOSER_RELEASE_JOIN_DATE = TRUE` means means `not H` can be evaluated for both, but would return true for `EMPTY_PROJECT_LINK` and false for `CLOSER_RELEASE_JOIN_DATE`. This would cause `quickUndetailed` to evaluate to false.
- `EMPTY_PROJECT_LINK = SKIP` and `CLOSER_RELEASE_JOIN_DATE = SKIP` means `not H` cannot be evaluated for either, meaning both would return false. This would cause `quickUndetailed` to evaluate to false.
- `EMPTY_PROJECT_LINK = TRUE` and `CLOSER_RELEASE_JOIN_DATE = TRUE` means means `not H` can be evaluated for both, but would return false for both of them. This would cause `quickUndetailed` to evaluate to false.

So `quickUndetailed` is only ever true when both heuristics fail.

behnazh-w (approved):

tromai (approved):
LGTM. Thanks for the fix.

PR #1067, opened 2025-04-24
feat: add pypi attestation discovery

### LINKED ISSUE #947: Obtain PyPI Publish Attestation
We need to obtain PyPI Publish Attestations when available to more accurately identify publishing workflows. We could obtain this info from deps.dev. See this package as an example: https://deps.dev/pypi/ultralytics

The provenance file should also be obtainable from PyPI: https://pypi.org/integrity/package-name/version/wheel-name/provenance, e.g., https://pypi.org/integrity/ultralytics/8.3.70/ultralytics-8.3.70-py3-none-any.whl/provenance

benmss:
## Summary

This PR adds discovery of PyPI attestation. URLs to these attestation files are sought via the deps.dev API.

## Description of changes

- `DepsDevRepoFinder` was updated to use the `DepsDevService`, ensuring consistent and easily configurable use of the API
- Tests were added for `DepsDevRepoFinder` functions (they were not added previously), including for the functions that PyPI attestation discovery relies upon.
- PyPI attestations do not have a predicate. The `pypi-attestation` is used to extract information from the attestation certificate. This information is coerced into a predicate for use elsewhere within Macaron.
- Addition of an integration test case using the `ultralytics` Python library as its target.

## Related issues

Closes #947

### DIFF
--- a/src/macaron/provenance/provenance_extractor.py
+++ b/src/macaron/provenance/provenance_extractor.py
@@ -43,8 +43,11 @@ def extract_repo_and_commit_from_provenance(payload: InTotoPayload) -> tuple[str
         If the extraction process fails for any reason.
     """
     predicate_type = payload.statement.get("predicateType")
-    if isinstance(payload, InTotoV1Payload) and predicate_type == "https://slsa.dev/provenance/v1":
-        return _extract_from_slsa_v1(payload)
+    if isinstance(payload, InTotoV1Payload):
+        if predicate_type == "https://slsa.dev/provenance/v1":
+            return _extract_from_slsa_v1(payload)
+        if predicate_type == "https://docs.pypi.org/attestations/publish/v1":
+            return _extract_from_pypi_v1(payload)
 
     if isinstance(payload, InTotoV01Payload):
         if predicate_type == "https://slsa.dev/provenance/v0.2":
@@ -195,6 +198,29 @@ def _extract_from_slsa_v1(payload: InTotoV1Payload) -> tuple[str | None, str | N
     return repo, commit or None
 
 
+def _extract_from_pypi_v1(payload: InTotoV1Payload) -> tuple[str | None, str | None]:
+    """Extract the repository and commit metadata from the pypi provenance file found at the passed path.
+
+    Parameters
+    ----------
+    payload: InTotoPayload
+        The payload to extract from.
+
+    Returns
+    -------
+    tuple[str, str]
+        The repository URL and commit hash if found, a pair of empty strings otherwise.
+    """
+    predicate: dict[str, JsonType] | None = payload.statement.get("predicate")
+    if not predicate:
+        logger.debug("No predicate in payload statement.")
+        return None, None
+
+    repo = json_extract(predicate, ["sourceUri"], str)
+    digest = json_extract(predicate, ["sourceDigest"], str)
+    return repo, digest
+
+
 def _extract_from_witness_provenance(payload: InTotoV01Payload) -> tuple[str | None, str | None]:
     """Extract the repository and commit metadata from the witness provenance file found at the passed path.
 
@@ -300,7 +326,7 @@ def check_if_input_purl_provenance_conflict(
     provenance_repo_url: str | None,
     purl: PackageURL,
 ) -> bool:
-    """Test if the input repository type PURL's repo and commit match the contents of the provenance.
+    """Test if the input repository type PURL's repo matches the contents of the provenance.
 
     Parameters
     ----------
@@ -620,6 +646,41 @@ class WitnessGitLabBuildDefinitionV01(ProvenanceBuildDefinition):
         return gl_workflow, gl_job_url
 
 
+class PyPICertificateDefinition(ProvenanceBuildDefinition):
+    """Class representing the derived PyPI certificate build definition.
+
+    This class implements the abstract methods from the `ProvenanceBuildDefinition`
+    to extract build invocation details specific to the GitHub Actions build type.
+    """
+
+    #: Determines the expected ``buildType`` field in the provenance predicate.
+    expected_build_type = "pypi_certificate"
+
+    def get_build_invocation(self, statement: InTotoV01Statement | InTotoV1Statement) -> tuple[str | None, str | None]:
+        """Retrieve the build invocation information from the given statement.
+
+        Parameters
+        ----------
+        statement : InTotoV1Statement | InTotoV01Statement
+            The provenance statement from which to extract the build invocation
+            details. This statement contains the metadata about the build process
+            and its associated artifacts.
+
+        Returns
+        -------
+        tuple[str | None, str | None]
+            A tuple containing two elements:
+            - The first element is the build invocation entry point (e.g., workflow name), or None if not found.
+            - The second element is the invocation URL or identifier (e.g., job URL), or None if not found.
+        """
+        if statement["predicate"] is None:
+            return None, None
+
+        gha_workflow = json_extract(statement["predicate"], ["workflow"], str)
+        invocation_url = json_extract(statement["predicate"], ["invocationUrl"], str)
+        return gha_workflow, invocation_url
+
+
 class ProvenancePredicate:
     """Class providing utility methods for handling provenance predicates.
 
@@ -685,6 +746,7 @@ class ProvenancePredicate:
             SLSAGCBBuildDefinitionV1(),
             SLSAOCIBuildDefinitionV1(),
             WitnessGitLabBuildDefinitionV01(),
+            PyPICertificateDefinition(),
         ]
 
         for build_def in build_defs:
--- a/src/macaron/provenance/provenance_extractor.py
+++ b/src/macaron/provenance/provenance_extractor.py
@@ -201,6 +201,9 @@ def _extract_from_slsa_v1(payload: InTotoV1Payload) -> tuple[str | None, str | N
 def _extract_from_pypi_v1(payload: InTotoV1Payload) -> tuple[str | None, str | None]:
     """Extract the repository and commit metadata from the pypi provenance file found at the passed path.
 
+    This payload represents a custom predicate created from the certificate of a PyPI v1 attestation file.
+    By design, these attestations come without a predicate.
+
     Parameters
     ----------
     payload: InTotoPayload
--- a/src/macaron/provenance/provenance_finder.py
+++ b/src/macaron/provenance/provenance_finder.py
@@ -293,11 +293,6 @@ def find_pypi_provenance(purl: PackageURL) -> list[InTotoPayload]:
     -------
     list[InTotoPayload] | None
         The provenance payload if found, or an empty list otherwise.
-
-    Raises
-    ------
-    ProvenanceAvailableException
-        If the discovered provenance file size exceeds the configured limit.
     """
     attestation, verified = DepsDevRepoFinder.get_attestation(purl)
     if not attestation:
--- a/src/macaron/provenance/provenance_finder.py
+++ b/src/macaron/provenance/provenance_finder.py
@@ -2,6 +2,7 @@
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 """This module contains methods for finding provenance files."""
+import json
 import logging
 import os
 import tempfile
@@ -12,6 +13,7 @@ from pydriller import Git
 
 from macaron.config.defaults import defaults
 from macaron.repo_finder.commit_finder import AbstractPurlType, determine_abstract_purl_type
+from macaron.repo_finder.repo_finder_deps_dev import DepsDevRepoFinder
 from macaron.slsa_analyzer.analyze_context import AnalyzeContext
 from macaron.slsa_analyzer.checks.provenance_available_check import ProvenanceAvailableException
 from macaron.slsa_analyzer.ci_service import GitHubActions
@@ -76,6 +78,10 @@ class ProvenanceFinder:
                 return []
 
             discovery_functions = [partial(find_gav_provenance, purl, self.jfrog_registry)]
+            return self._find_provenance(discovery_functions)
+
+        if purl.type == "pypi":
+            discovery_functions = [partial(find_pypi_provenance, purl)]
             return self._find_provenance(discovery_functions)
 
         # TODO add other possible discovery functions.
@@ -273,6 +279,42 @@ def find_gav_provenance(purl: PackageURL, registry: JFrogMavenRegistry) -> list[
 
     # We assume that there is only one provenance per GAV.
     return provenances[:1]
+
+
+def find_pypi_provenance(purl: PackageURL) -> list[InTotoPayload]:
+    """Find and download the PyPI based provenance for the passed PURL.
+
+    Parameters
+    ----------
+    purl: PackageURL
+        The PURL of the analysis target.
+
+    Returns
+    -------
+    list[InTotoPayload] | None
+        The provenance payload if found, or an empty list otherwise.
+
+    Raises
+    ------
+    ProvenanceAvailableException
+        If the discovered provenance file size exceeds the configured limit.
+    """
+    attestation, verified = DepsDevRepoFinder.get_attestation(purl)
+    if not attestation:
+        return []
+
+    with tempfile.TemporaryDirectory() as temp_dir:
+        file_name = os.path.join(temp_dir, f"{purl.name}")
+        with open(file_name, "w", encoding="utf-8") as file:
+            json.dump(attestation, file)
+
+        try:
+            payload = load_provenance_payload(file_name)
+            payload.verified = verified
+            return [payload]
+        except LoadIntotoAttestationError as load_error:
+            logger.error("Error while loading provenance: %s", load_error)
+            return []
 
 
 def find_provenance_from_ci(
--- a/src/macaron/repo_finder/repo_finder_deps_dev.py
+++ b/src/macaron/repo_finder/repo_finder_deps_dev.py
@@ -188,7 +188,11 @@ class DepsDevRepoFinder(BaseRepoFinder):
         if not result:
             return None, False
 
-        result_attestations = json_extract(result, ["attestations"], list)
+        attestation_keys = ["attestations"]
+        if "version" in result:
+            attestation_keys.insert(0, "version")
+
+        result_attestations = json_extract(result, attestation_keys, list)
         if not result_attestations:
             logger.debug("No attestations in result.")
             return None, False
--- a/src/macaron/repo_finder/repo_finder_deps_dev.py
+++ b/src/macaron/repo_finder/repo_finder_deps_dev.py
@@ -181,6 +181,8 @@ class DepsDevRepoFinder(BaseRepoFinder):
                 return None, False
             purl = latest_purl
 
+        # Example of a PURL endpoint for deps.dev with '/' encoded as '%2F':
+        # https://api.deps.dev/v3alpha/purl/pkg:npm%2F@sigstore%2Fmock@0.7.5
         purl_endpoint = DepsDevService().get_purl_endpoint(purl)
         target_url = urllib.parse.urlunsplit(purl_endpoint)
 
--- a/src/macaron/repo_finder/repo_finder_deps_dev.py
+++ b/src/macaron/repo_finder/repo_finder_deps_dev.py
@@ -17,6 +17,7 @@ from macaron.repo_finder.repo_finder_base import BaseRepoFinder
 from macaron.repo_finder.repo_finder_enums import RepoFinderInfo
 from macaron.repo_finder.repo_validator import find_valid_repository_url
 from macaron.slsa_analyzer.git_url import clean_url
+from macaron.slsa_analyzer.package_registry import PyPIRegistry
 from macaron.slsa_analyzer.package_registry.deps_dev import DepsDevService
 from macaron.util import send_get_http, send_get_http_raw
 
@@ -55,7 +56,7 @@ class DepsDevRepoFinder(BaseRepoFinder):
             A tuple of the found URL (or an empty string), and the outcome of the Repo Finder.
         """
         try:
-            json_data = DepsDevService.get_package_info(encode(str(purl), safe=""))
+            json_data = DepsDevService.get_package_info(str(purl))
         except APIAccessError:
             return "", RepoFinderInfo.DDEV_API_ERROR
 
@@ -93,7 +94,7 @@ class DepsDevRepoFinder(BaseRepoFinder):
 
         project_key = clean_repo_url.hostname + clean_repo_url.path
 
-        api_endpoint = DepsDevService.get_endpoint(purl=False, path=f"projects/{encode(project_key, safe='')}")
+        api_endpoint = DepsDevService.get_endpoint(f"projects/{encode(project_key, safe='')}")
         request_url = urllib.parse.urlunsplit(api_endpoint)
 
         response = send_get_http_raw(request_url)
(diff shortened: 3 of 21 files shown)


src/macaron/provenance/provenance_extractor.py
```diff
@@ -195,6 +198,29 @@ def _extract_from_slsa_v1(payload: InTotoV1Payload) -> tuple[str | None, str | N
...
+    -------
+    tuple[str, str]
+        The repository URL and commit hash if found, a pair of empty strings otherwise.
+    """
+    predicate: dict[str, JsonType] | None = payload.statement.get("predicate")
+    if not predicate:
+        logger.debug("No predicate in payload statement.")
+        return None, None
+
+    repo = json_extract(predicate, ["sourceUri"], str)
```
tromai, line 219: From the documentation for the predicate of `https://docs.pypi.org/attestations/publish/v1/`, currently its content is not defined. I wonder if the `sourceUri` and `sourceDigest` was copied from an existing implementation or it's purposely put here to catch these data if they are available in the future?
I have decoded the statement of https://pypi.org/integrity/ultralytics/8.3.119/ultralytics-8.3.119.tar.gz/provenance and got 
```
{
  "_type": "https://in-toto.io/Statement/v1",
  "subject": [
    {
      "name": "ultralytics-8.3.119.tar.gz",
      "digest": {
        "sha256": "497bdcf3eb1beb082f451d42e5af2a6af944693a5991c78a9b9b0ce538593153"
      }
    }
  ],
  "predicateType": "https://docs.pypi.org/attestations/publish/v1",
  "predicate": null
}
```

benmss: I may have confused things by converting the certificate information into a predicate and attaching it to the pypi spec that has none. The type and the related function could be renamed to better distinguish them from the actual spec. The actual field names within the substitute predicate were chosen for simplicity, and loosely based on existing predicates I suppose.

src/macaron/slsa_analyzer/package_registry/deps_dev.py
```diff
@@ -20,28 +22,24 @@ class DepsDevService:
     """The deps.dev service class."""
 
     @staticmethod
-    def get_package_info(purl: str) -> dict | None:
-        """Check if the package identified by the PackageURL (PURL) exists and return its information.
+    def get_endpoint(purl: bool = True, path: str | None = None) -> Any:
```
tromai, line 25: Is there any reason why we have to use `Any` here instead of `urllib.parse.SplitResult`

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -11,7 +11,7 @@
 import zlib
 from urllib.parse import urlparse
 
-from pypi_attestations import Attestation
+from cryptography import x509
```
tromai, line 14: I think we should add this into the dependency section in `pyproject.toml`.

tromai: Am I correct to assume that we don't need the `pypi_attestations` library anymore ? 
If that is True we might want to remove `pypi_attestations` from `pyproject.toml` too.

src/macaron/provenance/provenance_finder.py
```diff
@@ -275,6 +281,42 @@ def find_gav_provenance(purl: PackageURL, registry: JFrogMavenRegistry) -> list[
...
+
+    Returns
+    -------
+    list[InTotoPayload] | None
+        The provenance payload if found, or an empty list otherwise.
+
+    Raises
+    ------
+    ProvenanceAvailableException
+        If the discovered provenance file size exceeds the configured limit.
```
tromai, line 300: I don't think this exception is raised anywhere in the body of this function.

src/macaron/slsa_analyzer/package_registry/deps_dev.py
```diff
@@ -50,34 +47,77 @@ def get_package_info(purl: str) -> dict | None:
...
+        dict
+            The package metadata.
+
+        Raises
+        ------
+        APIAccessError
+            If the service is misconfigured, the API is invalid, a network error happens,
+            or unexpected response is returned by the API.
+        """
+        if "%" in purl:
```
tromai, line 108: It would be good to have a comment or an example here to explain why we need to convert all "%xx" to their character equivalent. 
If I understand it correctly, if we don't convert "%xx", the call to quote / encode below will convert % to "%25", which is not what we want.

src/macaron/slsa_analyzer/package_registry/deps_dev.py
```diff
@@ -50,34 +47,77 @@ def get_package_info(purl: str) -> dict | None:
...
+            purl = decode(purl)
+        purl = encode(purl, safe="")
```
tromai, line 110: I wonder if we can use `decode` even if `purl` doesn't have any `%xx`? 
If it's possible to do so it can be simplified to
```
convert_escape_char_purl = decode(purl)
encoded_purl = encode(convert_escape_char_purl, safe="")
```
If not, I'm happy to keep the if statement, but perhaps it's better if we use a different variable to stored the mutated version of the original purl string
```
        convert_escape_char_purl = decode(purl) if "%" in purl else purl
        encoded_purl = encode(convert_escape_char_purl, safe="")
```

src/macaron/slsa_analyzer/package_registry/deps_dev.py
```diff
@@ -20,28 +21,24 @@ class DepsDevService:
...
         -------
-        dict | None
-            The package metadata or None if it doesn't exist.
-
-        Raises
-        ------
-        APIAccessError
-            If the service is misconfigured, the API is invalid, a network error happens,
-            or unexpected response is returned by the API.
+        Any
```
tromai, line 36: This doc string part should be updated with the new return type hint too.

src/macaron/slsa_analyzer/package_registry/deps_dev.py
```diff
@@ -20,28 +21,24 @@ class DepsDevService:
     """The deps.dev service class."""
 
     @staticmethod
-    def get_package_info(purl: str) -> dict | None:
-        """Check if the package identified by the PackageURL (PURL) exists and return its information.
+    def get_endpoint(purl: bool = True, path: str | None = None) -> urllib.parse.SplitResult:
```
tromai, line 24: I have a couple of comments about this function
- Because we have two types of end point for different purposes, the PURL endpoint `https://api.deps.dev/v3alpha/purl/` and the BASE endpoint `https://api.deps.dev/v3alpha/`. I wonder if we could separate them into 2 functions?
```
def get_base_endpoint(...)
def get_purl_endpoint(...)
```
- Could we add a small comment (may be as a doc string) to explicitly say that `path` is added directly to the end of the URL endpoint without any modification? - This is just to prevent using this method without percent encoding `path` before hand.

src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -177,61 +167,80 @@ def _create_urls(self, purl: PackageURL) -> tuple[list[str], RepoFinderInfo]:
...
-        str :
-            The retrieved file data or an empty string.
-        """
-        response = send_get_http_raw(url, {})
-
-        if not response:
-            return ""
-
-        return response.text
+        api_endpoint = DepsDevService.get_endpoint(
```
tromai, line 182: Could we add a small example for what the api end point should look like here ?
```
https://api.deps.dev/v3alpha/systems/pypi/packages/ultralytics/versions/8.3.70
```

src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -164,11 +156,9 @@ def get_latest_version(purl: PackageURL) -> tuple[PackageURL | None, RepoFinderI
             RepoFinderInfo.FOUND_FROM_LATEST,
         )
 
-    def _create_urls(self, purl: PackageURL) -> tuple[list[str], RepoFinderInfo]:
-        """
-        Create the urls to search for the metadata relating to the passed artifact, and report on that process.
-
-        If a version is not specified, remote API calls will be used to try and find one.
+    @staticmethod
+    def get_attestation(purl: PackageURL) -> tuple[dict | None, bool]:
```
tromai, line 160: Is this function intended to support extracting the attestation for all type of PURL or only PYPI one ? 
I think this function assumes that the link to the attestation provided by deps.dev (e.g. https://pypi.org/integrity/ultralytics/8.3.70/ultralytics-8.3.70-py3-none-any.whl/provenance) to a PYPI provenance.

src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -177,61 +167,80 @@ def _create_urls(self, purl: PackageURL) -> tuple[list[str], RepoFinderInfo]:
...
+            return None, False
+
+        return attestations[0], json_extract(result_attestations, [0, "verified"], bool) or False
 
-    def _read_json(self, json_data: str) -> tuple[list[str], RepoFinderInfo]:
+    @staticmethod
+    def extract_links(json_data: dict) -> tuple[list[str], RepoFinderInfo]:
         """
-        Parse the deps.dev json file and extract the repository links.
+        Extract the repository links from  the deps.dev json data.
```
tromai, line 230: I have a naive question. Does the BASE endpoint and the PURL endpoint returns the JSON data with similar schema and content ?

src/macaron/repo_finder/repo_finder_deps_dev.py
```diff
@@ -177,61 +167,80 @@ def _create_urls(self, purl: PackageURL) -> tuple[list[str], RepoFinderInfo]:
...
         """
-        try:
-            parsed = json.loads(json_data)
-        except ValueError as error:
-            logger.debug("Failed to parse response from deps.dev: %s", error)
-            return [], RepoFinderInfo.DDEV_JSON_FETCH_ERROR
-
-        links_keys = ["version", "links"] if "version" in parsed else ["links"]
-        links = json_extract(parsed, links_keys, list)
+        links_keys = ["version", "links"] if "version" in json_data else ["links"]
```
tromai, line 242: I'm curious why there is some different in where the links are stored in the JSON data? Is that because different endpoints returns different schema or depending on the package we are searching 🤔 ?

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -83,6 +103,13 @@ def _load_provenance_file_content(
         # Some provenances, such as Witness may not include the DSSE envelope `dsseEnvelope`
         # property but contain its value directly.
         provenance_payload = provenance.get("payload", None)
+    if not provenance_payload:
+        # GitHub Attestation.
+        # TODO Check if old method (above) actually works.
```
tromai, line 108: I wonder why is there a change in the way we extract Github attestation?

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -103,9 +130,90 @@ def _load_provenance_file_content(
     if not isinstance(json_payload, dict):
         raise LoadIntotoAttestationError("The provenance payload is not a JSON object.")
 
+    if json_payload["predicate"]:
+        return json_payload
+
+    # For provenance without a predicate (e.g. PyPI), try to use the provenance certificate instead.
```
tromai, line 136: I think it would be good to add an if statement to check for predicateType being "https://docs.pypi.org/attestations/publish/v1". 
This is for recognizing if `predicate` being empty is expected (because it's a PYPI attestation) or not (e.g the content if malformed).

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -103,9 +130,90 @@ def _load_provenance_file_content(
...
+    json_payload["predicate"] = pypi_predicate
     return json_payload
 
 
+def get_x509_certificate_values(decoded_certificate: bytes) -> dict:
+    """Retrieve the values of interest from an x509 certificate.
+
+    Parameters
+    ----------
+    decoded_certificate: bytes
```
tromai, line 164: We assume that the format of this certificate is Distinguished Encoding Rules (DER) following https://peps.python.org/pep-0740/#provenance-objects. I think we should mention this assumption here for extra information. 
I also think we can rename `decoded_certificate` to `x509_der_certificate`, because the decoded seems to not be in scope of this function (it has been handled in line `_load_provenance_file_content`)

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -103,9 +130,90 @@ def _load_provenance_file_content(
...
+        The decoded certificate bytes.
+
+    Returns
+    -------
+    dict
```
tromai, line 169: Even though the return type of this function is a dictionary, it only contains a fix number of keys:
- source_repo
- source_digest
- workflow
- invocation
Do you think it would be worth capturing this in a `TypedDict`. Or instead we could return a instance of `PyPICertificatePredicate` from this function if the dictionary is not used elsewhere.

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -103,9 +130,90 @@ def _load_provenance_file_content(
...
+        If the values could not be extracted.
+    """
+    certificate = x509.load_der_x509_certificate(decoded_certificate)
+    certificate_claims = {}
+    for extension in certificate.extensions:
+        if extension.oid.dotted_string not in _OID_IDS:
+            continue
+
+        claim_name = _OID_IDS[extension.oid.dotted_string]
+        certificate_claims[claim_name] = extension.value.value
```
tromai, line 184: Are we assuming that `extension` here is always of type [cryptography.x509.UnrecognizedExtension](https://cryptography.io/en/latest/x509/reference/#cryptography.x509.UnrecognizedExtension) because of the OIDs we are looking for are not part of the standard X.509 extension type. 
If true, I think it's good to have a small comment about it.

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -103,9 +130,90 @@ def _load_provenance_file_content(
...
+
+    for name in _OID_NAMES:
```
tromai, line 186: I currently only see `_OID_NAMES`'s keys being used, and not its values. I wonder if we can either simplify `_OID_NAMES` to a list, or reuse `_OID_IDS.values()` for this loop ?

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -103,9 +130,90 @@ def _load_provenance_file_content(
...
+        A dictionary of the extracted values.
+
+    Raises
+    ------
+    ValueError
+        If the values could not be extracted.
+    """
+    certificate = x509.load_der_x509_certificate(decoded_certificate)
+    certificate_claims = {}
+    for extension in certificate.extensions:
```
tromai, line 179: I think assessing `certificate.extensions` can raise Error - https://cryptography.io/en/latest/x509/reference/#cryptography.x509.Certificate.extensions
- [cryptography.x509.DuplicateExtension](https://cryptography.io/en/latest/x509/reference/#cryptography.x509.DuplicateExtension)
- [cryptography.x509.UnsupportedGeneralNameType](https://cryptography.io/en/latest/x509/reference/#cryptography.x509.UnsupportedGeneralNameType)

src/macaron/slsa_analyzer/provenance/loader.py
tromai, line 186: I wonder if `_OID_NAMES` is ever used as a dictionary? If not, we can turn it into a list or reuse `_OID_IDS.values`.

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -103,9 +130,90 @@ def _load_provenance_file_content(
...
+
+        claim_name = _OID_IDS[extension.oid.dotted_string]
+        certificate_claims[claim_name] = extension.value.value
+
+    for name in _OID_NAMES:
+        if name not in certificate_claims:
+            raise ValueError(f"Missing certificate value: {name}")
+
+        # Values are DER encoded UTF-8 strings. Removing the first two bytes seems to be sufficient.
+        value: str = certificate_claims[name][2:].decode("UTF-8")
```
tromai, line 191: Could we perform this operation in the first loop ?
```
for extension in certificate.extensions:
    ...
```
I think the purpose of this loop is for validating `certificate_claims` content, so in my opinion, it's best if we don't do extra mutation.

tests/repo_finder/test_repo_finder_deps_dev.py
```diff
@@ -0,0 +1,192 @@
...
+        + f"{'/'.join(['systems', purl.type, 'packages', purl.name, 'versions', purl.version or ''])}"
+    )
+    attestation_url = f"{deps_dev_service_mock['base_scheme']}://{deps_dev_service_mock['base_netloc']}{target_url}"
+    data = """
+        {
+            "attestations": [{"url": "*replace_url*", "verified": true}],
+            "attestation_bundles": [{"attestations": [{"foo": "bar"}]}]
+        }
+    """
+    data = data.replace("*replace_url*", attestation_url)
```
tromai, line 188: I wonder if we can use a formatted multi-line string for `data`?

tests/repo_finder/test_repo_finder_deps_dev.py
```diff
@@ -0,0 +1,192 @@
...
+    result, _ = DepsDevRepoFinder().get_attestation(purl)
+    assert not result
+
+
+def test_get_attestation_success(httpserver: HTTPServer, deps_dev_service_mock: dict) -> None:
+    """Test get attestation success."""
+    purl = PackageURL.from_string("pkg:pypi/test@3")
+    target_url = (
+        f"/{deps_dev_service_mock['api']}/"
+        + f"{'/'.join(['systems', purl.type, 'packages', purl.name, 'versions', purl.version or ''])}"
```
tromai, line 179: I wonder why we have `purl.version or ''` here. I think `purl` is not modified at all in this test.

src/macaron/provenance/provenance_extractor.py
tromai: Oh okay, I can see the big picture now. I'm okay with some extra documentation here to indicate that we are relying on a Predicate that we built ourselves, and that it's not reflecting the real predicate format from Pypi (as there is None).

src/macaron/repo_finder/repo_finder_deps_dev.py
benmss: It is currently only intended for PyPI. I have moved the PyPI specific extraction code to the related registry, and will report unsupported for other types.

src/macaron/repo_finder/repo_finder_deps_dev.py
benmss: They return slightly different results in terms of the schema.
https://api.deps.dev/v3alpha/purl/pkg%3Anpm%2F%40colors%2Fcolors@1.4.0
https://api.deps.dev/v3alpha/systems/npm/packages/%40colors%2Fcolors/versions/1.4.0

src/macaron/repo_finder/repo_finder_deps_dev.py
benmss: As above, yes.

src/macaron/slsa_analyzer/provenance/loader.py
benmss: Perhaps @behnazh-w could weigh in on this?

tests/repo_finder/test_repo_finder_deps_dev.py
```diff
@@ -0,0 +1,192 @@
...
+) -> None:
+    """Test get project info failures."""
+    if not repo_url:
+        repo_url = f"{deps_dev_service_mock['base_scheme']}://{deps_dev_service_mock['base_netloc']}{server_url}"
+
+    if server_url:
+        target_url = f"/{deps_dev_service_mock['api']}/projects/{deps_dev_service_mock['base_hostname']}{server_url}"
+        httpserver.expect_request(target_url).respond_with_data(data)
+
+    assert not DepsDevRepoFinder().get_project_info(repo_url)
```
tromai, line 79: I think there are 2 cases of failures we are testing in this function
- The Repository URL is given, and not valid
- The repository URL is available (from pytest-httpserver) but the result JSON is invalid

Could we have 2 separated test functions for each case, so that we can simplify the if conditions on the test inputs.

tromai (changes requested):
I have finished my round of review. Thank you.

src/macaron/repo_finder/repo_finder_deps_dev.py
tromai: Ah, I see. Thanks

tests/repo_finder/test_repo_finder_deps_dev.py
benmss: I have separated the two test cases.

tests/repo_finder/test_repo_finder_deps_dev.py
benmss: This has been removed as part of using PURL for attestation instead of the BASE endpoint.

src/macaron/slsa_analyzer/provenance/loader.py
benmss: It was redundant and has been removed.

src/macaron/slsa_analyzer/provenance/loader.py
```diff
@@ -96,16 +118,106 @@ def _load_provenance_file_content(
...
     except (json.JSONDecodeError, TypeError) as error:
-        raise LoadIntotoAttestationError(
-            "Cannot deserialize the provenance payload as JSON.",
-        ) from error
+        raise LoadIntotoAttestationError("Cannot deserialize the provenance payload as JSON.") from error
 
     if not isinstance(json_payload, dict):
         raise LoadIntotoAttestationError("The provenance payload is not a JSON object.")
 
+    if json_payload["predicate"]:
```
tromai, line 126: This could raise an Exception if `predicate` doesn't exist (which is can be the case for a malformed provenance document). 
I wonder if we could rewrite it into something like
```python
observed_predicate = json_extract(json_payload, ["predicate"], dict)
observed_predicate_type =  json_extract(json_payload, ["predicateType"], str)

if not observed_predicate_type:
    raise LoadIntotoAttestationError("Missing predicateType in payload.")

if observed_predicate_type is None: # no key "predicate"
    raise ...

# observed_predicate is not None but it could be empty.
if observed_predicate:
    if observed_predicate_type == "pypi attestation":
        raise ...

    return json_payload

# For provenance with an empty predicate (e.g. PyPI), try to use the provenance certificate instead.
...
```
This is with my assumption that the predicate object is available, but it's an empty dictionary *only* in the case that `predicate_type == "https://docs.pypi.org/attestations/publish/v1"`

tromai, line 126: Would there be a case where the `predicate` key doesn't exist within `json_payload`, as it could raise an Exception here.

src/macaron/slsa_analyzer/provenance/loader.py
tromai: Thanks. It looks okay to me. Just a small observation that the logic for checking whether `certificate_claims` contains enough claim names (e.g. source_repo, source_digest, etc.) has been removed in the latest commit. Just want to make sure that this is intended.

benmss: Good catch. I'll add it back in.

src/macaron/slsa_analyzer/provenance/loader.py
benmss: Probably best to be sure.

tromai (dismissed):
LGTM. Thanks for addressing the feedbacks.

behnazh-w (approved):

src/macaron/slsa_analyzer/provenance/loader.py
benmss: This was actually not relevant for this PR, so I've removed it.

behnazh-w (approved):

PR #1096, opened 2025-06-03
build: replace shared library with standalone cuevalidator binary

behnazh-w:
## Summary
This PR replaces the Go shared library previously used via C-bindings in Python with a standalone binary for the cuevalidator component. The binary can now be invoked as a `subprocess`, simplifying integration and improving portability.

## Description of changes

- `cuevalidator` is now compiled as a standalone Go binary instead of a shared library.
- Removed dependency on ctypes/C-bindings in Python code. Python now calls the `cuevalidator` binary using `subprocess`.
- This change removes the need for platform-specific shared library builds, reducing complexity in packaging and deployment.
- Updated the `Makefile` and instructions to build and distribute the binary as part of the project tooling.


src/macaron/config/global_config.py
```diff
@@ -94,12 +94,12 @@ def load_expectation_files(self, exp_path: str) -> None:
         exp_files = []
         if os.path.isdir(exp_path):
             for policy_path in os.listdir(exp_path):
-                policy_file_path = os.path.join(exp_path, policy_path)
+                policy_file_path = os.path.abspath(os.path.join(exp_path, policy_path))
```
tromai, line 97: I'm curious about the reason we need an absolute path here.

src/macaron/slsa_analyzer/provenance/expectations/cue/cue_validator.py
```diff
@@ -78,20 +76,36 @@ def validate_expectation(expectation: str | None, prov: JsonType) -> bool:
...
+            timeout=defaults.getint("cue_validator", "timeout", fallback=30),
+        )
+    except (
+        subprocess.CalledProcessError,
+        subprocess.TimeoutExpired,
+        FileNotFoundError,
+    ) as error:
+        raise CUERuntimeError("Unable to process CUE expectation or provenance.") from error
+
+    if result.returncode == 0:
```
tromai, line 105: From the doc string in `cmd/cuevalidator/cuevalidator.go`, the return code can indicates whether the validation passed or failed. I wonder if we could rely on that ?

tromai (commented):
I have leave some minor questions about the Pull Request.

src/macaron/slsa_analyzer/provenance/expectations/cue/cue_validator.py
behnazh-w: The return code will be `0` if no error happens. The value logged in the stdout, i.e., `True` and `False` will report the validation result.

tromai (approved):

PR #1098, opened 2025-06-04
docs: make minor formatting fixes

benmss:
## Summary

This PR fixes various formatting issues in comments throughout the code base.


art1f1c3R:
The changes in this PR seem fine to me, but I would like to try get an idea of how to reproduce the problem and how the rule ID path caused errors in the unit tests?

benmss:
> The changes in this PR seem fine to me, but I would like to try get an idea of how to reproduce the problem and how the rule ID path caused errors in the unit tests?

This comment relates to the semgrep fix that was previously included in this PR. Further investigation led to removal of this "fix".

behnazh-w (approved):

PR #1106, opened 2025-06-16
feat(heuristics): add Fake Email analyzer to validate maintainer email domain

AmineRaouane:
## Summary

This PR adds a new heuristic analyzer called `FakeEmailAnalyzer`. It verifies the validity of maintainer email addresses listed in a PyPI package by checking both the format and the existence of MX records for their domains. This helps detect packages with fake or throwaway emails, which are often indicative of malicious intent.

## Description of changes

- Implemented `FakeEmailAnalyzer` that:
  - Validates email format using a regex.
  - Verifies the existence of MX records for the email domain via DNS resolution.
- Updated `detect_malicious_metadata_check.py` to include and invoke this new analyzer.
- The analyzer handles DNS errors and skips analysis if no email is present.
- The logical reason for combining quickUndetailed with a failed(Heuristics.FAKE_EMAIL.value) is that a package that is rushed onto a platform by someone using a fake email address points to an actor who may be trying to quickly distribute a package while obscuring their identity and avoiding being investigated.

## Related issues

None

## Checklist

### DIFF
--- a/THIRD_PARTY_LICENSES.txt
+++ b/THIRD_PARTY_LICENSES.txt
@@ -13059,3 +13059,4403 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 Google deps.dev
 
 As well as aggregating data, deps.dev generates additional data, including resolved dependencies, advisory statistics, associations between entities, etc. This generated data is available under a CC-BY 4.0 license - https://creativecommons.org/licenses/by/4.0/.
+
+
+================================================================================
+Semgrep
+
+
+Copyright: Semgrep Inc.
+
+License: LGPL v.2.1
+
+ ./LICENSE
+
+GNU LESSER GENERAL PUBLIC LICENSE
+		       Version 2.1, February 1999
+
+ Copyright (C) 1991, 1999 Free Software Foundation, Inc.
+ 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+[This is the first released version of the Lesser GPL.  It also counts
+ as the successor of the GNU Library Public License, version 2, hence
+ the version number 2.1.]
+
+			    Preamble
+
+  The licenses for most software are designed to take away your
+freedom to share and change it.  By contrast, the GNU General Public
+Licenses are intended to guarantee your freedom to share and change
+free software--to make sure the software is free for all its users.
+
+  This license, the Lesser General Public License, applies to some
+specially designated software packages--typically libraries--of the
+Free Software Foundation and other authors who decide to use it.  You
+can use it too, but we suggest you first think carefully about whether
+this license or the ordinary General Public License is the better
+strategy to use in any particular case, based on the explanations below.
+
+  When we speak of free software, we are referring to freedom of use,
+not price.  Our General Public Licenses are designed to make sure that
+you have the freedom to distribute copies of free software (and charge
+for this service if you wish); that you receive source code or can get
+it if you want it; that you can change the software and use pieces of
+it in new free programs; and that you are informed that you can do
+these things.
+
+  To protect your rights, we need to make restrictions that forbid
+distributors to deny you these rights or to ask you to surrender these
+rights.  These restrictions translate to certain responsibilities for
+you if you distribute copies of the library or if you modify it.
+
+  For example, if you distribute copies of the library, whether gratis
+or for a fee, you must give the recipients all the rights that we gave
+you.  You must make sure that they, too, receive or can get the source
+code.  If you link other code with the library, you must provide
+complete object files to the recipients, so that they can relink them
+with the library after making changes to the library and recompiling
+it.  And you must show them these terms so they know their rights.
+
+  We protect your rights with a two-step method: (1) we copyright the
+library, and (2) we offer you this license, which gives you legal
+permission to copy, distribute and/or modify the library.
+
+  To protect each distributor, we want to make it very clear that
+there is no warranty for the free library.  Also, if the library is
+modified by someone else and passed on, the recipients should know
+that what they have is not the original version, so that the original
+author's reputation will not be affected by problems that might be
+introduced by others.
+
+  Finally, software patents pose a constant threat to the existence of
+any free program.  We wish to make sure that a company cannot
+effectively restrict the users of a free program by obtaining a
+restrictive license from a patent holder.  Therefore, we insist that
+any patent license obtained for a version of the library must be
+consistent with the full freedom of use specified in this license.
+
+  Most GNU software, including some libraries, is covered by the
+ordinary GNU General Public License.  This license, the GNU Lesser
+General Public License, applies to certain designated libraries, and
+is quite different from the ordinary General Public License.  We use
+this license for certain libraries in order to permit linking those
+libraries into non-free programs.
+
+  When a program is linked with a library, whether statically or using
+a shared library, the combination of the two is legally speaking a
+combined work, a derivative of the original library.  The ordinary
+General Public License therefore permits such linking only if the
+entire combination fits its criteria of freedom.  The Lesser General
+Public License permits more lax criteria for linking other code with
+the library.
+
+  We call this license the "Lesser" General Public License because it
+does Less to protect the user's freedom than the ordinary General
+Public License.  It also provides other free software developers Less
+of an advantage over competing non-free programs.  These disadvantages
+are the reason we use the ordinary General Public License for many
+libraries.  However, the Lesser license provides advantages in certain
+special circumstances.
+
+  For example, on rare occasions, there may be a special need to
+encourage the widest possible use of a certain library, so that it becomes
+a de-facto standard.  To achieve this, non-free programs must be
+allowed to use the library.  A more frequent case is that a free
+library does the same job as widely used non-free libraries.  In this
+case, there is little to gain by limiting the free library to free
+software only, so we use the Lesser General Public License.
+
+  In other cases, permission to use a particular library in non-free
+programs enables a greater number of people to use a large body of
+free software.  For example, permission to use the GNU C Library in
+non-free programs enables many more people to use the whole GNU
+operating system, as well as its variant, the GNU/Linux operating
+system.
+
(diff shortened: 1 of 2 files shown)


src/macaron/malware_analyzer/pypi_heuristics/metadata/fake_email.py
```diff
@@ -0,0 +1,91 @@
...
+    """Analyze the email address of the package maintainers."""
+
+    def __init__(self) -> None:
+        super().__init__(
+            name="fake_email_analyzer",
+            heuristic=Heuristics.FAKE_EMAIL,
+            depends_on=None,
+        )
+
+    def is_valid_email(self, email: str) -> bool:
```
benmss, line 30: Validating email addresses is a complex task. What we have here is more like a sanity check, verifying that the address is vaguely of the right format. That may be enough for the purpose of this check, in which case this method should be renamed and re-documented to make that clear. Alternatively, if we do really want to ensure that email addresses are valid, this method will need to be expanded considerably. @behnazh-w 
See the top two answers on this stackoverflow thread for more information: https://stackoverflow.com/questions/2049502/what-characters-are-allowed-in-an-email-address

benmss: As part of the above, I think the regex checking and dns resolution steps should be split into separate functions. This could also simplify the related tests.

benmss: Another alternative is to use the Python library `email-validator` to more formally validate emails. It also uses `dnspython`, but handles some of the more complicated validation aspects.

tests/malware_analyzer/pypi/test_fake_email.py
```diff
@@ -0,0 +1,142 @@
...
+    assert result == HeuristicResult.SKIP
+    assert info["message"] == "No author or maintainer email available."
+
+
+def test_analyze_skip_no_info_key(analyzer: FakeEmailAnalyzer, pypi_package_json_asset_mock: MagicMock) -> None:
+    """Test the analyzer skips if 'info' key is missing in PyPI data."""
+    pypi_package_json_asset_mock.package_json = {}  # No 'info' key
+    result, info = analyzer.analyze(pypi_package_json_asset_mock)
+    assert result == HeuristicResult.SKIP
+    assert info["message"] == "No author or maintainer email available."
```
benmss, line 52: I think it would be good to have the analyzer report on this case properly.

benmss: I mean adding a new message. So we have a case where the info is missing, and then a case where both expected email properties are missing from it.

pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
     "problog >= 2.2.6,<3.0.0",
     "cryptography >=44.0.0,<45.0.0",
     "semgrep == 1.113.0",
+    "email_validator >=2.2.0",
```
benmss, line 41:
```suggestion
    "email_validator >=2.2.0,<3.0.0",
```

benmss (dismissed):

src/macaron/malware_analyzer/pypi_heuristics/metadata/fake_email.py
```diff
@@ -0,0 +1,93 @@
...
+            The validated email object if the email is valid, otherwise None.
+
+        Raises
+        ------
+        HeuristicAnalyzerValueError
+            if the failure is due to DNS resolution.
+        """
+        emailinfo = None
+        try:
+            emailinfo = validate_email(email, check_deliverability=True)
```
behnazh-w, line 48: Can you please make `check_deliverability` configurable via `defaults.ini` so that you can turn it off in unit tests? We try to avoid network calls in our unit tests.

tests/malware_analyzer/pypi/test_fake_email.py
```diff
@@ -0,0 +1,142 @@
...
+    assert result == HeuristicResult.FAIL
+    assert info == {"email": invalid_email}
+    mock_validate_email.assert_called_once_with(invalid_email, check_deliverability=True)
+
+
+def test_analyze_pass_only_maintainer_email_valid(
+    analyzer: FakeEmailAnalyzer, pypi_package_json_asset_mock: MagicMock, mock_validate_email: MagicMock
+) -> None:
+    """Test analyzer passes when only maintainer_email is present and valid."""
+    email = "maintainer@example.net"
```
behnazh-w, line 74: is `example.net` supposed to be a valid domain?

tests/malware_analyzer/pypi/test_fake_email.py
```diff
@@ -0,0 +1,142 @@
...
+        "info": {"author_email": "author@example.com", "maintainer_email": "maintainer@example.net"}
+    }
+    result, info = analyzer.analyze(pypi_package_json_asset_mock)
+    assert result == HeuristicResult.PASS
+    assert mock_validate_email.call_count == 2
+
+    validated_emails = info.get("validated_emails")
+    assert isinstance(validated_emails, list)
+    assert len(validated_emails) == 2
+    assert {"normalized": "author@example.com", "local_part": "author", "domain": "example.com"} in validated_emails
```
behnazh-w, line 116: Same question, if `example.com` is supposed to be a valid domain.

behnazh-w (changes requested):

tests/malware_analyzer/pypi/test_fake_email.py
AmineRaouane: Yes and no. The domain is technically valid, but it’s reserved by IANA and not intended for real-world use. So to ensure the email is actually usable by a real user, Should I add a list of reserved domains and TLDs and check against them before proceeding with validation or acceptance.

pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
     "problog >= 2.2.6,<3.0.0",
     "cryptography >=44.0.0,<45.0.0",
     "semgrep == 1.113.0",
+    "email_validator >=2.2.0,<3.0.0",
```
art1f1c3R, line 41: Is this the package [email-validator](https://pypi.org/project/email-validator/)?  I can find a package spelled with a hiphen, `email-validator`, but I can't find one spelled with an underscore like you have here.

src/macaron/malware_analyzer/pypi_heuristics/metadata/fake_email.py
```diff
@@ -0,0 +1,93 @@
...
+        Returns
+        -------
+        tuple[HeuristicResult, dict[str, JsonType]]:
+            The result and related information collected during the analysis.
+        """
+        package_json = pypi_package_json.package_json
+        if not package_json.get("info", {}):
+            return HeuristicResult.SKIP, {"message": "No package info available."}
+
+        author_email = json_extract(package_json, ["info", "author_email"], str)
```
art1f1c3R, line 76: I see here we're extracting the email fields and then running the `is_valid_email` function directly on what is present in that field. Have you tested if this works on fields where the email is not the only text present? I've got some PyPI JSON data from Django where `author_email` looks like this:
```
"author_email": "Django Software Foundation <foundation@djangoproject.com>"
```
So the text includes more than just the email. Here's another example from the `black` package:
```
"author_email": "\u0141ukasz Langa <lukasz@langa.pl>"
```
It may also be a string of multiple emails as well, like the ultralytics package:
```
"author_email": "Glenn Jocher <glenn.jocher@ultralytics.com>, Jing Qiu <jing.qiu@ultralytics.com>"
```

tests/malware_analyzer/pypi/test_fake_email.py
AmineRaouane: Yeah, but I tried to test both cases when `deliverability` is enabled and when it's not. Setting `test_environment=False` actually makes `.test` and similar test domains valid, so it's similar to the case when `check_deliverability=False`.

src/macaron/malware_analyzer/pypi_heuristics/metadata/fake_email.py
AmineRaouane: done

tests/malware_analyzer/pypi/test_fake_email.py
```diff
@@ -0,0 +1,133 @@
...
+    author_email = "example@gmail.com"
+    author_local_part, author_domain = author_email.split("@")
+    maintainer_email = "maintainer@example.net"
+    maintainer_local_part, maintainer_domain = maintainer_email.split("@")
+
+    pypi_package_json_asset_mock.package_json = {
+        "info": {"author_email": author_email, "maintainer_email": maintainer_email}
+    }
+    result, info = analyzer.analyze(pypi_package_json_asset_mock)
+    if analyzer.check_deliverability:
```
art1f1c3R, line 99: What controls `analyzer.check_deliverability` here? It defaults to `True` right? So won't this branch always be run?

tests/malware_analyzer/pypi/test_fake_email.py
art1f1c3R: Right. `check_deliverability=True` will mean it'll make a network connection right? This could cause some issues with running the unit tests offline. Would it be possible to either mock a network response, or otherwise put this in an integration test?

tests/malware_analyzer/pypi/test_fake_email.py
AmineRaouane: Yes, it defaults to `True`, but I tried to handle both cases depending on what the user sets. So if the user explicitly sets `check_deliverability=False`, it will follow the other path and skip that branch.

behnazh-w: @AmineRaouane Thanks for adding the configuration. We should also make sure that the unit tests don't make a network call. That means we cannot test the functionality of this heuristic fully, and can only partially test it. We use integration tests for any scenario that requires network call.

So, let's turn off the `check_deliverability` for the unit tests. For this we need to overwrite the configuration and call `load_defaults()` as you can see here: https://github.com/oracle/macaron/blob/3548af099fb2a034627d9e5870e2d3514b3e4957/tests/slsa_analyzer/checks/test_detect_malicious_metadata_check.py#L82-L106

behnazh-w (dismissed):

art1f1c3R (approved):

behnazh-w (approved):

benmss (approved):

PR #1110, opened 2025-06-25
docs: update provenance tutorial

benmss:
## Summary

This PR updates the npm provenance discovery tutorial to include the newer discovery methods: PyPI and GitHub.

## Description of changes

- The tutorial is updated to provide examples of provenance discovery for PyPI and GitHub using the `toga` PyPI library, which has provenance of either type available for discovery depending on the exact version specified.
- For GitHub attestation, a small code update is included to allow the repository URL and commit digest to be found from provenance.
- The npm part of the tutorial is updated to use a newer version of the semver library `7.7.2`


docs/source/pages/tutorials/provenance.rst
```diff
@@ -5,9 +5,11 @@
...
 
-For npm artifacts, Macaron makes use of available features provided by `npm <https://npmjs.com/>`_. Most importantly, npm allows developers to generate provenance files when publishing their artifacts. The `semver <https://www.npmjs.com/package/semver>`_ package is chosen as an example for this tutorial.
+Currently, Macaron supports discovery of attestation for: npm artifacts using features provided by `npm <https://npmjs.com/>`_; PyPI artifacts using features provided by `Open Source Insights <https://deps.dev/>`_; and artifacts that have been published attestations to `GitHub <https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-attestations>`_. This tutorial uses two example packages to demonstrate these three discovery methods: The `semver <https://www.npmjs.com/package/semver>`_ npm package, and the `toga <https://pypi.org/pypi/toga>`_ PyPI package.
```
behnazh-w, line 10: 1. We also support provenances published as release assets on GitHub. Please add that as well. You can use [urllib3@2.0.0a1](https://github.com/oracle/macaron/tree/benmss/update-provenance-tutorial/tests/integration/cases/urllib3_expectation_file) which also has a CUE expectation.
2. You should also mention that we support provenance files if they’re provided as input, as long as they use one of the supported `predicateType` and `buildType` (see the full list here). For reference, witness provenance is one supported example. You can find more details in the command-line options documentation.
3. I think it would be clearer to list the supported ecosystems as bullet points for better readability.

docs/source/pages/tutorials/provenance.rst
```diff
@@ -34,46 +36,111 @@ Skip this section if you already know how to install Macaron.
...
+
+During this analysis, Macaron will attempt to discover a GitHub attestation by computing the hash of the relevant artifact. This is a requirement of GitHub's API to view artifact attestation, see the `GitHub Attestation API <https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-attestations>`_. The hash is computed by downloading the artifact and analysing it with the SHA256 algorithm. With the hash, the GitHub API can be called to find the related attestation.
+
+In this particular case, the SHA256 hash of the toga 0.4.8 artifact is 0814a72abb0a9a5f22c32cc9479c55041ec30cdf4b12d73a0017aee58f9a1f00. A GitHub attestation can be found for this artifact `here <https://api.github.com/repos/beeware/toga/attestations/sha256:0814a72abb0a9a5f22c32cc9479c55041ec30cdf4b12d73a0017aee58f9a1f00>`_.
+
+Attestation discovered through GitHub cannot be ascertained as verified at this time. However, we can still be sure that the repository URL and commit digest associated with the user provided PURL match what is found within the attestation. This is reported by Macaron in two checks: ``mcn_provenance_derived_repo_1`` and ``mcn_provenance_derived_commit_1``.
```
behnazh-w, line 123: All attestations published on GitHub are signed, and their signatures are verified. Therefore, the verification check should pass for GitHub attestations, as long as we retrieve them directly from GitHub.

docs/source/pages/tutorials/provenance.rst
```diff
@@ -34,46 +36,111 @@ Skip this section if you already know how to install Macaron.
...
+
+.. code-block:: shell
+
+  open output/reports/pypi/toga/toga.html
+
+**************************************
+Run ``verify-policy`` command (semver)
+**************************************
 
 Another feature of Macaron is policy verification. This allows Macaron to report on whether an artifact meets the security requirements specified by the user. Policies are written using `Soufflé Datalog <https://souffle-lang.github.io/index.html>`_ , a language similar to SQL. Results collected by the ``analyze`` command can be checked via declarative queries in the created policy, which Macaron can then automatically check.
```
behnazh-w, line 141:
```suggestion
Another feature of Macaron is policy verification, which allows it to assess whether an artifact meets user-defined security requirements. This feature can also be integrated into CI/CD pipelines to automatically check policy compliance by returning appropriate error codes based on pass or fail status. Policies are written using `Soufflé Datalog <https://souffle-lang.github.io/index.html>`_ , a language similar to SQL. Results collected by the ``analyze`` command can be checked via declarative queries in the created policy, which Macaron can then automatically check.
```

src/macaron/provenance/provenance_finder.py
```diff
@@ -490,3 +499,96 @@ def download_provenances_from_ci_service(ci_info: CIInfo, download_path: str) ->
...
+        if artifact_hash:
+            return artifact_hash
+
+        source_url = pypi_asset.get_sourcecode_url("bdist_wheel")
+        if not source_url:
+            return None
+
+        return pypi_registry.get_artifact_hash(source_url)
+
+    logger.debug("Purl type '%s' not yet supported for GitHub attestation discovery.", purl.type)
```
tromai, line 593: I wonder if we should change this debug message as this function doesn't handle attestation?

src/macaron/provenance/provenance_verifier.py
```diff
@@ -249,8 +249,7 @@ def _find_subject_asset(
     for item in all_assets:
         item_path = os.path.join(download_path, item["name"])
         # Make sure to download an archive just once.
-        if not Path(item_path).is_file():
-            # TODO: check that it's not too large.
+        if not Path(item_path).exists():
```
tromai, line 252: Could you let me know why we moved away from accepting only files ?

src/macaron/slsa_analyzer/analyzer.py
```diff
@@ -546,14 +544,13 @@ def run_single(
...
-                        for ci_info in analyze_ctx.dynamic_data["ci_services"]:
-                            verified.append(verify_ci_provenance(analyze_ctx, ci_info, temp_dir))
-                            if not verified:
-                                break
-                        if verified and all(verified):
-                            provenance_l3_verified = True
+                    verified = []
+                    for ci_info in analyze_ctx.dynamic_data["ci_services"]:
+                        verified.append(verify_ci_provenance(analyze_ctx, ci_info, temp_dir))
+                        if not verified:
```
tromai, line 550: I have a small comment on this part, this might not be relevant to this PR because you only remove the `if verify_provenance` condition and keep the body intact. 
`verify_ci_provenance` returns a bool, therefore `verified` at line 550 is always not empty. I wonder in which case is `verified` empty at line 550 ?

benmss: It seems this should have been changed a while ago. I will update it so it checks the most recently added `bool` result and `breaks` based on that.

src/macaron/util.py
```diff
@@ -272,6 +274,104 @@ def send_post_http_raw(
...
+        response.raise_for_status()
+    except requests.exceptions.HTTPError as http_err:
+        logger.debug("HTTP error occurred when trying to stream source: %s", http_err)
+        return False
+
+    if response.status_code != 200:
+        return False
+
+    data_processed = 0
+    for chunk in response.iter_content(chunk_size=1048576):
```
tromai, line 359: Could you let me know more about why you picked this `chunk_size` ?

src/macaron/provenance/provenance_finder.py
benmss: Technically this function is only used for GitHub attestation purposes at the moment, but I agree it should be changed.

src/macaron/util.py
```diff
@@ -272,6 +274,104 @@ def send_post_http_raw(
...
+    ----------
+    url: str
+        The target of the request.
+    headers: dict
+        The headers to use in the request.
+    chunk_function: Callable[[bytes], None]
+        The function to use with each downloaded chunk.
+    timeout: int
+        The timeout in seconds for the request.
+    size_limit: int
```
tromai, line 339: At the moment, `size_limit` is an optional parameter with the defaults value of 0. Does this mean that you should never use `stream_file_with_size_limit` without providing a value for `size_limit` into the function call ?

PR #1115, opened 2025-07-02
feat: add reproducible central buildspec generation

tromai:
## Summary
This Pull Request adds a new command called `gen-build-spec`.
```
usage: macaron gen-build-spec [-h] -purl PACKAGE_URL --database DATABASE
                              [--output-format OUTPUT_FORMAT]

options:
  -h, --help            show this help message and exit
  -purl PACKAGE_URL, --package-url PACKAGE_URL
                        The PURL string of the software component to generate build
                        spec for.
  --database DATABASE   Path to the database.
  --output-format OUTPUT_FORMAT
                        The output format. Can be rc-buildspec (Reproducible-central
                        build spec) (default "rc-buildspec")
```

This command generates a buildspec, which contains the build related information for a PURL that Macaron has analyzed. The output file will be stored within `output/<purl_based_path>/macaron.buildspec`, where `<purl_based_path>` being the directory structure according to the input PackageURL.

An example

```
macaron analyze -purl pkg:maven/org.apache.hugegraph/computer-k8s@1.0.0
macaron gen-build-spec -purl pkg:maven/org.apache.hugegraph/computer-k8s@1.0.0 --database output/macaron.db
```

In this example, the final path to the output buildspec file is `output/maven/org_apache_hugegraph/computer-k8s/macaron.buildspec`

The content of `output/maven/org_apache_hugegraph/computer-k8s/macaron.buildspec`
, which uses the [Reproducible Central buildspec format](https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/doc/BUILDSPEC.md).

```bash
# Copyright (c) 2025, Oracle and/or its affiliates.
# Generated by Macaron version 0.15.0

# Input PURL - pkg:maven/org.apache.hugegraph/computer-k8s@1.0.0
# Initial default JDK version 8 and default build command [['mvn', '-DskipTests=true', '-Dmaven.test.skip=true', '-Dmaven.site.skip=true', '-Drat.skip=true', '-Dmaven.javadoc.skip=true', 'clean', 'package']].

groupId=org.apache.hugegraph
artifactId=computer-k8s
version=1.0.0

gitRepo=https://github.com/apache/hugegraph-computer

gitTag=d2b95262091d6572cc12dcda57d89f9cd44ac88b

tool=mvn
jdk=8

newline=lf

command="mvn -DskipTests=true -Dmaven.test.skip=true -Dmaven.site.skip=true -Drat.skip=true -Dmaven.javadoc.skip=true clean package"

buildinfo=target/computer-k8s-1.0.0.buildinfo
```

This Buildspec ideally can be used directly as part of the Reproducible Central rebuild infrastructure.

### Macaron database extractor

The first step to generate a buildspec is to extract the build related information from an existing Macaron SQLite database. The module `macaron_db_extractor.py` added in this [commit](https://github.com/oracle/macaron/pull/1115/commits/b6d83db20e7c6db4174bdfc23f12920fe9d64923) does just that. 

It uses [sqlalchemy SELECT statement for ORM Mapped Classes](https://docs.sqlalchemy.org/en/20/orm/queryguide/select.html) to extract the data from the database into equivalent ORM Mapped instances that we defined in [src/macaron/database/table_definitions.py](https://github.com/oracle/macaron/blob/6a712af1ebdcb435bd5b7199dc1b4f5473663090/src/macaron/database/table_definitions.py) for example.

### Maven and Gradle CLI Command Parser

We use the build commands obtained in CI/CD configuration (e.g. from github action workflow yaml file) for the final buildspec. However, those build commands cannot be used as is and they requires some additional patching to work as a rebuild command.

A proper way to patch any maven and gradle CLI build command is to first parse is. The maven and gradle CLI command parsers added in this [commit](https://github.com/oracle/macaron/pull/1115/commits/e268f6fc1ffc588d3da2a02264a491c508e19a80) leverage Python's builtin `argparse` library. 

### CLI Build Command Patcher

The modules added in this [commit](https://github.com/oracle/macaron/pull/1115/commits/6a0c2187575026315cb58590149398d642a43c11) uses the Maven and GRadle CLI Command Parser to parse and patch the build commands obtained from the Macaron database.

### Jdk version finding from java Maven Central artifacts 

Macaron can obtain the JDK version for a given build command obtained from CI/CD configuration. In some cases, the CI/CD configuration doesn't have enough information for us to obtain the JDK version. Therefore, we also rely on the JDK version included in `META-INF/MANIFEST.MF` in java artifacts from Maven Central `https://repo1.maven.org/`.

The module `jdk_finder.py` added in this [commit](https://github.com/oracle/macaron/pull/1115/commits/bdef76c9ecf1ef559248f6a961d87d46f609b59c) help download the java artifacts from Maven Central given a maven type PURL, then returns the JDK version if it is available in `META-INF/MANIFEST.MF`.

In some cases, the JDK version string from `META-INF/MANIFEST.MF` don't only contain the JDK major version. For example:
- pkg:maven/org.apache.druid.integration-tests/druid-it-cases@25.0.0
    - https://repo1.maven.org/maven2/org/apache/druid/integration-tests/druid-it-cases/25.0.0/
    - "8 (Azul Systems Inc. 25.282-b08)"

Because Reproducible Central Buildspec requires only the major version of JDK, we need to extract that major version only. The `jdk_version_normalizer.py` module contains the logic to do just that. It is added this in [commit](https://github.com/oracle/macaron/pull/1115/commits/556f435d3bfe039fcd97f6da9e4f534be281672e).

### Generating the Reproducible Central Buildspec

The two commits 
- https://github.com/oracle/macaron/pull/1115/commits/f71dfcf954b7d82a7175dc39dddb18abf4887fbc
- https://github.com/oracle/macaron/pull/1115/commits/ddbcc5534efe5aa213e8e1310a1150246a02ecd4

use all components listed above to generate the final Reproducible Central Buildspec

### Testing

This feature includes unit tests for all components used in RC Buildspec generation (e.g. CLI parsers, CLI patchers, JDK version finder, etc.)

For integration tests, a new script called [compare_rc_build_spec.py](https://github.com/oracle/macaron/blob/01f235de48ba94ef0614a7abfd9d61c5f66f9ead/tests/build_spec_generator/reproducible_central/compare_rc_build_spec.py) is added to compare the result Buildspec in the integration tests.

## Checklist

- [x] Come up with proper patching values for Reproducible Central format. Right now it's empty.
- [x] Added integration tests for the feature. Make sure to test on Docker image.
- [x] Added unit test for the reproducible central build spec generation format
- [x] Revise all remaining TODOs in the PR
- [x] Rebase the branch to properly split the changes into different commit, making it easier for review.
- [ ] Update the api docs in the Sphinx documentation (could be in a different PR)
- [ ] Add a tutorial on how to support generating build spec of a different format (e.g. support new build tool patching, etc.) (could also be in a different PR)
- [x] Generate the build spec to a PURL based path instead of `output/macaron.buildspec` (could also be in a different PR).
- [ ] Add a test case for prioritizing JDK version obtained from Maven Central JAR than the one from look up build command 
- [ ] Consider removing the "extra_comments" part in the generated build spec ?


tests/build_spec_generator/reproducible_central/compare_rc_build_spec.py
```diff
@@ -0,0 +1,165 @@
...
+
+def skip_compare(_result: object, _expected: object) -> bool:
+    """Return ``True`` always.
+
+    This compare function is used when we want to skip comparing a field.
+    """
+    return True
+
+
+def compare_rc_build_spec(
```
tromai, line 52: Most of the logic of this function is taken from https://github.com/oracle/macaron/blob/6a712af1ebdcb435bd5b7199dc1b4f5473663090/tests/vsa/compare_vsa.py

I think the comparing functions within `compare_vsa.py` could be refactored out into a `tests_util.py` module so that all "compare" scripts used in integration tests ([here](https://github.com/oracle/macaron/blob/b63eb9be22da84e595339aeaf291c81bd674d630/tests/integration/run.py#L77-L84)) could use if needed. 
Please let me know if this should be done in a subsequent PR or this PR.

src/macaron/build_spec_generator/reproducible_central/reproducible_central.py
```diff
@@ -0,0 +1,355 @@
...
+    extra_comments = []
+
+    purl = build_info.purl
+    logger.debug(
+        "Generating build spec for %s with command patches:\n%s",
+        purl,
+        pformat(patches),
+    )
+
+    group = purl.namespace
```
tromai, line 204: A Reproducible Central Buildspec file needs the GAV coordinate of the component. This will only make sense for an input `maven` type PackageURL. 
At the moment, we accept any type of PackageURL. Please let me know if it makes sense to enforce the input PackageURL to be of type maven when we are generating a Reproducible Central Buildspec ? (Ideally each type of Buildspec format might have different requirements on the input PURL)

src/macaron/build_spec_generator/reproducible_central/reproducible_central.py
```diff
@@ -0,0 +1,355 @@
...
+        macaron_build_tool_name=macaron_build_tool_name,
+        patches=patches,
+    )
+    if not final_build_command_seq:
+        logger.critical(
+            "Cannot generate a default build command for %s",
+            purl,
+        )
+        return None
+    final_jdk_version = "8"
```
tromai, line 244: I think we should always try to obtain the jdk version from maven central jar, and only fall back to `8` if we cannot do so. Please feel free to let me know if this is desireable.

behnazh-w:
@tromai Instead of storing the build spec in `output/macaron.buildspec`, what do you think about using the standard PURL-based path? That would help prevent overwrites and improve archiving outputs.

tromai:
@behnazh-w I only use `output/macaron.buildspec` as I didn't really know the best way to implement the output path for this feature. Because this feature is similar to the VSA generation, in the sense that it generates one file given an PURL from input.
I can definitely change it to a PURL based path format. The only downside of using a PURL based path here is that it would be a bit challenging for scripting in a CI/CD environment, if that is the targeted use case (you might need to compute the PURL based path from outside of Macaron).

src/macaron/build_spec_generator/reproducible_central/reproducible_central.py
tromai: From this commit - https://github.com/oracle/macaron/pull/1115/commits/9e60c860c86208e77e8ff1644c10c1659044f24b we will always look for JDK version from Maven Central Jar.

src/macaron/build_spec_generator/build_command_patcher.py
```diff
@@ -0,0 +1,137 @@
...
+
+        if not effective_cli_parser:
+            result.append(UnparsedCLICommand(original_cmds=cmds))
+            continue
+
+        try:
+            cli_command = effective_cli_parser.parse(cmds)
+        except CommandLineParseError as error:
+            logger.error(
+                "Failed to parse the mvn command %s. Error %s.",
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle/macaron/pull/1115/commits/68c4a6dd9676e1def7542a3626c3e04c9d6eb106

src/macaron/build_spec_generator/cli_command_parser/gradle_cli_parser.py
```diff
@@ -0,0 +1,701 @@
...
+class GradleCLICommandParser:
+    """A Gradle CLI Command Parser."""
+
+    ACCEPTABLE_EXECUTABLE = ["gradle", "gradlew"]
+
+    def __init__(self) -> None:
+        """Initialize the instance."""
+        self.arg_parser = argparse.ArgumentParser(
+            description="Parse Gradle CLI command",
+            prog="mvn",
```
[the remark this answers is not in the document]
tromai: Fixed in https://github.com/oracle/macaron/pull/1115/commits/68c4a6dd9676e1def7542a3626c3e04c9d6eb106

src/macaron/__main__.py
```diff
@@ -523,6 +571,30 @@ def main(argv: list[str] | None = None) -> None:
...
+        "--database",
+        help="Path to the database.",
+        required=True,
+    )
+
+    gen_build_spec_parser.add_argument(
+        "--output-format",
+        type=str,
+        help=('The output format. Can be rc-buildspec (Reproducible-central build spec) (default "rc-buildspec")'),
+        default="rc-buildspec",
```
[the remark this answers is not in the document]
tromai: Yes, I made this option with the intention that we can allow different formats. Right now we only support `rc-buildspec`.
Future format can be specified with
```
macaron gen-build-spec ... --output-format some-other-format
```

src/macaron/build_spec_generator/build_command_patcher.py
```diff
@@ -0,0 +1,137 @@
...
+            continue
+
+        try:
+            new_cli_command = effective_cli_parser.apply_patch(
+                cli_command=cli_command,
+                options_patch=patch,
+            )
+        except PatchBuildCommandError as error:
+            logger.error(
+                "Failed to patch the mvn command %s. Error %s.",
```
benmss, line 80: Looks like this should be changed from "mvn".

src/macaron/build_spec_generator/build_command_patcher.py
```diff
@@ -0,0 +1,137 @@
...
+                error,
+            )
+            return None
+
+        result.append(new_cli_command)
+
+    return result
+
+
+def patch_commands(
```
benmss, line 91: I'm not sure I see the value of having the two separated functions here. The public function only seems to pass the parsers and perform a small extra step on the results. I'd suggest combining them, making the parsers arg override the default list, if provided.

src/macaron/build_spec_generator/build_spec_generator.py
```diff
@@ -0,0 +1,142 @@
...
+        },
+    },
+}
+
+
+def gen_build_spec_for_purl(
+    purl: PackageURL,
+    database_path: str,
+    build_spec_format: BuildSpecFormat,
+    output_path: str,
```
benmss, line 72: This appears to be missing from the doc string.

src/macaron/build_spec_generator/build_spec_generator.py
```diff
@@ -0,0 +1,142 @@
...
+        The format of the final build spec content.
+
+    Returns
+    -------
+    int
+        The exit code for this function. ``os.EX_OK`` if everything is fine, ``os.EX_OSERR`` if the
+        buildspec file cannot be created in the local filesystem, ``os.EX_DATAERR`` if there was an
+        error in generate the content for the buildspec file.
+    """
+    db_engine = create_engine(f"sqlite+pysqlite:///{database_path}", echo=False)
```
benmss, line 92: I don't think this is strictly required, but could the database be opened in read-only mode?
A stackoverflow thread suggests adding "?mode=ro" after the path:
https://stackoverflow.com/questions/10205744/opening-sqlite3-database-from-python-in-read-only-mode

src/macaron/build_spec_generator/build_spec_generator.py
```diff
@@ -0,0 +1,142 @@
...
+        )
+
+        os.makedirs(
+            name=os.path.dirname(build_spec_filepath),
+            exist_ok=True,
+        )
+
+        try:
+            with open(build_spec_filepath, mode="w", encoding="utf-8") as file:
+                logger.info(
```
benmss, line 128: I think this log message could be moved outside the try block.

src/macaron/build_spec_generator/cli_command_parser/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+
+P = TypeVar("P")
+
+
+@dataclass
+class OptionDef(Generic[P]):
+    """This class represent a definition of a CLI option for argparse.ArgumentParser.
+
+    This class also contains the information for validating a patch value.
+    The generic type T is the patch expected type (if it's not None).
```
benmss, line 73: This should be "generic type P"?

src/macaron/build_spec_generator/cli_command_parser/gradle_cli_command.py
```diff
@@ -0,0 +1,388 @@
...
+        Only enabled options are returned.
+
+        Returns
+        -------
+        list[str]
+            The enabled options.
+        """
+        result = []
+
+        if self.help_:
```
benmss, line 197: How would you feel about grouping these append operations into functions based on the categories like with the class parameters? This just seems like a very long single function.

src/macaron/build_spec_generator/cli_command_parser/gradle_cli_parser.py
```diff
@@ -0,0 +1,701 @@
...
+
+    def is_valid_patch_option(self, patch: Any) -> TypeGuard[bool]:
+        """Return True if the provide patch value is compatible with the internal type of this option."""
+        return isinstance(patch, bool)
+
+    def add_itself_to_arg_parser(self, arg_parse: argparse.ArgumentParser) -> None:
+        """Add a new argument to argparser.ArgumentParser representing this option."""
+        kwargs: dict[str, Any] = {}
+
+        kwargs["action"] = "store_true"
```
benmss, line 57:
```suggestion
        kwargs: dict[str, Any] = {"action": "store_true"}
```

src/macaron/build_spec_generator/cli_command_parser/gradle_cli_parser.py
```diff
@@ -0,0 +1,701 @@
...
+                **kwargs,
+            )
+
+    def get_patch_type_str(self) -> str:
+        """Return the expected type for the patch value as string."""
+        return "bool"
+
+
+@dataclass
+class GradleOptionalNegateableFlag(OptionDef[bool]):
```
benmss, line 78:
```suggestion
class GradleOptionalNegatableFlag(OptionDef[bool]):
```

src/macaron/build_spec_generator/cli_command_parser/gradle_cli_parser.py
```diff
@@ -0,0 +1,701 @@
...
+        )
+
+    def get_patch_type_str(self) -> str:
+        """Return the expected type for the patch value as string."""
+        return "dict[str, str | None]"
+
+
+@dataclass
+class GradleTask(OptionDef[list[str]]):
+    """This option represents the positional task option in Maven CLI command.
```
benmss, line 178:
```suggestion
    """This option represents the positional task option in Gradle CLI command.
```

src/macaron/build_spec_generator/cli_command_parser/gradle_cli_parser.py
```diff
@@ -0,0 +1,701 @@
...
+                self.long_name,
+            )
+
+    def get_patch_type_str(self) -> str:
+        """Return the expected type for the patch value as string."""
+        return "str"
+
+
+@dataclass
+class GradlePropeties(OptionDef[dict[str, str | None]]):
```
benmss, line 147:
```suggestion
class GradleProperties(OptionDef[dict[str, str | None]]):
```

src/macaron/build_spec_generator/cli_command_parser/gradle_cli_parser.py
```diff
@@ -0,0 +1,701 @@
...
+    """This option represents an option used to define properties values of a Gradle CLI command.
+
+    This option can be defined multiple times and the values are appended into a list of string in argparse.
+    However, it's stored internally as a dictionary mapping between the system property name to its value.
+
+    In Gradle there are 2 options of this type:
+        - -D/--system-prop
+        - -P/--project-prop
```
benmss, line 155: I think an example of how these are used would help distinguish this class from `GradleAppendedList`. Perhaps a mention of key value pairs or similar would also be of use.

src/macaron/build_spec_generator/cli_command_parser/gradle_cli_parser.py
```diff
@@ -0,0 +1,701 @@
...
+        Raises
+        ------
+        CommandLineParseError
+            If an error happens when parsing the Gradle CLI Command.
+        """
+        if not cmd_list:
+            raise CommandLineParseError("The provided cmd list is empty.")
+
+        exe_path = cmd_list[0]
+        options = cmd_list[1:]
```
benmss, line 530: Is it possible for there to be no options in the command list? Is this a case we want to watch out for?

src/macaron/build_spec_generator/build_spec_generator.py
tromai: I think this is a good idea. I will add it in.

src/macaron/build_spec_generator/cli_command_parser/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+
+class CLICommand(Protocol):
+    """Interface of a CLI Command."""
+
+    def to_cmds(self) -> list[str]:
+        """Return the CLI Command as a list of strings."""
+
+
+T = TypeVar("T", bound="CLICommand")
+Y_contra = TypeVar("Y_contra", contravariant=True)
```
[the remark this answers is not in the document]
tromai: I agree. Will do.

src/macaron/build_spec_generator/cli_command_parser/gradle_cli_parser.py
tromai: Definitely, I will add the examples.

src/macaron/build_spec_generator/cli_command_parser/gradle_cli_parser.py
```diff
@@ -0,0 +1,701 @@
...
+
+    def get_patch_type_str(self) -> str:
+        """Return the expected type for the patch value as string."""
+        return "list[str]"
+
+
+# TODO: some value option only allows you to provide certain values
+# For example: --console allows "plain", "auto", "rich" or "verbose".
+# They are right now not enforced. We need to think whether we want to enforce them.
+GRADLE_OPTION_DEF: list[OptionDef] = [
```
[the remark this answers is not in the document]
tromai: I think an Enum would help us from using the same literal strings at different places. 
Would you think that I should handle `short_names`, `dest` and `long_name` or `long_name` only is sufficient (`short_names` and `dest` are optional so some `OptionDef` might not have them) ?

src/macaron/build_spec_generator/build_spec_generator.py
```diff
@@ -0,0 +1,142 @@
...
+
+    with Session(db_engine) as session, session.begin():
+        build_spec_content = None
+        match build_spec_format:
+            case BuildSpecFormat.REPRODUCIBLE_CENTRAL:
+                build_spec_content = gen_reproducible_central_build_spec(
+                    purl=purl,
+                    session=session,
+                    patches=CLI_COMMAND_PATCHES,
+                )
```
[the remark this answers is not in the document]
tromai: That's a good point. I will change it.

src/macaron/build_spec_generator/build_command_patcher.py
tromai: I wonder if this align with what you have in mind ?
```
def patch_commands(
    cmds_sequence: Sequence[list[str]],
    patches: Mapping[
        PatchCommandBuildTool,
        Mapping[str, PatchValueType | None],
    ],
    cli_parsers: Sequence[CLICommandParser] | None = None,
) -> list[list[str]] | None:
    effective_cli_parsers = cli_parsers if cli_parsers else [MVN_CLI_PARSER, GRADLE_CLI_PARSER]
    # The body of _patch_commands using effective_cli_parsers  
```

src/macaron/build_spec_generator/cli_command_parser/gradle_cli_parser.py
tromai: > Is it possible for there to be no options in the command list?

Yes, my intention is that we allow `gradle` as a completed and valid build commmand. I based on the fact that you can run `gradle` in a project that use gradle and the command doesn't failed. You can see the corresponding test case here - https://github.com/oracle/macaron/blob/6a3fdb4ec3991b8c7d071652f279311ae8bc4f73/tests/build_spec_generator/cli_command_parser/test_gradle_cli_parser.py#L15-L22.

It's not the same for Maven, however, where I didn't allow `mvn` or `mvnw` as a valid build command. You can see a test case here - https://github.com/oracle/macaron/blob/6a3fdb4ec3991b8c7d071652f279311ae8bc4f73/tests/build_spec_generator/cli_command_parser/test_maven_cli_parser.py#L172-L173. This is because if you don't provide any phase or goal to Maven, it will raise an error 

```
[ERROR] No goals have been specified for this build. You must specify a valid lifecycle phase or a goal in the format <plugin-prefix>:<goal> or <plugin-group-id>:<plugin-artifact-id>[:<plugin-version>]:<goal>. ...
```

@nicallen I also want to know your opinion on these behaviors, and whether we should make it consistent with the original `mvn` and `gradle` or be a bit relaxed and allow no options.

src/macaron/build_spec_generator/cli_command_parser/gradle_cli_parser.py
```diff
@@ -0,0 +1,701 @@
...
+        - For other option that expects a value (e.g `-c/--setting-file <path/to/settings/file>``), a string is
+          expected.
+
+        None can be provided to ANY type of option to forcefully remove it from the original build command.
+
+        Parameters
+        ----------
+        cli_command : GradleCLICommand
+            The original Gradle command, as a ``GradleCLICommand`` object from ``GradleCLICommandParser.parse(...)``
+        patch_options : Mapping[str, GradleOptionPatchValueType | None]
```
benmss, line 615:
```suggestion
        options_patch : Mapping[str, GradleOptionPatchValueType | None]
```

src/macaron/build_spec_generator/cli_command_parser/gradle_cli_parser.py
nicallen: It is actually possible to successfully run `mvn` with no args, since you can set `<defaultGoal>` in the pom file. So we shouldn't disallow that.

src/macaron/build_spec_generator/cli_command_parser/maven_cli_command.py
```diff
@@ -0,0 +1,324 @@
...
+
+    executable: str
+    options: MavenCLIOptions
+
+    def to_cmds(self) -> list[str]:
+        """Return the CLI Command as a list of strings."""
+        result = []
+        result.append(self.executable)
+        result.extend(self.options.to_option_cmds())
+        return result
```
benmss, line 324: Same comment as with GradleCLICommand.

src/macaron/build_spec_generator/cli_command_parser/maven_cli_parser.py
```diff
@@ -0,0 +1,594 @@
...
+            *(self.short_name, self.long_name),
+        )
+
+    def get_patch_type_str(self) -> str:
+        """Return the expected type for the patch value as string."""
+        return "str"
+
+
+@dataclass
+class MavenCommaDelimList(OptionDef[list[str]]):
```
benmss, line 95: I think I suggest just going with `MavenList`. It seems this is the only list type option?

src/macaron/build_spec_generator/cli_command_parser/maven_cli_parser.py
```diff
@@ -0,0 +1,594 @@
...
+    def get_patch_type_str(self) -> str:
+        """Return the expected type for the patch value as string."""
+        return "dict[str, str | None]"
+
+
+@dataclass
+class MavenGoalPhase(OptionDef[list[str]]):
+    """This option represents the positional goal/plugin-phase option in Maven CLI command.
+
+    argparse.Namespace stores this as a list of string. This is stored internally as a list of string.
```
benmss, line 160: An example of this option would still be useful.

src/macaron/build_spec_generator/cli_command_parser/maven_cli_parser.py
```diff
@@ -0,0 +1,594 @@
...
+    ),
+    MavenGoalPhase(
+        long_name="goals",
+    ),
+    # TODO: we need to confirm whether one can provide
+    # -P or -pl multiple times and the values will be aggregate into a list of string
+    # The current implementation only consider one instance of -P or -pl.
+    # Where to begin:
+    # https://github.com/apache/maven/blob/maven-3.9.x/maven-embedder/src/main/java/org/apache/maven/cli/CLIManager.java
+    # https://github.com/apache/commons-cli/blob/master/src/main/java/org/apache/commons/cli/Parser.java
```
benmss, line 287: It looks like this comment should be moved down a bit to where the related options are.

src/macaron/build_spec_generator/cli_command_parser/maven_cli_parser.py
```diff
@@ -0,0 +1,594 @@
...
+
+        return patch_mapping(
+            original=original_props,
+            patch=patch_value,
+        )
+
+    def apply_patch(
+        self,
+        cli_command: MavenCLICommand,
+        options_patch: Mapping[str, MavenOptionPatchValueType | None],
```
benmss, line 487:
```suggestion
        patch_options: Mapping[str, MavenOptionPatchValueType | None],
```
I think this is actually more appropriate than `options_patch`. Also applies to the Gradle counterpart.

src/macaron/build_spec_generator/jdk_finder.py
```diff
@@ -0,0 +1,345 @@
...
+        logger.error("Failed why trying to download jar file. Error: %s", error)
+        return None
+    except OSError as os_error:
+        logger.critical("Critical %s", os_error)
+        return None
+
+    return get_jdk_version_from_jar(local_artifact_path)
+
+
+def find_jdk_version_from_central_maven_repo(
```
benmss, line 292: A few things for this function and the things it relies on:

- The Java extension enum seems to only be used here, and contains only a single value. Maybe it should just be hardcoded? Or stored not as an enum?
- The caching strategy also seems to be somewhat static; however I see some value of having that option.
- The two `find_jdk...` functions are very similar. I would suggest having a single entry point that diverges based on the caching logic (assuming that's being kept), as it is mostly the download part that differs.

src/macaron/build_spec_generator/jdk_version_normalizer.py
```diff
@@ -0,0 +1,84 @@
...
+    >>> normalize_jdk_version("19-ea")
+    '19'
+    >>> normalize_jdk_version("11.0.1")
+    '11'
+    >>> normalize_jdk_version("1.8")
+    '8'
+    >>> normalize_jdk_version("25.0.1")
+    """
+    first, _, after = jdk_version_str.partition(".")
+    jdk_major_ver = None
```
benmss, line 61:
```suggestion
```

src/macaron/build_spec_generator/jdk_version_normalizer.py
```diff
@@ -0,0 +1,84 @@
+# Copyright (c) 2025 - 2025, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
+
+"""This module contains the logic to nomarlize a JDK version string to a major version number."""
+
+SUPPORTED_JAVA_VERSION = [
```
benmss, line 6: This should probably be a set.

src/macaron/build_spec_generator/macaron_db_extractor.py
```diff
@@ -0,0 +1,665 @@
...
+    )
+
+
+def get_sql_stmt_build_tools(component_id: int) -> Select[tuple[BuildToolFacts]]:
+    """Return an SQLAlchemy SELECT statement to query the BuildToolFacts for a given PackageURL.
+
+    Parameters
+    ----------
+    purl_string : str
+        The PackageURL string to find the BuildToolFacts.
```
benmss, line 179: Should be updated to match the function parameter.

src/macaron/build_spec_generator/macaron_db_extractor.py
```diff
@@ -0,0 +1,665 @@
...
+    )
+
+
+def get_sql_stmt_build_script_check(component_id: int) -> Select[tuple[BuildScriptFacts]]:
+    """Return an SQLAlchemy SELECT statement to query the BuildScriptFacts for a given PackageURL.
+
+    Parameters
+    ----------
+    purl_string : str
+        The PackageURL string to find the BuildScriptFacts.
```
benmss, line 315: Needs to be updated.

src/macaron/build_spec_generator/macaron_db_extractor.py
```diff
@@ -0,0 +1,665 @@
...
+    return latest_component
+
+
+def lookup_build_tools_check(component_id: int, session: Session) -> Sequence[BuildToolFacts]:
+    """Return the sequence of BuildToolFacts instances for given PackageURL string.
+
+    Parameters
+    ----------
+    purl_string : str
+        The PackageURL string to look for the BuildToolFacts.
```
benmss, line 399: Needs to be updated.

src/macaron/build_spec_generator/macaron_db_extractor.py
```diff
@@ -0,0 +1,665 @@
...
+    return build_as_code_check_facts
+
+
+def lookup_build_service_check(component_id: int, session: Session) -> Sequence[BuildServiceFacts]:
+    """Return the sequence of BuildServiceFacts instances for given PackageURL string.
+
+    Parameters
+    ----------
+    purl_string : str
+        The PackageURL string to look for the BuildServiceFacts.
```
benmss, line 469: Needs to be updated.

src/macaron/build_spec_generator/reproducible_central/reproducible_central.py
```diff
@@ -0,0 +1,471 @@
...
+
+    https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/doc/BUILDSPEC.md
+    """
+
+    MAVEN = "mvn"
+    GRADLE = "gradle"
+    SBT = "sbt"
+
+
+def format_build_command_infos(build_command_infos: list[GenericBuildCommandInfo]) -> str:
```
benmss, line 86:
```suggestion
def format_build_command_info(build_command_info: list[GenericBuildCommandInfo]) -> str:
```

src/macaron/build_spec_generator/reproducible_central/reproducible_central.py
```diff
@@ -0,0 +1,471 @@
...
+        The SQLAlchemy Session opened for the database to extract build information.
+
+    Returns
+    -------
+    GenericBuildCommandInfo | None
+        The GenericBuildCommandInfo object for the highest confidence build command or None if there was
+        an error, or no build command is found from the database.
+    """
+    try:
+        lookup_build_command_infos = lookup_any_build_command(component_id, session)
```
benmss, line 282:
```suggestion
        lookup_build_command_info = lookup_any_build_command(component_id, session)
```

src/macaron/build_spec_generator/reproducible_central/reproducible_central.py
```diff
@@ -0,0 +1,471 @@
...
+        "Attempted to find build command from the database. Result: %s",
+        lookup_build_command_info or "Cannot find any.",
+    )
+    lookup_build_command_jdk = (
+        get_lookup_build_command_jdk(
+            lookup_build_command_info,
+        )
+        if lookup_build_command_info
+        else None
+    )
```
benmss, line 424: I think this should only be executed if the `jdk_from_jar` is `None`.

src/macaron/build_spec_generator/reproducible_central/reproducible_central.py
```diff
@@ -0,0 +1,471 @@
...
+    patched_build_commands = patch_commands(
+        cmds_sequence=[selected_build_command],
+        patches=patches,
+    )
+    if not patched_build_commands:
+        logger.error(
+            "Failed to patch command sequences %s.",
+            [selected_build_command],
+        )
+        return None
```
benmss, line 455: As the JDK discovery can rely on the build information via `lookup_build_command_info`, I'd suggest moving all of the build command related code before the JDK related code. Up to and including the highlighted block here.
Also might be a good idea to extract some of this out to standalone functions. E.g. JDK, Patched Build Commands, etc.

tests/build_spec_generator/cli_command_parser/test_gradle_cli_command.py
```diff
@@ -0,0 +1,156 @@
...
+    ],
+)
+def test_comparing_gradle_cli_command_unequal_types(
+    gradle_cli_parser: GradleCLICommandParser,
+    command: str,
+    that: Any,
+) -> None:
+    """Test comparing MavenCLICommand with another incompatible type oject."""
+    this_command = gradle_cli_parser.parse(command.split())
+    assert not this_command == that
```
benmss, line 107: What is this testing? It seems to be comparing a GradleCLICommand object against a bool, list, and set.

tests/build_spec_generator/reproducible_central/compare_rc_build_spec.py
```diff
@@ -0,0 +1,165 @@
...
+) -> bool:
+    """Compare two dictionaries obatained from 2 Reproducible Central build spec.
+
+    Parameters
+    ----------
+    result : dict[str, str]
+        The result object.
+    expected : dict[str, str]
+        The expected object.
+    compare_fn_map : str
```
benmss, line 65:
```suggestion
    compare_fn_map : dict[str, CompareFn]
```

tests/build_spec_generator/test_macaron_db_extractor.py
```diff
@@ -0,0 +1,309 @@
...
+    assert lookup_repo.remote_path == "https://github.com/oracle/macaron"
+    assert lookup_repo.commit_sha == "d2b95262091d6572cc12dcda57d89f9cd44ac88b"
+
+
+def test_lookup_any_build_command_empty_db(macaron_db_session: Session) -> None:
+    """Test the lookup_any_build_command function with an empty database."""
+    assert not lookup_any_build_command(component_id=1, session=macaron_db_session)
+
+
+def test_invalid_input_databse(invalid_db_session: Session) -> None:
```
benmss, line 285:
```suggestion
def test_invalid_input_database(invalid_db_session: Session) -> None:
```

tests/build_spec_generator/reproducible_central/compare_rc_build_spec.py
benmss: Either is fine for me.

behnazh-w (approved):

src/macaron/build_spec_generator/cli_command_parser/maven_cli_parser.py
```diff
@@ -0,0 +1,611 @@
...
+        if os.path.basename(exe_path) not in MavenCLICommandParser.ACCEPTABLE_EXECUTABLE:
+            raise CommandLineParseError(f"{exe_path} is not an acceptable mvn executable path.")
+
+        # TODO: because our parser is not completed for all cases, should we be more relaxed and use
+        # parse_unknown_options?
+        try:
+            parsed_opts = self.arg_parser.parse_args(options)
+        except argparse.ArgumentError as error:
+            raise CommandLineParseError(f"Failed to parse command {' '.join(options)}.") from error
+        # Even though we have set `exit_on_error`, argparse still exists unexpectedly in some
```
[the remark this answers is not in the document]
behnazh-w: Fixed in commit [b73c59c](https://github.com/oracle/macaron/pull/1115/commits/b73c59cb7fd992a13473c171611b4e16baa439df).

nicallen (approved):

PR #1133, opened 2025-07-22
refactor: remove depends_on fields from heuristics and handle skips in problog

### LINKED ISSUE #1052: Error and Skip Handling in Heuristic Malware Analysis
The use of the `SKIP` result for heuristics is not clear. It is used in some places for when errors occur, and for some places where the heuristic does not apply. Separating the `SKIP` result from error handling will make results clearer. 

Proposed definition of a `SKIP` result: a heuristic should return `HeuristicResult.SKIP` when this heuristic analysis is not applicable to the package. An example would be when the `SUSPICIOUS_SETUP` heuristic is run on a package with no `setup.py` file. In this case, `SKIP` is an appropriate result as this heuristic is not applicable to the package, but the package information is not malformed. 

Identified appropriate and inappropriate uses of the `SKIP` result currently in the codebase are listed below:
**Appropriate uses:**
- `anomalous_version.py`: returns a `SKIP` if the version cannot be interpreted as per PEP 440. This is fine as the package is not malformed, the heuristic  does not apply here. 
- `suspicious_setup.py`: returns a `SKIP` if there is no setup.py, which is appropriate as the heuristic does not apply in this case. This does need to be refactored to error when setup.py is found, but there is a problem trying to analyse it.

**Inappropriate uses:**
- `closer_release_join_date.py`: `SKIP` is returned if there are no maintainers or no latest release information. This would be a result of a malformed metadata file, or a problem in parsing the HTML page, and is such an error. Release and maintainer information must exist.
- `high_release_frequency.py`: `SKIP` is returned if there are no releases or if there is only one release. If there are no releases, then this is malformed metadata. If there is 1 release, then this heuristic should not have been run. These both constitute errors.
- `one_release.py`: `SKIP` is returned if there are no releases. If there are no releases, then this is malformed metadata, implying an error has occurred.
- `unchanged_release.py`: `SKIP` is returned if there are no digests. This would occur if there are no releases, or if there are no digest fields in the releases. In both cases, this would be malformed metadata, so an error has occurred.

### LINKED ISSUE #1108: Use a temporary file for testing defaults.ini content loading
# Description
In this test module - https://github.com/oracle/macaron/blob/612e27ec347e7e073311b65a32b9eec8c6ba7d5c/tests/malware_analyzer/pypi/test_pypi_sourcecode_analyzer.py, we are mocking the `defaults` object to test some behaviors relating to some source code analysis configuration in `defaults.ini`.

# Expectation
In this module https://github.com/oracle/macaron/blob/44dbf0acbd3dcb9f577db97b5c6876cfb8c87957/tests/config/test_defaults.py we put the defaults.ini content into a temporary file and load from it.
This is an example https://github.com/oracle/macaron/blob/44dbf0acbd3dcb9f577db97b5c6876cfb8c87957/tests/config/test_defaults.py#L103-L109

which can remove the need for mocking and keep the defaults.ini in the test definition itself. Perhaps we can rewrite the tests in `test_pypi_sourcecode_analyzer.py` to use this approach instead.

art1f1c3R:
## Summary
The purpose of this PR is to address #1052, and implement appropriate skip and error handling in all existing heuristics. This PR removes the `depends_on` field from heuristics and makes it a responsibility of the heuristic code to determine if it should be skipped or not. All skip handling is then done my the ProbLog logic.

## Description of changes
The `BaseHeuristicAnalyzer` no longer has the `depends_on` field, and this has propagated to all other analyzers. The description for a `SKIP` result has been updated in line with issue #1052, and `HeuristicAnalyzerValueError`s have replaced some `SKIP` results previously used when malformed data was encountered. Unit tests have been updated accordingly.

Some refactoring to the `test_pypi_sourcecode_analyzer.py` unit tests have also been done in this PR to address #1108, adding in temporary `.ini` config files to the unit tests as opposed to the existing mock objects.

## Related issues
Closes #1052 and #1108.

## Blockers
Currently blocked by #1147, as this is treated as a `HeuristicAnalyzerValueError` in this PR and causes integration tests to fail.

## Checklist

### DIFF
--- a/tests/malware_analyzer/pypi/test_pypi_sourcecode_analyzer.py
+++ b/tests/malware_analyzer/pypi/test_pypi_sourcecode_analyzer.py
@@ -5,17 +5,34 @@
 
 import json
 import os
-from unittest.mock import MagicMock, patch
+from pathlib import Path
+from unittest.mock import MagicMock
 
 import pytest
 import yaml
 
 from macaron import MACARON_PATH
+from macaron.config.defaults import defaults, load_defaults
 from macaron.errors import ConfigurationError, HeuristicAnalyzerValueError
 from macaron.malware_analyzer.pypi_heuristics.heuristics import HeuristicResult
 from macaron.malware_analyzer.pypi_heuristics.sourcecode.pypi_sourcecode_analyzer import PyPISourcecodeAnalyzer
 
 RESOURCES_PATH = os.path.join(MACARON_PATH, "resources")
+
+
+def config_defaults(tmp_path: Path, content: str) -> None:
+    """Set up the global defaults variable with the given defaults.ini content.
+
+    Parameters
+    ----------
+    tmp_path: Path
+        Pytest temporary path fixture.
+    content: str
+        The content of the defaults.init file.
+    """
+    defaults_file = Path(os.path.join(tmp_path, "config.ini"))
+    defaults_file.write_text(content, encoding="utf-8")
+    assert load_defaults(str(defaults_file)) is True
 
 
 def get_rule_ids_list(path: str) -> set[str]:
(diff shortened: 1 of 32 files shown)


tests/malware_analyzer/pypi/test_pypi_sourcecode_analyzer.py
```diff
@@ -175,10 +166,7 @@ def test_rules(
         expected_results = json.loads(file.read())
 
     # Test with none of the defaults.ini settings used, to ensure this ruleset is run
-    mock_defaults.has_section.side_effect = lambda section: section == "heuristic.pypi"
-    mock_defaults.__getitem__.side_effect = lambda section: (
-        MagicMock(get=lambda _: None) if section == "heuristic.pypi" else None
-    )
+    defaults["heuristic.pypi"].clear()
```
tromai, line 169: What is the scenario where this case happen? For instance, what would be the values in the user provided `defaults.ini` if any.

art1f1c3R: Our `defaults.ini` usually looks like this:
```
disabled_default_rulesets = exfiltration
disabled_rules =
custom_semgrep_rules_path =
disabled_custom_rulesets =
```
The reason I used `.clear()` here is to make sure that the rules that are being tested aren't disabled. Particularly here, by default our `exfiltration` ruleset is disabled, but it is tested in this unit test, so the test would fail if I didn't remove that setting.

PR #1170, opened 2025-09-05
chore: improving efficiency of similar projects analyzer

art1f1c3R:
## Summary
The goal is to improve the efficiency of the `SimilarProjectsAnalyzer`, which currently downloads the sourcecode tarball for every package of every maintainer it finds. The solution to this is to use the structure of the tarball/wheel provided on the `inspector.pypi.io` page of the package, making web requests to extract the structure instead of downloading the package.

## Description of changes
This PR modifies the way PyPI inspector links are handled by created a separate `PyPIInspectorAsset` object, container information about the PyPI inspector URLs and with the ability to extract the project structure from a package URL. The `WheelAbsenceAnalyzer` is then modified to use this, simplifying it, and the `SimilarProjectsAnalyzer` then makes use of it for analyzing the package structure.

The `SimilarProjectsAnalyzer` normalizes the structure by doing the following:
- Only considering python files.
- Removing the `<package_name>-<version>` prefix.
- Removing the `<package_name` from the top-level folder, resulting in a folder structure that does not contain the package name at the top level.
- Removing `setup.py` from tarballs.

This makes it so that wheels and tarballs are comparable when looking at the package structure. A unit test is written to demonstrate this. The `SimilarProjectsAnalyzer` then extracts the hash for these folder structures, and compares them against other projects made by the maintainers of the analyzed package. If at least one is similar, the analyzer fails, but it does continue to loop and collect all similar projects.

A known complication with this is the fact that PyPI uses the Fastly CDN, returning a JavaScript challenge response. Since PyPI inspector uses URLs rerouted from PyPI, this means those JavaScript challenges are received when making programmatic requests in python to a PyPI inspector URL. This does not always happen, but is a frequent occurrence. To accommodate for this, this analyzer is written such that it does not raise `HeuristicAnalyzerValueError`s, and will return `SKIP` results when unable to obtain package information. 

## Checklist

### DIFF
--- a/tests/slsa_analyzer/checks/test_detect_malicious_metadata_check.py
+++ b/tests/slsa_analyzer/checks/test_detect_malicious_metadata_check.py
@@ -34,7 +34,7 @@ RESOURCE_PATH = Path(__file__).parent.joinpath("resources")
         # from reaching the network.
         pytest.param("pkg:pypi/zlibxjson", CheckResultType.PASSED, False, id="test_malicious_pypi_package"),
         pytest.param("pkg:pypi/test", CheckResultType.UNKNOWN, False, id="test_unknown_pypi_package"),
-        pytest.param("pkg:maven:test/test", CheckResultType.UNKNOWN, False, id="test_non_pypi_package"),
+        pytest.param("pkg:maven/test", CheckResultType.UNKNOWN, False, id="test_non_pypi_package"),
         # TODO: including source code analysis that detects flow from a remote point to a file write may assist in resolving
         # the issue of this false negative.
         pytest.param(


src/macaron/slsa_analyzer/package_registry/pypi_registry.py
```diff
@@ -906,6 +971,91 @@ def get_sha256(self) -> str | None:
         logger.debug("Found sha256 hash: %s", artifact_hash)
         return artifact_hash
 
+    def get_inspector_links(self) -> bool:
```
behnazh-w:
```suggestion
    def get_inspector_src_preview_links(self) -> bool:
```

behnazh-w (commented):
Please begin all comments with a capital letter.

behnazh-w (approved):

PR #1185, opened 2025-09-23
chore: fix a bug for uncaught SourceCodeError exception

behnazh-w:
## Summary
This PR addresses a bug related to the uncaught `SourceCodeError` exception in the PyPI type stub file analyzer. An additional integration test has been added to ensure correct handling for Python packages distributed only as wheels.

## Description of changes

- Bug fix: handles situations where source code files cannot be downloaded. In this case, the `TypeStubFileAnalyzer` fails, instead of raising an exception.
- A TODO comment is added to note that `.pyi` stub files may reside in both `sdist` and `wheel` distributions.
- Unit test updated.
- Added a new integration test case and a policy for packages only distributed as `wheels` and not as `sdists` (e.g., `ibm-agent-analytics-common`).

### DIFF
--- a/src/macaron/malware_analyzer/pypi_heuristics/metadata/type_stub_file.py
+++ b/src/macaron/malware_analyzer/pypi_heuristics/metadata/type_stub_file.py
@@ -6,7 +6,6 @@
 import logging
 import os
 
-from macaron.errors import SourceCodeError
 from macaron.json_tools import JsonType
 from macaron.malware_analyzer.pypi_heuristics.base_analyzer import BaseHeuristicAnalyzer
 from macaron.malware_analyzer.pypi_heuristics.heuristics import HeuristicResult, Heuristics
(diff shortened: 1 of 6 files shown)


src/macaron/malware_analyzer/pypi_heuristics/metadata/type_stub_file.py
```diff
@@ -40,11 +39,12 @@ def analyze(self, pypi_package_json: PyPIPackageJsonAsset) -> tuple[HeuristicRes
...
         """
+        # TODO: .pyi stub files may be present in both source distributions (sdist) and wheels.
+        # Currently, we only check the sdist, which can lead to false positives in this heuristic.
+        # To improve accuracy, we should also check for stub files in the wheel distribution.
         result = pypi_package_json.download_sourcecode()
         if not result:
-            error_msg = "No source code files have been downloaded"
-            logger.debug(error_msg)
-            raise SourceCodeError(error_msg)
+            return HeuristicResult.FAIL, {"message": "No source code files have been downloaded.", "pyi_files": 0}
```
art1f1c3R, line 47: Would it be more appropriate here to return `HeuristicResult.SKIP`? In this case, we cannot definitively say there are 0 `.pyi` files, and going off the #1052 issue, a skip should be returned when a heuristic does not apply. In this case, since we have no visibility into the source code, this heuristic cannot be applied.

behnazh-w: Thanks for the suggest. Fixed in [63b671a](https://github.com/oracle/macaron/pull/1185/commits/63b671af20ecac131a93a43cc911caf4ebe0c7b8).

art1f1c3R (approved):

PR #1203, opened 2025-10-08
feat: add basic support for Python in gen-build-spec

behnazh-w:
## Summary
This PR adds basic support for Python projects to the `gen-build-spec` feature in Macaron. It also introduces a new approach for generic specification generation.

## Description of changes

- Refactors gen-build-spec to first generate a base specification in JSON format.  
- Enables conversion from the generic JSON spec to other supported formats, such as Reproducible Central (rc-buildspec).  
- By default, the gen-build-spec command now outputs the specification in JSON; users can still specify rc-buildspec if they require the Reproducible Central format.  
- Adds initial support for Python, allowing the tool to generate build specifications for Python projects.  
- Adjusts and extends unit tests.
- Adds an integration test and compare script for the base build spec.


AbhinavPradeep (approved):
Looks good to me!

PR #1256, opened 2025-12-03
feat: add support to use inferred build tools and to extract tool-specific build dependency information

AbhinavPradeep:
## Summary
Added support to use inferred build tools and to extract tool-specific build dependency information.

## Description of changes
1. Modified build command generation and added functionality to scan  `[tool.hatch.build.hooks.*]` and `[tool.flit.metadata]` in`pypi_spec.py`.
2. Modified `pypi_dockerfile_output.py` to be compatible with this.
3. Modified relevant unit and integration tests to reflect this.

### DIFF
--- a/src/macaron/build_spec_generator/common_spec/pypi_spec.py
+++ b/src/macaron/build_spec_generator/common_spec/pypi_spec.py
@@ -74,9 +74,7 @@ class PyPIBuildSpec(
                 case "flit":
                     # We might also want to deal with existence flit.ini, we can do so via
                     # "python -m flit.tomlify"
-                    default_build_commands.append(
-                        'if test -f "flit.ini"; then python -m flit.tomlify; fi && flit build'.split()
-                    )
+                    default_build_commands.append("flit build".split())
                 case "hatch":
                     default_build_commands.append("hatch build".split())
                 case "conda":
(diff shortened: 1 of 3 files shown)


src/macaron/build_spec_generator/common_spec/pypi_spec.py
```diff
@@ -67,15 +67,20 @@ def get_default_build_commands(
 
             match build_tool_name:
                 case "pip":
-                    default_build_commands.append("python -m build".split())
+                    default_build_commands.append("python -m build --wheel -n".split())
                 case "poetry":
-                    default_build_commands.append("poetry build".split())
+                    default_build_commands.append("pip install poetry && poetry build".split())
```
behnazh-w: What do you think about keeping the default install commands separate from the build commands? For example, we could add a new field to the spec specifically for install commands. I believe this would make the spec easier for downstream users to consume.

src/macaron/build_spec_generator/common_spec/pypi_spec.py
```diff
@@ -156,18 +161,35 @@ def resolve_fields(self, purl: PackageURL) -> None:
...
                     with pypi_package_json.sourcecode():
                         try:
+                            # Get the build time requirements from ["build-system", "requires"]
                             pyproject_content = pypi_package_json.get_sourcecode_file_contents("pyproject.toml")
                             content = tomli.loads(pyproject_content.decode("utf-8"))
                             requires = json_extract(content, ["build-system", "requires"], list)
                             if requires:
                                 build_requires_set.update(elem.replace(" ", "") for elem in requires)
+                            # If we have hatch as a build_tool, we will
+                            if "hatch" in self.data["build_tools"]:
```
behnazh-w: These build tool–specific analyses could be refactored to improve maintainability.

AbhinavPradeep: Yep will do that!

src/macaron/build_spec_generator/common_spec/pypi_spec.py
behnazh-w: Yes,  that makes sense.

PR #1336, opened 2026-03-25
feat: change dockerfile generation for Python rebuild to always default to standard build command

AbhinavPradeep:
## Summary
Change dockerfile generation to always default to standard Python build command

## Description of changes
Only affects dockerfile generation. Note that the dockerfile still decides to use `python setup.py bdist_wheel` if it detects that a legacy build command might be required (pre Python 3.6 before `pyproject.toml` was introduced).

### DIFF
--- a/tests/integration/cases/pypi_pytesseract/expected_dockerfile.buildspec
+++ b/tests/integration/cases/pypi_pytesseract/expected_dockerfile.buildspec
@@ -64,7 +64,7 @@ RUN <<EOF
 EOF
 
 # Run the build
-RUN source /deps/bin/activate &&  python -m build --wheel -n
+RUN source /deps/bin/activate && /deps/bin/pip install wheel && python -m build --wheel -n
 
 # Validate script
 RUN cat <<'EOF' >/validate


src/macaron/build_spec_generator/dockerfile/pypi_dockerfile_output.py
```diff
@@ -46,24 +46,11 @@ def gen_dockerfile(buildspec: BaseBuildSpecDict) -> str:
...
-    elif buildspec["build_tools"][0] == "flit":
-        build_tool_install = (
-            f"pip install {buildspec['build_tools'][0]} && if test -f \"flit.ini\"; then python -m flit.tomlify; fi && "
-        )
-
-    modern_build_command = build_tool_install + " ".join(x for x in buildspec["build_commands"][0])
+
+    backend_install_commands = " && ".join(build_backend_commands(buildspec))
+
+    modern_build_command = "python -m build --wheel -n"
```
behnazh-w: Don't we need to install `wheel` when we pass this option?

PR #1355, opened 2026-04-09
feat(ci): add Simple Index to serve distribution packages; add sdist package that can be used to build binary and other distribution artifacts; fix minor build issues

jenstroeger:
@behnazh-w it probably would make sense to break up this change into multiple feature PRs?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ share/python-wheels/
 .installed.cfg
 *.egg
 MANIFEST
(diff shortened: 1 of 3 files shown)


behnazh-w:
> @behnazh-w it probably would make sense to break up this change into multiple feature PRs?

Thanks for the PR. Yes, it would be good to submit multiple PRs for these changes.

PR #1356, opened 2026-04-10
build: use pip's dependency cooldown feature

behnazh-w:
## Summary
Use `pip`'s `--uploaded-prior-to` option to avoid installing packages that have very recently been published to PyPI. By doing so, this allows for human operators like security researchers and PyPI admins a chance to respond to reports of malware. See: https://blog.pypi.org/posts/2026-04-02-incident-report-litellm-telnyx-supply-chain-attack/

### DIFF
--- a/.github/workflows/_build.yaml
+++ b/.github/workflows/_build.yaml
@@ -1,4 +1,4 @@
-# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2026, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 # This is a trusted builder implemented as a reusable workflow that can be called by other
@@ -73,10 +73,11 @@ jobs:
     # Instead we create an empty .venv folder so that the Makefile continues to function
     # while Python operates within the runner's global environment. It is safe to ignore
     # warnings from the Makefile about the missing virtual environment.
+    # Remove PYPI_UPLOAD_DELAY_DAYS=1 when the security patches do not need it.
     - name: Create empty virtual environment for Actions
       run: mkdir .venv
     - name: Install dependencies
-      run: make setup
+      run: make setup PYPI_UPLOAD_DELAY_DAYS=1
 
     # Audit all currently installed packages for security vulnerabilities.
     - name: Audit installed packages
--- a/Makefile
+++ b/Makefile
@@ -227,12 +227,24 @@ gnu-sed:
 # here instead of `go get -u` to avoid updating indirect dependencies
 # and creating a broken state:
 # https://github.com/golang/go/issues/28424#issuecomment-1101896499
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -227,12 +227,24 @@ gnu-sed:
...
+# have very recently been published to PyPI. By doing so, this allows
+# for human operators like security researchers and PyPI admins a
+# chance to respond to reports of malware. See:
+# https://blog.pypi.org/posts/2026-04-02-incident-report-litellm-telnyx-supply-chain-attack/
+#
+# Override PYPI_UPLOAD_DELAY_DAYS for urgent security updates that were
+# released very recently, for example:
+# make setup PYPI_UPLOAD_DELAY_DAYS=1
+PYPI_UPLOAD_DELAY_DAYS ?= 2
+PYPI_UPLOADED_PRIOR_TO ?= $$(date -d '-$(PYPI_UPLOAD_DELAY_DAYS)days' -Idate)
```
jenstroeger, line 241: Might not apply here, but the default macOS `date` command will fail:
```bash
jens@pooh ~ > which date  # GNU date command.
/opt/local/libexec/gnubin/date
jens@pooh ~ > date -d '2days' -Idate
2026-04-12
jens@pooh ~ > /bin/date -d '2days' -Idate  # macOS default date command
/bin/date: illegal option -- d
usage: date [-jnRu] [-I[date|hours|minutes|seconds]] [-f input_fmt]
            [-r filename|seconds] [-v[+|-]val[y|m|w|d|H|M|S]]
            [[[[mm]dd]HH]MM[[cc]yy][.SS] | new_date] [+output_fmt]
```

.github/workflows/_build.yaml
```diff
@@ -73,10 +73,11 @@ jobs:
     # Instead we create an empty .venv folder so that the Makefile continues to function
     # while Python operates within the runner's global environment. It is safe to ignore
     # warnings from the Makefile about the missing virtual environment.
+    # Remove PYPI_UPLOAD_DELAY_DAYS=1 when the security patches do not need it.
     - name: Create empty virtual environment for Actions
       run: mkdir .venv
     - name: Install dependencies
-      run: make setup
+      run: make setup PYPI_UPLOAD_DELAY_DAYS=1
```
jenstroeger, line 80: I keep looking for the documentation of this notation 🤔 `PYPI_UPLOAD_DELAY_DAYS=1` _after_ the target is a variable passed to `make`?

If I use
```bash
PYPI_UPLOAD_DELAY_DAYS=1 make setup
```
then `PYPI_UPLOAD_DELAY_DAYS` is an environment variable passed to make ([discussion](https://unix.stackexchange.com/questions/56444/how-do-i-set-an-environment-variable-on-the-command-line-and-have-it-appear-in-c)) so it’s available in the Makefile ([docs](https://www.gnu.org/software/make/manual/html_node/Environment.html)).

But I can’t find the above notation documented?

jenstroeger (commented):
This is great and very useful to avoid immediate CI breakage with new packages!

PR #1358, opened 2026-04-10
chore: add a new Makefile target `simple-index` which builds all distribution artifacts and then creates a PEP-503 compatible Simple Index in the dist/ folder

jenstroeger:
## Summary

Add a new Makefile target `simple-index` which builds all distribution artifacts and then creates a PEP-503 compatible Simple Index in the dist/ folder.

## Description of changes

The Simple Index structure is defined in [PEP 503](https://peps.python.org/pep-0503/) and this change works as follows:
```bash
~ > mkdir /tmp/macaron-index/
~ > cd /tmp/macaron-index/
/tmp/macaron-index > python -m venv .venv  # Create a new virtual environment and activate it.
/tmp/macaron-index > . .venv/bin/activate
/tmp/macaron-index > pip install --extra-index-url file:///path/to/macaron/dist/simple-index/ macaron  # Install the Macaron package.
Looking in indexes: https://pypi.org/simple, file:///path/to/macaron/dist/simple-index/
Processing /path/to/macaron/dist/simple-index/macaron/macaron-0.23.0.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
WARNING: Location '/path/to/macaron/dist/simple-index/requests/' is ignored: it is either a non-existing path or lacks a specific scheme.
Collecting requests<3.0.0,>=2.32.3 (from macaron)
  Using cached requests-2.33.1-py3-none-any.whl.metadata (4.8 kB)
```
We can ignore the warnings because the Simple Index contains only the Macaron package; all other packages fall back to the global PyPI default index.

## Related issues

No related issue, but the changes are cherry-picked from PR https://github.com/oracle/macaron/pull/1355.

## Checklist

- [X] I have reviewed the [contribution guide](../CONTRIBUTING.md).
- [X] My PR title and commits follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) convention.
- [X] My commits include the "Signed-off-by" line.
- [X] I have signed my commits following the instructions provided by [GitHub](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits). Note that we run [GitHub's commit verification](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) tool to check the commit signatures. A green `verified` label should appear next to **all** of your commits on GitHub.
- [ ] I have updated the relevant documentation, if applicable.
- [ ] I have tested my changes and verified they work as expected.

PR #1377, opened 2026-04-15
build: make the Docker image building reproducible by using Macaron’s pinned and hashed requirements.txt

jenstroeger:
## Summary

Following up on comment https://github.com/oracle/macaron/pull/1358#issuecomment-4234829139, this change makes the Docker image building reproducible by using Macaron’s pinned and hashed requirements.txt.

## Description of changes

Building on the Simple Index artifact (see PR https://github.com/oracle/macaron/pull/1358) this change uses that local index and Macaron’s own generated requirements.txt to install Macaron into the Docker container using only the pinned and hashed dependencies.

## Related issues

Issues: n/a
Pulls: https://github.com/oracle/macaron/pull/1358

## Checklist

- [X] I have reviewed the [contribution guide](../CONTRIBUTING.md).
- [X] My PR title and commits follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) convention.
- [X] My commits include the "Signed-off-by" line.
- [X] I have signed my commits following the instructions provided by [GitHub](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits). Note that we run [GitHub's commit verification](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) tool to check the commit signatures. A green `verified` label should appear next to **all** of your commits on GitHub.
- [ ] I have updated the relevant documentation, if applicable.
- [ ] I have tested my changes and verified they work as expected.

### DIFF
--- a/.github/workflows/_build.yaml
+++ b/.github/workflows/_build.yaml
@@ -1,4 +1,4 @@
-# Copyright (c) 2022 - 2023, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2022 - 2026, Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/.
 
 # This is a trusted builder implemented as a reusable workflow that can be called by other
@@ -115,12 +115,13 @@ jobs:
         set -euo pipefail
         TARBALL_PATH=$(find dist/ -type f -name "*.tar.gz")
         WHEEL_PATH=$(find dist/ -type f -name "*.whl")
+        SIMPLE_INDEX_PATH=$(find dist/ -type f -name "*-simple-index.tar")
         REQUIREMENTS_PATH=$(find dist/ -type f -name "*-requirements.txt")
         SBOM_PATH=$(find dist/ -type f -name "*-sbom.json")
         SBOM_GO_PATH=$(find dist/ -type f -name "*-sbom-go.json")
         HTML_DOCS_PATH=$(find dist/ -type f -name "*-docs-html.zip")
         BUILD_EPOCH_PATH=$(find dist/ -type f -name "*-build-epoch.txt")
-        DIGEST=$(sha256sum "$TARBALL_PATH" "$WHEEL_PATH" "$REQUIREMENTS_PATH" "$SBOM_PATH" \
+        DIGEST=$(sha256sum "$TARBALL_PATH" "$WHEEL_PATH" "$SIMPLE_INDEX_PATH" "$REQUIREMENTS_PATH" "$SBOM_PATH" \
           "$SBOM_GO_PATH" "$HTML_DOCS_PATH" "$BUILD_EPOCH_PATH" | base64 -w0)
         echo "Digest of artifacts is $DIGEST."
         echo "artifacts-sha256=$DIGEST" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/_build.yaml
+++ b/.github/workflows/_build.yaml
@@ -107,7 +107,8 @@ jobs:
 
     # Find the paths to the artifact files that will be included in the release, compute
     # the SHA digest for all the release files and encode them using Base64, and export it
-    # from this job.
+    # from this job. Note that using -maxdepth 1 prevents searching through the generated
+    # Simple Index, which contains duplicated sdist and wheel files.
     - name: Compute package hash
       if: matrix.os == env.ARTIFACT_OS && matrix.python == env.ARTIFACT_PYTHON
       id: compute-hash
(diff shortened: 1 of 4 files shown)


.github/workflows/_build.yaml
```diff
@@ -113,14 +114,15 @@ jobs:
       shell: bash
       run: |
         set -euo pipefail
-        TARBALL_PATH=$(find dist/ -type f -name "*.tar.gz")
-        WHEEL_PATH=$(find dist/ -type f -name "*.whl")
+        TARBALL_PATH=$(find dist/ -maxdepth 1 -type f -name "*.tar.gz")
+        WHEEL_PATH=$(find dist/ -maxdepth 1 -type f -name "*.whl")
```
jenstroeger, line 118: Shall we replicate the `-maxdepth 1` argument to all `find`s here just for consistency? It’s not really necessary for the other files, but would also avoid `find` descending into the simple index directory…

Makefile
```diff
@@ -423,6 +423,8 @@ simple-index: dist/$(PACKAGE_WHEEL_DIST_NAME).whl dist/$(PACKAGE_SDIST_NAME).tar
 	echo -e "<!-- https://peps.python.org/pep-0503/ -->\n<!DOCTYPE html><html lang='en'><head><meta name='pypi:repository-version' content='1.3'><title>Simple Index: $(PROJECT_NAME)</title></head><body><ul><li><a href='$(PACKAGE_WHEEL_DIST_NAME).whl#sha256="$$(python -c "with open('dist/$(PACKAGE_WHEEL_DIST_NAME).whl', 'rb') as f: import hashlib; print(hashlib.sha256(f.read()).hexdigest());")"'>$(PACKAGE_WHEEL_DIST_NAME).whl</a></li><li><a href='$(PACKAGE_SDIST_NAME).tar.gz#sha256="$$(python -c "with open('dist/$(PACKAGE_SDIST_NAME).tar.gz', 'rb') as f: import hashlib; print(hashlib.sha256(f.read()).hexdigest());")"'>$(PACKAGE_SDIST_NAME).tar.gz</a></li></ul></body></html>" > dist/simple-index/$(PROJECT_NAME)/index.html
 	cp -f dist/$(PACKAGE_WHEEL_DIST_NAME).whl dist/simple-index/$(PROJECT_NAME)/
 	cp -f dist/$(PACKAGE_SDIST_NAME).tar.gz dist/simple-index/$(PROJECT_NAME)/
+	python -m zipfile --create dist/$(PACKAGE_SDIST_NAME)-pep503-simple-index.zip dist/simple-index/
+	python -m tarfile --create dist/$(PACKAGE_SDIST_NAME)-pep503-simple-index.tar dist/simple-index/
```
jenstroeger, line 427: We use the `.tar` for the Docker image because the tar command is available from the base image; alternatively we’d use `.zip` and use python to unzip. Any preferences?

.github/workflows/_build.yaml
behnazh-w: Trying to understand the options here. Was `-maxdepth 1` necessary for `WHEEL_PATH`?

jenstroeger: Yes, because `make simple-index` copies the sdist and wheel into the `dist/simple-index/` folder so that `find` then finds both:
```
(macaron) jens@pooh ~/path/to/macaron > find dist/ -type f -name "*.tar.gz"
dist/macaron-0.23.0.tar.gz
dist/simple-index/macron/macaron-0.23.0.tar.gz
```
whereas `--maxdepth` limits how “deep” the search goes:
```
(macaron) jens@pooh ~/path/to/macaron > find dist/ -maxdepth 1 -type f -name "*.tar.gz"
dist/macaron-0.23.0.tar.gz
```
From the [docs](https://linux.die.net/man/1/find):
```
-maxdepth levels
  Descend at most levels (a non-negative integer) levels of directories below the command line arguments. -maxdepth 0 means only apply the tests and actions to the command line arguments.
```
I thought that adding the maxdepth argument to all of the `find`s here avoids unnecessarily searching the  simple index folder for files we know aren’t there. Practically, it probably makes no difference though.

jenstroeger:
Oh dear, a timely discussion: [What would it look like to deprecate PEP 503?](https://discuss.python.org/t/what-would-it-look-like-to-deprecate-pep-503/106959)

They make a good point there. [PEP 503](https://peps.python.org/pep-0503/) is an HTML based simple index whereas [PEP 691](https://peps.python.org/pep-0691/) is JSON based. Given the supply chain developments of the past months it probably makes sense to switch to JSON… 🤔 For more details, see [Simple repository API](https://packaging.python.org/en/latest/specifications/simple-repository-api/) which discusses both formats.

.github/workflows/_build.yaml
behnazh-w: OK, thanks for the clarification. Then it's good to add the option to all the commands to be consistent. Also, please add a comment to explain why `-maxdepth 1` is used.

behnazh-w:
> Oh dear, a timely discussion: [What would it look like to deprecate PEP 503?](https://discuss.python.org/t/what-would-it-look-like-to-deprecate-pep-503/106959)
> 
> They make a good point there. [PEP 503](https://peps.python.org/pep-0503/) is an HTML based simple index whereas [PEP 691](https://peps.python.org/pep-0691/) is JSON based. Given the supply chain developments of the past months it probably makes sense to switch to JSON… 🤔 For more details, see [Simple repository API](https://packaging.python.org/en/latest/specifications/simple-repository-api/) which discusses both formats.

Thanks for finding and sharing the PEP. Yes, agreed to use `PEP 691` instead.

Makefile
behnazh-w: Is there a benefit for using `.zip`?

jenstroeger:
> Thanks for finding and sharing the PEP. Yes, agreed to use `PEP 691` instead.

I’ve not played around with serving JSON instead of HTML for the simple index, so that change would probably take a couple of weeks. Do you want to do that in a separate PR (similar to PR https://github.com/oracle/macaron/pull/1358)? Do you want to continue with this PR and a PEP 503 compatible simple index for now, or wait?

Makefile
jenstroeger: I think both sdist and wheel are already compressed data so using `zip` probably won’t help much beyond compressing the two html files; with `tar` you just bundle up the entire simple index, which is what we really want.[^1] Considering we’re talking about four files in total I think the difference is negligible and either `zip` or `tar` is just fine 🤷🏻‍♂️

Maybe flip a coin 😁

[^1]: Maybe see also [What is the difference between tar and zip?](https://stackoverflow.com/questions/10540935/what-is-the-difference-between-tar-and-zip) which links to further discussions.

.github/workflows/_build.yaml
jenstroeger: Commit https://github.com/oracle/macaron/pull/1377/changes/e7e9633e91ef7264d2b5c457614859a97977eb12.

behnazh-w:
> > Thanks for finding and sharing the PEP. Yes, agreed to use `PEP 691` instead.
> 
> I’ve not played around with serving JSON instead of HTML for the simple index, so that change would probably take a couple of weeks. Do you want to do that in a separate PR (similar to PR #1358)? Do you want to continue with this PR and a PEP 503 compatible simple index for now, or wait?

OK, in that case let's continue with this PR and switch to `PEP 691` in a follow up PR.

Makefile
behnazh-w: Then let's just create the `.tar` file.

jenstroeger: Commit https://github.com/oracle/macaron/pull/1377/changes/a8e59b32b95c933239922f435ebb5ae3e09629c3.

behnazh-w (approved):

PR #1408, opened 2026-05-15
feat: add inference for Python wheels that bundle Rust binaries.

AbhinavPradeep:
## Summary
Added inference for non-pure Python wheels that bundle Rust binaries

## Description of changes
1. Modified the buildspec to introduce the concept of a `SpecBuildRequirementDict`. This provides a richer description of build dependencies.
2. Added inference for the `maturin` build tool in specific. Rust-related build dependancies are read from the `Cargo.lock/toml` and the right Rust toolchain versions are inferred.

## Checklist

### DIFF
--- a/src/macaron/resources/schemas/macaron_buildspec_schema.json
+++ b/src/macaron/resources/schemas/macaron_buildspec_schema.json
@@ -122,7 +122,7 @@
         "required": ["name", "installer"],
         "additionalProperties": false
       },
-      "description": "Required packages that must be available in the build environment."
+      "description": "Build environment requirements, including the installer and version constraint for each requirement."
     },
     "build_backends": {
       "type": "array",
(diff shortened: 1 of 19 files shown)


src/macaron/resources/schemas/macaron_buildspec_schema.json
behnazh-w, line 1: Please add this change to https://github.com/oracle/macaron/blob/abhinav/python-rust-inference/src/macaron/resources/schemas/macaron_buildspec_schema.md and update the description where needed.

tests/integration/cases/pypi_tree-sitter/expected_default.buildspec
behnazh-w, line 1: Can you please also add the dockerfile output for this test case and run it locally to make sure it actually builds a valid wheel? It would also be good to run the docker build for the other integration tests to make sure we don't have regressions.

AbhinavPradeep: Added validation for the Rust cases! Tree sitter is one that bundles C++, so we do not expect it to build a wheel at all. We do not generate a dockerfile for this case actually.

behnazh-w (approved):

PR #1438, opened 2026-07-25
chore(deps): update semgrep to the latest

behnazh-w:
## Summary
Update semgrep to the latest version and remove the ignored advisories found in its dependencies. This PR also improves some of the tests that would fail on Python 3.14.


tests/database/test_database_manager.py
```diff
@@ -37,11 +37,15 @@ def db_man() -> Iterable:
...
-        con.execute("drop table if exists _test_orm_table;")
-        con.execute("drop view if exists test_orm_table;")
-        con.commit()
+    try:
+        with con:
+            con.execute("drop table if exists _test_orm_table;")
+            con.execute("drop view if exists test_orm_table;")
+            con.commit()
+    finally:
+        con.close()
```
jenstroeger: Take a look at `contextlib.closing()` ([docs](https://docs.python.org/3/library/contextlib.html#contextlib.closing)).

jenstroeger (commented):
I noticed that `create_engine()` is called in different places for local instances of SQLA’s Engine though it should be a single global instance for the application’s lifetime, see also docs [here](https://docs.sqlalchemy.org/en/20/core/connections.html#basic-usage) and [here](https://docs.sqlalchemy.org/en/20/core/connections.html#engine-disposal). Similarly, I think the test fixture `macaron_db_session()` could use context managers for its session.

behnazh-w:
> I noticed that `create_engine()` is called in different places for local instances of SQLA’s Engine though it should be a single global instance for the application’s lifetime

We have three main CLI commands that need their own engines, but for the duration of a command, we only create one engine. The three CLI commands that are run independently are `analyze`, `verify-policy`, and `gen-build-spec`.

> Similarly, I think the test fixture macaron_db_session() could use context managers for its session.

Yes, we could improve that in a future PR perhaps.

tests/database/test_database_manager.py
behnazh-w: Thanks for the suggestion. See this commit [a393b3f](https://github.com/oracle/macaron/pull/1438/commits/a393b3fbe508abb57458a1fa8d835cbe845f2799).

tests/database/test_database_manager.py
```diff
@@ -37,11 +38,13 @@ def db_man() -> Iterable:
     """Set up the database and ensure it is empty."""
     db_manager = DatabaseManager(DB_PATH, base=Base)
     con = sqlite3.connect(DB_PATH)
-    with con:
+    with closing(con), con:
```
jenstroeger, line 41: Interesting, you’re not using `pylint`’s [code-style checker](https://pylint.readthedocs.io/en/stable/user_guide/checkers/extensions.html#code-style-checker) so you didn’t get the tip to change these into
```python
with closing(con := sqlite3.connect(DB_PATH)), con:
    ...
```

